PR #250, opened 2024-06-19
feat: Chat bubble design

Yonom:


Yonom:
Bugs to fix: 
image

ActionBar hover tracking is buggy

PR #346, opened 2024-06-28
feat: markdown math support

Yonom:


Acervans:
Hi, I've been trying this but couldn't get the mathematical formulas to show correctly (they were just shown as raw text). I checked the commits to the `markdown-text.tsx` component and saw that in dd09e3392dd43da56aa3b2aecdd64ab5891aae21 , for some reason the `remarkMath` and `rehypeKatex` for parsing math expressions were gone. Was there a reason to have them removed? Thanks

PR #597, opened 2024-07-28
feat: add streaming with `useLocalRuntime`

Rajaniraiyn:
add streaming support to local runtime via async iterator

Example

```ts
import { type ChatModelAdapter } from "@assistant-ui/react";

const CustomModelAdapter: ChatModelAdapter = {
  async run({ messages, signal }) {
    const res = await fetch("/api/chat", {
      method: "POST",
      body: JSON.stringify({ messages }),
      signal,
    });

    const reader = res.body.getReader();
    const decoder = new TextDecoder();
    let done = false;
    let content = '';

    return (async function* () {
      while (!done) {
        const { value, done: readerDone } = await reader.read();
        done = readerDone;
        if (value) {
          const chunk = decoder.decode(value, { stream: true });
          content += chunk;
          yield {
            content: [{ type: "text", text: content }],
            status: { type: "running" },
          };
        }
      }
    })();
  },
};

export default CustomModelAdapter;
```

### DIFF
--- a/.changeset/silly-otters-smash.md
+++ b/.changeset/silly-otters-smash.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react": patch
+---
+
+feat(local-runtime): AsyncGenerator support
--- a/apps/docs/content/docs/docs/runtimes/custom-rest.mdx
+++ b/apps/docs/content/docs/docs/runtimes/custom-rest.mdx
@@ -120,3 +120,28 @@ export default function RootLayout({
 
   </Step>
 </Steps>
+
+## Streaming
+
+Declare the `run` function as an `AsyncGenerator` (`async *run`). This allows you to `yield` the results as they are generated.
+
+```tsx {2, 11-13}
+const MyModelAdapter: ChatModelAdapter = {
+  async *run({ messages, abortSignal, config }) {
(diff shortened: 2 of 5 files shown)


Yonom:
@Rajaniraiyn this is an amazing contribution! thank you so much!

Yonom:
I made a small adjustment to your code:

```ts
import { type ChatModelAdapter } from "@assistant-ui/react";

const CustomModelAdapter: ChatModelAdapter = {
  async *run({ messages, signal }) { // made the outer function async iterable
    const res = await fetch("/api/chat", {
      method: "POST",
      body: JSON.stringify({ messages }),
      signal,
    });

    const reader = res.body.getReader();
    const decoder = new TextDecoder();
    let done = false;
    let content = '';

    while (!done) {
      const { value, done: readerDone } = await reader.read();
      done = readerDone;
      if (value) {
        const chunk = decoder.decode(value, { stream: true });
        content += chunk;
        yield { // now we can directly yield here!
          content: [{ type: "text", text: content }],
          status: { type: "running" },
        };
      }
    }
  },
};

export default CustomModelAdapter;
```

Yonom (approved):
LGTM

PR #598, opened 2024-07-28
feat: Images support to `VercelUseChatRuntime`

Rajaniraiyn:
images content type for user message when using `useVercelChatRuntime`

```ts
import { useAppendMessage } from "@assistant-ui/react";

const CustomButton = () => {
  const append = useAppendMessage();

  const handleClick = () => {
    append({
    role: "user",
    content: [{type: "image", image: "https://example.com/example.png"}]
    });
  };

  return <button onClick={handleClick}>Send</button>;
};
```


Yonom:
Hmm... I would love to add support for images, but... The Vercel useChat runtime doesn't seem to natively support appending images yet. The typescript types state that Message.content is a string. Your method only accidentally works if the backend endpoint does not use "convertToCoreMessages(messages)"...

Rajaniraiyn:
Thanks for the feedback,

Totally get it. Maybe the error message could mention the Edge runtime for more clarity.

Yonom:
@Rajaniraiyn  good idea - done https://github.com/Yonom/assistant-ui/pull/602

PR #643, opened 2024-08-04
Ensure Tool Results are Added for Multiple Parallel Tool Calls

Rajaniraiyn:
This update ensures that all tool calls in an assistant message receive a corresponding result before performing a round trip. Without this fix, most API providers will throw an error due to missing results from parallel tool calls.

### DIFF
--- a/packages/react/src/runtimes/local/LocalThreadRuntime.tsx
+++ b/packages/react/src/runtimes/local/LocalThreadRuntime.tsx
@@ -213,13 +213,11 @@ export class LocalThreadRuntime implements ThreadRuntime {
     if (message.role !== "assistant")
       throw new Error("Tried to add tool result to non-assistant message");
 
-    let added = false;
     let found = false;
     const newContent = message.content.map((c) => {
       if (c.type !== "tool-call") return c;
       if (c.toolCallId !== toolCallId) return c;
       found = true;
-      if (!c.result) added = true;
       return {
         ...c,
         result,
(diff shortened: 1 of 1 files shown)


packages/react/src/runtimes/local/LocalThreadRuntime.tsx
```diff
@@ -235,7 +233,12 @@ export class LocalThreadRuntime implements ThreadRuntime {
     };
     this.repository.addOrUpdateMessage(parentId, message);
 
-    if (added && shouldContinue(message)) {
+    // Check if all tool calls have results
+    const allToolCallsHaveResults = message.content.every(
+      (c) => c.type !== "tool-call" || c.result,
```
Yonom, line 238: @Rajaniraiyn shouldContinue already checks for all tool calls having a result. this seems to be a duplicate check of the same thing

https://github.com/Yonom/assistant-ui/blob/c92547de23b878ec088f7c888dab4d1b3db1756e/packages/react/src/runtimes/local/shouldContinue.tsx#L6

packages/react/src/runtimes/local/LocalThreadRuntime.tsx
```diff
@@ -213,13 +213,11 @@ export class LocalThreadRuntime implements ThreadRuntime {
     if (message.role !== "assistant")
       throw new Error("Tried to add tool result to non-assistant message");
 
-    let added = false;
```
Yonom, line 216: Why is this added check removed?

It checks for repeat calls to addToolResult() to ensure that a roundtrip is only triggered if a tool result is added (and not updated)

Yonom (changes requested):
I am not able to reproduce the bug you specify... 

Furthermore, this code removes the added check (why?) and the added check is already in shouldContinue (does that one not work?)

Yonom:
There is a adjacent bug that I found while testing for this. 

If a message has pending tool calls; one can send new messages via the composer. I am fixing this in #644

Rajaniraiyn:
Closing this PR as it has duplicate checks and there is a better way to do the tool execution via `useAssistantTool` instead of `useAssistantToolUI`

Yonom:
@Rajaniraiyn thank you for the submission! I really appreciate your efforts :)

PR #1065, opened 2024-10-24
feat: Edge Runtime Server Accessible Ids

Seanw265:
Implements a method of accessing ids of existing messages and new messages from the server on the Edge Runtime. This enables the server to maintain a copy of a data structure compatible with the `.import(...)` function available on some runtimes.

This PR consists of three changes:
1. The LocalRuntime's `performRoundtrip` now supplies its chat adapter's `*run` async generator with `unstable_assistantMessageId`.
2. The Edge Runtime includes `unstable_assistantMessageId` in the body of the request it makes to the backend.
3. `useEdgeRuntime` now provides an `unstable_sendMessageIds` option. When enabled, the EdgeRuntime will not strip the `id` field from messages before sending them to the server.

Usage:
```
// within runtime provider
const runtime = useEdgeRuntime({
  api: "/api/chat",
  unstable_sendMessageIds: true,
});



// within server route handler
const {
  messages, // messages may contain an `id` if `unstable_sendMessageIds` is enabled on the frontend
  unstable_assistantMessageId: newMessageId, // the id of the message that is about to generate
  ...rest
} = await req.json();

...

// extract the latest message and the message it replies to
const userLatestMessage = messages.at(-1);
const previousMessage = messages.at(-2);

// store new user message
storeMessage({
  ...newUserMessage,
  parentId: previousMessage?.id
})

...

const assistantResponse = await getEdgeRuntimeResponse({
  options: {
    async onFinish(finishResult) {
      // extract the assistant's generated message
      const assistantGeneratedMessage = finishResult.messages.at(-1);
      
      // store new assistant message
      storeMessage({
        ...assistantGeneratedMessage,
        id: unstable_assistantMessageId,
        parentId: newUserMessage.id,
      })
    }
  }
});
...
```

This change was discussed on Discord before implementation:
https://discord.com/channels/1251324227668283443/1251754454139539547/1298767870469083178

### DIFF
--- a/.changeset/fluffy-birds-burn.md
+++ b/.changeset/fluffy-birds-burn.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react": patch
+---
+
+feat: Edge Runtime Server Accessible Ids (temp)
--- a/packages/react/src/runtimes/edge/EdgeChatAdapter.ts
+++ b/packages/react/src/runtimes/edge/EdgeChatAdapter.ts
@@ -31,12 +31,22 @@ export type EdgeChatAdapterOptions = {
   credentials?: RequestCredentials;
   headers?: Record<string, string> | Headers;
   body?: object;
+
+  /**
+   * When enabled, the adapter will not strip `id` from messages in the messages array.
+   */
+  unstable_sendMessageIds?: boolean;
 };
 
 export class EdgeChatAdapter implements ChatModelAdapter {
   constructor(private options: EdgeChatAdapterOptions) {}
 
-  async *run({ messages, abortSignal, config }: ChatModelRunOptions) {
+  async *run({
+    messages,
+    abortSignal,
+    config,
+    unstable_assistantMessageId,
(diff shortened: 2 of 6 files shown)


Yonom:
@seanw265 I renamed "id" to "unstable_id" because I believe long term I do not want to add it to CoreMesages (trying to keep it as lightweight and close to LLM formats as possible)

another small bug I fixed:
```ts
type CoreMessageWithConditionalId<T extends boolean> = T extends true
  ? CoreMessage & { id?: string }
  : CoreMessage;
```

this would cause `CoreMessageWithConditionalId<boolean>` to have the type `CoreMesesage`

fixed

```ts
type CoreMessageWithConditionalId<T extends boolean> = T extends false
  ? CoreMessage
  : CoreMessage & { unstable_id?: string };
```

Yonom (approved):

Seanw265:
Awesome! Thanks for the help on this. This makes Assistant UI much more usable for my needs.

PR #1068, opened 2024-10-24
build: disable bundling in react component package releases

Yonom:


Yonom:
This PR is blocked because .cjs build outputs do not put "use client" on top of the file...

PR #1080, opened 2024-10-26
feature: add Trieve integration guide and use Trieve Search

densumesh:


Yonom (approved):

Yonom:
Thank you so much! This is great! will do some more testing and then merge

PR #1114, opened 2024-10-28
feature: hook into page source API to send chunks to Trieve and use group search for component

densumesh:


Yonom:
Thanks, this is great!

A few things:
- this repository is built somewhere between 50x-100x a day (on my own machine, in PRs, twice in prod (github actions + vercel deployment)
  - non-prod builds definitely should not push anything to trieve
  - perhaps the sync script should only run when anything in `/apps/docs/content` is changed (see `.github/workflows/registry.yaml` for a similar workflow)
- how are old chunks discarded?

apps/docs/content/docs/runtimes/langserve.mdx
```diff
@@ -34,7 +34,7 @@ npm install @assistant-ui/react @assistant-ui/react-ai-sdk ai ai/react @langchai
 ### Setup a backend route under `/api/chat`
 
 ```tsx twoslash title="@/app/api/chat/route.ts"
-// @errors: 2558 2345
+// @errors: 2558 2345 2739
```
Yonom: I don't think this is necessary

apps/docs/lib/update-index.mts
```diff
@@ -0,0 +1,76 @@
+import { ChunkReqPayload, TrieveSDK } from "trieve-ts-sdk";
+import * as fs from "node:fs";
+import { DocumentRecord } from "fumadocs-core/search/algolia";
+import { env } from "node:process";
+import "dotenv/config";
+
+let trieve = new TrieveSDK({
+  apiKey: env["NEXT_PUBLIC_TRIEVE_API_KEY"] as string,
```
Yonom: preferably we can configure a separate "TRIEVE_API_KEY" variable here and the public key has no write access to the index

apps/docs/components/ui/search.tsx
```diff
@@ -23,113 +24,52 @@ type ChunkWithHighlights = {
...
-  chunkHtmlHeadingsDiv.innerHTML = item.chunk.chunk_html || "";
-  const chunkHtmlHeadings = chunkHtmlHeadingsDiv.querySelectorAll(
-    "h1, h2, h3, h4, h5, h6",
-  );
-
-  const $firstHeading = chunkHtmlHeadings[0] ?? document.createElement("h1");
-  $firstHeading?.querySelectorAll(":not(mark)")?.forEach((tag) => {
-    return tag.replaceWith(tag.textContent || "");
+export default function TrieveSearchDialog(props: SharedProps) {
+  const trieveSDK = new TrieveSDK({
```
Yonom: currently I get an ESLint error because the useEffect hook does not have trieveSDK or groupresults etc as part of its dependencies 

move trieve SDK declaration + static functions outside of the render function or memoize them

apps/docs/components/ui/search.tsx
```diff
@@ -23,113 +24,52 @@ type ChunkWithHighlights = {
...
+    apiKey: process.env["NEXT_PUBLIC_TRIEVE_API_KEY"] as string,
```
Yonom: env variables need to be added to turbo.json:

```
"env": [<...>, "NEXT_PUBLIC_*", "TRIEVE_*"]
```

apps/docs/components/ui/search.tsx
```diff
@@ -23,113 +24,52 @@ type ChunkWithHighlights = {
...
-
-  return results as SortedResult[];
-};
 
-export default function TrieveSearchDialog(props: SharedProps) {
-  const [results, setResults] = useState<SortedResult[]>([]);
-  const [search, setSearch] = useState("");
-  const [isLoading, setIsLoading] = useState(false);
+      for (const c of result.chunks) {
+        let chunk = c.chunk;
```
Yonom: build error:

```suggestion
        const chunk = c.chunk;
```

densumesh:
>   * non-prod builds definitely should not push anything to trieve
>   * perhaps the sync script should only run when anything in `/apps/docs/content` is changed (see `.github/workflows/registry.yaml` for a similar workflow)
I think that I can just run the script from the github workflow then rather than the build command, as it just needs a built version of the app
> how are old chunks discarded?
We expose an API for clearing, that I realize we can just use here. We would just provide two api keys, one for the public facing search and the other for the actual modification and deletion of chunks

Yonom (changes requested):

apps/docs/lib/update-index.mts
densumesh: Yes the current one that is being used is read only

Yonom (approved):
LGTM

PR #1583, opened 2025-02-10
refactor: use generics with runtime typing

PieterT2000:


Yonom:
@PieterT2000 thank you so much for the PR.

I decided to go with a different approach: unifying all runtime APIs.
The reset method is mapped to:

```ts
runtime.reset({ initialMessages });
// --- to ---
runtime.threads.main.import(ExportedMessageRepository.fromArray(initialMessages);
```

PR #1601, opened 2025-02-12
Add 'data' type to ThreadMessageLike

brandonbevans:
For Vercel AI SDK compatibility with messages, ThreadMessageLike needs to be able to accept a 'data' role as well, otherweise we get a type error like the following:
Screenshot 2025-02-12 at 11 12 10 AM

Where "Message[]" is from the "ai" package

`import { AssistantRuntimeProvider } from '@assistant-ui/react';
import { useChatRuntime } from '@assistant-ui/react-ai-sdk';
import { Message } from 'ai';
export function MyRuntimeProvider({
  children,
  chatId,
  namespace,
  messages
}: Readonly<{
  children: React.ReactNode;
  messages: Message[];
}>) {
  // const initialMessages = messages.map((msg) => ({
  //   role: msg.role as 'system' | 'user' | 'assistant',
  //   content: msg.content,
  //   createdAt: msg.createdAt
  // }));

  const runtime = useChatRuntime({
    api: '/api/chat',
    initialMessages: messages
  });
  return (
      {children}
    </AssistantRuntimeProvider>
  );
}`

Currently for this code to work, the commented out code is required, because MessageThreadLike doesn't allow for the role: 'data' that is present in the Message type from the "ai" package

### DIFF
--- a/packages/react/src/runtimes/external-store/ThreadMessageLike.tsx
+++ b/packages/react/src/runtimes/external-store/ThreadMessageLike.tsx
@@ -1,17 +1,17 @@
 import { generateId } from "../../internal";
 import {
-  MessageStatus,
-  TextContentPart,
-  ImageContentPart,
-  UIContentPart,
-  ThreadMessage,
-  ThreadAssistantContentPart,
-  ThreadAssistantMessage,
-  ThreadUserContentPart,
-  ThreadUserMessage,
-  ThreadSystemMessage,
   CompleteAttachment,
   FileContentPart,
+  ImageContentPart,
+  MessageStatus,
+  TextContentPart,
+  ThreadAssistantContentPart,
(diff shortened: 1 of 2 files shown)


brandonbevans:
Hey @Yonom I would love to add this change but the bots bring up a good point about not propogating this "data" content type through the code. It seems like it may be a bit beyond me.

Yonom:
Hmm - ThreadMessageLike is not intended to be compatible with AI SDK's message type... it just happens to overlap enough for some cases

nevertheless - especially when using useChatRuntime, I'd want to let you pass in AI SDK UI message objects to initialMessages - so the solution is probably to add a ~`convertUIMessage` helper function which converts messages to ThreadMessage - and then have useChatRuntime support it

PR #1744, opened 2025-03-21
fix: no duplicate tool calls in edge runtime in case tool-deltas are enabled

stephg:
If tool-call-deltas are enabled, a `ToolCall` chunk will call the tool anyways, so there is no need for any fallback-completion for the "current" tool.

----

> [!IMPORTANT]
> Prevent duplicate tool calls in `assistantDecoderStream.ts` by checking for `ToolCall` and `ToolCallDelta` types.
> 
>   - **Behavior**:
>     - In `assistantDecoderStream.ts`, updated `transform` function to prevent duplicate tool calls by checking for `AssistantStreamChunkType.ToolCall` and `ToolCallDelta`.
>     - Ensures no fallback-completion for current tool when `ToolCall` chunk is present.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=assistant-ui%2Fassistant-ui&utm_source=github&utm_medium=referral)<sup> for 3af2e6744b0b7a9703b78515b627ab22d4edd325. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/packages/react/src/runtimes/edge/streams/assistantDecoderStream.ts
+++ b/packages/react/src/runtimes/edge/streams/assistantDecoderStream.ts
@@ -31,7 +31,8 @@ export function assistantDecoderStream() {
   >({
     transform({ type, value }, controller) {
       if (
-        type !== AssistantStreamChunkType.ToolCallDelta &&
+        type !== AssistantStreamChunkType.ToolCall &&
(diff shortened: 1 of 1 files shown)


stephg:
@Yonom do you need further details about the issue?
I don't see any tests or something that need to be adapted.

PR #2202, opened 2025-07-19
feat: messages-tuple support

Yonom:


eladlachmi:
@Yonom I'd really like to help make progress with `messages-tuple` support.  
I can test this change with LangGraph cloud, if you think this PR can still be salvaged.  
WDYT?

PR #2479, opened 2025-09-15
fix: argsText parsing

Yonom:


aiswaryasankar:
**🔒 Entelligence AI Vulnerability Scanner**

❌ Security analysis failed: Security analysis failed: {"sandboxId":"iey4z1561u3hogm5ymj31","message":"The sandbox is running but port is not open","port":49999,"code":502}: This error is likely due to sandbox timeout. You can modify the sandbox timeout by passing 'timeout' when starting the sandbox or calling '.set_timeout' on the sandbox with the desired timeout.

PR #2718, opened 2025-11-11
fix(react-ai-sdk): correctly initialize history loading state

### LINKED ISSUE #2714: ThreadPrimitive.If empty never matches when useChatRuntime lacks a history adapter
After upgrading to @assistant-ui/react@0.11.37, `<ThreadPrimitive.If empty>` never renders its empty branch when the assistant runtime is created via useChatRuntime without a history adapter. Even with zero messages in the current thread, thread.isLoading stays true, so the condition `messages.length === 0 && !isLoading` never passes.

### Observation:
Logging useAssistantState(({ thread }) => thread) right after mounting shows:
```
{
  "messages": [],
  "isLoading": true,
  "isRunning": false,
  "suggestions": []
}
```
Because ThreadPrimitive.If implements the empty guard as `thread.messages.length === 0 && !thread.isLoading`, the welcome view is permanently hidden.

### Minimal repro:
```
import {
  AssistantRuntimeProvider,
  ThreadPrimitive,
} from "@assistant-ui/react";
import { useChatRuntime } from "@assistant-ui/react-ai-sdk";

export function Demo() {
  const runtime = useChatRuntime({
    // No history adapter supplied
    transport: { api: "/api/demo" },
  });

  return (
      <ThreadPrimitive.Root>
        <ThreadPrimitive.Viewport>
          <ThreadPrimitive.If empty>
            <div data-testid="welcome">Welcome!</div>
          </ThreadPrimitive.If>
          <ThreadPrimitive.Messages />
        </ThreadPrimitive.Viewport>
      </ThreadPrimitive.Root>
    </AssistantRuntimeProvider>
  );
}
```

Running this in a fresh environment (no messages yet, backend not called) never renders the `<div data-testid="welcome">` node because thread.isLoading stays true.

### Root cause
`useChatRuntime` now calls `useExternalHistory`. That hook initializes `const [isLoading, setIsLoading] = useState(true)` and only toggles it to false inside `loadHistory()`. When no history adapter is provided, the hook returns early (`if (!historyAdapter || loadedRef.current) return;`), so `setIsLoading(false)` is never invoked. The runtime therefore reports `thread.isLoading === true` indefinitely.

Suggested fix: default `isLoading` to false and only set it to true while the history adapter is actually fetching data. Either approach restores the expected contract so that ThreadPrimitive.If empty works without forcing downstream apps to register a no-op history adapter.

ShobhitPatra:
The useExternalHistory hook could initialize its isLoading state to true even when no history adapter was provided. This 
caused the loading state to get stuck, preventing UI components like ThreadPrimitive.If empty from rendering correctly. This change defaults the isLoading state to false, ensuring it only becomes true when a history fetch is actively in  progress.
Fixes #2714

----

> [!IMPORTANT]
> Fixes `isLoading` state initialization in `useExternalHistory` to prevent stuck loading state when no `historyAdapter` is provided, resolving issue #2714.
> 
>   - **Behavior**:
>     - Fixes `isLoading` state initialization in `useExternalHistory` to `false` when no `historyAdapter` is provided.
>     - Ensures `isLoading` only becomes `true` during an active history fetch.
>   - **Bug Fix**:
>     - Resolves issue #2714 where UI components like `ThreadPrimitive.IfEmpty` were not rendering correctly due to stuck loading state.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=assistant-ui%2Fassistant-ui&utm_source=github&utm_medium=referral)<sup> for 53151646036346d90f079353de0904a0914d6f36. You can [customize](https://app.ellipsis.dev/assistant-ui/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/.changeset/clean-tips-obey.md
+++ b/.changeset/clean-tips-obey.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react-ai-sdk": patch
+---
+
+fix(react-ai-sdk): correctly initialize history loading state


AVGVSTVS96:
> ✅ No prettier issues found in this PR!

prettier failure coming from issue in unrelated file, merging anyway

Yonom:
This will cause the initial UI to flash when we have next router integration. The check will need to be re-added, until then, this works though

PR #2750, opened 2025-11-19
feat: visually anchor new user messages to the top of the thread viewport

petekp:


packages/react/src/primitives/thread/useThreadViewportAutoScroll.tsx
```diff
@@ -36,37 +40,173 @@ export const useThreadViewportAutoScroll = <TElement extends HTMLElement>({
...
-  const handleScroll = () => {
-    const div = divRef.current;
-    if (!div) return;
+  const updateSpacerForAnchor = useCallback(
+    (element: HTMLElement) => {
+      const div = divRef.current;
+      if (!div) return;
+
+      const spacer = div.querySelector<HTMLElement>(
+        ".aui-thread-viewport-spacer",
```
Yonom, line 49: Using a hardcoded CSS class name in our Radix UI components is an antipattern / the class name is hard to discover from the JS API surface. I would rather we find a more elegant way to do this. Let's go through this tomorrow to cleanly separate the API between the library and shadcn templates

packages/react/src/primitives/thread/useThreadViewportAutoScroll.tsx
```diff
@@ -15,6 +15,9 @@ export namespace useThreadViewportAutoScroll {
   };
 }
 
+const USER_MESSAGE_TOP_PADDING = 16;
```
Yonom, line 18: hardcoded paddings in the the primitives is an anitpattern, needs a different solution

packages/react/src/primitives/thread/useThreadViewportAutoScroll.tsx
```diff
@@ -1,7 +1,7 @@
 "use client";
```
Yonom, line 1: +190 lines of extra logic in this file is a lot! my gut feeling says it can be simplified to not rely on so many magic values & specific hardcoded names and styles. Happy to pair and audit this but I have a hard time doing so by just reading the file on my own

petekp:
Superceded by a new approach in #2759

Issue #2736, opened 2025-11-17
useRemoteThreadList cannot wrap useChatRuntime


Yonom:
useChatRuntime should be nestable in another useRemoteThreadList, in which case the inner threadlist will become a no-op

eladlachmi:
@Yonom Same issue with `useLangGraphRuntime`, and likely the same reason as I mentioned in #2734

gertsteyn:
Hi @Yonom 

I'm struggling to get around #2681 , am I correct in my conclusion that it is just another symptom of this issue?


_PS: Thanks for all your great work on assistant-ui, I can see it's going to be the foundation of numerous products!_

lucascurti:
@Yonom I've just opened a PR (https://github.com/assistant-ui/assistant-ui/pull/2752) related it to this. It adds the ability to pass a custom `remoteThreadListadapter` to useLangGraphRuntime. Let me know if you are thinking about a different approach and I'll adapt it.

PR #2752, opened 2025-11-19
Add custom remoteThreadListadapter for useLangGraphRuntime()

lucascurti:
### Summary

This PR adds support for custom thread list adapters in `useLangGraphRuntime`, allowing developers to bring their own thread storage implementations instead of being limited to the cloud adapter.

Closes https://github.com/assistant-ui/assistant-ui/issues/2736

#### New Feature: Custom Thread List Adapter

- Added a new optional `remoteThreadListAdapter` parameter to `UseLangGraphRuntimeOptions`
- When provided, the custom adapter takes precedence over the default cloud adapter
- This enables use cases where developers need to manage thread storage through their own LangGraph backend or other custom solutions

#### Bug Fix: Prevent Redundant Thread Loading

- Fixed an issue where threads were being loaded multiple times unnecessarily
- Added `lastLoadedIdRef` to track the last loaded `externalId` and prevent duplicate loads
- Moved to using `useAssistantState` to reactively track `externalId` changes instead of directly accessing runtime state
- The `loadThread` function now only executes when the `externalId` actually changes

### Problems Solved

1. **Limited flexibility**: Previously, users were forced to use the cloud adapter for thread list management. This change allows developers to use custom backends (e.g., their own LangGraph server) for thread storage and management.

2. **Performance issue**: The hook was triggering redundant `loadThread` calls even when the `externalId` hadn't changed, causing unnecessary network requests and potential race conditions. The new implementation ensures threads are only loaded once per unique `externalId`.

3. **Better reactivity**: By using `useAssistantState` to track `externalId` changes, the hook now properly responds to state updates in a more React-idiomatic way.

### API Addition

```typescript
remoteThreadListAdapter?: unstable_RemoteThreadListAdapter
```

Use this when you need custom thread storage instead of the default cloud storage.

---

### DIFF
--- a/.changeset/empty-parrots-cross.md
+++ b/.changeset/empty-parrots-cross.md
@@ -2,4 +2,4 @@
 "@assistant-ui/react-langgraph": patch
 ---
 
-Add the ability to use a custom remoteThreadListadapter for useLangGraphRuntime()
+Add the ability to use a custom remoteThreadListAdapter for useLangGraphRuntime()


elvenking:
Really appreciate this fix. Can we get it merged ?

PR #2757, opened 2025-11-19
feat(react-ai-sdk): mark tool call as complete once user send new message

### LINKED ISSUE #2756: Messaging instead of interacting with a tool call, breaks the chat
After updating to 
```
 "@assistant-ui/react": "^0.11.41",
"@assistant-ui/react-ai-sdk": "^1.1.12",
```
The Tool Calling behaviour changed in a breaking way. Before, when the assistant called a frontend tool that is interactive and for example adds a user input via addResult(), the user could decide not to interact with that tool ui but write a message instead and the conversation flow continued (toot call immediately was set to output available). After the update, the tool call message part remains as input-available and when sending a new message instead of interacting with the tool and triggering the addResult() function, i recieve the error `No tool output found for function call call_XYZ` .
This is probably the more correct behaviour however it can easily break the assistant when the user writes something like "No i dont want that" or similar instead of interacting with the tool UI.

Could we not mark all tools as completed once the user sends a new message or find another way of handling this?

Maybe there is a recommended way that I am just missing?

Appreciate the support!

okisdev:
This PR marks tool call as complete once user send new message when tool calling is processing.

This PR closes #2756.

Features:
- [x] logic: handle user sends message that cancelling toolcall
- [x] logic: can config as `cancelPendingToolCallsOnSend`, default as `true`
- [x] ui: adding state of cancelled to tool-fallback

### DIFF
--- a/packages/react-ai-sdk/src/ui/use-chat/useAISDKRuntime.tsx
+++ b/packages/react-ai-sdk/src/ui/use-chat/useAISDKRuntime.tsx
@@ -151,13 +151,6 @@ export const useAISDKRuntime = <UI_MESSAGE extends UIMessage = UIMessage>(
 
     // Mark tools as errored in the message history
     pendingHumanTools.forEach(({ toolCallId }) => {
-      // chatHelpers.addToolOutput({
-      //   state: "output-error",
-      //   tool: toolCallId,
-      //   toolCallId,
-      //   errorText: "User cancelled tool call by sending a new message.",
-      // });
-
       chatHelpers.setMessages(
         chatHelpers.messages.map((message) => {
           if (message.id === toolCallId) {


Yonom (changes requested):
This is a great addition. Feature requests before we merge:

- cancelPendingToolCallsOnSend: default true
- addToolOutput should add an error output (instead of success), the errorText should say "User cancelled tool call by sending a new message." or similar.

Follow up feature request in AUI: we should have a extra tool call status for cancelled tools, so that the Tool UI can show a special state for when tool calls are cancelled like this

okisdev:
@Yonom 

thanks for your feedback

i have completed the feature you requested for your further review. thx.

packages/react-ai-sdk/src/ui/use-chat/useAISDKRuntime.tsx
```diff
@@ -129,6 +182,8 @@ export const useAISDKRuntime = <UI_MESSAGE extends UIMessage = UIMessage>(
       toolInvocations.abort();
     },
     onNew: async (message) => {
+      completePendingToolCalls();
```
Yonom, line 195: AI SDK has a option called
sendAutomaticallyWhen: lastAssistantMessageHasCompleteToolCalls

with this option, the AI SDK immediately sends a request, so we end up with maybe two requests suddenly?

PR #2791, opened 2025-11-26
feat:  add shimmer-bg, eased gradients, shimmer-container, and enhanced inheritability to tw-shimmer

petekp:
## Summary

  Adds `shimmer-bg` utility for skeleton loaders with sine-eased gradients that eliminate the visible banding common in shimmer effects.

  **Key features:**
  - `shimmer-bg` utility for skeleton/background shimmer (non-text elements)
  - **Sine-eased gradients** with 13-17 stops following a sine ease-in-out curve for buttery-smooth highlights
  - OKLCH color space for perceptually uniform color mixing
  - `--shimmer-bg-x/y` CSS variables for pixel-perfect diagonal sync across elements

----

> [!IMPORTANT]
> Enhance `tw-shimmer` with new utilities for background shimmer, position sync, and angle adjustments, along with updated documentation and examples.
> 
>   - **New Features**:
>     - Add `shimmer-bg` utility for skeleton loaders and background shimmer effects.
>     - Introduce sine-eased gradients for smooth shimmer highlights.
>     - Add `shimmer-x-*` and `shimmer-y-*` for position synchronization across elements.
>     - Add `shimmer-angle-*` for diagonal shimmer sweeps.
>     - Implement `shimmer-container` for auto-width, auto-speed, and auto-spread.
>   - **Defaults**:
>     - Background shimmer defaults to 800px width and 1000px/s speed.
>     - Text shimmer defaults to 200px width and 150px/s speed.
>   - **Documentation**:
>     - Update `README.md` with new features and usage examples.
>     - Add comprehensive testing and examples in `page.tsx` and `spread-test/page.tsx`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=assistant-ui%2Fassistant-ui&utm_source=github&utm_medium=referral)<sup> for 6de30869394ac42e031faa9d8b093530a438d355. You can [customize](https://app.ellipsis.dev/assistant-ui/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/apps/docs/app/tw-shimmer/page.tsx
+++ b/apps/docs/app/tw-shimmer/page.tsx
@@ -617,11 +617,11 @@ function CodeBlock({
         transformerMetaWordHighlight(),
       ]}
       components={{
-        // biome-ignore lint/correctness/noNestedComponentDefinitions: <explanation>
+        // biome-ignore lint/correctness/noNestedComponentDefinitions: false positive
         Pre: ({ className, ...props }: any) => (
           <pre className={className} {...props} />
         ),
-        // biome-ignore lint/correctness/noNestedComponentDefinitions: <explanation>
+        // biome-ignore lint/correctness/noNestedComponentDefinitions: false positive
         Code: ({ className, ...props }: any) => (
           <code className={className} {...props} />
         ),
--- a/apps/docs/app/tw-shimmer/spread-test/page.tsx
+++ b/apps/docs/app/tw-shimmer/spread-test/page.tsx
@@ -19,7 +19,7 @@ export default function ShimmerSandboxPage() {
       <div className="mx-auto max-w-7xl px-4 py-12">
         {/* Header - full width */}
         <header className="mb-12 space-y-4">
-          <h1 className="text-4xl font-bold">Shimmer Sandbox</h1>
+          <h1 className="font-bold text-4xl">Shimmer Sandbox</h1>
           <p className="max-w-2xl text-muted-foreground">
             Comprehensive testing playground for shimmer text and shimmer-bg
             utilities. Use this to tune values, test edge cases, and experiment
@@ -71,7 +71,7 @@ function TableOfContents() {
 
   return (
     <nav className="sticky top-24">
-      <h2 className="mb-3 text-xs font-semibold tracking-wide text-muted-foreground uppercase">
+      <h2 className="mb-3 font-semibold text-muted-foreground text-xs uppercase tracking-wide">
         Sections
       </h2>
       <ul className="space-y-1">
@@ -79,7 +79,7 @@ function TableOfContents() {
           <li key={id}>
             <a
               href={`#${id}`}
-              className="block rounded-md px-3 py-1.5 text-sm text-muted-foreground transition-colors hover:bg-muted hover:text-foreground"
+              className="block rounded-md px-3 py-1.5 text-muted-foreground text-sm transition-colors hover:bg-muted hover:text-foreground"
             >
               {label}
             </a>
@@ -104,7 +104,7 @@ function Section({
   return (
     <section id={id} className="scroll-mt-8 space-y-6">
       <div className="space-y-2">
-        <h2 className="text-2xl font-bold">{title}</h2>
+        <h2 className="font-bold text-2xl">{title}</h2>
         {description && <p className="text-muted-foreground">{description}</p>}
       </div>
       {children}
@@ -126,7 +126,7 @@ function TestCard({
       <div>
         <h3 className="font-semibold">{title}</h3>
         {description && (
-          <p className="text-sm text-muted-foreground">{description}</p>
+          <p className="text-muted-foreground text-sm">{description}</p>
         )}
       </div>
       {children}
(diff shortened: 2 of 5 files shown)


packages/tw-shimmer/src/index.css
```diff
@@ -51,3 +86,86 @@
...
+    );
+  }
+
+  @variant dark {
+    --shimmer-bg-base: oklch(0.25 0 0);
+    --shimmer-bg-highlight: oklch(0.45 0 0);
+  }
+}
+
+@utility shimmer-bg-angle-* {
```
Yonom: why new css variables instead of reusing existing ones? e.g. choose shimmer-text vs shimmer-bg + any of the shimmer-x utilities

use same terminology
- width should be spread
- travel  should be `--shimmer-width-x` with no utility
- base should just be the bg color on the element?
- heighlight should be color

packages/tw-shimmer/src/index.css
```diff
@@ -15,11 +24,37 @@
   --shimmer-color: black;
   --shimmer-spread: 6ch;
   --shimmer-angle: 90deg;
+  /* Eased gradient: sine ease-in-out for smooth falloff at both center and edges */
+  --shimmer-highlight: color-mix(
```
Yonom: highlight should be same as shimmer color? unsure why it should be 90% of the user defined color
make it default to 90% of current color & black if you want, sure

packages/tw-shimmer/src/index.css
petekp: Great catch, can't believe I missed this 🤦‍♂️

Yonom:
@petekp CSS `anchor()` has decent support (only firefox is missing)

Screenshot 2025-11-26 at 5 38 49 PM

I think its good enough to ship an anchor implementation for --shimmer-bg-x and --shimmer-bg-y (side note: can we rename to --shimmer-offset-x --shimmer-offset-y?)

```
.shimmer-container {
  anchor-name: --shimmer-anchor;
}

.shimmer-bg {
  --shimmer-offset-x: anchor(--shimmer-anchor left);
  --shimmer-offset-y: anchor(--shimmer-anchor top);
}
```

packages/tw-shimmer/src/index.css
```diff
@@ -7,19 +7,54 @@
...
+    100% {
+      translate: calc(var(--shimmer-width, 600) * 1px) 0;
+    }
+  }
 }
 
 @utility shimmer {
   --shimmer-speed: 100;
-  --shimmer-width-x: 200;
+  --shimmer-width: 200;
```
petekp: Since width always applies to the `x` dimension, having a `-x` suffix isn't necessary.

apps/docs/app/tw-shimmer/page.tsx
```diff
@@ -29,10 +30,51 @@ const HIGHLIGHT_STYLES = `
     background: rgba(147, 197, 253, 0.3);
     color: rgb(165, 180, 252);
   }
+  .shimmer-metallic {
```
Yonom: it's not a good look if we aren't relying on the shimmer utility for the shimmer docs site! 😆

packages/tw-shimmer/src/index.css
```diff
@@ -51,3 +92,79 @@
...
+
+@utility -shimmer-y-* {
+  --shimmer-y: calc(-1 * --value(integer));
+}
+
+@utility shimmer-x-* {
+  --shimmer-x: --value(number);
+}
+
+@utility -shimmer-x-* {
```
Yonom: not needed I believe?

packages/tw-shimmer/src/index.css
```diff
@@ -51,3 +92,79 @@
...
+
+@utility shimmer-width-* {
+  --shimmer-width: --value(number);
+}
+
+@utility shimmer-y-* {
+  --shimmer-y: --value(number);
+}
+
+@utility -shimmer-y-* {
```
Yonom: not needed I believe?

packages/tw-shimmer/src/index.css
```diff
@@ -51,3 +92,79 @@
...
+        var(--shimmer-speed, 500) * -1s
+    );
+  }
+}
+
+@utility shimmer-width-* {
+  --shimmer-width: --value(number);
+}
+
+@utility shimmer-y-* {
```
Yonom: See my comment about the anchor() implementation that would remove the need to.manually set the x/y -- then most users would only need to set the shimmer-container class and not think about x/y

IMO the utilities for x/y can be skipped (instead rely on explicit CSS var setting if the user wants to set it manually)

packages/tw-shimmer/src/index.css
```diff
@@ -51,3 +92,79 @@
...
+      (
+          var(--shimmer-y, 0) * tan(calc(90deg - var(--shimmer-angle, 90deg))) -
+            var(--shimmer-x, 0)
+        ) /
+        var(--shimmer-speed, 500) * -1s
+    );
+  }
+}
+
+@utility shimmer-width-* {
```
Yonom: if we tell users to wrap all bg shimmers, we can calculate both x/y location as well as the width. 

for the width, we can use container queries + cqw units

.shimmer-container {
  container-name: shimmer;
  container-type: inline-size;
}

// only if this component is wrapping a shimmer container, grab the shimmer width from the container query (no manual setting needed in this case!)
   @container shimmer (min-width: 0px) {
     .shimmer {
       --shimmer-width: 100cqw;
     }
   }

Yonom (changes requested):

packages/tw-shimmer/src/index.css
petekp: @Yonom I added `shimmer-container` to provide auto-width handling (also auto-spread and auto-speed for `shimmer-bg`!)

I attempted what you suggested with `anchor()` but I don't believe it's possible to get `x/y` using `anchor()`, unfortunately. `anchor()` doesn’t expose element offsets for in-flow children, and it’s only valid inside inset properties on absolutely/fixed-positioned elements.

Feel free to give it a shot!

packages/tw-shimmer/src/index.css
petekp: yeah not really

packages/tw-shimmer/src/index.css
```diff
@@ -1,38 +1,233 @@
 @theme inline {
-  @keyframes shimmer {
+  @keyframes tw-shimmer-text {
```
Yonom: conventionL should keyframes use "tw-" prefix? I dont think so?

packages/tw-shimmer/src/index.css
```diff
@@ -41,13 +236,82 @@
 }
 
 @utility shimmer-color-* {
-  --shimmer-color: --value(--color-*);
+  --shimmer-color: --value([color], --color-*);
```
Yonom: ooo what's this?

packages/tw-shimmer/src/index.css
```diff
@@ -41,13 +236,82 @@
...
+
+/*
+  Container-based auto-width for shimmer animations.
+  Uses CSS container queries to automatically set --shimmer-width
+  based on the container's width, without needing JS.
+
+  In older browsers, the shimmer-container class has no effect and 
+  shimmers fall back to their default widths or manually set values.
+*/
+@supports (width: 1cqw) {
```
Yonom: convert to utility so we dont polute if users dont use shimmer container

PR #2850, opened 2025-11-29
feat: expose `hasAttachments` and `isEmpty` state to Composer

okisdev:
This PR exposing `hasAttachments` and `isEmpty` state to Composer

----

> [!IMPORTANT]
> Expose `hasAttachments` and `isEmpty` states in `ComposerIf` to conditionally render children based on attachments and emptiness.
> 
>   - **Behavior**:
>     - Exposes `hasAttachments` and `isEmpty` states in `ComposerIf` to conditionally render children based on attachments and emptiness.
>     - Updates `useComposerIf` in `ComposerIf.tsx` to handle `hasAttachments` and `isEmpty` conditions.
>   - **Documentation**:
>     - Updates `Composer.mdx` to include `hasAttachments` and `isEmpty` in the API reference and examples.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=assistant-ui%2Fassistant-ui&utm_source=github&utm_medium=referral)<sup> for f44ab2d77e590d51111ede8e9cd427350b2bcb68. You can [customize](https://app.ellipsis.dev/assistant-ui/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/apps/docs/content/docs/api-reference/primitives/Composer.mdx
+++ b/apps/docs/content/docs/api-reference/primitives/Composer.mdx
@@ -227,9 +227,23 @@ Renders children if a condition is met.
       type: "boolean | undefined",
       description: "Render children if the message is being edited.",
     },
+    {
+      name: "hasAttachments",
+      required: false,
+      type: "boolean | undefined",
+      description: "Render children if the composer has attachments.",
+    },
+    {
+      name: "isEmpty",
+      required: false,
+      type: "boolean | undefined",
+      description: "Render children if the composer is empty (no text and no attachments).",
+    },
   ]}
 />
 
 ```tsx
 <Composer.If editing>{/* rendered if message is being edited */}</Composer.If>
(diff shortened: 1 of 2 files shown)


Yonom:
I've been thinking about replacing all the if components with a more generic:

```ts
<AssistantIf query={({ composer }) => composer.isEmpty}>
```

what do you think of this? (it would use `useAssistantState` under the hood)
this way we don't have to add all sorts of small checks to the Thread/Composer.If primitives

okisdev:
@Yonom 

That's an interesting refactor on `Composer.If`. Makes the simple things simple, and the complex things possible.

It becomes clear, an intuitively way to understand what it is doing through the `query`, so I strongly agree.

We also don't need to manually manage these things expose and they are entirely controlled by the users, which is a great practice.

Yonom:
replaced by the new AssistantIf API! thanks Harry for the idea! :)

PR #2874, opened 2025-12-04
fix(react): scope ESC keydown handler to input element

markusrubey:
## Summary

The `useEscapeKeydown` hook in `ComposerPrimitive.Input` registers a global document listener in the capture phase, which intercepts ESC key events before other UI elements can handle them.

This causes issues when:
- A Radix Dialog is open alongside a page containing a `ComposerPrimitive.Input`
- User presses ESC to close the dialog
- The composer's ESC handler fires first (capture phase) and calls `preventDefault()`
- The dialog never receives the ESC event and doesn't close

## Fix

Added a check to ensure the ESC handler only triggers when the event originated from within the composer input itself:

```tsx
useEscapeKeydown((e) => {
  if (!cancelOnEscape) return;

  // Only handle ESC if it originated from within this input
  if (!textareaRef.current?.contains(e.target as Node)) return;

  const composer = api.composer();
  if (composer.getState().canCancel) {
    composer.cancel();
    e.preventDefault();
  }
});
```

----

> [!IMPORTANT]
> Fix `useEscapeKeydown` in `ComposerPrimitive.Input` to only trigger ESC handler if event originates from input element, preventing interference with other UI elements.
> 
>   - **Behavior**:
>     - Fix `useEscapeKeydown` in `ComposerPrimitive.Input` to only trigger ESC handler if event originates from input element.
>     - Prevents ESC handler from interfering with other UI elements like Radix dialogs.
>   - **Code**:
>     - Add check in `useEscapeKeydown` to verify event target is within `textareaRef` before calling `preventDefault()`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=assistant-ui%2Fassistant-ui&utm_source=github&utm_medium=referral)<sup> for cd2e42795c366e7580620f6d5e4c5afa8b57b0ea. You can [customize](https://app.ellipsis.dev/assistant-ui/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/.changeset/scope-escape-keydown.md
+++ b/.changeset/scope-escape-keydown.md
@@ -1,0 +1,7 @@
+---
+"@assistant-ui/react": patch
+---
+
+Fix ESC keydown handler to only trigger when event originates from the composer input
+
+The `useEscapeKeydown` hook was intercepting ESC key events globally, preventing other UI elements (like Radix dialogs) from responding to ESC. The handler now checks if the event target is within the composer input before calling `preventDefault()`.


Yonom:
Thank you!

PR #2908, opened 2025-12-09
feat(useIsMobile): Add SSR check and optimize with mql.matches

hhhaan:
##  Improvements to `useIsMobile` Hook

This PR enhances the hook's stability in different rendering environments and optimizes its performance.

### 1. Prevent SSR Errors (Stability)

* Added the condition `if (typeof window === 'undefined') { return; }` inside `useEffect`.
* **Fixes:** Prevents the `window is not defined` runtime error that occurs when the hook module is processed during Server-Side Rendering (SSR) in frameworks like Next.js.

### 2. Performance Optimization (Efficiency)

* Switched the change handler logic from re-calculating `window.innerWidth < MOBILE_BREAKPOINT` to directly using the **`mql.matches`** property of the `MediaQueryList` object.
* **Benefits:** This avoids redundant recalculations of the viewport width, leveraging the built-in efficiency of the `matchMedia` API and minimizing unnecessary DOM/Window API calls.

----

> [!IMPORTANT]
> Enhances `useIsMobile` hook with SSR safety and performance optimization using `mql.matches`.
> 
>   - **Stability**:
>     - Add `if (typeof window === 'undefined') { return; }` in `useEffect` of `useIsMobile` to prevent SSR errors.
>   - **Performance**:
>     - Use `mql.matches` instead of `window.innerWidth < MOBILE_BREAKPOINT` in `useIsMobile` for efficiency.
>     - Set initial state with `mql.matches` to avoid redundant calculations.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=assistant-ui%2Fassistant-ui&utm_source=github&utm_medium=referral)<sup> for 70f55d0559b0b5e72c618d6156c207919fa39690. You can [customize](https://app.ellipsis.dev/assistant-ui/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


Yonom:
Thank you for opening a PR.

useEffect does not run on SSR, so I am confused about this fix. 

do you have a minimal reproducible version of the error you are seeing to test this behavior? 

@hhhaan

hhhaan:
I was mistaken. I'll keep only the performance improvement.

PR #2955, opened 2025-12-19
fix(react): prevent viewport slack feedback loop 

### LINKED ISSUE #2954: System message rendered with MessagePrimitive.Root can freeze the UI (main-thread lockup)
Hi team — I’m hitting a hard UI freeze when rendering a system role message with MessagePrimitive.Root.
I’ll attach a small repro project.

### Environment

- @assistant-ui/react: 0.11.51 and 0.11.52 (both repro)
- React: 19.2.3
- Vite: 5.4.x
- OS/Browser: Chrome (also repros in other Chromium browsers)

### Repro Steps

1. Use useExternalStoreRuntime with two messages:
    - a long user text message
    - a simple system message (e.g., “Switched to remote”)
2. Render both using ThreadPrimitive.Messages
3. Render the system message with MessagePrimitive.Root
4. Scroll down to the system message

### Actual Result

The UI becomes unresponsive:

- Inputs and buttons stop responding
- Browser back button stops responding
- Pausing script execution in DevTools immediately restores responsiveness

### Expected Result

The system message should render normally and the UI should stay responsive.

### Suspected Root Cause (Hypothesis)

MessagePrimitive.Root always wraps content in ThreadPrimitiveViewportSlack, which subscribes to the
viewport store and mutates inline styles (min-height/flex). The viewport itself uses a MutationObserver
(via useOnResizeContent) to react to DOM mutations. With a system message as the last message, the slack
component is attached to a “non-turn” message (not user/assistant). This appears to create a feedback
loop:

1. Slack updates inline styles on the system message.
2. MutationObserver fires due to style changes.
3. Viewport auto-scroll bookkeeping runs again.
4. Slack recomputes and writes styles again.

This loop pegs the main thread (hence the hard freeze). User/assistant messages don’t trigger this
because they align with the turn/anchor logic, so the layout stabilizes.

### Workaround

Render system messages without MessagePrimitive.Root (plain DOM rendering). This eliminates the freeze
entirely.

### Demo Project

I’ll attach a minimal repro project (Vite React TS) with the exact setup.

[assistant-ui-freeze-repro.zip](https://github.com/user-attachments/files/24258400/assistant-ui-freeze-repro.zip)

tiann:
fixes #2954

> [!IMPORTANT]
> `MessagePrimitiveRoot` in `MessageRoot.tsx` now skips wrapping system messages with `ThreadPrimitiveViewportSlack`.
> 
>   - **Behavior**:
>     - In `MessageRoot.tsx`, `MessagePrimitiveRoot` now skips wrapping system messages with `ThreadPrimitiveViewportSlack`.
>     - System messages are identified using `role` from `useAssistantState`.
>   - **Code Structure**:
>     - Introduces a conditional check for `role === "system"` to determine if wrapping should be skipped.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=assistant-ui%2Fassistant-ui&utm_source=github&utm_medium=referral)<sup> for 8bdb1edfc21d72e434333b63355fd32b94d3c0a3. You can [customize](https://app.ellipsis.dev/assistant-ui/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/.changeset/fuzzy-hotels-tie.md
+++ b/.changeset/fuzzy-hotels-tie.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react": patch
+---
+
+fix(react): prevent viewport slack feedback loop 


okisdev:
hey @tiann, thanks for the fix.

ideas on root fix on top of ur solution:
- tighten slack condition - only apply to last assistant message with valid turn
- filter style mutations in `useOnResizeContent` - prevent the feedback loop at source

what do u think?

PR #2994, opened 2026-01-06
[WIP] feat(assistant-stream): resumable

okisdev:
This PR adds support for resumable streaming.


okisdev:
closing in favor of #3979 which supersedes this WIP.

the new PR drops the hard `redis` dep and reshapes the surface so the same primitive works for in-memory dev, node-redis, ioredis, or any custom backend that satisfies a 6-method `ResumableStreamStore` interface. persistence stays byte-level (post-encoder) so any existing encoder works without modification, but the API now mirrors `vercel/resumable-stream` for migration and adds `createResumableAssistantStreamResponse` / `createResumeAssistantStreamResponse` helpers that bridge `AssistantStreamController` callbacks to the store. cluster-safe Redis keys (`{streamId}` hash tag), TTL persisted in meta so `acquire` / `append` / `finalize` stay consistent, validated streamId pattern, and 215 vitest cases including a live Redis suite gated by `REDIS_TESTS`.

client integration stays out of `packages/react` and `packages/react-ai-sdk` for this round; the example wires `useChat` + `prepareReconnectToStreamRequest` directly so adopters can copy the ~80-line transport shim without waiting on a runtime change. a follow-up PR can promote that into `react-ai-sdk` once the surface settles.

docs added under `docs/guides/resumable-streams` with a production checklist (auth, `waitUntil` on serverless, TTL).

PR #3004, opened 2026-01-09
fix(react): prevent crash on new thread creation

ShobhitPatra:
This PR fixes a  race condition in the useRemoteThreadListRuntime hook that occurs when a user is implementing a  thread list.

 ### Problem:

  When the runtime is initialized, it immediately creates a new thread for the user to interact with. However, the 
  switchToThread function did not wait for the new thread's runtime to be fully created before notifying the UI to 
  re-render. This resulted in the UI attempting to render an input field for a thread that wasn't ready, leading to a 
  crash with the error: `Uncaught Error: This is the empty thread, a placeholder for the main thread....`

Reference - https://canary.discord.com/channels/1251324227668283443/1268446361192497164/threads/1442287267602501783

----

> [!IMPORTANT]
> Fixes race condition in `switchToThread()` in `RemoteThreadListThreadListRuntimeCore.tsx` to prevent UI crash on new thread creation.
> 
>   - **Behavior**:
>     - Fixes race condition in `switchToThread()` in `RemoteThreadListThreadListRuntimeCore.tsx` by awaiting `task` to ensure thread runtime is fully initialized before UI re-renders.
>     - Prevents crash when rendering input field for uninitialized thread, resolving error: `Uncaught Error: This is the empty thread, a placeholder for the main thread...`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=assistant-ui%2Fassistant-ui&utm_source=github&utm_medium=referral)<sup> for 3f621792b2aa9eae5ac16a63cbee753652bca84f. You can [customize](https://app.ellipsis.dev/assistant-ui/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/.changeset/purple-bottles-do.md
+++ b/.changeset/purple-bottles-do.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react": minor
+---
+
+fix(react): prevent crash on new thread creation
--- a/packages/react/src/legacy-runtime/runtime-cores/remote-thread-list/RemoteThreadListThreadListRuntimeCore.tsx
+++ b/packages/react/src/legacy-runtime/runtime-cores/remote-thread-list/RemoteThreadListThreadListRuntimeCore.tsx
@@ -15,6 +15,7 @@ import { create } from "zustand";
 import { AssistantMessageStream } from "assistant-stream";
 import { ModelContextProvider } from "../../../model-context";
 import { RuntimeAdapterProvider } from "../adapters/RuntimeAdapterProvider";
+import { ThreadRuntimeCore } from "../core/ThreadRuntimeCore";
 
 type RemoteThreadData =
   | {
@@ -147,6 +148,7 @@ export class RemoteThreadListThreadListRuntimeCore
   private _loadThreadsPromise: Promise<void> | undefined;
 
   private _mainThreadId!: string;
+  private _pendingComposerText: string | undefined;
   private readonly _state = new OptimisticState<RemoteThreadState>({
     isLoading: false,
     newThreadId: undefined,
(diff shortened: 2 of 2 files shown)


ShobhitPatra:
@claude review

okisdev:
hey @ShobhitPatra, thanks for digging into this and putting up a fix. unfortunately i think this PR has been overtaken by later work, so i'd like to close it. wanted to give you the full picture first.

a few reasons:

1. the file you're patching has moved. PR #3325 extracted the remote thread list into `@assistant-ui/core`, so `packages/react/src/legacy-runtime/runtime-cores/remote-thread-list/RemoteThreadListThreadListRuntimeCore.tsx` is now a 4-line re-export. the real implementation lives at `packages/core/src/react/runtimes/RemoteThreadListThreadListRuntimeCore.tsx`. that's also why GitHub reports the PR as conflicting.

2. the root cause has been addressed. the symptom in the linked Discord thread and in issue #3678 came from `unstable_Provider` deferring `children` (e.g. behind a loading state or Suspense gate); in that case the runtime binder never ran and `getMainThreadRuntimeCore()` fell back to `EMPTY_THREAD_CORE`. PR #3891 fixed that. #3948 (shipped in `@assistant-ui/core@0.2.0` / `@assistant-ui/react@0.14.0`) then settled on the final shape: the binder stays inside the user's Provider so cloud persistence and the history adapter keep working, and a dev-mode warning fires within ~100ms if the Provider withholds children. the synchronous-children contract is now documented in `apps/docs/content/docs/runtimes/concepts/threads.mdx`, and the regression is locked in by `packages/react/src/tests/RemoteThreadListRuntime.deferredProvider.test.tsx`.

3. the buffering approach is composer-only. `_pendingComposerText` papers over `composer.setText`, but every other method on `EMPTY_THREAD_CORE` (`startRun`, `append`, `cancelRun`, attachments, etc.) still throws, so the same race on a different code path would still crash. the architectural fix in #3891 / #3948 covers all of those.

4. minor process note for next time: the changeset bumps `@assistant-ui/react` as `minor`, but per `AGENTS.md` everything lands as `patch` unless a maintainer signs off on a wider bump.

going to close this. if anyone is still hitting the error on a recent version, please open a fresh issue with a minimal repro of the `unstable_Provider`. nine times out of ten it's the synchronous-children rule and the dev warning should now make that obvious; if not, we'll want to see the exact Provider shape.

thanks again for the time you put into this!

PR #3010, opened 2026-01-11
fix(react-langgraph): correctly return externalId on thread creation

ShobhitPatra:
This PR fixes a `Thread not found` error  by correcting how th `externalId` is handled during thread creation in nested runtimes.
  
  ### The Bug
  
  The `create` function in `useLangGraphRuntime` was returning an object with an incorrect shape, causing the  `externalId` to be lost.
  
  ### The Fix
 
 In `useLangGraphRuntime`: The `create` function now correctly destructures the `externalId` from the result of api.threadListItem().initialize()` and returns it in the expected format.
 
In `cloud.tsx`: The `initialize` method now explicitly checks if adapter.create exists before calling it.
Reference - https://canary.discord.com/channels/1251324227668283443/1268446361192497164/threads/1447348641218957474

----

> [!IMPORTANT]
> Fix `Thread not found` error in `useLangGraphRuntime` by correctly handling `externalId` during thread creation.
> 
>   - **Bug Fix**:
>     - Fix `Thread not found` error in `useLangGraphRuntime` by correctly handling `externalId` during thread creation.
>   - **Implementation**:
>     - In `useLangGraphRuntime`, update `create` function to destructure `externalId` from `api.threadListItem().initialize()` and return it in the correct format.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=assistant-ui%2Fassistant-ui&utm_source=github&utm_medium=referral)<sup> for f87fa3ecb673cfb31b2845463cc9dcdde7476e40. You can [customize](https://app.ellipsis.dev/assistant-ui/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/.changeset/soft-roses-turn.md
+++ b/.changeset/soft-roses-turn.md
@@ -1,5 +1,9 @@
 ---
 "@assistant-ui/react-langgraph": minor
+"@assistant-ui/react": patch
 ---
 
-fix(react-langgraph): correctly return externalId on thread creation
+fix: resolve "Thread not found" error in thread creation
+
+- fix(react-langgraph): correctly return externalId shape from useLangGraphRuntime create function
+- fix(react): properly handle missing create function in cloud adapter initialize method
--- a/.changeset/soft-roses-turn.md
+++ b/.changeset/soft-roses-turn.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react-langgraph": minor
+---
+
+fix(react-langgraph): correctly return externalId on thread creation
--- a/packages/react-langgraph/src/useLangGraphRuntime.ts
+++ b/packages/react-langgraph/src/useLangGraphRuntime.ts
@@ -346,7 +346,8 @@ export const useLangGraphRuntime = ({
       }
 
       if (api.threadListItem.source) {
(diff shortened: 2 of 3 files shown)


itzhoujun:
This issue blocks development

okisdev:
thanks for the patch, @ShobhitPatra.

we dug into this and the real root cause turned out to be the `if (!cloud)` branch in `useCloudThreadListAdapter` returning a bare `InMemoryThreadListAdapter`, whose default `initialize` returned `{ remoteId, externalId: undefined }` without ever invoking the user-supplied `create` callback. that fallback was replaced in #3349 (`feat(react-langgraph): messages tuple`, merged 2026-02-18); the `inMemory.initialize` override added there is what actually unblocks the "Thread not found" scenario you and kayleigh reported on Discord.

on current main the failing path no longer reproduces. we just opened #4045 which adds a focused regression test in `useLangGraphRuntime.test.tsx` (mounts `useLangGraphRuntime` with a `create` callback but no `cloud`, sends a message, asserts the callback fires and `externalId` propagates) so this cannot silently regress again.

closing this one out since the underlying issue is already resolved. thanks again for surfacing it.

PR #3015, opened 2026-01-12
feat: add tool-ui-server package and with-openai-apps example

samdickson22:
## Summary

- Add new `@assistant-ui/tool-ui-server` package for rendering rich tool UIs in sandboxed iframes using the OpenAI Apps SDK protocol
- Add comprehensive `with-openai-apps` example demonstrating MCP servers with rich tool UI rendering
- Fix critical bug where HTML templates were being embedded in tool results (causing context_length_exceeded errors)
- Include working Node.js and Python server implementations

### New Package: `@assistant-ui/tool-ui-server`

A new package that provides:
- `RemoteToolUI` component for rendering tool UIs in sandboxed iframes
- Message bridge for parent-child iframe communication
- OAuth utilities for authenticated MCP servers
- Full OpenAI Apps SDK protocol compatibility

### Example: `with-openai-apps`

This example shows how to:

1. **Run multiple MCP servers** with a single chat client
2. **Render rich tool UIs** in sandboxed iframes using the OpenAI Apps SDK protocol
3. **Properly handle template URIs** without bloating LLM context
4. **Namespace tools** using `serverId__toolName` format for OpenAI compatibility

### Servers Included

| Server | Port | Language | Features |
|--------|------|----------|----------|
| Kitchen Sink (Node) | 8000 | TypeScript | All SDK features demo |
| Pizzaz (Node) | 8001 | TypeScript | Maps, carousels, lists |
| Pizzaz (Python) | 8002 | Python | Same features in Python |
| Kitchen Sink (Python) | 8003 | Python | SDK features in Python |
| Solar System (Python) | 8004 | Python | 3D Three.js visualization |
| Shopping Cart (Python) | 8006 | Python | Stateful cart management |

### Key Bug Fix

The original implementation fetched HTML via `readResource` and embedded it in tool results:
```typescript
// BEFORE (bad): 500KB+ HTML sent to LLM
result._template = { html: "<html>..." }
```

Now we only store the template URI and load HTML client-side:
```typescript
// AFTER (good): Only URI in tool result
result._templateUri = "ui://widget/pizza-map.html"
// Frontend loads HTML from assets server
```

## How to Test

```bash
cd examples/with-openai-apps
pnpm install
pnpm setup:python  # For Python servers
pnpm build:assets
pnpm dev           # Node servers only
pnpm dev:all       # All servers
```

Open http://localhost:3005 and try:
- "Show me pizza places"
- "Launch the kitchen sink demo"
- "Show the solar system"

## Architecture

```
Tool Execution Flow:
1. LLM calls tool → MCP server returns result with _meta["openai/outputTemplate"]
2. mcp-client-native.ts extracts templateUri (NOT the HTML!)
3. Result goes to LLM (small JSON)
4. Frontend sees _templateUri → converts to assets server URL
5. RemoteToolUI loads HTML from localhost:4444
6. Assets server injects bridge script → widget renders with window.openai protocol
```

## Tool Namespacing

When multiple servers are enabled, tools are namespaced using `serverId__toolName` to prevent collisions (e.g., `pizzaz-node__pizza-map`). This format is required for OpenAI compatibility (function names must match `^[a-zA-Z0-9_-]{1,64}$`).

## Checklist

- [x] New package builds and tests pass
- [x] Example runs with `pnpm dev`
- [x] Tool UIs render correctly in iframes
- [x] No HTML sent to LLM (prevents context overflow)
- [x] Python servers work with `pnpm setup:python`
- [x] Documentation updated in README


petekp (dismissed):

samdickson22:
Superseded by the cleaner approach in #3192 (Tool UI Runtime). Will close this once that lands. The server SDK portion (`@assistant-ui/tool-ui-server`) will be a separate, focused PR based on main.

okisdev:
first off, thank you for this. it's a lot of genuinely careful work (6 servers across node and python, the widget bundles, the oauth flow), and it surfaced a real need: rendering rich, server provided tool UIs inside the chat. i want to be transparent about why we're going to close it instead of merging, and point you at where the same goal now lives.

## why we're not adding the package

shipping a new published package (`@assistant-ui/tool-ui-server`) is a long term commitment to everyone who installs it: semver, security patches, docs, and a public API we have to keep stable for years. we treat adding a package as a deliberate decision, not a default. the bar is roughly "a durable capability that our existing primitives genuinely cannot express." this PR doesn't clear that bar, because the capability already exists in the library without a new package.

## the capability is already native

- rendering an MCP server's own `ui://` HTML widget in a sandboxed iframe, with a postMessage bridge (call tools, set widget state, request display mode, send follow up messages, and so on) is exactly what `McpAppRenderer` + `McpAppsRemoteHost` do today (`packages/react/src/mcp-apps/`, exported from `@assistant-ui/react`). you compose it with `Tools({ mcpApp: McpAppRenderer({ host: McpAppsRemoteHost({ url }) }) })`.
- attaching custom UI to a specific tool by name is `makeAssistantToolUI`, and a per name registration overrides the MCP fallback.
- authoring the widgets themselves (the html bundles, `useToolInput`, `useCallTool`, theming, a local workbench, ChatGPT compatibility) is `@assistant-ui/mcp-app-studio`.

between those three, the `RemoteToolUI` component, the message bridge, the widget hooks, the registry client, and the copied `button.tsx` / `cn` in this PR are all things we already maintain. a parallel package would mean two copies of the same surface, which is the cost we're trying to avoid.

## to be fair on timing

when you opened this on jan 12, none of that existed yet. `mcp-app-studio` landed jan 28 (#3132), and the host side `McpAppRenderer` only merged in #4024. you were early to a real problem. the project then built the native version aligned to the official MCP Apps spec (the open standard Anthropic shipped in late january, which ChatGPT, Claude, Goose, and VS Code have all adopted). so this isn't "you reinvented something that existed," it's "the standard landed while this sat, and it landed in a shape we have to converge on."

## the one piece that is genuinely new, and where it belongs

the only thing here the native path doesn't already cover is supporting widgets written for OpenAI's specific convention: the `ui://` template is named by `_meta["openai/outputTemplate"]` on the tool definition (the standard uses `_meta.ui.resourceUri`) and served as `text/html+skybridge`. `@ai-sdk/mcp` doesn't recognize that key, so the pointer never reaches the renderer.

the correct fix for that is about three lines, upstream, not a package: teach `@ai-sdk/mcp`'s `getMCPAppToolMeta` to also read `openai/outputTemplate`. after that, the existing `McpAppRenderer` renders OpenAI authored widgets with zero changes on our side. for anyone who wants it today without waiting on that, we just documented a small route level bridge in the MCP Apps guide (#4084) that copies the pointer onto the result `_meta["ui/resourceUri"]` the renderer already reads.

## for full transparency, the diff also has blocking issues

independent of the architecture decision, a number of items here would be yours to resolve before this could land in any form:

- CI is red. the changeset is `minor`, which trips the semver breaking check (0.0.1 to 0.1.0 is outside the caret range). our policy is patch by default; anything larger needs explicit maintainer sign off.
- merge conflicts in `.gitignore`, `biome.json`, and `pnpm-lock.yaml`. the branch is about four months behind main and needs a rebase.
- `src/remote/message-bridge.ts` assigns `this.expectedOrigin` (lines 67, 70, 79) to a field that is never declared and never read. that's a TypeScript error under a strict config, and the origin check it implies isn't enforced anywhere.
- the build diverges from every other published package: it uses `tsup` and a hand rolled `tsconfig.json` instead of `aui-build` and extending `@assistant-ui/x-buildutils/ts/base`.
- dependency hygiene: `zod` is pinned to v3 in a v4 workspace; `react-dom` is imported (`createPortal`) but missing from `peerDependencies`; `@modelcontextprotocol/sdk` is an optional peer that isn't externalized in the build, so it gets bundled; and 8 of the 13 `dependencies` (`shiki`, `lucide-react`, `ansi-to-react`, `zod-to-json-schema`, and four `@radix-ui/*` packages) are never imported.
- tests: the two headline exports, `createToolUIServer` and `createToolUIRuntime`, have no tests; `__tests__/protocol-types.test.ts` asserts literals against themselves (it would pass with the types deleted); and several files are copy paste duplicates. note that no build, typecheck, lint, or unit test job ran on the PR, so the passing checks don't actually tell you the package compiles.

## suggested path

we'll close this, but the need was real and we'd genuinely welcome the small upstream piece: a PR to `@ai-sdk/mcp` adding `openai/outputTemplate` recognition to `getMCPAppToolMeta`, plus a case in `mcp-apps.test.ts`. if you have widgets to share, `mcp-app-studio` is the home for them. thanks again for pushing on this; it's a real part of why the native support and the docs exist today.

PR #3099, opened 2026-01-22
fix(react): handle tool argsText key reordering during streaming

### LINKED ISSUE #2613: [BUG] Tool call argsText can only be appended, not updated
After upgrading `@assistant-ui/react-ai-sdk` from 1.1.3 to anything above (1.1.4, 1.1.5, 1.1.6), when the model is making a tool call, I get the following error

```
Tool call argsText can only be appended, not updated: {...} does not start with {}

Uncaught Error: Tool call argsText can only be appended, not updated: {...} does not start with {}
    at useToolInvocations.useEffect.processMessages (useToolInvocations.ts:164:25)
```

I am using gemini model with streamText and minimal setup 

```
const result = streamText({
      model: vertex('gemini-2.5-flash'),
      system: 'Some simple system prompt...',
      messages: convertToModelMessages(messages),
      tools,
    });

    return result.toUIMessageStreamResponse();
```

AVGVSTVS96:
When transitioning from streaming to complete state, some providers (like Gemini)
may return tool args with keys in a different order than they were streamed.
This caused a false-positive error because the new argsText didn't start with
the previous value.

The fix checks if both old and new argsText are complete JSON. If so, it's just
key reordering - close the stream and continue. This is safe because the parsed
args object is always correct.

Fixes #2613

----

> [!IMPORTANT]
> Fixes false-positive error in `useToolInvocations.ts` by handling JSON key reordering during streaming.
> 
>   - **Behavior**:
>     - Fixes false-positive error in `useToolInvocations.ts` when transitioning from streaming to complete state with reordered JSON keys.
>     - Checks if `argsText` is complete JSON for both old and new states to identify key reordering.
>     - Closes stream and continues if key reordering is detected.
>   - **Functions**:
>     - Adds `isArgsTextComplete()` to check JSON completeness in `useToolInvocations.ts`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=assistant-ui%2Fassistant-ui&utm_source=github&utm_medium=referral)<sup> for 8aa68a0e9dfe1c89fd918cab5e3149124405aa2b. You can [customize](https://app.ellipsis.dev/assistant-ui/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/.changeset/brown-pandas-juggle.md
+++ b/.changeset/brown-pandas-juggle.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react": patch
+---
+
+fix(react): handle tool argsText key reordering during streaming


urjitc:
I'm still experiencing this despite the fix. This is what claude code is saying, this may be true or not, I didin't find the root cause myself yet:
 
"**Issue:** The current fix only handles key reordering when both `lastState.argsText` and `content.argsText` are complete JSON. However, when tool arguments are large the previous argsText may still be streaming (incomplete) when a complete reordered version arrives. In this case, `isArgsTextComplete(lastState.argsText)` returns `false`, so the check fails and the error is thrown.

**Additional concerns** (as noted by reviewers):
1. The fix doesn't verify JSON equivalence - it only checks parseability, which could silently accept completely different arguments
2. The early return skips result/nested-message handling"

upsidejames:
Experiencing the same issue here w/ gemini models, and exactly as described here. ie. `lastState.argsText` is not fully formed 

> "... However, when tool arguments are large the previous argsText may still be streaming (incomplete) when a complete reordered version arrives. In this case, `isArgsTextComplete(lastState.argsText)` returns `false`, so the check fails and the error is thrown.

urjitc:
@AVGVSTVS96

Creabine:
## Incomplete fix: streaming (incomplete) argsText + complete reordered argsText still throws

Thanks for the key reordering fix! However, we are still hitting the `argsText can only be appended, not updated` error in `@assistant-ui/react@0.12.10` when:

1. The LLM streams tool call args with keys in one order (e.g. `instruction` before `category`)
2. Before the streaming argsText is complete JSON, the final argsText arrives with keys in a different order (`category` before `instruction`)

The current fix at L123-124 requires **both** `lastState.argsText` and `content.argsText` to be complete JSON:

```js
if (isArgsTextComplete(lastState.argsText) && isArgsTextComplete(content.argsText)) {
```

But in this scenario, `lastState.argsText` is still incomplete (streaming in progress), so `isArgsTextComplete(lastState.argsText)` returns `false`, and the code falls through to the `throw`.

### Reproduction

- Use Anthropic Claude with `streamText()` + tools that have multiple optional params
- Add a new optional parameter (e.g. `category`) to an existing tool schema
- The LLM may output keys in a different order than the final serialized JSON

### Workaround

We are using a pnpm patch that relaxes the condition to only check the new argsText:

```diff
- if (isArgsTextComplete(lastState.argsText) && isArgsTextComplete(content.argsText)) {
+ if (isArgsTextComplete(content.argsText)) {
```

This accepts the reordered args whenever the new value is valid JSON, regardless of whether the old value was complete.

### Environment

- `@assistant-ui/react@0.12.10`
- `@assistant-ui/react-ai-sdk@1.3.7`
- `ai@6.0.79` / `@ai-sdk/react@3.0.81`
- Anthropic Claude via `@ai-sdk/anthropic`

Issue #3127, opened 2026-01-26
React 18 compatibility broken - useEffectEvent import from React fails


isawalhi:
**Problem**

The package claims React 18 support in peerDependencies, but @assistant-ui/store directly imports useEffectEvent from React, which is only available in React 19.

// package.json peerDependencies
  "react": "^18 || ^19"

// node_modules/@assistant-ui/store/dist/useAuiEvent.js
  import { useEffect, useEffectEvent } from "react";

**Steps to Reproduce**

  1. Create a React 18 project
  2. Install @assistant-ui/react (latest version)
  3. Use any component that depends on @assistant-ui/store
  4. Application crashes with: useEffectEvent is not a function or similar error

**Expected Behavior**

The library should work with React 18 as stated in peerDependencies.

**Actual Behavior**

The application crashes because useEffectEvent is not exported from React 18.

**Environment**

- @assistant-ui/react: 0.12.1
- @assistant-ui/react-ai-sdk: 1.3.1
- React: 18.3.1
- Node: 20.x

**Possible Solutions**

  1. Use a polyfill like use-effect-event internally for React 18 compatibility
  2. Update peerDependencies to require React 19 only if React 18 support is no longer planned

**Additional Context**

  We attempted to polyfill useEffectEvent by patching the React object, but this doesn't work in module federation setups where React is a singleton shared across microfrontends.

Yonom:
Thanks! We're evaluating our timeline for React 18 support, is there any official end of life timestamp? 
Otherwise we might do something where we support the React versions that cover >90% of real world usage (currently 18+19)

dflynn15:
I ran into this same issue. While Assistant UI is evaluating when to drop React 18 support, thoughts on just using this approach for current consumers?

cc @Yonom @AVGVSTVS96 

https://github.com/assistant-ui/assistant-ui/pull/3130

PR #3130, opened 2026-01-26
chore: add useEffectEvent polyfill to preserve React support

dflynn15:
# What?

Fixes React 18 compatibility by replacing the direct `useEffectEvent` import from React (only available in React 19.2+) with the `use-effect-event` polyfill package.

# Why?
`@assistant-ui/store` lists support for React 18 but used `useEffectEvent`, which doesn't exist in React 18, causing runtime errors.

this provides an easily removable polyfill for consumers while Assistant UI can figure out when they want to holistically deprecate and remove React 18 for their support. 

Fixes: #3127

### DIFF
--- a/.changeset/fix-react-18-effect-event.md
+++ b/.changeset/fix-react-18-effect-event.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/store": patch
+---
+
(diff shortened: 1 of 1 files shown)


Yonom:
Thank you @dflynn15 !

Issue #3197, opened 2026-02-03
Duplicate key toolCallId error when calling addResult in HITL tools


chinkan:
# Duplicate toolCallId Error When Calling addResult in HITL Tools

## Summary

When using `makeAssistantToolUI` with `addResult` for Human-in-the-Loop (HITL) tools, calling `addResult()` throws a "Duplicate key toolCallId-XXX in tapResources" error. This prevents HITL tools from completing successfully.

## Environment

| Package | Version |
|---------|---------|
| `@assistant-ui/react` | 0.12.3 |
| `@assistant-ui/react-ai-sdk` | 1.3.3 |
| `@assistant-ui/react-markdown` | 0.12.1 |
| `@assistant-ui/store` | 0.1.2 (with PR #3177 patch applied) |
| `@assistant-ui/tap` | 0.4.2 |
| `ai` (Vercel AI SDK) | 6.0.67 |
| `react` | 19.2.4 |
| `next` | 16.0.10 |

**Note:** We have applied the PR #3177 patch (`fix(store): allow destructuring proxy methods`) to fix the previous "Destructuring the client method is not supported" error. This is a **different** error that occurs after the patch.

## Error

```
Error: Duplicate key toolCallId-call_CIw22cJmfIEYCrMJPaXpZyhy in tapResources
    at tapMemo (tap-resources.js)
    at tapResources (tap-resources.js)
    at callResourceFn (...)
    at tapInlineResource (...)
    ... (React component stack)
```

## Reproduction Steps

1. Create a HITL tool using `makeAssistantToolUI`:

```tsx
"use client";

import { makeAssistantToolUI } from "@assistant-ui/react";

export const ApprovalTool = makeAssistantToolUI<
  { title: string; description: string },
  "approved" | "denied"
>({
  toolName: "requestApproval",
  render: ({ args, result, addResult }) => {
    if (result !== undefined) {
      return <div>Decision: {result}</div>;
    }

    return (
      <div>
        <h3>{args.title}</h3>
        <p>{args.description}</p>
        <button onClick={() => addResult("approved")}>Approve</button>
        <button onClick={() => addResult("denied")}>Deny</button>
      </div>
    );
  },
});
```

2. Register the tool in your component tree:

```tsx
<AssistantRuntimeProvider runtime={runtime}>
  <Thread />
  <ApprovalTool />
</AssistantRuntimeProvider>
```

3. Have the agent call the `requestApproval` tool

[...]

Issue #3173, opened 2026-01-29
HITL Tool addResult Broken with AI SDK v6 and @assistant-ui/react v0.12.x


chinkan:
## Summary

Human-in-the-Loop (HITL) tool UIs using `makeAssistantToolUI` with `addResult` prop are broken when using `@assistant-ui/react-ai-sdk` with Vercel AI SDK v6. Calling `addResult()` or `props.addResult()` throws proxy-related errors, making HITL tools completely unusable.

## Environment

| Package | Version |
|---------|---------|
| `@assistant-ui/react` | 0.12.3 |
| `@assistant-ui/react-ai-sdk` | 1.3.3 |
| `@assistant-ui/react-markdown` | 0.12.1 |
| `ai` (Vercel AI SDK) | 6.0.57 |
| `react` | 19.1.1 |
| `react-dom` | 19.1.1 |
| Node.js | 22.x |
| pnpm | 10.28.1 |

## Issue Description

When using `makeAssistantToolUI` to create a HITL tool that collects user input (e.g., OTP verification, approval dialogs), calling `addResult()` from the render function throws errors from the `@assistant-ui/store` proxy mechanism.

### Error 1: Destructuring addResult

When destructuring `addResult` from props:

```tsx
render: ({ args, result, addResult }) => {
  // ...
  addResult({ value: "test" });
}
```

**Error:**
```
Uncaught (in promise) Error: Destructuring the client method "addResult" is not supported.
    at template (tapClientResource.ts:53:15)
```

### Error 2: Using props.addResult

When using `props.addResult()` without destructuring:

```tsx
render: function MyTool(props) {
  // ...
  props.addResult({ value: "test" });
}
```

**Error:**
```
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'methods')
    at Object.template [as addResult] (tapClientResource.ts:57:46)
```

### Error 3: Using useAui() in Inner Component

When using `useAui().part().addToolResult()` in a separate inner component:

```tsx
function ToolInner() {
  const aui = useAui();
  const handleSubmit = () => {
    aui.part().addToolResult({ value: "test" });
  };
  return <Button onClick={handleSubmit}>Submit</Button>;
}
```

**Error:**
```
Uncaught Error: Duplicate key toolCallId-call_XXX in tapResources
    at tap-resources.ts:63:15
```

## Root Cause Analysis

[...]

PR #3198, opened 2026-02-03
fix: prevent duplicate tool call rendering and handle argsText reset gracefully

chinkan:
## Summary

This PR fixes two issues that occur during Human-in-the-Loop (HITL) tool flows and multi-step tool execution:

- **Duplicate tool call UI rendering** - When message content contains multiple entries with the same `toolCallId`, both are rendered causing duplicate UI elements
- **"argsText can only be appended" error** - When tool call arguments are reset during streaming (e.g., transitioning from incomplete to complete state), an error is thrown that crashes the UI

### Issue 1: Duplicate Tool Call Rendering

When using HITL tools with `makeAssistantToolUI` and `addResult`, the message content array can contain duplicate entries with the same `toolCallId`. The current implementation renders all entries, causing duplicate tool UIs to appear in the chat.

**Reproduction:**
1. Create a HITL tool using `makeAssistantToolUI` with `addResult`
2. Trigger the tool and call `addResult` to complete it
3. Observe that the same tool UI appears multiple times

### Issue 2: argsText Reset Error

During multi-step tool execution or HITL flows, the `argsText` can be reset from an incomplete streaming state to a new complete value. The current implementation throws an error:

```
Tool call argsText can only be appended, not updated: {} does not start with {"incomplete":"json...
```

This crashes the UI when the agent continues after HITL completion.

### Fix 1: Skip duplicate toolCallId entries

Modified `ThreadMessageClient.tsx` and `MessageRuntimeClient.ts` to track seen `toolCallId` values and skip rendering duplicates entirely (instead of just assigning different React keys).

### Fix 2: Handle argsText reset gracefully

Modified `useToolInvocations.ts` to accept argsText transitions from incomplete to complete JSON, which commonly occurs during HITL flows when the agent continues after tool completion.

**Changes:**
1. If old argsText is incomplete and new argsText is complete JSON, accept the reset gracefully
2. Log a warning in development mode for debugging
3. Never throw errors that crash the UI - gracefully recover instead

## Test plan

- [ ] Create a HITL tool with `makeAssistantToolUI` that uses `addResult`
- [ ] Trigger the tool and complete it with approve/deny action
- [ ] Verify no duplicate tool UIs appear
- [ ] Verify no "argsText can only be appended" errors
- [ ] Verify the multi-step flow completes successfully

## Related Issues

Potentially related to #3173 #3197

----

> [!IMPORTANT]
> Fixes duplicate tool call rendering and argsText reset errors in HITL tool flows by updating `ThreadMessageClient.tsx`, `MessageRuntimeClient.ts`, and `useToolInvocations.ts`.
> 
>   - **Behavior**:
>     - Prevents duplicate tool call UI rendering by tracking `toolCallId` in `ThreadMessageClient.tsx` and `MessageRuntimeClient.ts`.
>     - Handles `argsText` reset from incomplete to complete JSON in `useToolInvocations.ts` to prevent UI crashes.
>   - **Files**:
>     - `ThreadMessageClient.tsx`: Skips rendering duplicate `toolCallId` entries.
>     - `MessageRuntimeClient.ts`: Skips rendering duplicate `toolCallId` entries and maintains correct runtime index mapping.
>     - `useToolInvocations.ts`: Accepts `argsText` transitions from incomplete to complete JSON and logs warnings in development mode.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=assistant-ui%2Fassistant-ui&utm_source=github&utm_medium=referral)<sup> for 8247bc9e390243850ed7a83c9b702116ce9692e9. You can [customize](https://app.ellipsis.dev/assistant-ui/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/.changeset/fix-duplicate-toolcall-rendering.md
+++ b/.changeset/fix-duplicate-toolcall-rendering.md
@@ -1,0 +1,8 @@
+---
+"@assistant-ui/react": patch
+---
+
+fix: prevent duplicate tool call rendering and handle argsText reset gracefully
+
+- Skip rendering duplicate `toolCallId` entries in message content to prevent duplicate tool call UIs during HITL flows
+- Handle argsText reset from incomplete to complete JSON gracefully instead of throwing errors
--- a/packages/react/src/client/ThreadMessageClient.tsx
+++ b/packages/react/src/client/ThreadMessageClient.tsx
@@ -77,15 +77,18 @@ export const ThreadMessageClient = resource(
     const [isHoveringState, setIsHovering] = tapState(false);
 
     const parts = tapClientLookup(() => {
-      // Track seen toolCallIds to skip duplicates entirely
-      // This prevents duplicate tool call UI rendering during HITL flows
+      // Track seen toolCallIds to skip duplicates entirely.
+      // This prevents duplicate tool call UI rendering during HITL flows where
+      // the same toolCallId may appear multiple times in message.content.
+      // Note: After filtering, part({ index: N }) returns the Nth non-duplicate part,
+      // not necessarily message.content[N]. Use part({ toolCallId }) for tool calls.
       const seenToolCallIds = new Set<string>();
       return message.content
         .map((part, idx) => {
           if ("toolCallId" in part && part.toolCallId != null) {
             const toolCallId = part.toolCallId;
             if (seenToolCallIds.has(toolCallId)) {
-              // Skip duplicate toolCallId entirely to prevent duplicate UI rendering
+              // Skip duplicate toolCallId - only render the first occurrence
               return null;
             }
             seenToolCallIds.add(toolCallId);
--- a/packages/react/src/legacy-runtime/client/MessageRuntimeClient.ts
+++ b/packages/react/src/legacy-runtime/client/MessageRuntimeClient.ts
@@ -71,19 +71,22 @@ export const MessageClient = resource(
     );
 
     const parts = tapClientLookup(() => {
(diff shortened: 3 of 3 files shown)


okisdev:
thanks for investigating this and putting together a detailed PR!

we've addressed the duplicate `toolCallId` crash in #3249 with a different approach that targets the root causes:

1. **race condition fix in `useToolInvocations.ts`** — moved `hasResult = true` before `setResponse()`/`close()` to prevent the stream pipeline's `onResult` callback from firing a duplicate `add-tool-result` command.
2. **ai sdk source-level dedup in `convertMessage.ts`** — filters duplicate `toolCallId` parts at the converter boundary where ai sdk data enters the system.

on this PR:

- **rendering-layer dedup (`ThreadMessageClient` / `MessageRuntimeClient`)**: this creates an index semantic mismatch — after filtering, `parts.get({ index: N })` returns the Nth *filtered* part, but `MessagePartByIndex` still uses the *original* content array index. we opted to fix the root causes instead so duplicates never reach `tapResources`.
- **argsText reset handling**: the `argsText` accumulation is guaranteed to be append-only by both `AssistantMessageAccumulator` (assistant-stream path) and `stripClosingDelimiters` (ai sdk path). the existing key-reordering check (both-complete-JSON case) covers the known edge case. the throw is an intentional invariant - if it fires, it indicates an upstream bug that should be investigated rather than silently swallowed.

closing in favor of #3249. thanks again for the report and analysis!

PR #3237, opened 2026-02-06
feat(store): add useAuiStateShallow hook for object selectors

cewinharhar:
## Summary

When using `useAuiState` with selectors that return objects or arrays, React's `useSyncExternalStore` triggers infinite re-renders because each selector call creates a new object reference that fails `Object.is` comparison.

This PR adds `useAuiStateShallow` which uses shallow equality comparison to maintain stable references when the object contents are identical.

## Problem

```typescript
// ❌ Causes infinite re-render loop
const { count, isActive } = useAuiState(({ thread }) => ({
  count: thread.messages.length,
  isActive: thread.status?.type === "running",
}));
```

The selector returns a new `{}` object on every call. `useSyncExternalStore` uses `Object.is` to compare snapshots:
- New object `{count: 5, isActive: true}` !== previous object `{count: 5, isActive: true}`
- React thinks state changed → re-render → new selector call → new object → infinite loop

## Solution

```typescript
// ✅ Works correctly with shallow comparison  
const { count, isActive } = useAuiStateShallow(({ thread }) => ({
  count: thread.messages.length,
  isActive: thread.status?.type === "running",
}));
```

`useAuiStateShallow` caches the previous result and returns it if shallowly equal to the new result, maintaining reference stability.

## Changes

- **`shallow.ts`** — New shallow equality utility function
- **`useAuiState.tsx`** — Added `useAuiStateShallow` hook alongside existing `useAuiState`
- **`index.ts`** — Export `useAuiStateShallow` and `shallow`

## Prior Art

This follows the same pattern as [Zustand's `useShallow`](https://github.com/pmndrs/zustand#using-zustand-without-react) which solves the identical problem with `useSyncExternalStore`.

## Real-world Impact

We hit this bug while building a production LangGraph multi-agent chat app with assistant-ui. The `ChainOfThoughtPrimitive` component crashed with infinite loops when trying to count reasoning/tool-call parts:

```typescript
// Our workaround: split into separate primitive selectors
const reasoningCount = useAuiState(({ chainOfThought }) =>
  chainOfThought.parts.reduce((n, p) => n + (p.type === "reasoning" ? 1 : 0), 0)
);
const toolCount = useAuiState(({ chainOfThought }) =>
  chainOfThought.parts.reduce((n, p) => n + (p.type === "tool-call" ? 1 : 0), 0)
);

// With this PR, this natural pattern would work:
const { reasoningCount, toolCount } = useAuiStateShallow(({ chainOfThought }) => ({
  reasoningCount: chainOfThought.parts.filter(p => p.type === "reasoning").length,
  toolCount: chainOfThought.parts.filter(p => p.type === "tool-call").length,
}));
```

----

> [!IMPORTANT]
> Adds `useAuiStateShallow` hook and `shallow` utility to prevent infinite re-renders with object/array selectors.
> 
>   - **New Hook**:
>     - Adds `useAuiStateShallow` in `useAuiState.tsx` to prevent infinite re-renders by using shallow equality comparison for selectors returning objects or arrays.
>   - **Utility Function**:
>     - Introduces `shallow` function in `shallow.ts` for shallow comparison of objects and arrays.
>   - **Exports**:
>     - Updates `index.ts` to export `useAuiStateShallow` and `shallow`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=assistant-ui%2Fassistant-ui&utm_source=github&utm_medium=referral)<sup> for ac66206b08058472e93cd052f1f3e63d32c8b4fd. You can [customize](https://app.ellipsis.dev/assistant-ui/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/.changeset/shallow-state-selector.md
+++ b/.changeset/shallow-state-selector.md
@@ -1,0 +1,25 @@
+---
+"@assistant-ui/store": minor
+---
+
+feat: add useAuiStateShallow hook to prevent infinite re-renders
+
+When using `useAuiState` with selectors that return objects or arrays, React's `useSyncExternalStore` triggers infinite re-renders because each selector call creates a new object reference that fails `Object.is` comparison.
+
+This adds `useAuiStateShallow` which uses shallow equality comparison to maintain stable references when the object contents are identical.
+
+```typescript
+// ❌ Causes infinite re-render loop
+const { count, isActive } = useAuiState(({ thread }) => ({
+  count: thread.messages.length,
+  isActive: thread.status?.type === "running",
+}));
+
+// ✅ Works correctly with shallow comparison
+const { count, isActive } = useAuiStateShallow(({ thread }) => ({
+  count: thread.messages.length,
+  isActive: thread.status?.type === "running",
+}));
+```
+
+Also exports the `shallow` utility function for custom equality comparisons.
--- a/packages/store/src/index.ts
+++ b/packages/store/src/index.ts
@@ -1,7 +1,10 @@
 // hooks
 export { useAui } from "./useAui";
-export { useAuiState } from "./useAuiState";
+export { useAuiState, useAuiStateShallow } from "./useAuiState";
 export { useAuiEvent } from "./useAuiEvent";
+
+// utilities
+export { shallow } from "./shallow";
 
 // components
 export { AuiIf } from "./AuiIf";
--- a/packages/store/src/shallow.ts
+++ b/packages/store/src/shallow.ts
@@ -1,0 +1,46 @@
+/**
+ * Shallow comparison function for objects and arrays.
+ * Returns true if both values are shallowly equal.
+ *
+ * Used by useAuiStateShallow to prevent infinite re-renders
+ * when selectors return new object references with identical values.
+ */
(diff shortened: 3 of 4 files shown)


okisdev:
thanks for the patch and the detailed write-up. the real-world friction is genuine, but we're going to close this one.

the design we want for `useAuiState` is one call per leaf value (a primitive or stable reference). the docs at `apps/docs/content/docs/guides/context-api.mdx:109` show this PR's exact "Bad" example, and the troubleshooting section at line 584 calls out infinite re-renders specifically. adding `useAuiStateShallow` as a parallel API would soften that guidance and create a "which hook do i pick?" decision on every selector. the workaround you used in your own description (separate `reasoningCount` / `toolCount` selectors) is the pattern we want to keep recommending.

separately, the current implementation also has a memoization bug (`useCallback` deps include the fresh-per-render `proxiedState`), so it would need a rewrite + tests before it could land even if we did want the API.

if you hit a place where the leaf-per-call pattern is awkward enough that splitting feels worse than bundling, please open an issue with the specific component. there might be a `RenderChildrenWithAccessor`-style helper that fits better than a shallow hook.

closing for now. thanks again!

PR #3240, opened 2026-02-06
feat: add cloudflare agents integration

ShobhitPatra:
> [!IMPORTANT]
> Introduces `@assistant-ui/react-cloudflare-agents` for integrating Cloudflare Agents with assistant-ui, including documentation and example setup.
> 
>   - **New Package**:
>     - Adds `@assistant-ui/react-cloudflare-agents` for integrating Cloudflare Agents with assistant-ui.
>     - Provides `useCloudflareAgentRuntime` hook for connecting to Cloudflare Workers.
>   - **Documentation**:
>     - Adds `react-cloudflare-agents.mdx` and updates `meta.json` in `api-reference/integrations`.
>     - Adds `cloudflare-agents.mdx` and updates `meta.json` in `runtimes`.
>   - **Examples**:
>     - Adds `with-cloudflare-agents` example with setup instructions in `README.md`.
>     - Includes `worker` and `app` directories for backend and frontend setup.
>   - **Backend**:
>     - Implements `AIChatAgent` in `worker/chat.ts` for handling chat messages.
>     - Configures CORS and routing in `worker/index.ts`.
>   - **Frontend**:
>     - Uses `useCloudflareAgentRuntime` in `app/page.tsx` to connect to the worker.
>     - Configures Tailwind CSS in `app/globals.css`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=assistant-ui%2Fassistant-ui&utm_source=github&utm_medium=referral)<sup> for 047545b1b0729892e34e6b5c08330cc381751807. You can [customize](https://app.ellipsis.dev/assistant-ui/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


okisdev:
hey @ShobhitPatra, circling back to close this one properly and share some longer-form context, since this is the second integration PR that hit the same wall (after #3192). writing it out so the pattern is explicit, not because the work wasn't appreciated.

## why this PR specifically is closing

after doing a deep audit of the integration (full compat trace through `useAISDKRuntime`, both type-level and behavioural), the conclusion is that Cloudflare Agents doesn't need its own adapter package. `@cloudflare/ai-chat`'s `useAgentChat` returns `Omit<ReturnType<typeof useChat>, "addToolOutput"> & { extras }`. it's literally `useChat` with a WebSocket transport and a handful of CF-specific helpers. `useAISDKRuntime` already consumes that shape end to end (status, messages, sendMessage, regenerate, stop, setMessages, addToolOutput, addToolResult, error, history, edit, reload, suggestions, attachments). the wrapper hook in this PR re-implements maybe 10% of what `useAISDKRuntime` already does, and in the process had to expose `@assistant-ui/react-ai-sdk/internal` so the new package could borrow `AISDKMessageConverter`. the fact that the new package couldn't stand up without breaking a sibling package's encapsulation is itself the strongest signal that the boundary is in the wrong place.

the rule of thumb that came out of this: if the upstream library returns AI-SDK-shape values, or can be converted to `UIMessageStream` in the user's API route, it's a docs guide, not a package. dedicated adapter packages exist only for protocols that genuinely aren't AI SDK shape (react-langgraph, react-ag-ui, react-a2a, react-google-adk, react-opencode). Cloudflare Agents and Mastra both sit on the docs side of that line.

the replacement landed in #4042. it covers everything this PR aimed at ("users can build chat UIs against Cloudflare Agents") with zero new package surface, zero `internal.ts` re-exports, and the full power of `useAISDKRuntime` instead of a hand-rolled subset.

## three things worth internalizing for next time

**1. a new package is a big call. default to "no".**

every published package in this repo is a permanent commitment: a version range to maintain, peer-dep matrix to track, a semver story to keep clean, docs to keep current, and an API surface that has to stay backward compatible. adding one for an integration that already routes through an existing adapter creates ongoing maintenance overhead forever. before opening a new package, the question to answer first is: *what does the upstream library do that the existing adapters cannot wrap?* if the honest answer is "nothing, it's just a different transport for the same protocol", it's a docs page. if it's "yes, here is a specific event or protocol or UI semantic with no equivalent in AI SDK or LocalRuntime", then a package may be justified, and the PR should lead with that gap, not bury it under scaffolding.

**2. research depth before code volume.**

both #3192 and #3240 shipped a substantial implementation before the architectural question was answered. that's the hardest failure mode to recover from: the code becomes sunk cost, conflicts accumulate against main, CI rots, and the eventual "actually, this should be done a different way" feedback lands on top of weeks of polish. the cheaper version: open a short issue or design note first with one paragraph: "here is what I want to integrate; here is the upstream API shape; my best guess is X adapter / Y new package, but I want to confirm before investing." 10 minutes of upfront alignment saves a thousand lines of churn.

**3. precision over size.**

a 14k-line, 95-file PR (#3192) and a 3k-line, 34-file PR (#3240) are genuinely hard to review carefully even when the maintainer wants to. small focused PRs land in days; big ones spend weeks in queue, accumulate conflicts, lose reviewer attention, and eventually go stale. when a feature genuinely needs that much code, splitting into independent landing surfaces (core adapter / docs / example app) lets reviewers move on each one separately, lets you ship the foundational piece while iterating on the rest, and acts as a natural filter: if the core adapter alone doesn't justify itself, the docs and example pieces don't either.

## use draft mode

related to all of the above: `gh pr create --draft` (or "Convert to draft" in the UI) is the right state for any PR where the architectural fit isn't already confirmed. draft mode signals "feedback welcome, not asking for merge yet", which is exactly what you want when the approach is still being figured out. both of these PRs were opened as ready-for-review with full polish, which framed the conversation as "this should be merged, what's blocking?" rather than "is this the right shape?". the latter framing would have surfaced the AI-SDK-compat observation in a comment thread before any worker scaffold, example app, or `internal.ts` export was written.

## if you want to keep going

if there's still a Cloudflare Agents use case that #4042 doesn't cover (CF-specific UI affordances that don't fit AI SDK's UIMessage model, native subagent tracking visualisation, server-side persistence reset UI, etc.), open an issue describing the specific gap. that might end up as a small set of helpers rather than a full package, and the conversation starts from "here's the missing primitive" instead of "here's a 3k-line PR".

appreciate the persistence and the depth of thought visible in both attempts. closing here.

PR #3300, opened 2026-02-15
feat(react): add message-level component scope clients

petekp:
## Summary
- add `component` scope schema and store augmentation for message-level component clients
- add `message.component({ index | instanceId })` lookup API
- hydrate component state from `metadata.unstable_state.components[instanceId]` in ThreadMessageClient, MessageRuntimeClient, and ExternalThread
- emit scoped `component.lifecycle` and `component.state` events on seq-forward transitions
- add contract tests for lookup ordering, hydration, and scoped emits

## Validation
- `pnpm --filter @assistant-ui/react test -- src/tests/message-component-client.test.tsx`


petekp:
Superseded by consolidated Stage 1 rollup PR #3303 to reduce PR noise. Follow-up Stage 2 work continues on #3302 (rebased onto #3303).

PR #3392, opened 2026-02-22
feat: add support for data-* prefixed types in messages and attachments

### LINKED ISSUE #3358: Support DataUIPart (data-* prefixed types) in user messages and attachments
Hi! Following up on #3311 and #3312. PR #3316 added `DataMessagePart` (type `"data"`) to `ThreadUserMessagePart`, which is great. However, `data-*` prefixed types (e.g. `data-workflow`, `data-reference`) are still not supported. Here are the issues I'm running into:

## 1. Cannot use `data-*` parts in user messages

When sending a user message with a custom `data-*` part, TypeScript rejects it because `ThreadUserMessagePart` doesn't include `data-*` prefixed types:

```tsx
const message: ThreadMessageLike = {
  role: "user",
  content: [
    { type: "text", text: "Run this workflow" },
    { type: "data-workflow", data: { id: "wf-1", name: "Summarize" } },
    //       ^^^^^^^^^^^^^^ Type error: not assignable to ThreadMessageLike content
  ],
};
```

At runtime, `fromThreadMessageLike()` throws `Unsupported user message part type: data-workflow`.

## 2. Cannot create `data-*` attachments

When appending a message with a `data-*` attachment via the runtime API, TypeScript rejects it because `BaseAttachment.type` only allows `"image" | "document" | "file"`:

```tsx
api.thread().append({
  role: "user",
  content: [{ type: "data-workflow", data: { id: "wf-1" } }],
  attachments: [{
    id: "att-1",
    type: "data-workflow",
    //    ^^^^^^^^^^^^^^^ Type error: not assignable to "image" | "document" | "file"
    name: "My Workflow",
    status: { type: "complete" },
    content: [{ type: "data-workflow", data: { id: "wf-1" } }],
  }],
});
```

## 3. `contentType` is required but meaningless for data attachments

`BaseAttachment` requires `contentType: string`, which represents a MIME type for files. For `data-*` attachments there is no meaningful MIME type since they are structured data, not files. This forces consumers to pass a dummy value:

```tsx
{
  type: "data-workflow",
  name: "My Workflow",
  contentType: "", // forced to provide this even though it has no meaning here
  // ...
}
```

Ideally, `contentType` would not exist at all on `data-*` attachments (not even as optional).

## 4. Runtime crashes on unknown types instead of graceful fallback

Several components throw on unrecognized types at runtime instead of degrading gracefully:

- `ComposerAttachments` and `MessageAttachments` throw `Unknown attachment type`, crashing the UI even when a generic `Attachment` component is provided
- `MessageParts` and `MessagePartsGrouped` throw `Unknown message part type`, crashing the UI instead of silently skipping unrecognized parts
- `fromThreadMessageLike()` throws for `data-*` parts on assistant messages instead of filtering them out

It would be great if these could fall back to a generic component or return null rather than crash.

Thanks!

adi-bradwell:
## Summary

Adds full support for `data-*` prefixed types in user messages and attachments, addressing all issues from #3358.

### Changes (17 files, +66 / -29)

**New types (`packages/core/src/types/`):**
- Added `DataUIPart` type: `{ type: \`data-${string}\`; data: unknown; id?: string }`
- Updated `ThreadUserMessagePart` union to include `DataUIPart`
- Split `BaseAttachment` into a discriminated union: `BaseFileAttachment` (with required `contentType: string`) and `BaseDataAttachment` (with optional `contentType?: string`)

**Graceful degradation in switch statements:**
- `thread-message-like.ts`: user handler returns `data-*` parts as-is (using `isDataUIPart` type guard); assistant handler returns `null` (skips)
- `MessageParts.tsx` / `MessagePartsGrouped.tsx`: returns `null` for `data-*` parts
- `ComposerAttachments.tsx` / `MessageAttachments.tsx`: falls back to `components?.Attachment`
- `auiV0.ts`: skips `data-*` parts via `return null` + `.filter(Boolean)`
- All downstream packages (`react-a2a`, `react-langgraph`, `react-ai-sdk`, `ui`, `docs`) updated

**Adapter guards:**
- `CompositeAttachmentAdapter.remove()`: early return for `data-*` attachments
- `CloudFileAttachmentAdapter.send()`: graceful handling when `contentType` is absent
- `vercelAttachmentAdapter.send()`: runtime check for `contentType`

**Exports:**
- `DataUIPart` exported from both `@assistant-ui/core` and `@assistant-ui/react`

### Design principles
- **Add-only**: no existing behavior is changed; guards are added in `default` cases before existing throws
- **Uses `type.startsWith("data-")`** consistently (not `"data"`) to avoid changing behavior for the existing bare `"data"` type
- **No `as` assertions** — uses proper type guard (`isDataUIPart`) instead

### Verified
- All 44 build tasks pass
- All 35 test tasks pass

## Test plan
- [x] Full monorepo build passes (44/44 tasks)
- [x] Full test suite passes (35/35 tasks)
- [ ] Verify `data-*` parts in user messages don't crash rendering
- [ ] Verify `data-*` attachments can be created and removed without errors
- [ ] Verify existing `"data"`, `"image"`, `"file"`, etc. types continue working unchanged

Resolves #3358

----

> [!IMPORTANT]
> Adds support for `data-*` prefixed types in messages and attachments, ensuring graceful degradation and proper handling across components and adapters.
> 
>   - **Behavior**:
>     - Adds `DataUIPart` type to `types/message.ts` and updates `ThreadUserMessagePart` union to include it.
>     - Splits `BaseAttachment` into `BaseFileAttachment` and `BaseDataAttachment` in `types/attachment.ts`.
>     - Updates `fromThreadMessageLike()` in `thread-message-like.ts` to handle `data-*` parts, returning `null` for assistant role.
>     - Graceful degradation in `MessageParts.tsx`, `MessagePartsGrouped.tsx`, and `auiV0.ts` by returning `null` for `data-*` parts.
>   - **Adapters**:
>     - `CompositeAttachmentAdapter.remove()` in `attachment.ts` returns early for `data-*` attachments.
>     - `CloudFileAttachmentAdapter.send()` in `CloudFileAttachmentAdapter.ts` handles missing `contentType` gracefully.
>     - `vercelAttachmentAdapter.send()` in `vercelAttachmentAdapter.ts` checks for `contentType` at runtime.
>   - **Exports**:
>     - `DataUIPart` exported from `@assistant-ui/core` and `@assistant-ui/react`.
>   - **Misc**:
>     - Consistent use of `type.startsWith("data-")` across files to avoid affecting existing `"data"` type.
>     - No `as` assertions used; proper type guards are implemented.
>   - **Verification**:
>     - All 44 build tasks and 35 test tasks pass.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=assistant-ui%2Fassistant-ui&utm_source=github&utm_medium=referral)<sup> for d6afa30359cdd83b5c7b26023346f50b7acafec8. You can [customize](https://app.ellipsis.dev/assistant-ui/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/packages/react-langgraph/src/useLangGraphRuntime.ts
+++ b/packages/react-langgraph/src/useLangGraphRuntime.ts
@@ -83,8 +83,12 @@ const getMessageContent = (msg: AppendMessage) => {
       default:
         if (type.startsWith("data-"))
           throw new Error(`Unsupported append message part type: ${type}`);
-        const _exhaustiveCheck: "reasoning" | "source" | "audio" | "data" | `data-${string}` =
-          type;
+        const _exhaustiveCheck:
+          | "reasoning"
+          | "source"
+          | "audio"
+          | "data"
+          | `data-${string}` = type;
         throw new Error(
           `Unsupported append message part type: ${_exhaustiveCheck}`,
         );


petekp:
@adi-bradwell Hi, just want to confirm, did https://github.com/assistant-ui/assistant-ui/pull/3376 not address this fully?

okisdev:
hi @adi-bradwell, thanks for the contribution and for tackling #3358.

this was addressed by #3376 which was merged shortly before this PR was opened. the approach we went with normalizes `data-*` parts at the boundary (`data-foo` → `DataMessagePart { type: "data", name: "foo" }`) rather than passing them through as a first-class type — this keeps the internal type system clean with fixed string literals so switch statements can exhaustively match.

closing in favor of #3376. thanks again for the effort here.

PR #3473, opened 2026-03-02
fix: harden and simplify npm release workflows

JustAnOkapi:
Improves release workflow safety and operational behavior for npm publishing while simplifying failure handling.

## Changes
- `changeset.yaml`
  - Uses pinned action SHAs.
  - Detects version-bumped public packages via `actions/github-script` and only dispatches npm publish when release packages exist.
  - Handles zero/missing `before` SHA safely.
  - Dispatches `repository_dispatch` with `releaseSha`.

- `npm-publish.yaml`
  - Uses a one-approval flow:
    - `approve` resolves and validates `releaseSha` plus repository `default_branch`.
    - `cancel_pending` best-effort cancels older queued/waiting runs for this workflow before publish.
    - `publish` is the environment-gated job (`npm Publish`).
  - Supports `repository_dispatch` and `workflow_dispatch`:
    - manual runs may pass `releaseSha`,
    - manual fallback defaults to default-branch HEAD.
  - Validates checkout SHA and ancestry against `origin/<default_branch>`.
  - Resolves a single target set from package manifests vs npm:
    - target package = `latest` tag does not match repo version.
    - fails on prerelease versions in default-branch releases.
    - refuses downgrading `latest` to an older version.
  - Runs `changeset publish --tag next` with one retry after 20 seconds.
  - Promotes the same target set to `latest` using npm trusted publishing OIDC exchange.
  - Uses one GitHub OIDC token by default; refreshes token and retries exchange once on `401`.
  - Uses explicit npm registry for dist-tag promotion.

## Outcome
- Publish execution is pinned to an explicit commit SHA.
- Approval remains required via `npm Publish` environment before execution.
- Older waiting publish runs are canceled so only the newest pending request proceeds.
- Reruns continue partial `next` publishes and then promote the release target set to `latest`.

### DIFF
--- a/.github/workflows/npm-publish.yaml
+++ b/.github/workflows/npm-publish.yaml
@@ -222,19 +222,6 @@ jobs:
           fetch-depth: 0
           ref: ${{ env.RELEASE_SHA }}
 
-      - name: Install pnpm
-        uses: pnpm/action-setup@41ff72655975bd51cab0327fa583b6e92b6d3061 # v4
-
-      - name: Setup node.js
-        uses: actions/setup-node@6044e13b5dc448c55e2357c09f80417699197238 # v6
-        with:
-          node-version: 24
-          # No cache - hardened against cache poisoning for sensitive publish operations
-          registry-url: "https://registry.npmjs.org"
-
-      - name: Install dependencies
-        run: pnpm install --frozen-lockfile
-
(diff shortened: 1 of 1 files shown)


JustAnOkapi:
https://github.com/assistant-ui/assistant-ui/pull/3473 replaced

PR #3480, opened 2026-03-03
fix(core): respect attachment changes in edit composer

### LINKED ISSUE #3476: [BUG] DefaultEditComposerRuntimeCore: original FileMessagePart content is always re-injected on edit, making file-bearing message edits broken
## Package

`@assistant-ui/core` / `@assistant-ui/react`

## Summary

When a user edits a message that contains `FileMessagePart` content (files that arrived in `message.content` after a round-trip through the backend), `DefaultEditComposerRuntimeCore` captures all original non-text parts in `_nonTextParts` and **unconditionally re-appends them** to the edited message's content on every save — regardless of what the user changed in the edit composer.

This causes two serious problems:

1. **Duplicate document names** — Backends like AWS Bedrock reject a conversation where the same file appears in multiple messages. After an edit, the original message's file is still in thread history AND the edited message also contains the re-injected file, resulting in the error: `"Messages can't contain duplicate document names."` 

2. **Attachment removal in edit composer has no effect** — If a user explicitly removes a file attachment from the edit composer, it is silently re-added to the saved message from `_nonTextParts`. There is no way for a user to remove a file from an edited message.

## Steps to Reproduce

1. Send a message with a file attachment (PDF or image)
2. Receive an AI response
3. Click "Edit" on the original message
4. Change the text and click "Send" (or just click Send without changing text)
5. Observe the error from the backend

## Error (AWS Bedrock via LangGraph)

```json
{
  "code": "VALIDATION_EXCEPTION",
  "error": "ValidationException",
  "message": "An error occurred (ValidationException) when calling the ConverseStream operation: Messages can't contain duplicate document names. Rename the document and retry your request."
}
```

## Root Cause

In `DefaultEditComposerRuntimeCore`:

```ts
constructor(runtime, endEditCallback, { parentId, message }) {
  // ...
  this._nonTextParts = message.content.filter((part) => part.type !== "text")
}

async handleSend(message) {
  const text = getThreadMessageText(message)
  if (text !== this._previousText) {
    this.runtime.append({
      ...message,
      content: [...message.content, ...this._nonTextParts], // ← always re-injects!
      parentId: this._parentId,
      sourceId: this._sourceId,
    })
  }
  this.handleCancel()
}
```

`_nonTextParts` contains the original file content parts. They are appended to the new message unconditionally, even if:
- The user removed the file from the edit composer
- The file was already in the thread history at the checkpoint being forked from

## Additional Problem: Attachment-Only Edits Are Silently Dropped

`handleSend` only calls `this.runtime.append()` if `text !== this._previousText`. If the user only modifies attachments (e.g. removes a file) without changing any text, `append` is never called, and the edit is discarded with no feedback to the user.

## Expected Behavior

1. `_nonTextParts` should only be re-injected if the user has not explicitly modified attachments in the edit composer. If the edit composer's attachment list differs from the original, the new attachment list (not `_nonTextParts`) should be used.

2. Alternatively, `handleSend` should also trigger `append` when attachments changed (not only when text changed).

3. At minimum, when `_nonTextParts` are re-injected, the runtime should deduplicate file names to avoid backend errors.

## Workaround

We work around this in two ways:

**1. Checkpoint rewinding**: Pass the correct `checkpointId` (the state before the original message was added) to `client.runs.stream` so LangGraph no longer has the original message in its thread history, preventing the duplicate:

```ts
getCheckpointId: async (threadId, messages) => {
  // For messages.length === 0 (editing first message), find the pre-first-message checkpoint
  if (messages.length === 0) {
    const history = await client.threads.getHistory(threadId, { limit: 100 })
    // Find an empty-messages checkpoint or the parent of the oldest entry
    for (const state of history) {
      const stateMsgs = state.values?.messages ?? []
      if (stateMsgs.length === 0) return state.checkpoint.checkpoint_id
    }
    const oldest = history[history.length - 1]
    return oldest?.parent_checkpoint?.checkpoint_id ?? null
  }
  // ...
}
```

**2. Filename deduplication**: Remove duplicate file blocks by filename in the message batch before sending to the backend.

## Environment

- `@assistant-ui/react`: 0.12.12
- `@assistant-ui/core`: 0.1.1
- `@langchain/langgraph-sdk`: 1.5.4
- LangGraph Cloud backend with AWS Bedrock

ShobhitPatra:
closes https://github.com/assistant-ui/assistant-ui/issues/3476

----

> [!IMPORTANT]
> `DefaultEditComposerRuntimeCore` now respects attachment changes in the edit composer by tracking and comparing attachment states.
> 
>   - **Behavior**:
>     - `DefaultEditComposerRuntimeCore` now respects attachment changes in `handleSend()`.
>     - Introduces `attachmentsEqual()` to compare attachment arrays.
>     - If attachments change, original non-text parts are not re-injected.
>   - **State Management**:
>     - Adds `_previousAttachments` to track previous attachment state.
>     - Updates `_previousAttachments` in constructor and `handleSend()`.
>   - **Misc**:
>     - Imports `CompleteAttachment` type in `default-edit-composer-runtime-core.ts`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=assistant-ui%2Fassistant-ui&utm_source=github&utm_medium=referral)<sup> for 2d291d0f4d67512042f707a8e67c00811a4111a5. You can [customize](https://app.ellipsis.dev/assistant-ui/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/.changeset/sixty-rings-end.md
+++ b/.changeset/sixty-rings-end.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/core": patch
+---
+
+fix(core): respect attachment changes in edit composer
--- a/packages/core/src/runtime/base/default-edit-composer-runtime-core.ts
+++ b/packages/core/src/runtime/base/default-edit-composer-runtime-core.ts
@@ -1,9 +1,21 @@
-import type { AppendMessage, ThreadMessage } from "../../types";
+import type {
+  AppendMessage,
+  CompleteAttachment,
+  ThreadMessage,
+} from "../../types";
 import { getThreadMessageText } from "../../utils/text";
 import type { AttachmentAdapter } from "../../adapters/attachment";
 import type { DictationAdapter } from "../../adapters/speech";
 import type { ThreadRuntimeCore } from "../interfaces/thread-runtime-core";
 import { BaseComposerRuntimeCore } from "./base-composer-runtime-core";
+
+const attachmentsEqual = (
+  a: readonly CompleteAttachment[],
+  b: readonly CompleteAttachment[],
+): boolean => {
+  if (a.length !== b.length) return false;
+  return a.every((att, i) => att.id === b[i]!.id);
+};
 
 export class DefaultEditComposerRuntimeCore extends BaseComposerRuntimeCore {
   public get canCancel() {
@@ -20,6 +32,7 @@ export class DefaultEditComposerRuntimeCore extends BaseComposerRuntimeCore {
 
   private _nonTextParts;
   private _previousText;
+  private _previousAttachments;
   private _parentId;
   private _sourceId;
   constructor(
@@ -41,7 +54,8 @@ export class DefaultEditComposerRuntimeCore extends BaseComposerRuntimeCore {
     this.setText(this._previousText);
 
     this.setRole(message.role);
-    this.setAttachments(message.attachments ?? []);
+    this._previousAttachments = message.attachments ?? [];
+    this.setAttachments(this._previousAttachments);
 
     this._nonTextParts = message.content.filter((part) => part.type !== "text");
 
@@ -52,10 +66,22 @@ export class DefaultEditComposerRuntimeCore extends BaseComposerRuntimeCore {
     message: Omit<AppendMessage, "parentId" | "sourceId">,
   ) {
     const text = getThreadMessageText(message as AppendMessage);
-    if (text !== this._previousText) {
+    const currentAttachments = message.attachments ?? [];
+    const didAttachmentsChange = !attachmentsEqual(
+      currentAttachments,
+      this._previousAttachments,
+    );
+
+    if (text !== this._previousText || didAttachmentsChange) {
+      // Only re-inject original non-text content parts if the user
+      // did not modify attachments; otherwise respect their changes.
+      const content = didAttachmentsChange
+        ? message.content
+        : [...message.content, ...this._nonTextParts];
+
       this.runtime.append({
         ...message,
-        content: [...message.content, ...this._nonTextParts] as any,
+        content: content as any,
         parentId: this._parentId,
         sourceId: this._sourceId,
       });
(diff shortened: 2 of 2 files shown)


okisdev:
thanks for the work on this @ShobhitPatra. closing in favor of #3857, which takes a structurally different approach. sharing the reasoning here so the trade-offs are on record.

**why this PR isn't landing as-is:**

1. **unresolved P2 from cubic**: when `didAttachmentsChange` is true, `_nonTextParts` is dropped entirely. that's data loss. adding a new attachment to an existing message silently removes the original file/image content parts. the `attachmentsEqual` guard fixes the primary bug but introduces a regression in a neighboring scenario, and the comment has been open for weeks.

2. **two parallel states**: `_nonTextParts` and `_attachments` represent the same thing (attached files) but live in separate fields with separate update paths. `DefaultThreadComposerRuntimeCore` only manages `_attachments`, so `DefaultEditComposerRuntimeCore` is the only composer with this asymmetry. the inline comment `Future non-attachment content types would need separate tracking here` flags that this design will keep needing more parallel tracking.

3. **`as any` remains**: the merge is `[...message.content, ..._nonTextParts] as any`. the type system is signaling that the two arrays don't belong in the same place.

4. **no test coverage**: `DefaultEditComposerRuntimeCore` has no existing unit tests, which is likely why the original bug survived as long as it did. this PR doesn't add any, so the same class of regression can recur silently.

**what #3857 does differently:**

- for user messages, lifts non-text content parts (`FileMessagePart`, `ImageMessagePart`, etc.) into `_attachments` during construction. single source of truth, no `_nonTextParts`, symmetric with `DefaultThreadComposerRuntimeCore`.
- for assistant and system messages, keeps a content pass-through (those roles can't carry `attachments` because `fromThreadMessageLike` throws), so reasoning, tool-call, and source parts are preserved intact.
- extracts `attachmentsEqual`, `partToCompleteAttachment`, `liftNonTextParts` into `adapters/attachment.ts` next to `fileMatchesAccept`.
- 25 unit tests covering role branches, attachment removal, `startRun`, lifecycle, and all part-type conversions.
- `as any` becomes `as AppendMessage["content"]` since the merged shape is now role-consistent.

appreciate the initial signal on the bug. your `attachmentsEqual` helper carried over into #3857 under the same name.

PR #3485, opened 2026-03-03
ci: harden npm publish workflow reliability

JustAnOkapi:
## Problem

When a new package isn't configured for npm trusted publishing, `changeset publish` publishes configured packages but fails on the unconfigured one — leaving published packages with broken peer dependencies on the missing version. Recovery requires manual `npm dist-tag` fixes.

The existing GitHub release creation step uses `gh release create --verify-tag`, which silently fails because `changeset publish` creates tags locally but never pushes them. Release detection shells out to `npm view` on every package, which can race against registry propagation after a version merge.

Action refs use mutable version tags, and `cancel-in-progress: true` on the changeset workflow can cancel a run mid-version-PR creation.

## What changed

**changeset.yaml**
- SHA-pin all actions; set `cancel-in-progress: false`
- Replace shell `npm view` loop with git-diff version comparison (`fetch-depth: 0`); only dispatch when a public package's version actually changed
- Pass `releaseSha: context.sha` in dispatch payload

**npm-publish.yaml** (near-complete rewrite)
- New `approve` job resolves release SHA from dispatch payload or `workflow_dispatch` input, validates 40-char format
- New `cancel_pending` job cancels stale queued runs while preserving in-progress publishes
- Publish job checks out exact `RELEASE_SHA` and validates it is an ancestor of the default branch
- New "Resolve publish targets" step queries npm registry to determine which packages actually need publishing; warns (not fails) on stale dist-tags; rejects prerelease versions and downgrades
- New "Verify npm trusted publishing" step does pre-flight OIDC token exchange for every target — fails the entire run before any publish if any package is misconfigured
- Run `pnpm ci:publish` through `changesets/action` instead of a bare `run:` step, adding `createGithubReleases: true` for tag pushing and release creation
- Remove `NPM_TOKEN`/`NODE_AUTH_TOKEN` reference and `registry-url` from setup-node (OIDC handles auth directly)
- Scope permissions per-job; pin `ubuntu-24.04`; SHA-pin all actions
- Add `workflow_dispatch` trigger with optional `releaseSha` for manual recovery
- Recovery guidance step on failure

## Why this design

- **Pre-flight over rollback**: verifying OIDC config before publish is simpler and safer than trying to unpublish or retag after a partial failure
- **All-or-nothing**: if any package is misconfigured, no packages publish — correct because changesets versions packages together with cross-dependencies
- **Warn on stale dist-tags**: packages already on npm with wrong `latest` get a warning with a manual fix command, rather than blocking unrelated releases
- **changesets/action for releases**: it detects published packages from stdout, pushes git tags, and creates GitHub releases — replaces the silently broken shell approach

## Known tradeoffs

- New public packages require npm trusted publishing configuration before the version PR merges; misconfigured packages block the entire release (by design)
- No automatic retry on transient npm failures — rerun the workflow or use `workflow_dispatch` with the failed `releaseSha`
- Stale dist-tag warnings require manual `npm dist-tag add`

## How to recover / test

- **Failed publish**: rerun the workflow run, or trigger `workflow_dispatch` with `releaseSha=<sha>`
- **New package not configured**: the OIDC verification step lists failing packages with a link to npm trusted publishing docs
- **Stale dist-tag**: follow the `npm dist-tag add` command in the workflow warning output

### DIFF
--- a/.github/workflows/changeset.yaml
+++ b/.github/workflows/changeset.yaml
@@ -11,7 +11,7 @@ permissions:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
+  cancel-in-progress: false
 
 env:
   CI: true
@@ -22,15 +22,15 @@ jobs:
     runs-on: blacksmith-4vcpu-ubuntu-2404
     steps:
       - name: Checkout code repository
-        uses: actions/checkout@v6
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
         with:
-          fetch-depth: 1
+          fetch-depth: 0
 
       - name: Install pnpm
-        uses: pnpm/action-setup@v4
+        uses: pnpm/action-setup@41ff72655975bd51cab0327fa583b6e92b6d3061 # v4
 
       - name: Setup node.js
-        uses: actions/setup-node@v6
+        uses: actions/setup-node@6044e13b5dc448c55e2357c09f80417699197238 # v6
         with:
           node-version: 24
           cache: "pnpm"
@@ -40,7 +40,7 @@ jobs:
 
       - name: Create version PR
         id: changesets
-        uses: changesets/action@v1
+        uses: changesets/action@6a0a831ff30acef54f2c6aa1cbbc1096b066edaf # v1.7.0
         with:
           commit: "chore: update versions"
           title: "chore: update versions"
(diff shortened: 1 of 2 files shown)


.github/workflows/changeset.yaml
```diff
@@ -11,7 +11,7 @@ permissions:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
```
Yonom, line 14: why?

PR #3486, opened 2026-03-04
Copy templates into monorepo

JustAnOkapi:
## Summary

Copies the 6 starter templates (default, minimal, cloud, cloud-clerk, langgraph, mcp) into the monorepo and adapts them for monorepo development.

### Changes

**Monorepo integration**
- Added `templates/*` to `pnpm-workspace.yaml`
- Converted all `@assistant-ui/*` deps to `workspace:*`
- Removed stale `pnpm-lock.yaml` files from each template

**Linter/formatter cleanup**
- Removed Prettier config + deps from cloud, cloud-clerk, langgraph, mcp (monorepo uses Biome)
- Removed unused ESLint deps from same 4 templates (no eslint config files existed)
- Removed `packageManager` field from default template

**Bug fixes**
- Fixed `AssistantCloud` being recreated on every render in cloud-clerk template (wrapped in `useMemo`)
- Fixed `.gitignore` `.env*` pattern blocking `.env.example` from being tracked (added `!.env.example`)
- Added missing `.env.example` for minimal template
- Removed debug `console.log` that leaked API key to stdout in langgraph proxy route


biome.json
```diff
@@ -184,7 +184,11 @@
       }
     },
     {
-      "includes": ["packages/ui/src/components/ui/**"],
+      "includes": [
```
Yonom, line 187: Should probably update the ui files in the future to be compliant here

PR #3556, opened 2026-03-09
fix(core): add typeof process guard in useCloudThreadListAdapter

### LINKED ISSUE #3555: Client-side runtime depends on process.env
The changes at https://github.com/assistant-ui/assistant-ui/pull/3534 have introduced a subtle but important change.

The [original in packages/react](https://github.com/assistant-ui/assistant-ui/blob/167ece2d8699d58c2601c642bb4cc90000a705c8/packages/react/src/legacy-runtime/runtime-cores/remote-thread-list/adapter/cloud.tsx#L30-L31) (before the refactor) read:
```ts
const baseUrl =
  typeof process !== "undefined" &&
  process?.env?.["NEXT_PUBLIC_ASSISTANT_BASE_URL"];
```

The new [broken copy in packages/core](https://github.com/assistant-ui/assistant-ui/blob/88081edbcd55f48be8f5307fc6405ee4d855040c/packages/core/src/react/runtimes/cloud/useCloudThreadListAdapter.tsx#L29) (after commit 88081ed):
```ts
declare const process: { env: Record<string, string | undefined> };
// ...
const baseUrl = process?.env?.["NEXT_PUBLIC_ASSISTANT_BASE_URL"];
```

The `typeof process !== "undefined"` guard was replaced with a TypeScript `declare const process` which only satisfies the compiler, but it crashes on client runtimes.

julen:
The refactor in #3534 replaced the runtime `typeof process !== "undefined"` guard with a TypeScript `declare const process`, which only satisfies the compiler but crashes in browser environments where `process` is undeclared.

Restores the original guard from packages/react.

Fixes #3555.

### DIFF
--- a/.changeset/fix-process-env-guard.md
+++ b/.changeset/fix-process-env-guard.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/core": patch
+---
+
+fix: restore `typeof process` runtime guard in useCloudThreadListAdapter
--- a/packages/core/src/react/runtimes/cloud/useCloudThreadListAdapter.tsx
+++ b/packages/core/src/react/runtimes/cloud/useCloudThreadListAdapter.tsx
@@ -26,7 +26,9 @@ type CloudThreadListAdapterOptions = {
   delete?: ((threadId: string) => Promise<void>) | undefined;
 };
 
-const baseUrl = process?.env?.["NEXT_PUBLIC_ASSISTANT_BASE_URL"];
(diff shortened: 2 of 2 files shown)


Yonom:
Thank you!

julen:
@Yonom any chance this could be released soon? 🙏

JustAnOkapi:
tomorrow morning @julen

julen:
Thanks!

PR #3572, opened 2026-03-10
fix(react-langgraph): set thread.isLoading during load handler in use…

### LINKED ISSUE #3571: useLangGraphRuntime fails to set "thread.isLoading" to "true "when "load" method is calling
# `thread.isLoading` is never `true` during `load` handler execution in `useLangGraphRuntime`

## Summary

When using `useLangGraphRuntime` with the `load` option (for loading thread history on thread switch), `thread.isLoading` (accessible via `useAuiState(s => s.thread.isLoading)`) remains `false` for the entire duration of the `load` call. This makes it impossible to show a loading state while thread messages are being fetched.

## Environment

- `@assistant-ui/react`: `0.12.x`
- `@assistant-ui/react-langgraph`: `0.13.3`
- React: `18.x`

## Steps to Reproduce

1. Set up `useLangGraphRuntime` with a `load` handler that performs an async fetch:

```tsx
const runtime = useLangGraphRuntime({
  stream: streamHandler,
  load: async (threadId) => {
    const messages = await fetchThreadMessages(threadId); // async API call
    return { messages };
  },
});
```

2. Subscribe to `thread.isLoading` anywhere in the component tree:

```tsx
const isThreadLoading = useAuiState((s) => s.thread.isLoading);
```

3. Trigger a thread switch (e.g. via `api.threads().switchToThread(remoteId)`).

4. Observe that `isThreadLoading` remains `false` throughout the entire `load` execution, even while the async fetch is in progress.

## Expected Behavior

`thread.isLoading` should be `true` from the moment the `load` handler is invoked until it resolves, so that consuming components can render an appropriate loading UI.

## Actual Behavior

`thread.isLoading` is always `false`. It never transitions to `true` during the `load` call.

## Root Cause Analysis

The issue has two compounding causes:

### 1. `useLangGraphRuntimeImpl` never passes `isLoading` to `useExternalStoreRuntime`

Inside `useLangGraphRuntimeImpl`, the `ExternalStoreRuntime` is created with only `isRunning` (which tracks message streaming), but no `isLoading`:

```ts
// useLangGraphRuntime.ts
  isRunning: effectiveIsRunning, // ✅ tracks message streaming
  // ❌ isLoading is never set — no mechanism to signal "loading thread history"
  messages: threadMessages,
  ...
});
```

`ExternalStoreThreadRuntimeCore.isLoading` reads directly from `this._store.isLoading ?? false`. Since `isLoading` is never written to the store, it permanently resolves to `false`.

```ts
// @assistant-ui/core — external-store-thread-runtime-core.ts
public get isLoading() {
  return this._store.isLoading ?? false; // always false
}
```

### 2. The `load` handler is invoked inside a `useEffect` with no prior loading state

```ts
// useLangGraphRuntime.ts
useEffect(() => {
  const load = loadRef.current;
  if (!load) return;

  const externalId = aui.threadListItem().getState().externalId;
  if (externalId == null) return;

  // load() is called here, after render, with no `isLoading: true` set beforehand
  load(externalId).then(({ messages, interrupts }) => {
    setMessages(messages);
    setInterrupt(interrupts?.[0]);
  });
}, [aui, setMessages, setInterrupt]);
```

Even if a consumer tried to derive loading state externally, there is no notification boundary — the `load` call starts silently after the first render of a thread's runtime hook instance, with no corresponding state transition.

## Workaround

Currently, the only workaround is to maintain a separate loading flag outside the assistant-ui state (e.g. via a custom Zustand store or context), set it manually at the beginning of the `load` handler, and clear it at the end:

```ts
load: async (threadId) => {
  myStore.setThreadDetailLoading(true); // manual workaround
  try {
    const messages = await fetchThreadMessages(threadId);
    return { messages };
  } finally {
    myStore.setThreadDetailLoading(false);
  }
}
```

This is fragile and forces consumers to maintain state that should be owned by the runtime.

## Proposed Fix

`useLangGraphRuntimeImpl` should track the loading state of the `load` call and propagate it to `useExternalStoreRuntime`:

```ts
// useLangGraphRuntime.ts (proposed change)
const [isLoadingThread, setIsLoadingThread] = useState(false);

useEffect(() => {
  const load = loadRef.current;
  if (!load) return;

  const externalId = aui.threadListItem().getState().externalId;
  if (externalId == null) return;

  setIsLoadingThread(true); // signal loading start
  load(externalId)
    .then(({ messages, interrupts }) => {
      setMessages(messages);
      setInterrupt(interrupts?.[0]);
    })
    .finally(() => {
      setIsLoadingThread(false); // signal loading end
    });
}, [aui, setMessages, setInterrupt]);

  isRunning: effectiveIsRunning,
  isLoading: isLoadingThread, // ✅ propagate to store
  messages: threadMessages,
  ...
});
```

This would make `useAuiState(s => s.thread.isLoading)` behave correctly and allow consumers to declaratively show loading UI during thread switches without any workarounds.

## Additional Context

`ExternalStoreAdapterBase` already defines `isLoading` as an optional field:

```ts
// @assistant-ui/core — external-store-adapter.ts
type ExternalStoreAdapterBase<T> = {
  isDisabled?: boolean | undefined;
  isRunning?: boolean | undefined;
  isLoading?: boolean | undefined; // ✅ field exists, just never populated
  ...
};
```

The plumbing is in place — it just needs to be wired up in `useLangGraphRuntime`.

ShobhitPatra:
closes https://github.com/assistant-ui/assistant-ui/issues/3571


Malcolm-Ma:
Hi there, is there any update of the PR, like when will be merged and published? Waiting for this feature to resolve our bugs. 

Thanks.

PR #3584, opened 2026-03-10
fix: normalize attachment-only sends for langgraph

### LINKED ISSUE #3577: Bug: Sending attachment with empty text triggers two `stream()` calls and renders two user messages
**Package:** `@assistant-ui/core` / `@assistant-ui/react-langgraph`  
**Version:** `@assistant-ui/react` `0.12.15` · `@assistant-ui/react-langgraph` `0.13.3`

---

## Description

When a user submits the composer with **no text and one or more file attachments**, two separate HTTP requests are sent to the LangGraph backend instead of one, and two user message bubbles are rendered in the thread UI.

The first request contains only the file content (no text block). The second request contains an empty human message. This causes downstream failures with models like **AWS Bedrock** that require a text block to accompany any document/file block:

```
ValidationException: A text block must be included when using documents.
Add a text block and retry your request.
```

---

## Steps to Reproduce

1. Use `useLangGraphRuntime` with a `CompositeAttachmentAdapter` (or any adapter that returns `content: [{ type: "file", ... }]`).
2. Open the composer, attach a file (e.g. PDF), and leave the text input **empty**.
3. Press Send.

**Expected:** One HTTP request with a single human message containing both a text block and the file block.  
**Actual:** Two HTTP requests fired, two user bubbles rendered.

---

### 1. `base-composer-runtime-core.js` — empty `content` when text is blank

Inside the `send()` method of `BaseComposerRuntimeCore`:

```js
// @assistant-ui/core — base-composer-runtime-core.js
const message = {
  content: text ? [{ type: "text", text }] : [],  // ← [] when text is empty
  attachments: await attachments,                  // ← file lives here
  ...
};
this.handleSend(message);
```

When `text` is empty, `content` is set to `[]`. The file attachment is resolved separately in the `attachments` field. This means the internal message object has `content: []` and `attachments: [{ content: [{ type: "file", ... }] }]`.

### 2. `useLangGraphRuntime.js` — `getMessageContent` triggers a stream call with content-less first message

The LangGraph runtime's `onNew` handler calls `getMessageContent(msg)`:

```js
// @assistant-ui/react-langgraph — useLangGraphRuntime.js
const getMessageContent = (msg) => {
  const allContent = [
    ...msg.content,                                         // [] — empty
    ...(msg.attachments?.flatMap((a) => a.content) ?? [])  // [{ type: "file", ... }]
  ];
  // ...
};
```

The adapter's `send()` (which reads the file as base64) is **async**. Because of the timing between when the text message is dispatched and when the attachment promise resolves, two separate `onNew` / `stream()` invocations are triggered:

- **Call 1:** `messages: [{ type: "human", content: [], id: "..." }]` — the empty text message fires immediately
- **Call 2:** `messages: [{ type: "human", content: [{ type: "file", ... }], id: "..." }]` — fires once the attachment resolves

Both calls result in separate user message bubbles in the thread.

---

## Expected Behavior

When text is empty and attachments are present, the library should either:

- **Option A:** Automatically include a minimal text block (e.g. `{ type: "text", text: " " }`) so the message is well-formed for document-capable models.
- **Option B:** Wait for all attachment promises to resolve before calling `handleSend`, so only one `onNew` fires with the complete content merged.

---

## Workaround

Until this is fixed upstream, intercept the composer's `send` at the UI level using `useComposerRuntime`:

```tsx
import { useComposerRuntime, useComposer } from '@assistant-ui/react'

const ComposerSendButton = () => {
  const composerRuntime = useComposerRuntime()
  const text = useComposer((s) => s.text)
  const attachmentCount = useComposer((s) => s.attachments.length)

  const handleSend = () => {
    // Inject a space so content is never empty when attachments are present.
    // Without this, assistant-ui sends two separate messages (empty text + file),
    // causing two stream() calls and two user bubbles in the UI.
    if (text.trim() === '' && attachmentCount > 0) {
      composerRuntime.setText(' ')
    }
    composerRuntime.send()
  }

  return <button onClick={handleSend}>Send</button>
}
```

And as a server-side defence-in-depth, normalise the messages array before sending to the backend:

```ts
// In your useLangGraphRuntime stream() callback:
stream: async (messages, config) => {
  const normalizedMessages = normalizeMessagesForBedrock(messages)
  return client.runs.stream(threadId, ASSISTANT_ID, {
    input: { messages: normalizedMessages },
    ...
  })
}

function normalizeMessagesForBedrock(messages) {
  // Ensure every human message with file blocks also has a text block
  return messages.map((msg) => {
    const isHuman = msg.type === 'human' || msg.role === 'user'
    if (!isHuman || !Array.isArray(msg.content)) return msg

    const hasFile = msg.content.some((c) => c?.type === 'file')
    const hasText = msg.content.some((c) => c?.type === 'text')

    if (hasFile && !hasText) {
      return { ...msg, content: [{ type: 'text', text: ' ' }, ...msg.content] }
    }
    return msg
  })
}
```

---

## Environment

| | |
|---|---|
| `@assistant-ui/react` | `0.12.15` |
| `@assistant-ui/react-langgraph` | `0.13.3` |
| `@assistant-ui/core` | (bundled with react) |
| LangGraph backend | Python LangGraph + AWS Bedrock (ConverseStream) |
| Attachment type | PDF via custom `AttachmentAdapter` returning `{ type: "file", data: "<base64>", mimeType: "application/pdf" }` |

samdickson22:
Fixes #3577

## Summary
- normalize composer sends so attachment-only messages include a placeholder text part (`" "`)
- add defensive normalization in `useLangGraphRuntime` to prepend placeholder text when file content exists without text
- add regressions for attachment-only composer send and file-only thread append in LangGraph runtime
- add core composer regression for empty-text + attachment send

## Why
Fixes duplicate/invalid attachment-only message behavior and ensures Bedrock-compatible text+file payloads for human messages.

## Validation
- `pnpm test` (full turbo test matrix)
- `pnpm --filter @assistant-ui/react-langgraph test`

### DIFF
--- a/.changeset/fix-normalize-attachment-only-sends--20260310134223.md
+++ b/.changeset/fix-normalize-attachment-only-sends--20260310134223.md
@@ -1,0 +1,8 @@
+---
+"@assistant-ui/core": patch
+"@assistant-ui/react-langgraph": patch
+---
+
+fix: normalize attachment-only sends for langgraph
+
+Attachment-only user sends now normalize to include a minimal text part alongside file content. This prevents duplicate/invalid message payloads and improves compatibility with backends that require text + document blocks (for example, Bedrock Converse).
--- a/packages/core/src/runtime/base/base-composer-runtime-core.ts
+++ b/packages/core/src/runtime/base/base-composer-runtime-core.ts
@@ -173,12 +173,21 @@ export abstract class BaseComposerRuntimeCore
     const quote = this._quote;
     this._quote = undefined;
     this._emptyTextAndAttachments();
+    const resolvedAttachments = await attachments;
+    const hasAttachments = resolvedAttachments.some(
+      (a) => (a.content?.length ?? 0) > 0,
+    );
+    const hasUserText = text.trim().length > 0;
 
     const message: Omit<AppendMessage, "parentId" | "sourceId"> = {
       createdAt: new Date(),
       role: this.role,
-      content: text ? [{ type: "text", text }] : [],
-      attachments: await attachments,
+      content: hasUserText
+        ? [{ type: "text", text }]
+        : hasAttachments
+          ? [{ type: "text", text: " " }]
+          : [],
+      attachments: resolvedAttachments,
       runConfig: this.runConfig,
       metadata: { custom: { ...(quote ? { quote } : {}) } },
     };
--- a/packages/core/src/runtime/base/base-composer-runtime-core.ts
+++ b/packages/core/src/runtime/base/base-composer-runtime-core.ts
@@ -170,25 +170,20 @@ export abstract class BaseComposerRuntimeCore
         : [];
 
     const text = this.text;
+    const createdAt = new Date();
+    const role = this.role;
+    const runConfig = this.runConfig;
     const quote = this._quote;
     this._quote = undefined;
     this._emptyTextAndAttachments();
     const resolvedAttachments = await attachments;
-    const hasAttachments = resolvedAttachments.some(
-      (a) => (a.content?.length ?? 0) > 0,
-    );
-    const hasUserText = text.trim().length > 0;
 
     const message: Omit<AppendMessage, "parentId" | "sourceId"> = {
-      createdAt: new Date(),
-      role: this.role,
-      content: hasUserText
-        ? [{ type: "text", text }]
-        : hasAttachments
-          ? [{ type: "text", text: " " }]
-          : [],
+      createdAt,
+      role,
+      content: text ? [{ type: "text", text }] : [],
       attachments: resolvedAttachments,
-      runConfig: this.runConfig,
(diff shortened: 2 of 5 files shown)


okisdev:
thanks for the work here, @samdickson22.

#3577 has since been closed by #3730, which lands the same approach as the react-langgraph half of this PR (prepend `{ type: "text", text: " " }` to attachment-only human messages inside `getMessageContent`). the second half of the issue (two `stream()` calls) was already fixed earlier by #3549, so there's nothing left for this PR to address.

the core snapshot change isn't needed either: `_emptyTextAndAttachments()` only resets `text` and `attachments`, not `role` or `runConfig`, so reading them back after the call returns the same values. main still doesn't snapshot them.

going to close this out for those reasons. appreciate the work and the regression tests.

PR #3601, opened 2026-03-12
Enhance AG-UI RUN_ERROR event handling with visible error messages in chat responses

weikao:
1. RUN_ERROR Event Handling Enhancement (packages/react-ag-ui/src/runtime/adapter/run-aggregator.ts)
2. Event Type Definition Update (packages/react-ag-ui/src/runtime/types.ts)
3. Event Parser Enhancement (packages/react-ag-ui/src/runtime/event-parser.ts)
4. Subscriber Interface Extension (packages/react-ag-ui/src/runtime/adapter/subscriber.ts)

These modifications enhance the error handling and user feedback mechanism in assistant-ui:
Key Improvements:
Event System Extension: Capture and handle RUN_ERROR events in AG-UI runtime
Error Visualization: Display backend error messages directly in AI response bubbles
Type Safety: Complete event type definitions with support for error messages and codes
Subscriber Pattern: Provide error event listening capability via onRunErrorEvent callback


packages/react-ag-ui/src/runtime/AgUiThreadRuntimeCore.ts
```diff
@@ -272,7 +272,29 @@ export class AgUiThreadRuntimeCore {
...
+      this.messages = messages.map((msg) => {
+        if (
+          msg.role === "assistant" &&
+          (msg as ThreadAssistantMessage).status?.type === "running"
+        ) {
+          return {
+            ...(msg as ThreadAssistantMessage),
+            status: {
+              type: "incomplete" as const,
+              reason: "cancelled" as const,
```
Yonom, line 289: can we set reason here to be error with the error set here?

packages/core/src/runtime/base/default-edit-composer-runtime-core.ts
```diff
@@ -51,15 +51,12 @@ export class DefaultEditComposerRuntimeCore extends BaseComposerRuntimeCore {
   public async handleSend(
     message: Omit<AppendMessage, "parentId" | "sourceId">,
   ) {
-    const text = getThreadMessageText(message as AppendMessage);
-    if (text !== this._previousText) {
```
Yonom, line 55: new code looks correct, but im not sure why this code was here in the first place.. curious, do you know why?

okisdev:
thanks! a few of these have moved on since this PR was opened, so closing as superseded:

- `default-edit-composer-runtime-core.ts`: the text-changed guard was kept on purpose to avoid re-running on no-op edits. it was later extended in #3775 and #3857 to also respect attachment changes and an explicit `startRun` opt-in. removing the guard now would regress that.
- `run-aggregator.ts`: `RUN_ERROR` is already handled by storing the message on `status.error` (`status: { type: "incomplete", reason: "error", error }`) rather than appending to the message text. that avoids the partial-stream concatenation issue cubic/greptile flagged here.
- `subscriber.ts`: `onRunFailed` already dispatches a `RUN_ERROR` event with `message` and `code`, so the extra `onRunErrorEvent` slot is redundant.
- `AgUiThreadRuntimeCore.ts` (race fix in `applyExternalMessages`): the underlying race is real and not yet fixed in main, but as Yonom and the bots noted, hardcoding `reason: "cancelled"` mislabels error runs. the locally-committed status already carries the correct `reason: "error"`, so the fallback should read from that instead of overwriting blindly.

if the snapshot race is still reproducible, a smaller follow-up PR with only that fix (using the correct error reason) would be the cleanest path.

PR #3621, opened 2026-03-13
feat(core) add thread list pagination support

### LINKED ISSUE #3604: [Feature Request] Infinite Load of ThreadsList with LocalRuntime
For a Thread, I can understand that "infinite load" is not a priority: normally, threads should not be "enormous", and it is a bit complex to manage branching when you do not have all the messages yet (but it is completely possible; I did another chat interface from scratch, and I implemented this feature)

Bur for the ThreadList, it is by nature supposed to grow indefinitely.

I therefore tried to implement an infinite load, but the only way I found implied to access a "_core" feature not officially exposed, to reset the "_loadThreadsPromise"

Could it be possible to:
1. At least expose an official way to update the ThreadList
2. Ideally add a Primitive (the "LoadSentinel") and callback ("loadMore()") to officialy support the feature without the developper having to manage by himself implementation details?

For now, I use the following code to achieve the feature. _Note that the current version of the API I use does not really support "pagination", so it is a bit dumb, but of course the same idea could be adapted for a true infinite load API_

ThreadsListPagination Provider (with access to internal _core; btw, is this hack still feasible with useAui()?)

```tsx
import {
  createContext,
  useCallback,
  useContext,
  useRef,
  useState,
  type MutableRefObject,
  type ReactNode,
} from "react"
import { useAssistantRuntime } from "@assistant-ui/react"

const PAGE_SIZE = 20

type ThreadListPaginationContextValue = {
  loadMore: () => Promise<void>
  hasMore: boolean
  isLoadingMore: boolean
}

const ThreadListPaginationContext =
  createContext<ThreadListPaginationContextValue>({
    loadMore: async () => {},
    hasMore: true,
    isLoadingMore: false,
  })

export const useThreadListPagination = () =>
  useContext(ThreadListPaginationContext)

/**
 * Provider that wires infinite-load pagination into the AssistantUI runtime.
 *
 * Because AUI caches the result of `adapter.list()` in a private
 * `_loadThreadsPromise`, the only way to append more threads is to:
 *   1. bump the limit ref that `list()` reads,
 *   2. reset the cached promise on the runtime core, and
 *   3. await a fresh `getLoadThreadsPromise()` which re-calls `list()`.
 *
 * This is admittedly coupling to an implementation detail of AUI,
 * but the library currently offers no public pagination API.
 */
export function ThreadListPaginationProvider({
  limitRef,
  hasMoreRef,
  children,
}: {
  limitRef: RefObject<number>
  hasMoreRef: RefObject<boolean>
  children: ReactNode
}) {
  const runtime = useAssistantRuntime()
  const [isLoadingMore, setIsLoadingMore] = useState(false)
  const [hasMore, setHasMore] = useState(true)
  const lockRef = useRef(false)

  const loadMore = useCallback(async () => {
    if (lockRef.current || !hasMoreRef.current) return
    lockRef.current = true
    setIsLoadingMore(true)

    try {
      limitRef.current += PAGE_SIZE

      // Access AUI internals to force a re-fetch
      // eslint-disable-next-line @typescript-eslint/no-explicit-any
      const core = (runtime as any)._core
      const threadListCore = core.threads
      threadListCore._loadThreadsPromise = undefined
      await threadListCore.getLoadThreadsPromise()

      setHasMore(hasMoreRef.current)
    } finally {
      lockRef.current = false
      setIsLoadingMore(false)
    }
  }, [runtime, limitRef, hasMoreRef])

  return (
    <ThreadListPaginationContext.Provider
      value={{ loadMore, hasMore, isLoadingMore }}
    >
      {children}
    </ThreadListPaginationContext.Provider>
  )
}

export { PAGE_SIZE }
```

Runtime-Provider 

```tsx

/**
 * Per-thread runtime hook.
 *
 * Called by the RemoteThreadListRuntime for each thread instance.
 * It sets up the LocalRuntime with our ChatModelAdapter (OpenAI streaming)
 * and HistoryAdapter (Memory API persistence).
 */
function ThreadRuntimeHook() {
  const memoryApi = useMemoryApi()
  const aui = useAui()

  const chatModelAdapter = useChatModelAdapter()
  const historyAdapter = useMemoryApiHistoryAdapter(memoryApi, aui)

  // useLocalRuntime detects it's nested inside useRemoteThreadListRuntime
  // and becomes just useLocalThreadRuntime (no-op on the thread list layer)
  return useLocalRuntime(chatModelAdapter, {
    adapters: { history: historyAdapter },
  })
}

export function RuntimeProvider({ children }: { children: ReactNode }) {
  const { adapter, limitRef, hasMoreRef } = useMemoryApiThreadListAdapter()

  const runtime = unstable_useRemoteThreadListRuntime({
    runtimeHook: ThreadRuntimeHook,
    adapter,
  })

  return (
      <ThreadListPaginationProvider
        limitRef={limitRef}
        hasMoreRef={hasMoreRef}
      >
        {children}
      </ThreadListPaginationProvider>
    </AssistantRuntimeProvider>
  )
}
```

Remote Threads List adapter:

```tsx
/**
 * RemoteThreadListAdapter backed by the Memory API.
 *
 * Handles thread CRUD, title generation, and integrates with the
 * `unstable_useRemoteThreadListRuntime` from assistant-ui.
 *
 * The `list()` method uses a dynamic `limitRef` so the pagination
 * provider can progressively increase the number of threads fetched.
 */
export function useMemoryApiThreadListAdapter() {
  // Manage models
  const { data: models } = useModels()
  const modelsRef = useRef<LLMModel[] | undefined>(models)
  modelsRef.current = models
  // Get Memory API client
  const memoryApi = useMemoryApi()
  // Navigate after thread creation
  const navigate = useNavigate()

  // Pagination refs — mutated by ThreadListPaginationProvider
  const limitRef = useRef(PAGE_SIZE)
  const hasMoreRef = useRef(true)

  const adapter = useMemo<RemoteThreadListAdapter>(() => {
    console.log("[MemoryApiThreadListAdapter] Creation")
    return {
      async list() {
        const threads = await memoryApi.threads.list({
          limit: limitRef.current,
        })
        hasMoreRef.current = threads.length >= limitRef.current
        return {
          threads: threads.map((t) => ({
            remoteId: t.id,
            title: t.name ?? undefined,
            status: "regular" as const,
          })),
        }
      },

      async initialize(_threadId: string) {
        const thread = await memoryApi.threads.create({})
        navigate(`/chat/${thread.id}`)
        return { remoteId: thread.id, externalId: undefined }
      },

      async rename(remoteId: string, newTitle: string) {
        await memoryApi.threads.update(remoteId, { name: newTitle })
      },

      async archive(_remoteId: string) {
        // Not supported by Memory API yet
      },

      async unarchive(_remoteId: string) {
        // Not supported by Memory API yet
      },

      async delete(remoteId: string) {
        await memoryApi.threads.delete(remoteId)
      },

      async generateTitle(remoteId, messages) {
        return createAssistantStream(async (controller) => {
          const title = "plop"; // I spare you the code to generate the title ^^ 
          controller.appendText(title)
          controller.close()
        })
      },

      async fetch(threadId: string) {
        const thread = await memoryApi.threads.get(threadId)
        if (!thread) throw new Error("Thread not found")
        return {
          remoteId: thread.id,
          title: thread.name ?? undefined,
          status: "regular" as const,
        }
      },
    }
  }, [memoryApi, navigate])

  return { adapter, limitRef, hasMoreRef }
}
```

thread-list.tsx

```tsx

/** Sentinel at the bottom of the list — triggers loadMore when it becomes visible. */
const LoadMoreSentinel: FC = () => {
  const { loadMore, hasMore, isLoadingMore } = useThreadListPagination()
  const sentinelRef = useRef<HTMLDivElement>(null)

  useEffect(() => {
    const el = sentinelRef.current
    if (!el || !hasMore) return

    const observer = new IntersectionObserver(
      ([entry]) => {
        if (entry?.isIntersecting) {
          loadMore()
        }
      },
      { threshold: 0 },
    )

    observer.observe(el)
    return () => observer.disconnect()
  }, [loadMore, hasMore])

  if (!hasMore) return null

  return (
    <div ref={sentinelRef} className="flex h-9 items-center justify-center">
      {isLoadingMore && (
        <Skeleton className="aui-thread-list-skeleton h-4 w-24" />
      )}
    </div>
  )
}

export const ThreadList = () => {
  return (
    <ThreadListPrimitive.Root>
      {/* Initial full-page skeleton (no threads loaded yet) */}
      <AuiIf
        condition={({ threads }) =>
          threads.isLoading && threads.threadIds.length === 0
        }
      >
        <ThreadListSkeleton />
      </AuiIf>

      {/* Thread items — always visible once we have at least one thread */}
      <AuiIf condition={({ threads }) => threads.threadIds.length > 0}>
        <ThreadListPrimitive.Items components={{ ThreadListItem }} />
        <LoadMoreSentinel />
      </AuiIf>
    </ThreadListPrimitive.Root>
  )
}
```

ShobhitPatra:
closes #3604

### DIFF
--- a/.changeset/eighty-canyons-roll.md
+++ b/.changeset/eighty-canyons-roll.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/core": patch
+---
+
+feat(core): add thread list pagination support  
--- a/apps/docs/content/docs/(reference)/api-reference/runtimes/thread-list-runtime.mdx
+++ b/apps/docs/content/docs/(reference)/api-reference/runtimes/thread-list-runtime.mdx
@@ -15,6 +15,7 @@ import { useAui } from "@assistant-ui/react";
 const aui = useAui();
 aui.threads().switchToNewThread();
 aui.threads().switchToThread(threadId);
+aui.threads().loadMore();
 ```
 
 <ParametersTable {...ThreadListRuntime} />
@@ -28,6 +29,7 @@ import { useAuiState } from "@assistant-ui/react";
 
 const threads = useAuiState((s) => s.threads.threadIds);
 const isLoading = useAuiState((s) => s.threads.isLoading);
+const canLoadMore = useAuiState((s) => s.threads.canLoadMore);
 ```
 
 <ParametersTable {...ThreadListState} />
--- a/apps/docs/content/docs/runtimes/custom/custom-thread-list.mdx
+++ b/apps/docs/content/docs/runtimes/custom/custom-thread-list.mdx
@@ -70,9 +70,11 @@ When the hook mounts it calls `list()` on your adapter, hydrates existing thread
     import { myModelAdapter } from "./model-adapter"; // your chat model adapter
 
     const threadListAdapter: RemoteThreadListAdapter = {
-      async list() {
-        const response = await fetch("/api/threads");
-        const threads = await response.json();
+      async list(params) {
+        const url = new URL("/api/threads", window.location.origin);
+        if (params?.cursor) url.searchParams.set("cursor", params.cursor);
+        const response = await fetch(url);
+        const { threads, cursor } = await response.json();
         return {
           threads: threads.map((thread: any) => ({
             remoteId: thread.id,
@@ -80,6 +82,7 @@ When the hook mounts it calls `list()` on your adapter, hydrates existing thread
             status: thread.is_archived ? "archived" : "regular",
             title: thread.title ?? undefined,
           })),
+          cursor,
         };
       },
       async initialize(localId) {
@@ -155,9 +158,9 @@ When the hook mounts it calls `list()` on your adapter, hydrates existing thread
   parameters={[
     {
       name: "list",
-      type: "() => Promise<{ threads: RemoteThreadMetadata[] }>",
+      type: "(params?: { cursor?: string }) => Promise<{ threads: RemoteThreadMetadata[]; cursor?: string }>",
       description:
-        "Return the current threads. Each thread must include status, remoteId, and any metadata you want to show immediately.",
+        "Return threads for the current page. If your backend supports pagination, pass the cursor from the previous response. Return a cursor to indicate more pages are available.",
       required: true,
     },
     {
@@ -209,11 +212,53 @@ When the hook mounts it calls `list()` on your adapter, hydrates existing thread
 
 ## Thread Lifecycle Cheatsheet
 
-- `list()` hydrates threads on mount and during refreshes.
+- `list()` hydrates threads on mount and during refreshes. If it returns a `cursor`, the runtime enables pagination.
+- `loadMore()` fetches the next page by calling `list({ cursor })` and appending the results. Duplicate threads are automatically skipped.
 - Creating a new conversation calls `initialize()` once the user sends the first message.
 - `archive`, `unarchive`, and `delete` are called optimistically; throw to revert the UI.
 - `generateTitle()` powers the automatic title button and expects an `AssistantStream`.
 - Provide a `runtimeHook` that always returns a fresh runtime instance per active thread.
+
+## Pagination
+
+If your thread list can grow large, implement cursor-based pagination in your `list()` method. Return a `cursor` string alongside threads to indicate more pages are available.
+
+```tsx
+const threadListAdapter: RemoteThreadListAdapter = {
+  async list(params) {
+    const response = await fetch(
+      `/api/threads?cursor=${params?.cursor ?? ""}`,
+    );
+    const { threads, nextCursor } = await response.json();
+    return {
+      threads: threads.map((t: any) => ({
+        remoteId: t.id,
+        status: t.is_archived ? "archived" : "regular",
+        title: t.title,
+      })),
+      cursor: nextCursor, // undefined when no more pages
+    };
+  },
+  // ...other methods
+};
+```
+
+On the UI side, use `canLoadMore` and `loadMore()` to build infinite scroll or a "load more" button:
+
+```tsx
+import { useAui, useAuiState } from "@assistant-ui/react";
+
+function LoadMoreButton() {
+  const aui = useAui();
+  const canLoadMore = useAuiState((s) => s.threads.canLoadMore);
+
+  if (!canLoadMore) return null;
+
+  return (
+    <button onClick={() => aui.threads().loadMore()}>Load more</button>
+  );
+}
+```
 
 ## Avoiding Race Conditions in History Adapters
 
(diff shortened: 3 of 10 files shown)


okisdev:
@ShobhitPatra thanks for the work here. closing in favour of #3953, which lands cursor pagination for the same issue (#3604) with a different shape. logging the divergence so the choice is on the record:

**design choices**
- adapter signature: `list(params?: { after?: string })` returning `nextCursor`. `after` disambiguates request cursor from response cursor; `nextCursor` matches Stripe / GitHub conventions.
- `loadMore?()`, `hasMore?`, `isLoadingMore?` are optional on `ThreadListRuntimeCore` (sibling to the existing `reload?()`), so `LocalThreadListRuntimeCore` and `ExternalStoreThreadListRuntimeCore` need no changes. this PR makes them required and adds `canLoadMore = false; loadMore() = RESOLVED_PROMISE` boilerplate to every non-paginating core.
- separate `isLoadingMore` state instead of reusing `isLoading`, so the load-more spinner does not fight the initial skeleton.
- tap layer wired through (`aui.threads().loadMore()`, `s.threads.hasMore`, `s.threads.isLoadingMore`); `@assistant-ui/react` ships `ThreadListPrimitive.LoadMore` plus `useThreadListLoadMore` so consumers do not roll their own button or sentinel hook.

**review findings**
the Cubic / Claude bot review here flagged five P1/P2 issues that are still unresolved. #3953 addresses each one structurally:

| issue here | how #3953 handles it |
|---|---|
| cursor leaks across adapter swap (P1) | `__internal_setOptions` clears the cursor; `loadMore()` captures the active adapter so a swap mid-flight cannot leak a stale page |
| `threadIdMap` / `threadData` retain stale entries on fresh load (P1) | fresh-load reducer rebuilds the maps cleanly; `loadMore` appends with O(1) dedup against the existing map |
| cursor advances when `_applyListResponse` throws (P1/P2) | cursor and data are committed atomically inside the `OptimisticState` reducer, so a throwing reducer leaves the cursor untouched |
| missing dedup when `append=true` and a thread shifts between pages (P2) | dedup against `threadIdMap` inside the append reducer |
| truthiness vs undefined cursor check (P2) | `nextCursor \|\| undefined` normalises empty string to "no more pages" |

**coverage**
- 15 tests covering dedup, reload-while-loadMore race, adapter swap, retry after rejection, empty cursor, and reducer throw
- docs across `concepts/threads.mdx`, `primitives/thread-list.mdx` (with an `IntersectionObserver` sentinel sketch), `api-reference/*`, and `ui/thread-list.mdx`
- `examples/with-custom-thread-list` demos pagination end-to-end with 25 pre-seeded threads paged 10 at a time

PR #3622, opened 2026-03-13
feat(react-ink): add direct named exports for Bun compile compatibility

cewinharhar:
## Summary

- Add direct named exports for all 11 primitive groups in `@assistant-ui/react-ink`, alongside the existing `export * as XPrimitive` namespace exports
- Fixes symbol elimination when consuming the package via `bun build --compile`, where the tree-shaker cannot trace dynamic property access (`MessagePrimitive.Root`) through namespace re-export chains

## Problem

`@assistant-ui/react-ink` only exports namespace objects:

```ts
export * as MessagePrimitive from "./primitives/message"
```

When a consumer accesses `MessagePrimitive.Root`, Bun's compile-time tree-shaker sees a dynamic property access on a re-exported namespace and drops the symbol. This causes `MessageRoot is not defined` at runtime in any CLI tool built with `bun build --compile`.

This affects the exact target audience for a terminal UI library.

## Solution

Add direct named exports from each primitive barrel, re-aliasing short names back to their full prefixed form:

```ts
// Namespace (existing, unchanged)
export * as MessagePrimitive from "./primitives/message";

// Named (new, bundler-safe)
export {
  Root as MessageRoot,
  Content as MessageContent,
  // ...
} from "./primitives/message";
```

Both import patterns now work:

```ts
import { MessagePrimitive } from "@assistant-ui/react-ink"   // namespace
import { MessageRoot } from "@assistant-ui/react-ink"         // named (new)
```

All named exports re-export from the existing barrel files (no deep imports), so the barrel remains the single source of truth.

## Scope

- **11 primitive groups**: Thread, Message, Composer, ThreadList, ActionBar, BranchPicker, Attachment, ThreadListItem, ChainOfThought, Suggestion, ToolCall
- **All value and type exports** included (components + props types)
- **Zero breaking changes** — purely additive, no existing exports removed or renamed

## Verification

- `pnpm build --filter @assistant-ui/react-ink` passes
- `vitest run` — 2/2 tests pass
- Both `MessagePrimitive` (namespace) and `MessageRoot` (named) resolve in dist output
- Changeset included (patch bump)

## Context

We hit this building [Lobster AI](https://github.com/the-omics-os/lobster)'s terminal interface with `bun build --compile` + `@assistant-ui/react-ink`.

### DIFF
--- a/.changeset/react-ink-named-exports.md
+++ b/.changeset/react-ink-named-exports.md
@@ -1,0 +1,15 @@
+---
+"@assistant-ui/react-ink": patch
+---
+
+Add direct named exports for all primitives alongside existing namespace exports.
+
+Bundlers that use static analysis for tree-shaking (notably Bun `--compile`) cannot trace dynamic property access through `export * as Namespace` re-export chains. This causes symbols like `MessagePrimitive.Root` to be dropped at compile time.
+
+Both import patterns now work:
+```ts
+import { MessagePrimitive } from "@assistant-ui/react-ink"   // namespace (existing)
+import { MessageRoot } from "@assistant-ui/react-ink"         // named (new)
+```
+
+Zero breaking changes — purely additive.
--- a/packages/react-ink/src/index.ts
+++ b/packages/react-ink/src/index.ts
@@ -101,7 +101,7 @@ export {
   useRemoteThreadListRuntime,
 } from "./runtimes";
 
-// Primitives
+// Primitives — namespace exports (existing API)
 export * as ThreadPrimitive from "./primitives/thread";
 export * as ComposerPrimitive from "./primitives/composer";
 export * as MessagePrimitive from "./primitives/message";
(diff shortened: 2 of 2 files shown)


okisdev:
hey @cewinharhar, thanks for the detailed writeup and for digging into the Bun --compile behaviour. unfortunately going to push back on landing this upstream.

the diff itself is clean, but it doubles the public API of @assistant-ui/react-ink (every primitive gains a parallel long-name export) to work around a static-analysis quirk in `bun build --compile`. a few concerns with that as a framework-level fix:

- it's asymmetric. @assistant-ui/react and @assistant-ui/react-native use the same `export * as Namespace` shape and would, by the same argument, need the same parallel exports. either we double the surface across all three distribution packages, or react-ink ships an export pattern the others don't.
- once shipped, the named exports become a public contract. even if Bun's analyzer learns to trace namespace re-exports later (which is the right fix and where this really belongs), we can't remove them without a breaking change.
- the consumer can resolve this locally without us shipping anything. a short re-export file at module-eval time gives Bun direct value bindings to trace:

  ```ts
  // your-app/aui-shims.ts
  import { MessagePrimitive, ThreadPrimitive } from "@assistant-ui/react-ink";
  export const MessageRoot = MessagePrimitive.Root;
  export const MessageContent = MessagePrimitive.Content;
  export const ThreadRoot = ThreadPrimitive.Root;
  // ...one line per symbol you actually use
  ```

  that decays cleanly the day Bun fixes the analyser, and keeps the workaround scoped to the one place that needs it.

going to close this. for Lobster specifically, the shim file above should unblock you with no framework changes. if Bun ships an analyser fix or this pattern hits a wall you can't shim around, happy to revisit. appreciate the thorough investigation either way.

PR #3633, opened 2026-03-14
feat(react-ink): add LoadingPrimitive

ShobhitPatra:


packages/react-ink/src/primitives/loading/LoadingElapsedTime.tsx
```diff
@@ -0,0 +1,68 @@
...
+  return `${minutes}m ${remainingSeconds}s`;
+};
+
+export type LoadingElapsedTimeProps = {
+  color?: ComponentProps<typeof Text>["color"];
+  format?: (seconds: number) => string;
+};
+
+export const LoadingElapsedTime = ({
+  color = "gray",
```
okisdev: we like to unstyled components, so the primitives should never have any base (default) styles.

okisdev (approved):

okisdev (approved):

Issue #3644, opened 2026-03-16
The webpage at https://www.assistant-ui.com/ won't open


vci-123:
The webpage at https://www.assistant-ui.com/ won't open.

Application error: a client-side exception has occurred while loading www.assistant-ui.com (see the browser console for more information).

lucasverra:
same

CHlNlTO:
same

masao-cc:
same

okisdev:
thanks for ur report, and we are sorry about the interruption.

this issue has now been fixed. happy coding!

PR #3645, opened 2026-03-16
fix(store): stabilize AuiForEach keys selector to prevent infinite re…

ShobhitPatra:
closes https://github.com/assistant-ui/assistant-ui/issues/3644


ShobhitPatra:
closing in favour of https://github.com/assistant-ui/assistant-ui/pull/3647 which includes the same core fix plus additional optimizations

PR #3646, opened 2026-03-16
feat: add localStorage persistence, thread list reload, and thread metadata

samdickson22:
## Summary

- Wire `createLocalStorageAdapter` into `useLocalRuntime` via new `storage`, `storagePrefix`, and `titleGenerator` options
- Export `createLocalStorageAdapter`, `AsyncStorageLike`, `TitleGenerationAdapter`, and `createSimpleTitleAdapter` publicly from `@assistant-ui/core/react` and `@assistant-ui/react`
- Add `reload()` method to `ThreadListRuntime` for re-fetching the thread list from the adapter
- Add optional `metadata` field to `RemoteThreadMetadata`, `RemoteThreadData`, `ThreadListItemCoreState`, and `ThreadListItemState`

## Test plan

- [x] Build passes (`pnpm turbo build --filter=@assistant-ui/core --filter=@assistant-ui/react`)
- [x] All 52 core tests pass (`pnpm --filter @assistant-ui/core test`)
- [x] No new lint errors introduced
- [ ] Manual test: `useLocalRuntime(model, { storage: localStorage })` persists threads across page reloads
- [ ] Manual test: `runtime.threads.reload()` re-fetches thread list from adapter

### DIFF
--- a/.changeset/feat-localstorage-adapter.md
+++ b/.changeset/feat-localstorage-adapter.md
@@ -1,0 +1,11 @@
+---
+"@assistant-ui/core": patch
+"@assistant-ui/react": patch
+---
+
+feat: add localStorage persistence support and thread list reload
+
+- Wire `createLocalStorageAdapter` into `useLocalRuntime` via `storage` option
+- Export `createLocalStorageAdapter`, `AsyncStorageLike`, `TitleGenerationAdapter` publicly
+- Add `reload()` method to `ThreadListRuntime` for re-fetching thread list
+- Add optional `metadata` field to `RemoteThreadMetadata` and `ThreadListItemState`
--- a/packages/core/src/react/adapters/LocalStorageThreadListAdapter.tsx
+++ b/packages/core/src/react/adapters/LocalStorageThreadListAdapter.tsx
@@ -17,9 +17,9 @@ import { RuntimeAdapterProvider } from "../runtimes/RuntimeAdapterProvider";
 import type { TitleGenerationAdapter } from "./TitleGenerationAdapter";
 
 export type AsyncStorageLike = {
-  getItem(key: string): Promise<string | null>;
-  setItem(key: string, value: string): Promise<void>;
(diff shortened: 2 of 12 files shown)


okisdev:
thanks for pushing on this. closing in favor of #3873, which landed the `reload()` part of your proposal as a focused change. wanted to leave a note on what went into that decision rather than just closing silently.

the shape in #3873 is the same as yours: `reload?()` as an optional method on `ThreadListRuntimeCore`, an explicit `reload()` on `ThreadListRuntime` with a `Promise.resolve()` fallback when the core does not implement it, and a per-load generation counter so a superseded in-flight response cannot overwrite state. the credit for that shape goes to this PR.

what i couldn't merge as-is was the rest of the bundle. four things were entangled: `reload()`, the `createLocalStorageAdapter` wiring on `useLocalRuntime`, the new `metadata` field on `RemoteThreadMetadata`/`RemoteThreadData`/`ThreadListItemState`, and a rewrite of the `then` reducer from merge to clobber. each has its own review surface. the reducer change in particular is a behavior change on the *initial* load (it drops `threadIdMap`/`threadData` entries written between construction and the first `list()`, which is why the PR needed a separate preservation block for `newThreadId` and `mainThreadId`), and it needs its own tests and discussion separate from the auth-race fix that issue #3815 describes.

the other gaps were smaller: the tap-only `ThreadsMethods` / `ThreadListClient` weren't updated, so `aui.threads().reload()` wasn't reachable; the changeset only patched `@assistant-ui/react` rather than `@assistant-ui/core` where the change lives; and there were no tests for the new reload path (stale-success dropped via generation, stale-rejection guarded in `.catch`, recovery after a failed initial load, multiple overlapping reloads). #3873 covers all of those.

if you'd like to open a focused PR for `createLocalStorageAdapter` (wiring + the `storage` / `storagePrefix` / `titleGenerator` options) and a second one for the `metadata` field, i'd review both on their own.

PR #3647, opened 2026-03-16
fix(core, store): stabilize AuiForEach selectors to prevent useSyncExternalStore loops

samdickson22:
## Summary

Selectors passed to `AuiForEach` that create new arrays on every call (e.g. `ids.map((_, i) => i)`) can trigger infinite re-render loops with `useSyncExternalStore`. Fixed by:

- **Returning stable array references from state** where possible (ThreadListItems, ComposerAttachments, MessageAttachments)
- **Using ID-based keys** instead of index-based keys where items have IDs (ThreadMessages, attachments)
- **Adding `useShallowArray` safety net** in `AuiForEach` for cases where selectors must derive new arrays (suggestions, parts)

## Changed files

**`packages/store/src/AuiForEach.tsx`** — Added `useShallowArray` between `useAuiState` and `useMemo` to stabilize array references when contents are identical

**`packages/core/src/react/primitives/`** — Fixed 7 selectors:
| File | Fix |
|------|-----|
| `threadList/ThreadListItems.tsx` | Return `threadIds` directly (stable state ref) |
| `thread/ThreadMessages.tsx` | `.map(m => m.id)` (stable IDs) |
| `thread/ThreadSuggestions.tsx` | Kept index mapping, stabilized by `useShallowArray` |
| `composer/ComposerAttachments.tsx` | `.map(a => a.id)` |
| `message/MessageAttachments.tsx` | `.map(a => a.id)` |
| `chainOfThought/ChainOfThoughtParts.tsx` | Kept index mapping, stabilized by `useShallowArray` |
| `message/MessageParts.tsx` | Kept index mapping, stabilized by `useShallowArray` |

## Why not fix `useAuiState` directly?

The proxy-based state system depends on **object identity changes** to signal state updates. Adding `shallowEqual` caching inside `useAuiState`'s `getSnapshot` suppresses legitimate state transitions (e.g., message sending breaks). The fix must be at the selector/consumer level, not in the store hook.

## Test plan

- [x] All 52 core tests pass
- [x] Build succeeds
- [ ] Verify ThreadListPrimitive.Items renders without infinite loop
- [ ] Verify ThreadPrimitive.Messages renders correctly
- [ ] Verify message sending still works

### DIFF
--- a/packages/store/src/AuiForEach.tsx
+++ b/packages/store/src/AuiForEach.tsx
@@ -32,7 +32,7 @@ function useStableArraySelector<TKey extends string | number>(
   // useRef is safe here — getSnapshot reads the ref but never calls setState.
   const cacheRef = useRef<readonly TKey[]>(undefined!);
 
-  const arr = useSyncExternalStore(aui.subscribe, () => {
+  const getSnapshot = () => {
     const next = selector(proxiedState);
 
     // Fast path: same reference (selector already returns stable array)
@@ -49,7 +49,9 @@ function useStableArraySelector<TKey extends string | number>(
 
     cacheRef.current = next;
     return next;
-  });
+  };
+
+  const arr = useSyncExternalStore(aui.subscribe, getSnapshot, getSnapshot);
 
   return arr;
 }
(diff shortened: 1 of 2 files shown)


okisdev:
this has been fixed, closing this PR, thanks!

Issue #3653, opened 2026-03-16
RemoteThreadListThreadListRuntimeCore.delete() crashes with 'Entry not available in the store'


iznogoud-01:
## Bug

Deleting a thread via `ThreadListItemPrimitive.Delete` followed by creating a new thread crashes with:

```
Error: Entry not available in the store
  at new ShallowMemoizeSubject (subscribable.js:109)
  at new ThreadListRuntimeImpl (thread-list-runtime.js:42)
  at RemoteThreadListHookInstanceManager._OuterActiveThreadProvider (RemoteThreadListHookInstanceManager.js:103)
```

## Root Cause

`RemoteThreadListThreadListRuntimeCore.delete()` (line 399-415) removes the thread from state via the `"deleted"` status reducer, but **never calls `this._hookManager.stopThreadRuntime(data.id)`**.

The `detach()` method (line 416-424) correctly calls `stopThreadRuntime`, but `delete()` is missing it.

This means the React component for the deleted thread remains in `HookInstanceManager.instances`. On the next re-render, `_OuterActiveThreadProvider` tries to construct a `ShallowMemoizeSubject` for the deleted thread ID, but `getThreadListItemState` returns `SKIP_UPDATE` because the data was already removed from the store.

## Suggested Fix

Add `this._hookManager.stopThreadRuntime(data.id)` to the `delete()` method after `_ensureThreadIsNotMain`, mirroring what `detach()` does.

## Versions

- `@assistant-ui/react`: 0.12.17
- `@assistant-ui/core`: 0.1.5

## Reproduction

1. Use `useRemoteThreadListRuntime` with a `RemoteThreadListAdapter`
2. Create and use a thread
3. Click `ThreadListItemPrimitive.Delete` on any thread
4. Observe crash

## Workaround

Use `ThreadListItemPrimitive.Archive` instead of `Delete`. The archive status keeps the data in the store (moved to `archivedThreadIds` rather than removed), avoiding the `SKIP_UPDATE` crash path.

PR #3661, opened 2026-03-17
fix(core): stop thread runtime on delete to prevent store crash

ShobhitPatra:
closes https://github.com/assistant-ui/assistant-ui/issues/3653


okisdev (approved):
LGTM!

Issue #3652, opened 2026-03-16
React 17: Switching threads crashes when message components are mounted with `ThreadPrimitive.Messages`  ("tapClientLookup: Index 0 out of bounds")


clocasto-wandb:
> [!NOTE] 
> **@Yonom's Note**
>
> This issue is about a zombie child problem which only occurs in the following cases:
>
> - You are using React 17 or lower with a `"use-snyc-external-store/shim"` polyfill
> - You are using React 18 with the unsupported `ReactDOM.render` API (which makes React behave as if it was React 17)
>
> `useSyncExternalStore` in React 18+ prevents the zombie child problem (zombie child problem = child components rerender before their parent has the chance to unmount them, breaking the store subscription selectors).
>
> As this issue only occurs on old React versions, we recommend upgrading to React 19.
> 
> We are looking into an elegant way of solving this issue for users on older react versions. We are unlikely to directly merge a PR that solves this problem due to the added complexity. We would instead provide a `patch-package` which affected users can install (similar to the use-sync-external-store/shim polyfill)
>
> **If you are using React 19 and see this error, please open a new Issue! This error should never appear.**

---
First off, thanks for the great library!

I'm encountering an error when attempting to start a new thread after having closed and re-opened our chat window (i.e. unmounted it and then mounted it again):

```
tapClientLookup: Index 0 out of bounds (length: 0)

getSnapshot                  @ useAuiState.js:22
checkIfSnapshotChanged       @ react-dom:16147
handleStoreChange            @ react-dom:16132
notifySubscribers            @ NotificationManager.js
(anonymous)                  @ useAui.js:26
(anonymous)                  @ tapResourceRoot.js:54
runTask                      @ scheduler.js:22
flushScheduled               @ scheduler.js:45
```

I think what's happening is that when the thread switches, message components from the old thread are still subscribed to the `NotificationManager`. They get notified about the store change before they can be unmounted, and their selectors hit the bounds check in `tapClientLookup.get()` because the new thread has no messages.

I think this only happens when `ThreadPrimitive.Messages` mounts with existing messages - mounting with zero messages and then receiving them later doesn't trigger it.

I'm not sure what the right fix is here, but we're relying on the following workaround for the time being to delay messages until the second render:

<details>
  <summary>Workaround</summary>

  ```tsx
  const DeferredMessages = () => {
    const messagesLength = useAuiState(
      s => s.thread.messages.length as number
    );

    const [ready, setReady] = useState(false);
[...]

PR #3663, opened 2026-03-17
fix(store): prevent crash on thread switch in useAuiState

ShobhitPatra:
closes https://github.com/assistant-ui/assistant-ui/issues/3652


Yonom:
let's fix the underlying issue instead of resorting to this. I'm investigating the root cause

PR #3666, opened 2026-03-17
fix(store,core): stabilize useSyncExternalStore snapshots

ShobhitPatra:
This PR fixes the React 19 snapshot/render loop (I hit while testing `react-ink` session persistence), where the UI showed `getSnapshot should be cached` and could crash with a max update depth error. This stabilizes `useAuiState` snapshots and removes render-phase runtime updates in the remote thread runtime flow.

### DIFF
--- a/.changeset/good-seas-kiss.md
+++ b/.changeset/good-seas-kiss.md
@@ -1,0 +1,7 @@
+---
+"@assistant-ui/core": patch
+"@assistant-ui/store": patch
+---
+
+fix(store,core): stabilize useSyncExternalStore snapshots  
+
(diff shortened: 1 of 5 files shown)


Yonom:
Thanks for the careful investigation — going to pass on this one.

The render-phase `updateRuntime()` in `_InnerActiveThreadProvider` is load-bearing for first-paint correctness on React DOM. Without it, the first client paint reads `EMPTY_THREAD_CORE` (`isLoading: true`, all `capabilities` false, `messages: []`, `composer.isEditing: true`) until the post-commit `useEffect` binds the real runtime — that's a visible flash of loading spinners / disabled actions / empty message list on every navigation.

We don't see this manifest on React DOM today, which means React-Ink must be reconciling something differently — I'd rather fix the underlying React-Ink behavior than relax the binding contract. Separately, we're planning a tap-based thread-switching architecture that will reshape this code path entirely.

The `useAuiState` snapshot caching half of this PR is a sensible fix on its own; happy to take that as a standalone PR if you'd like.

PR #3673, opened 2026-03-18
fix(store,react): handle derived scope invalidation on thread switch

ShobhitPatra:
This PR adds explicit derived-scope invalidation and teaches useAuiState to reuse the previous hook-local value only during that invalidation window, while preserving normal failures for real bugs and first-render errors. Derived scopes could remain mounted after their parent lookup became invalid during thread switches, causing zombie-child crashes. 

Also fixes MessageRoot to avoid holding a stale aui.message() handle.

### DIFF
--- a/.changeset/derived-scope-invalidation.md
+++ b/.changeset/derived-scope-invalidation.md
@@ -1,0 +1,6 @@
+---
+"@assistant-ui/store": patch
+"@assistant-ui/react": patch
+---
+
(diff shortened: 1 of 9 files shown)


ShobhitPatra:
closing - see https://github.com/assistant-ui/assistant-ui/issues/3652#issuecomment-4136488891 , fix is overkill for something that can't be reproduced in practice

PR #3675, opened 2026-03-18
feat(react-ink): add DiffPrimitive

ShobhitPatra:
This PR adds composable diff primitives and DiffView for terminal diff rendering
Screenshot from 2026-03-18 16-11-27


okisdev:
hey @ShobhitPatra, thanks for the contribution! i've pushed some follow-up changes on top of your commit. here's a summary of what was adjusted and why — good patterns to keep in mind for future PRs:

**architecture**

- **`Root` primitives should be thin context providers.** `DiffRoot` was doing parsing + diffing internally. in this codebase, `*Root` components (like `ThreadRoot`, `MessageRoot`) are just context + layout — no business logic. parsing now lives in `DiffView`, the convenience component.

- **don't export internal utils from primitive barrels.** `parsePatch`/`computeDiff`/`foldContext` were exported from `index.ts`, which made them part of the `DiffPrimitive.*` public API. only components and types belong there. if they need to be exposed for advanced use later, they'd go under a separate subpath (e.g. `./internal`).

- **primitives should be unstyled, styled layer should be separate.** primitives in this codebase are structural building blocks — no `color`, `bold`, or `dimColor`. for CLI/terminal packages this is a special case where we do provide a styled convenience component (`DiffView`), but the styling must be cleanly separated from the primitives. `DiffPrimitive.*` now outputs plain structural content; all colors and emphasis live in `DiffView` via `renderLine`/`renderFold` and an inline styled header.

**React patterns**

- **`useMemo` deps: destructure objects into primitives.** when a prop is an object like `oldFile: { content, name }`, passing it directly as a `useMemo` dep means inline JSX (`oldFile={{ content: "..." }}`) re-computes every render. destructure into `oldContent`, `oldName` etc. first.

- **compose existing primitives, don't duplicate markup.** `DiffHeader` was inlining the same `+N / -N` JSX that `DiffStats` already renders. now it composes `<DiffStats />` instead.

**edge cases**

- **handle CRLF and diff markers.** `\r\n` line endings and `\ No newline at end of file` markers are common in real-world diffs. both `parsePatch` and `computeDiff` now strip these, with tests covering the cases.

- **`DiffViewProps` should omit `children`.** `DiffView` renders its own children internally, so accepting `children` in the type was misleading.

**types**

- **public prop types belong in `types.ts`.** `DiffFileInput` (the `oldFile`/`newFile` shape) was defined locally in `DiffView.tsx`. moved to `types.ts` and exported since it's part of the public API surface.

**documentation**

- **new primitives need docs.** added DiffPrimitive and DiffView documentation to `apps/docs/content/docs/ink/primitives.mdx`, covering all sub-components and the convenience `DiffView` with full props tables. future PRs adding new primitives should include a corresponding docs section in the same file.

PR #3679, opened 2026-03-19
fix(core): prevent EMPTY_THREAD_ERROR when unstable_Provider delays c…

### LINKED ISSUE #3678: [Bug] EMPTY_THREAD_ERROR for useChatRuntime
### Checklist

### Environment info

```json
"dependencies": {
    ...
    "@ai-sdk/react": "^3.0.80",
    "@assistant-ui/react": "^0.12.19",
    "@assistant-ui/react-ai-sdk": "^1.3.6",
    "@assistant-ui/react-devtools": "^0.2.3",
    "@assistant-ui/react-markdown": "^0.12.6",
    ...
  }
```

### Affected area(s)

Runtime / Core

### Runtime adapter in use

### Steps to reproduce

if use the `useChatRuntime` directly, it can work correctly.

please refer to the code snippet.

Expected: I can see the history and input for each thread history
Actual: Cannot input, got EMPTY_THREAD_ERROR during typing in text field, nothing happen during switching thread

### Browser (if applicable)

_No response_

### Code snippet or error log

```tsx
const threadListAdapter: RemoteThreadListAdapter = {
  async list() {
    const threads = await listThreads()
    return {
      threads: threads.map((t) => ({
        remoteId: t.id,
        status: t.is_archived ? ('archived' as const) : ('regular' as const),
        title: t.title,
      })),
    }
  },

  async initialize(_localId: string) {
    const result = await createThread()
    return { remoteId: result.id, externalId: undefined }
  },

  async rename(remoteId: string, newTitle: string) {
    await renameThread({ data: { threadId: remoteId, title: newTitle } })
  },

  async archive(remoteId: string) {
    await archiveThread({ data: { threadId: remoteId } })
  },

  async unarchive(remoteId: string) {
    await unarchiveThread({ data: { threadId: remoteId } })
  },

  async delete(remoteId: string) {
    await deleteThread({ data: { threadId: remoteId } })
  },

  async fetch(remoteId: string) {
    const thread = await getThread({ data: { threadId: remoteId } })
    return {
      remoteId: thread.id,
      status: thread.is_archived
        ? ('archived' as const)
        : ('regular' as const),
      title: thread.title,
    }
  },

  async generateTitle(_remoteId, messages) {
    // Extract a short title from the first user message
    const firstUser = messages.find((m) => m.role === 'user')
    let title = 'New Chat'

    if (firstUser) {
      const textPart = firstUser.content.find((p) => p.type === 'text')
      if (textPart && 'text' in textPart) {
        title = textPart.text.slice(0, 80)
      }
    }

    // Persist the title
    await renameThread({ data: { threadId: _remoteId, title } })

    // Return as an AssistantStream (required by the adapter contract)
    return createAssistantStream(async (controller) => {
      controller.appendText(title)
    })
  },

  // Wrap every thread runtime with the HistoryProvider
  unstable_Provider: HistoryProvider,
}

export function ChatProvider({ children }: { children: ReactNode }) {
  const runtime = useRemoteThreadListRuntime({
    runtimeHook: useMyRuntime,
    adapter: threadListAdapter,
  })

  return (
      {children}
    </AssistantRuntimeProvider>
  )
}

function useMyRuntime() {
  return useChatRuntime({
    transport: new AssistantChatTransport({ api: '/api/chat' }), adapters: {
      dictation: new WebSpeechDictationAdapter({
        language: 'zh-HK',
        continuous: true,
        interimResults: true,
      })
    }
  })
}
```

### Additional context

it is work with `@assistant-ui/react": "^0.12.9`, I got this error after upgrade to latest version

Similar with this [issue](https://github.com/assistant-ui/assistant-ui/issues/2681), but this mark as resolved

ShobhitPatra:
Closes #3678. Moves runtime binding outside the user's `unstable_Provider` so it executes immediately, even if the Provider delays  rendering children.

### DIFF
--- a/.changeset/tender-ducks-write.md
+++ b/.changeset/tender-ducks-write.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/core": patch
+---
+
+fix(core): prevent EMPTY_THREAD_ERROR when unstable_Provider delays children
--- a/packages/core/src/react/runtimes/RemoteThreadListHookInstanceManager.tsx
+++ b/packages/core/src/react/runtimes/RemoteThreadListHookInstanceManager.tsx
@@ -82,9 +82,10 @@ export class RemoteThreadListHookInstanceManager extends BaseSubscribable {
     }
   }
 
-  private _InnerActiveThreadProvider: FC<{
+  private _RuntimeBinder: FC<{
     threadId: string;
-  }> = ({ threadId }) => {
+    provider: ComponentType<PropsWithChildren>;
+  }> = ({ threadId, provider: Provider }) => {
     const { useRuntime } = this.useRuntimeHook();
     const runtime = useRuntime();
 
@@ -110,6 +111,19 @@ export class RemoteThreadListHookInstanceManager extends BaseSubscribable {
       updateRuntime();
       return threadBinding.outerSubscribe(updateRuntime);
     }, [threadBinding, updateRuntime]);
+
+    return (
+      <Provider>
+        <this._InitializationHandler runtime={runtime} />
+      </Provider>
+    );
+  };
+
+  private _InitializationHandler: FC<{
+    runtime: AssistantRuntime;
+  }> = ({ runtime }) => {
+    const threadBinding = (runtime.thread as ThreadRuntimeImpl)
+      .__internal_threadBinding;
 
     const aui = useAui();
     const initPromiseRef = useRef<Promise<unknown> | undefined>(undefined);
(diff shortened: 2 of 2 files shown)


okisdev:
thanks for digging into this, @ShobhitPatra.

#3678 ended up landing as #3891 (and tweaked again in #3948), with a different shape than the split here:

- the binder was consolidated into a single `_RuntimeBinder` rather than split into a runtime binder + initialization handler. yonom called this out in the #3891 description: hoisting the runtime hook out while leaving the init listener inside the Provider registers it late, and the comment from @AlexChim1231 on the issue confirms it (history threads stop rendering messages after switching).
- #3948 then moved the consolidated binder back inside the Provider, because rendering it outside hides `RuntimeAdapterProvider`'s context from the runtime hook and breaks cloud message persistence, run telemetry, and attachments.
- the final fix surface for the original bug is a dev mode warning that tells users their `unstable_Provider` must render `children` synchronously, plus a docs callout.

going to close this out for those reasons. really appreciate the effort and the repro work on the issue.

PR #3681, opened 2026-03-19
feat: add checklist primitives

ShobhitPatra:
## Summary                                                                                                                                                                                  
                                                                                                                                                                                            
  Adds a three-layer checklist primitive for tracking tool-call progress:
          
  - **Core** (`@assistant-ui/core`): shared types (`ChecklistItemData`, `ChecklistItemStatus`, `ChecklistData`), `useToolActivityChecklist` hook, `mapToolCallToChecklistItem` utility,       
  `flattenChecklistItems`
  - **React** (`@assistant-ui/react`): unstyled Radix primitives (`ChecklistPrimitive.Root`, `.Item`, `.Progress`), convenience components (`LiveChecklist`, `ToolActivityChecklist`,         
  `DataChecklist`)                                                                                                                                                                            
  - **React-Ink** (`@assistant-ui/react-ink`): refactored to import shared logic from core, terminal-specific rendering stays (ASCII indicators, ink-spinner, colors)
  - **UI** (`@assistant-ui/ui`): styled `LiveChecklist` with variants (outline/ghost/muted), shimmer animation, status indicators    
 ### web 
https://github.com/user-attachments/assets/3e14b676-615b-4425-aed9-4ae5d26dd20b

 ### ink
https://github.com/user-attachments/assets/ee5e485e-5c89-4671-958c-e235966ab55d


ShobhitPatra:
Pushed three commits since the last round.

**Done**
- `b5bc73fb5`: progress reads `X/Y done`, not `complete` (errors are terminal).
- `038a529c3`: extracted `ChecklistView`, shared by `LiveChecklist` / `ToolActivityChecklist` / `DataChecklist`.
- `ec01381b7`: dropped dead `STATUS_INDICATORS.running` and removed type re-exports from the primitive barrel.

**Declined (with reasoning)**
- **Derivation in selector (perf):** core primitive-hooks are thin `useAuiState` readers, no memo machinery. The suggested fix returns a new array each call (per your own note), so it doesn't cut re-renders anyway. Memoization belongs in the component.
- **Conditional spreads:** `exactOptionalPropertyTypes` is on and ink props are `?: T` (no `| undefined`), so `maxDepth={maxDepth}` won't compile. The spread is required.
- **`Exclude` vs `Partial`:** `STATUS_INDICATORS[item.status]` is indexed with the full union (incl. `running`), so an `Exclude`-keyed record errors there. `Partial` keeps it safe.
- **`DataChecklist` props:** it's a data-part renderer, so inputs come from the payload and `renderItem` can't be serialized. Matches web.
- **`mapToolCallToChecklistItem` export:** test-only, not in the public index. Can mark `@internal` if preferred.
- **Shared resolver with `ToolFallback`:** deliberate divergence, better as a follow-up than this PR.

okisdev (approved):
thanks for this, @ShobhitPatra. the checklist itself is a genuinely nice feature, so i pushed a refactor to the branch to tighten the scope before we ship it. here's what changed:

- scoped the whole feature to @assistant-ui/react-ink. reverted @assistant-ui/core, react, ui, the registry entry, and the web docs page to their pre PR state, and moved the shared types and the hook into react-ink so the package is self contained (it leans on core only for the generic `MessagePartState` type and on @assistant-ui/store for `useAuiState`).
- dropped the data-part path (`DataChecklist` + `ChecklistDataUI`). that was a second data source with no consumer in the repo, and a different concept from the headline of deriving the checklist from the message's tool-call parts.
- folded `ChecklistView` into `LiveChecklist`, since it had one consumer left after the cut.
- renamed `useToolActivityChecklist` to `useToolCallChecklist`. "activity" isn't a term we use anywhere; the domain noun is tool call (`ToolCallMessagePart`, `useToolArgsStatus`, `ToolFallback`), so the new name joins the `useTool*` family.
- aligned the react-ink code with existing idioms: optional props typed as `?: T | undefined` and passed directly like `DiffView` instead of conditional spreads, and `Record<Exclude<...>>` status maps like `ToolFallback`.

net effect: from roughly four packages, ~18 files, two data sources, and a speculative web layer down to one self contained react-ink feature. the public surface is now `ChecklistPrimitive.Root` / `.Item` / `.Progress`, `LiveChecklist`, `useToolCallChecklist`, and the `ChecklistItemData` / `ChecklistItemStatus` types, each one earning its place. changeset is react-ink only; build, typecheck, tests, and lint are green.

on the why: the "optimization" here is subtraction toward the minimal correct surface plus alignment with how the repo is already written, not perf or new features. the cross platform layers and the data-part source had no consumer, and unused public surface is real recurring cost to review, document, version, and maintain, bought against a hypothetical payoff. doing the extraction upfront is also the harder bet, since you can't design the right shared abstraction before there's a second real use of it; lifting to core speculatively risks the wrong shape and dead code in a large PR for a feature only one distribution ships. keeping it in one package now is cheap and reversible, and assistant-ui already proves the promotion path: `ChainOfThought` lives in core and ships to react, react-native, and react-ink, so lifting this later is mechanical (move the types and hook up a level, add the per platform primitives) and can land alongside the web consumer and its tests in the same PR. the original ambition was reasonable, just early.

PR #3682, opened 2026-03-19
fix(react): ensure scroll to bottom is only triggered after the new message is mounted for anchor="top"

### LINKED ISSUE #3648: `turnAnchor="top"` scroll to bottom call is sometimes mistimed
When the turn-anchor is set to "top", the call to scroll the conversation to the bottom after the slack space as been added sometimes does not receive, use and set the right scroll position. This is because this scroll action relies on the `thread.runStart` event, which fires before the UI has been updated with the user message.

Even though the scroll action is called inside of `requestAnimationFrame`, this does not guarantee that react has rendered the user message, which means that sometimes the height used to calculate the "bottom" scroll position uses the viewport height before the user message + slack is rendered.

WoodyWoodsta:
Fixes https://github.com/assistant-ui/assistant-ui/issues/3648

Happy to formalise the new event pattern. I appreciate this is the first of its kind in this project. Also welcome suggestions for how to achieve this direction of communication in a different/more standard way.

### DIFF
--- a/.changeset/shaggy-boxes-stop.md
+++ b/.changeset/shaggy-boxes-stop.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react": patch
+---
+
+fix(react): ensure scroll to bottom is only triggered after the new message is mounted for anchor="top"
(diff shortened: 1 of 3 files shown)


AVGVSTVS96:
fixed this in #3908 which redesigns top anchoring and updates the api

PR #3716, opened 2026-03-27
fix(core): show loading state for empty parts children API

### LINKED ISSUE #3700: MessagePrimitive.Parts with children render function doesn't show loading indicator when parts is empty
### Checklist

### Environment info

```
Environment:
  OS:               macOS 26.3.1 (arm64)
  Node.js:          v20.20.0
  Package Manager:  pnpm 8.15.7
  Framework:        Vite 5.4.10

Packages:
  @assistant-ui/react           0.12.20
  assistant-stream              0.3.4
  @assistant-ui/react-ai-sdk    1.3.15
  @assistant-ui/react-markdown  0.12.6
```

### Affected area(s)

Runtime / Core

### Runtime adapter in use

### Steps to reproduce

1. Use MessagePrimitive.Parts with a children render function
2. Send a message that triggers an assistant response
3. Observe the brief period before the first token streams — no loading indicator is shown

Expected: The assistant message should show the loading indicator while waiting for the stream to happen
Actual: No loading indicator is displayed

### Browser (if applicable)

_No response_

### Code snippet or error log

```tsx

```

### Additional context

After upgrading to @assistant-ui/react@0.12.20 / @assistant-ui/core@0.1.8, the loading indicator no longer appears during the window between when an assistant message is created and when the first stream token arrives.

Expected behavior: A loading indicator (e.g. the pulsing dot from dot.css) should be visible while the message is in a running state but has no parts yet (i.e. before the first token streams in).

Actual behavior: Nothing renders. The loading indicator only appears once the first text part arrives and streaming begins.

Root cause:

When using the children render function pattern:

```typescript
<MessagePrimitive.Parts>
  {({ part }) => {
    if (part.type === 'text') return <MarkdownText />;
    return null;
  }}
</MessagePrimitive.Parts>
``` 

`MessagePrimitivePartsInner` iterates over `message.parts.length`:

```typescript
const contentLength = useAuiState((s) => s.message.parts.length);
return useMemo(() => Array.from({ length: contentLength }, ...), [contentLength, children]);
```

When parts is empty (before the first stream token), `Array.from({ length: 0 })` produces an empty array, so the render function is never invoked and nothing is rendered.

The components-based code path (MessagePrimitivePartsCompat) correctly handles this with an EmptyParts fallback:

```typescript
if (contentLength === 0) {
  return _jsx(EmptyParts, { components: components });
}
```

But the children render function path (MessagePrimitivePartsInner) has no equivalent fallback — this is the gap introduced in the refactor.

ShobhitPatra:
fixes #3700

### DIFF
--- a/.changeset/fair-apes-unite.md
+++ b/.changeset/fair-apes-unite.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/core": patch
+---
+
+fix(core): show loading state for empty parts children API
--- a/packages/core/src/react/primitives/message/MessageParts.tsx
+++ b/packages/core/src/react/primitives/message/MessageParts.tsx
@@ -442,6 +442,10 @@ const COMPLETE_STATUS: MessagePartStatus = Object.freeze({
   type: "complete",
 });
 
+const RUNNING_STATUS: MessagePartStatus = Object.freeze({
+  type: "running",
+});
+
 const EmptyPartsImpl: FC<MessagePartComponentProps> = ({ components }) => {
   const status = useAuiState(
     (s) => (s.message.status ?? COMPLETE_STATUS) as MessagePartStatus,
@@ -591,13 +595,17 @@ const MessagePrimitivePartsInner: FC<{
   children: (value: { part: EnrichedPartState }) => ReactNode;
 }> = ({ children }) => {
   const aui = useAui();
-  const contentLength = useAuiState((s) => s.message.parts.length);
-  const emptyRunningStatus = useAuiState((s) => {
-    if (s.message.parts.length > 0) return null;
+  const { contentLength, isEmptyRunning } = useAuiState(
+    useShallow((s) => {
+      const contentLength = s.message.parts.length;
+      const status = (s.message.status ?? COMPLETE_STATUS) as MessagePartStatus;
 
-    const status = (s.message.status ?? COMPLETE_STATUS) as MessagePartStatus;
-    return status.type === "running" ? status : null;
-  });
+      return {
+        contentLength,
+        isEmptyRunning: contentLength === 0 && status.type === "running",
+      };
+    }),
+  );
   // Subscribed (not snapshotted like `tools`) so fallbacks registered after
   // the first render trigger a re-render and `hasUI` re-evaluates.
   const dataRenderers = useAuiState((s) => s.dataRenderers);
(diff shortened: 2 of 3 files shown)


okisdev (approved):
looks good to ship as `patch`.

- `emptyRunningStatus` (MessageParts.tsx:594 in the first commit) returns the raw `MessagePartStatus` object reference, so any store update that re-allocates the status object invalidates the downstream `useMemo` dependency even when `status.type` hasn't changed.
- the same selector reads `s.message.status` on every store tick, including the hot path where `parts.length > 0` and the result is always `null`; the closure runs regardless.
- the synthetic empty-text payload `{ type: "text", text: "", status: emptyRunningStatus }` is built inline inside `useMemo`, reallocating on every empty-running render even though all three fields are constants.
- `COMPLETE_STATUS` exists at line 441 but no `RUNNING_STATUS` companion is added, so the empty-running branch can't share a frozen singleton with neighbouring code.
- `EnrichedPartState` (line 575) gets no JSDoc note about the new contract: consumer code like `if (part.type === "text") return <MarkdownText />` will now render with `text=""` during the loading window, and the type alone gives no hint that this synthetic invocation can occur.
- the test omits the negative case: no assertion that the loading indicator stays hidden when parts are empty and status is `complete` (or `incomplete`, `requires-action`), leaving `EmptyPartsImpl`'s `if (status.type !== "running") return null` guard untested for the children path.
- test imports go through the package root barrel `..` rather than the subpath source files used by the other suites under `packages/react/src/tests/`.

### what we modified

- added `RUNNING_STATUS = Object.freeze({ type: "running" })` adjacent to `COMPLETE_STATUS` so the running-state singleton mirrors the complete one.
- replaced the composite selector with two single-primitive `useAuiState` calls (`contentLength: number`, `isRunning: boolean`) and derived `isEmptyRunning` in JS, matching the pattern dominant across `packages/core/src/react/primitives/` (every selector returns one primitive or one stable ref).
- extracted the synthetic part as a frozen module-level constant `EMPTY_RUNNING_TEXT_PART: Extract<EnrichedPartState, { type: "text" }>` and pass it as `children({ part: EMPTY_RUNNING_TEXT_PART })`, removing the per-render allocation.
- documented the new behaviour on `EnrichedPartState` with a JSDoc paragraph explaining the synthetic-text invocation and how to disambiguate it (`part.status?.type === "running" && part.text === ""`).
- added a regression test covering empty parts plus `{ type: "complete", reason: "stop" }` status, asserting `screen.queryByTestId("loading-dot")` returns `null`.
- generalised the test fixture's `RuntimeProvider` and `renderThread` to accept a `messages` argument so the same setup serves both running and complete cases.
- switched test imports from the `..` barrel to subpath sources (`../primitives/messagePart`, `../legacy-runtime/runtime-cores/local/useLocalRuntime`, etc.) to match neighbouring tests.

PR #3749, opened 2026-04-02
fix: include scope registration types so AssistantState has all properties

### LINKED ISSUE #3731: Property 'message' does not exist on type 'AssistantState'.ts(2339)
### Problem statement

assistant-ui/react version : ": "^0.12.17"

I’m trying to access `message` from `useAssistantState` (for `id`, `content`, `status`)

Usage:
Image

But TypeScript shows:

```ts
Property 'message' does not exist on type 'AssistantState'. ts(2339)
```
Had to suppress it with:

```ts
// @ts-ignore TS2339 - message exists in MessagePrimitive context
```

Would appreciate a fix.

### Proposed solution

_No response_

### Area

None

### Alternatives considered

_No response_

### Additional context

_No response_

Yonom:
## Summary
- Add `/// <reference types="@assistant-ui/core/store" />` to `@assistant-ui/react`, `@assistant-ui/react-native`, and `@assistant-ui/react-ink`
- This ensures the `ScopeRegistry` module augmentation from `@assistant-ui/core/store` is loaded, so `AssistantState` correctly includes all properties (`message`, `thread`, `composer`, etc.)
- Without this, TypeScript users see `Property 'message' does not exist on type 'AssistantState'` when using `useAssistantState`

Closes #3731

## Test plan
- [x] `pnpm turbo build` succeeds for all three packages
- [x] `pnpm --filter @assistant-ui/react test` — 201 tests pass
- [x] Verified `dist/index.d.ts` includes the new reference directive for all three packages
- [x] `pnpm lint` clean (only pre-existing warning)

### DIFF
--- a/.changeset/fix-assistant-state-types.md
+++ b/.changeset/fix-assistant-state-types.md
@@ -1,0 +1,7 @@
+---
+"@assistant-ui/react": patch
+"@assistant-ui/react-native": patch
+"@assistant-ui/react-ink": patch
+---
+
+fix: include scope registration types so AssistantState has all properties
--- a/packages/react-ink/src/index.ts
+++ b/packages/react-ink/src/index.ts
@@ -1,3 +1,4 @@
+/// <reference types="@assistant-ui/core/store" />
 /// <reference types="@assistant-ui/core/react" />
 
 // Re-export core types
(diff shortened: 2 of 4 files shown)


Yonom:
I thought importing types from core already does this, is this not true? can you reproduce the original error and confirm that this issue fixes it?

PR #3761, opened 2026-04-05
fix(assistant-stream): handle CRLF line endings in LineDecoderStream

lawrence3699:
## Summary

`LineDecoderStream` splits on `\n` only, leaving trailing `\r` when servers use `\r\n` (CRLF) line endings — the standard in HTTP. This breaks all four stream decoders in the library:

- **`AssistantTransportDecoder`**: `[DONE]` marker becomes `"[DONE]\r"` → not matched → `JSON.parse("[DONE]\r")` throws `SyntaxError`
- **`SSEDecoder`**: Same `[DONE]` detection failure
- **`UIMessageStreamDecoder`**: Same `[DONE]` detection failure
- **`DataStreamDecoder`**: Trailing `\r` in parsed line values (benign for JSON but incorrect)

Empty-line detection also fails: `"\r" !== ""`, so SSE events are never dispatched incrementally — they batch into `flush()` instead of streaming.

## Fix

Strip trailing `\r` after splitting on `\n` in `LineDecoderStream`. One line change, handles the common CRLF case.

> **Note:** Lone `\r` (CR-only) as a line separator — while technically part of the SSE spec — is not handled. No modern SSE server uses CR-only line endings. Supporting it would require a cross-chunk state machine that conflicts with the project's pragmatic coding style. This can be added in a follow-up if ever needed.

## Test plan

- [x] 9 new unit tests for `LineDecoderStream`: LF, CRLF, mixed, cross-chunk CRLF, empty lines (LF and CRLF), multiple chunks, incomplete stream error, SSE-like CRLF data
- [x] 1 new integration test: `AssistantTransportDecoder` decodes CRLF SSE end-to-end (data + `[DONE]` marker)
- [x] All 169 existing tests pass (`pnpm --filter assistant-stream test`)
- [x] Linter passes (biome via lint-staged)

### DIFF
--- a/.changeset/fix-line-decoder-crlf.md
+++ b/.changeset/fix-line-decoder-crlf.md
@@ -1,0 +1,5 @@
+---
+"assistant-stream": patch
+---
+
+fix(assistant-stream): handle CRLF line endings in LineDecoderStream
--- a/packages/assistant-stream/src/core/serialization/assistant-transport/AssistantTransport.test.ts
+++ b/packages/assistant-stream/src/core/serialization/assistant-transport/AssistantTransport.test.ts
@@ -161,6 +161,31 @@ describe("AssistantTransportDecoder", () => {
     expect(decodedChunks).toEqual(originalChunks);
   });
 
+  it("should decode SSE with CRLF line endings", async () => {
+    // Simulate a server that uses \r\n line endings (common in HTTP)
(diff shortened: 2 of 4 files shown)


Yonom (approved):

Yonom:
Thank you!

PR #3790, opened 2026-04-08
feat: add a2ui declarative surface rendering integration

ShobhitPatra:
## Summary                                                                                                                                                                 
                                                                                                 
  - Add `onCustomEvent` callback and STATE_DELTA JSON Patch support to `@assistant-ui/react-ag-ui`                                                                           
  - New `@assistant-ui/react-a2ui` package for rendering declarative agent UI surfaces
  - Example app (`with-a2ui`) demonstrating chat + surface sidebar with custom styled components                                                                             
  - Runtime guide  docs                                                                                                                                     
                                                                                                                                                                             
  ## Package: `@assistant-ui/react-a2ui`                                                                                                                                     
                   
  - `A2uiProvider` — manages surfaces, data binding, and component resolution                                                                                                
  - Hooks: `useA2ui`, `useA2uiSurfaces`, `useA2uiSurface`, `useA2uiData`, `useA2uiActions`       
  - `A2uiSurfaceRenderer` — renders component trees with recursive child resolution                                                                                          
  - AG-UI bridge (`createA2uiMessageHandler`, `useA2uiAgUiBridge`) for wiring CUSTOM events                                                                                  
  - Protocol parser, `SurfaceManager`, `SurfaceDataStore` with granular path subscriptions                                                                                   
  - Transport-agnostic — works with AG-UI, WebSocket, HTTP, or any event source                                                                                              
                                                                                                                                                                             
  ## AG-UI changes                                                                                                                                                           
                                                                                                                                                                             
  - `onCustomEvent` callback on `UseAgUiRuntimeOptions` — intercepts CUSTOM events before they hit the aggregator                                                            
  - `STATE_DELTA` handling via minimal RFC 6902 JSON Patch implementation


ShobhitPatra:
closing to scope the gap against where the codebase is now before continuing. leaving the framing + open questions here as the public record, for me when i pick this back up or for anyone else hitting the same question.

the question this pr needs to answer: is a2ui's distinction from mcp apps load-bearing enough to warrant a separate package, given mcp apps is the canonical path for agent-controlled UI rendering and AGENTS.md calls out a tap-only migration?

distinction a2ui was scoped against:
- mcp apps: server ships html/js, host sandboxes in iframe
- a2ui: server ships declarative json, host renders from a pre-approved react component catalog. no code over the wire.

different trust posture (catalog vs sandbox), different render path (react tree vs iframe), different coupling (independent surfaces vs inline tool-call attachment). real distinction, but the bar isn't "different", it's "covers a use case the canonical path can't". not confident in that.

open architectural questions worth nailing down before reshaping:

1. **gap vs mcp apps.** the narrow cases (csp-restricted deployments, host-design-system parity, display-only surfaces without chat) are real but not load-bearing enough to answer yes confidently.

2. **structural analog.** comparing a2ui to a2a / langgraph / google-adk and mapping state onto runtime.thread.extras is the wrong frame. those are chat runtimes. the right analog is react-mcp: a stateful non-runtime rendering package mounted as a tap resource via useAui({ mcp: McpManagerResource(...) }). this pr ships a parallel A2uiContext + own SurfaceManager / SurfaceDataStore, which is neither. if a package is justified, tap-resource is the shape.

3. **standalone-no-runtime track.** the docs advertise "standalone websocket / http, no AssistantRuntime required" alongside the ag-ui track. included it because a2ui is transport-agnostic by design and display-only UIs (dashboards, agent monitors) may want surfaces without a chat side, which is a genuine use case the iframe-sandbox model fits poorly. tap resources work under AuiProvider alone so it can be supported idiomatically. open whether it should be, given no sibling does it.

taking those internal rather than continuing on this branch. if (1) resolves to "mcp apps + docs guide covers it", this pr was the wrong artifact regardless; if it resolves to "build as a tap resource", the new pr isn't this one. either way this branch isn't the next step.

**can be reopened if** the gap warrants a package *and* the parallel-context architecture is the right shape. otherwise the next step is a fresh pr against the agreed architecture, with this branch as a reference implementation for the catalog-render + json-pointer-binding model.

PR #3796, opened 2026-04-09
feat(core): add fallback data renderer for dynamic ui loading

ShobhitPatra:
closes #3791


rotem1850 (changes requested):
@ShobhitPatra  Thanks for the PR — `setFallbackDataUI` / `makeAssistantFallbackDataUI` / `useAssistantFallbackDataUI` is the right building block. A few things I'd want addressed before merge:

---

## Bug: late-registered fallbacks silently render nothing

In `MessagePrimitivePartsInner`, the `hasUI` check reads from a plain snapshot:

```ts
const drState = aui.dataRenderers().getState();
const hasUI =
  (Array.isArray(entry) && !!entry[0]) ||
  !!drState.fallback;
```

Since `setFallbackDataUI` is designed for *dynamic* loading — registering the fallback after the message first renders — `drState.fallback` will be `undefined` on first paint and `part.dataRendererUI` will be `null` in the `children` render path. The fallback only becomes visible after an unrelated re-render triggers a new snapshot read.

This is a pre-existing limitation for named renderers, but it's more impactful here: late registration is the entire point of this API. A `LoadExternalComponent` use case will hit this on every cold load.

**Fix:** replace the raw `getState()` call with a reactive `useAuiState` selector so `hasUI` updates when the fallback is registered.

---

## Missing: `loader` integration in the LangGraph adapter

This PR adds the core primitive but leaves the LangGraph adapter integration entirely to the user. To actually use `LoadExternalComponent` with `useLangGraphRuntime`, a user currently has to:

1. Call `useAssistantFallbackDataUI` manually
2. Use `useLangGraphUIMessages` to get the raw message list
3. Reconstruct the `name` + `stream` context that `LoadExternalComponent` expects

That's a lot of internal knowledge required. The adapter should handle this directly:

```ts
useLangGraphRuntime({
  stream: ...,
  uiComponents: {
    loader: (name, stream) => loadExternalComponent(name, stream),
    chart: ({ data }) => <Chart {...data} />, // static override
  },
});
```

When the adapter receives a `ui_message` with no matching static renderer, it calls `loader(name, stream)`. Static renderers take priority; the loader is the fallback. This keeps the dynamic path opt-in and composable — and it's what was asked for in #3791.

---

## Missing: docs

Neither this PR nor #3789 added a guide for using `LoadExternalComponent` inside an `AssistantRuntimeProvider` tree. At minimum, the docs should cover:

- Where to call `useAssistantFallbackDataUI` in the component tree
- How to pass the `stream` context `LoadExternalComponent` requires
- The `metadata.message_id` / message association model
- A minimal end-to-end example: LangGraph backend → `langgraph.json` → `LoadExternalComponent` rendering inside the thread

Without this, teams using LangSmith's colocated UI workflow have no official path to follow.

rotem1850 (approved):

PR #3850, opened 2026-04-17
feat(react-ink): rewrite react-ink composer input

ShobhitPatra:


okisdev (approved):
## review

LGTM as `patch`.

- cursor disappeared in the inverse cell when positioned on `\n` (after `End` / `Ctrl+E` in multi-line); rendering an inverse newline produces no visible block.
- `getLineStart` returned `1` instead of `0` when `cursorOffset === 0` on a leading `\n`, corrupting `kill-start` (`Ctrl+U`) into duplicating the newline.
- char-level navigation and deletion operated on UTF-16 code units, splitting surrogate pairs (emoji and other non-BMP characters got truncated).
- `ctrl+j` fell through to the `key.return` branch in single-line mode and submitted on terminals that report both `key.ctrl` and `key.return`.
- the hook exposed 13 per-action `useCallback` helpers that `ComposerInput` never called.
- `dispatchAction` wrapped `useReducer`'s already-stable `dispatch` in a redundant `useCallback`.
- `ctrl+k` at end-of-line in multi-line was a no-op; emacs / readline convention is to kill the trailing `\n` and join the next line.
- `ctrl+d` → `delete-forward` (the standard readline forward-delete) was not bound at all.
- pending-sync queue was a spread-array with O(n) `indexOf` per keystroke and no upper bound; a store that drops echoes could grow it indefinitely.
- `auiRef` held `useAui()`'s result in a ref despite `useAui()` already returning a stable reference.
- `applyAction` and `dispatchAction` shared confusingly similar names at the call site.

### what we modified

- render a space in the inverse cursor cell when on `\n`, keep the `\n` in `after` so the line break still happens.
- early return for `cursorOffset === 0` in `getLineStart`.
- grapheme stepping via `Intl.Segmenter` (covers ZWJ sequences, skin-tone modifiers, regional flag pairs, CJK ideographs).
- unconditional `return` after the ctrl+j branch.
- removed all 13 named helpers; aliased `dispatchAction = dispatch` directly.
- `ctrl+k` at EOL in multi-line now kills the trailing `\n`, joining the next line.
- added `ctrl+d` → `delete-forward`.
- pending-sync queue is a `Map<string, number>` capped at `PENDING_SYNC_CAP = 64`.
- removed the `auiRef` indirection.
- renamed `applyAction` → `commitAction` so the eager-then-dispatch wrapper reads distinct from raw `dispatchAction`.
- expanded tests to cover every binding plus unicode / CJK / edge cases (46 total).
- rewrote the keyboard reference as a key / action / context table with a terminal compatibility note (macOS Terminal.app meta requirement, CSI-u for `Shift+Enter`).

### potential further fixes

- ink's `Key` type does not declare `home`, `end`, or `shift`; the `extendedKey` cast is the workaround. file upstream so the cast can go.
- `moveVertical`'s `preferredColumn` is computed in code units, not visual width. wide CJK / wide emoji will misalign across vertical moves. would need a font-metrics integration to fix.
- `Intl.Segmenter` word operations scan the whole buffer per call. acceptable for composer-sized input; cache per text version if 10k+ char buffers become common.
- CJK word boundaries fall back to per-ideograph because there is no built-in dictionary segmentation. acceptable without an external segmenter.

Issue #3504, opened 2025-12-06
Image generation


Rajbir-Singh1334:
Please also add image generation in it to take it one step further. I hope you will take this idea seriously and give the best outcome for us.

okisdev:
thanks for the suggestion. just to make sure we're on the same page, assistant-ui is a UI and runtime library, the actual image generation needs to come from whichever model/backend you connect (OpenAI images, Gemini, etc.). what assistant-ui already ships on top of that is the full pipeline to display generated images:

- `ImageMessagePart` content part type, valid on both user and assistant messages (`packages/core/src/types/message.ts`)
- `MessagePart.Image` primitive and `useMessagePartImage` hook (`packages/react/src/primitives/messagePart/`)
- a shadcn-style `Image` component with preview, filename, and zoom (`packages/ui/src/components/assistant-ui/image.tsx`)
- attachment adapter that turns uploaded images into image parts, plus cloud history adapter that persists them

so a tool call (or stream) that returns an image URL or base64 will render out of the box once you emit it as an `image` part.

could you clarify what you'd like to see added on top of this? for example:
1. a worked docs example for a generate-image tool call
2. an input/attachment flow for sending images into the model
3. integration with a specific provider's image API
4. something else entirely

happy to scope it once we know which piece is missing for your use case. closing for now since the existing primitives already cover image rendering, please reopen or comment with more detail and we'll take another look.

PR #3867, opened 2026-04-20
feat: image generation guide, example, and Image component

samdickson22:
## summary

generation itself is the model/backend's job, and rendering generated images is already first-class in assistant-ui (`ImageMessagePart`, `MessagePart.Image`, the cloud history adapter). this PR documents that path and polishes the rendering, without adding a new published primitive:

- **docs + example**: a guide plus a worked `with-image-generation` example. generate in your backend (e.g. `ai.generateImage` in a route handler), return an `ImageMessagePart`, and render it with the `@assistant-ui/ui` Image component. the example falls back to a mock image when `OPENAI_API_KEY` is unset, so it runs without credentials.
- **`@assistant-ui/ui` Image component**: adds loading, content-filter, zoom, and action (download / copy / regenerate) states. regenerate is wired through an `onRegenerate` callback so the app owns the generation flow.
- **`@assistant-ui/core`**: relaxes `thread-message-like` image validation to accept `https://` and `blob:` URLs (and `svg+xml` data URIs) alongside base64 `data:` URIs, so assistant-authored images served from a URL render instead of being dropped.

addresses the image generation request in #3504.

## scope

this began as an "image generation primitive" with published hooks (`useImageGeneration`, `useImagePartRegenerate`, `useImagePartDownload`, `useImagePartCopy`), an AI SDK adapter, and an `ImageMessagePart` metadata kernel persisted through `auiV0`. that public API surface is not needed: generation belongs to the backend and rendering already works, so the hooks, the adapter, the metadata fields, and the persistence were dropped. `@assistant-ui/react` and `@assistant-ui/react-ai-sdk` are untouched by this PR.

## changeset

`@assistant-ui/core` patch (the validation relax). `@assistant-ui/ui` is private, so it carries no changeset.

## test plan

- [x] example verified in a browser with the mock fallback: generate renders, regenerate re-runs, and the download / copy actions appear; `POST /api/image` returns 200.
- [ ] reviewer: run it against a real provider with `OPENAI_API_KEY` set.


samdickson22:
Rebased onto current main (2e113b4d4) to clear the merge conflict. Conflicts resolved:

- Docs guides directory was renamed `(docs)/guides/` → `guides/` upstream; moved `image-generation.mdx` into the new path and added the `image-generation` slug to the new `meta.json` under `---Tools & Generative UI---`.
- `pnpm-lock.yaml`: took upstream's lockfile and re-ran `pnpm install` to register the `examples/with-image-generation` workspace.

All 4 cubic P2 findings remain addressed (verified after rebase):

1. `useImageGeneration.ts` — `inflightRef` claim, `lastCallRef.current = now`, and `windowRef.current.push(now)` all happen synchronously before `await confirm()`; an early `if (inflightRef.current) return null;` blocks concurrent callers from sneaking past the gates.
2. `thread-message-like.ts` — URL regex tightened to `^(https:\/\/|blob:)`; `http://` is rejected.
3. `image.tsx` — `onRegenerateStart` is wired through the hook's `observers.onStart`, which fires only after debounce/rate-limit/confirm gates pass.
4. `examples/with-image-generation/app/page.tsx` — `onRegenerated` uses `prev?.prompt ?? prompt` so the stored prompt comes from the part being regenerated, not the live textbox.

Local verification: `@assistant-ui/core`, `@assistant-ui/react`, and `@assistant-ui/react-ai-sdk` test suites all green; `pnpm exec biome check` clean on the touched files; turbo build of the four published packages green.

okisdev (approved):
thanks for this. the implementation was genuinely solid (the regenerate concurrency lock, the fake-timer tests, green CI), so this is a scope decision, not a quality one.

the core call: assistant-ui is a UI + runtime library. the generation itself comes from whatever model/backend you connect, and rendering generated images is already first-class here, `ImageMessagePart`, `MessagePart.Image` / `useMessagePartImage`, the `@assistant-ui/ui` Image component, and the cloud history adapter. a tool or route that returns an image already renders out of the box. that's also how #3504 was resolved: the existing primitives cover the rendering, and generation belongs to the backend. a new published primitive (the hooks + the AI SDK adapter + an `ImageMessagePart` metadata kernel) is public API surface, and a long-term maintenance commitment, that the goal doesn't require.

i've reshaped the scope on the branch directly. what it keeps:

- the `@assistant-ui/ui` `Image` component: loading, content-filter, zoom, and download / copy / regenerate actions. the download/copy/regenerate logic now lives inside that copy-in component, and regenerate is an `onRegenerate` callback so the app owns the generation flow.
- the `with-image-generation` example and a docs guide for the end-to-end path: generate in a route with `ai.generateImage`, return an `ImageMessagePart`, render it. the example has a mock fallback so it runs without a key.
- the `thread-message-like` validation relax to accept `https://` and `blob:` URLs. that one is a real gap; assistant-side URL images were being dropped before.

what it drops:

- `useImageGeneration` / `useImagePartRegenerate` / `useImagePartDownload` / `useImagePartCopy`. download and copy are generic browser utilities (now inlined in the component), and `useImageGeneration` was a thin wrapper that didn't integrate with the runtime, so none of it needed to be published API.
- `createAiSdkImageAdapter` / `createMockImageAdapter`. the adapter is a thin pass-through over `ai.generateImage`, which apps can call directly (the example does).
- the `ImageMessagePart` metadata fields (`prompt`, `model`, `seed`, …) and their `auiV0` persistence. nothing in the kept surface consumed most of them, so it was public type surface with no current consumer.

net result: `@assistant-ui/react` and `@assistant-ui/react-ai-sdk` are untouched, and the only published change is a `@assistant-ui/core` patch for the validation relax. i verified the example end to end in a browser (generate, regenerate, the actions, and the mock fallback all work).

if what you were after is a runtime-integrated regenerate, a first-class primitive that reads the image part from scope and writes the new result back into the thread, that's a genuinely useful feature and a different thing from these standalone hooks. i'd be glad to spec it out with you as a scoped, tap-aware primitive if that's the direction you want.

PR #3887, opened 2026-04-25
feat: add chat-test-kit (v1)

ShobhitPatra:
## Summary

This PR adds Chat Test Kit: a transcript-driven testing framework for assistant-ui chat applications.

The goal is to make chat UI tests deterministic. Instead of depending on live model calls, network timing, or hand-rolled mocks, tests define a conversation transcript and replay it through an assistant-ui runtime. This lets app developers assert streaming text, tool calls, cancellation, transport failures, and rendered UI states with repeatable timing.

This is not an eval framework for judging model quality. It is for UI behavior tests: “given this user message and this assistant/tool timeline, did the app render and behave correctly?”

This PR contains the initial v1 scope for the package pair.

  ## What This Adds

  - `@assistant-ui/chat-test-kit-core`
    - Portable transcript format and chainable `transcript()` DSL
    - Canonical JSON format with published JSON Schema
    - Runtime-agnostic replayer state machine
    - Virtual clock for deterministic stream timing
    - Failure injections for cancel, disconnect, transport error, interrupt, and abort/restart
    - `RuntimeAdapter` interface and headless reference adapter

  - `@assistant-ui/chat-test-kit-react`
    - React harness via `createChatTestHarness`
    - assistant-ui adapter built on public runtime APIs
    - A-tier testing API for normal test flows
    - C-tier controls for stepping chunks, tool calls, and failures
    - Seven chat-aware matchers for thread/message/tool-call assertions
    - `setupChatTestKit()` matcher registration helper
    - Core transcript APIs re-exported for single-package installs

  - Dogfood tests in examples
    - `examples/with-ai-sdk-v6`
    - `examples/with-langgraph`
    - `examples/with-assistant-transport`

  - Documentation
    - Quickstart guide in the docs app
    - Transcript format reference in the docs app
    - Package READMEs for core and react packages
    - Changesets for both published packages

  ## Testing

  - `pnpm --filter @assistant-ui/chat-test-kit-core test`
  - `pnpm --filter @assistant-ui/chat-test-kit-react test`
  - `pnpm --filter with-ai-sdk-v6 test`
  - `pnpm --filter with-langgraph test`
  - `pnpm --filter with-assistant-transport test`

### DIFF
--- a/.changeset/chat-test-kit-core.md
+++ b/.changeset/chat-test-kit-core.md
@@ -1,0 +1,7 @@
+---
+"@assistant-ui/chat-test-kit-core": patch
+---
+
+feat: add chat-test-kit-core package
+
+Adds the transcript format (v0 grammar, 12 verbs), the `transcript()` DSL, canonical JSON + JSON Schema, parser, replayer state machine, virtual clock, failure injection (cancel, interrupt, transportError, disconnect, abortAndRestart), `RuntimeAdapter` interface, and headless reference adapter.
--- a/.changeset/chat-test-kit-react-initial.md
+++ b/.changeset/chat-test-kit-react-initial.md
@@ -1,0 +1,7 @@
+---
+"@assistant-ui/chat-test-kit-react": patch
+---
+
+feat: initial release of chat-test-kit-react (M2).
+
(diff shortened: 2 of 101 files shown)


okisdev:
thanks for putting this together, Shobhit. closing with context so the reasoning is recorded rather than discovered on a second attempt.

testing infrastructure isn't where we're investing right now. our experience has been that chat UI quality is mostly determined by interaction feel, streaming pacing, and edge cases that surface incidentally; passing test suites and a good chat experience correlate weakly. we've watched green test suites coexist with regressed UX more than once, and a transcript replayer + matcher set tends to lock in the shape of what we test before we've learned what's worth defending.

the line we hold internally is smoke tests on individual functions only. anything that asserts on a UI component, on the runtime as a whole, or on an application flow gets cut, because it stops being a regression net and starts being a constraint on iteration.

none of this reflects on the implementation, which is solid. but landing it in the monorepo would signal a direction we're choosing not to commit to, and pull review and maintenance bandwidth away from where we want to spend it. if you'd like to keep developing it, a standalone community package would let you iterate freely; we're happy to link to it from docs once it stabilizes.

closing as not planned. thanks again for the effort and the writeup.

PR #3916, opened 2026-04-29
fix(core): drop "./*" wildcard export

Yonom:
## Summary

The `"./*"` wildcard in `@assistant-ui/core`'s `exports` field was leaking the entire `dist/` tree as public subpaths — including internal test helpers (`@assistant-ui/core/tests/...`) and internal type files (`@assistant-ui/core/types/...`). Removing it.

- **Drop the wildcard** from `packages/core/package.json`.
- **Re-export `PendingAttachmentStatus` and `CompleteAttachmentStatus`** from `@assistant-ui/core` root. Without this, three hooks in `packages/react/src/legacy-runtime/hooks/AttachmentContext.ts` would emit TS2883 ("inferred type cannot be named without a reference to internal path"). These two types were previously only reachable through the wildcard.
- **Inline `makeAdapter` into a local helper in `packages/react`**. It was the sole external consumer of the now-removed `@assistant-ui/core/tests/remote-thread-list-test-helpers` subpath; the function is small and self-contained, so duplicating it inside the package that needs it is cleaner than exposing test helpers as a public API surface.

## Test plan

- [x] `pnpm turbo build --filter=@assistant-ui/core --filter=@assistant-ui/react --filter=@assistant-ui/react-native --filter=@assistant-ui/react-ink --filter=@assistant-ui/store` — clean (verified in a worktree from main; the workspace had unrelated WIP from another virtual branch).
- [x] `pnpm --filter @assistant-ui/react test` — 17 files / 243 tests pass.
- [x] `pnpm --filter @assistant-ui/core test` — 20 files / 157 tests pass.
- [x] `pnpm lint` — clean.
- [x] Audited every cross-package import of `@assistant-ui/core/<subpath>` in the monorepo. The only wildcard-reliant consumer was the test helper now inlined; all other subpaths are explicit (`./internal`, `./store`, `./store/internal`, `./react`).

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -32,7 +32,9 @@ export type {
 export type {
   Attachment,
   PendingAttachment,
+  PendingAttachmentStatus,
   CompleteAttachment,
+  CompleteAttachmentStatus,
   AttachmentStatus,
   CreateAttachment,
 } from "./types/attachment";
--- a/packages/react/src/tests/remote-thread-list-test-helpers.ts
+++ b/packages/react/src/tests/remote-thread-list-test-helpers.ts
@@ -1,0 +1,33 @@
+import { vi } from "vitest";
+import type { RemoteThreadListAdapter } from "../index";
+
+export function makeAdapter(
+  overrides: Partial<RemoteThreadListAdapter> = {},
+): RemoteThreadListAdapter {
+  return {
+    list: vi.fn(async () => ({ threads: [] })),
+    initialize: vi.fn(async (threadId: string) => ({
+      remoteId: threadId,
+      externalId: threadId,
+    })),
+    rename: vi.fn(async () => {}),
+    archive: vi.fn(async () => {}),
+    unarchive: vi.fn(async () => {}),
+    delete: vi.fn(async () => {}),
+    generateTitle: vi.fn(
+      async () =>
+        new ReadableStream({
+          start(c) {
+            c.close();
+          },
+        }) as never,
+    ),
+    fetch: vi.fn(async (id: string) => ({
+      status: "regular" as const,
+      remoteId: id,
+      externalId: id,
+      title: "Test",
+    })),
+    ...overrides,
+  };
+}
--- a/.changeset/core-drop-wildcard-export.md
+++ b/.changeset/core-drop-wildcard-export.md
@@ -1,0 +1,9 @@
+---
+"@assistant-ui/core": patch
+---
+
+chore: drop `./*` wildcard export and surface internal attachment status types
+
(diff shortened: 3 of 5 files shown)


packages/core/src/index.ts
```diff
@@ -32,7 +32,9 @@ export type {
 export type {
   Attachment,
   PendingAttachment,
+  PendingAttachmentStatus,
```
Yonom, line 35: unrelated change and why even do this?

packages/react/src/tests/remote-thread-list-test-helpers.ts
```diff
@@ -0,0 +1,33 @@
...
+    ),
+    fetch: vi.fn(async (id: string) => ({
+      status: "regular" as const,
+      remoteId: id,
+      externalId: id,
+      title: "Test",
+    })),
+    ...overrides,
+  };
+}
```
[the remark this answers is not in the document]
Yonom: The previous iteration of this PR did exactly that — added `./tests/remote-thread-list-test-helpers` as an explicit subpath in `core/exports`. The maintainer pushed back: test helpers should not be part of the package's public API surface. Inlining a small (~30 LOC) helper into the consuming package is the lighter-weight option per project preference.

Duplication risk is bounded: both copies import `RemoteThreadListAdapter` from the same package, so any change to that interface produces a TypeScript error at both call sites — divergence is visible immediately rather than silent.

packages/core/src/index.ts
Yonom: Not unrelated — load-bearing for the wildcard removal. Without these two re-exports, the build of `@assistant-ui/react` fails with TS2883 in three places in `legacy-runtime/hooks/AttachmentContext.ts`:

```
src/legacy-runtime/hooks/AttachmentContext.ts:91 useThreadComposerAttachment
src/legacy-runtime/hooks/AttachmentContext.ts:94 useEditComposerAttachment
src/legacy-runtime/hooks/AttachmentContext.ts:97 useMessageAttachment
  → The inferred type cannot be named without a reference to
    'PendingAttachmentStatus' / 'CompleteAttachmentStatus' from
    '../../../node_modules/@assistant-ui/core/dist/types/attachment'.
    This is likely not portable. A type annotation is necessary.
```

Those two types were previously namable only through the `./*` wildcard at `@assistant-ui/core/types/attachment`. Re-exporting them from the package root makes the inferred return types portable as `import("@assistant-ui/core").PendingAttachmentStatus`.

Verified empirically: removing the wildcard with no re-export → 6 TS2883 errors; adding the re-exports → all builds clean. The alternative is annotating each of the three hooks with explicit return types, which is more invasive.

Yonom: LGTM

PR #3920, opened 2026-04-30
feat: add video message rendering support

ShobhitPatra:
## Summary

  Adds first-class video message part support across assistant-ui packages.

  - Adds `VideoMessagePart` types, exports, normalization, attachments, cloud serialization, and assistant-stream accumulation.
  - Adds web, React Native, and Ink video rendering support, including `MessagePartPrimitive.Video` and default/custom renderer wiring.
  - Adds protocol/runtime adapter support for A2A, AI SDK, Google ADK, LangChain, LangGraph, and OpenCode video parts.
  - Adds shadcn-style video UI, attachment labeling, docs page, sample video assets, and a patch changeset.


okisdev:
hi @ShobhitPatra, first, genuine thanks. this is a lot of careful work: 70+ files, real test coverage, adapters wired across every protocol, and you carried it through several review rounds with CI green. i don't want this to read as a knock on the execution, because the execution is solid.

we're going to close it though, and the reason is a direction call, not anything about the code.

we don't want to keep adding a dedicated message part per media type. image, audio, now video, every new primitive is permanent surface area we have to define, normalize, serialize, and maintain across core, react, react-native, ink, and every runtime adapter, forever. that cost compounds with each one, and it pushes the library toward trying to model every possible content shape ourselves, which isn't a game we can win.

the direction we landed on after discussion is the opposite: shrink the primitive set, ideally down to just `file` and `data`. a video already is a `file` part today; `mimeType: "video/*"`, `data`, and `filename` are all there. dropping a `<video>` on screen doesn't need its own part type, it's a `file` whose mime starts with `video/`, and a renderer can branch on that. anything more structured is a `data` part.

the thinking behind it: people's needs here are too varied for us to model each one as a built in. a small, simple hook API that lets you read a `file` or `data` part and render exactly what you want is more powerful than a growing pile of typed primitives that each lock you into our shape. we'd rather hand you the hook than the constraint.

so this won't get merged, but please don't read it as wasted effort or a reflection on you. the work is good; it's the surface area we're choosing not to take on. genuinely hope you'll send more.

PR #3926, opened 2026-04-30
fix: recover composer after dropped compositionend

### LINKED ISSUE #3923: ComposerPrimitive.Input freezes when compositionend is dropped (dead-key layouts)
### Checklist

### Environment info

Environment:
  OS:               macOS 15.4.1 (x64)
  Node.js:          v23.11.0
  Package Manager:  yarn 4.13.0
  Framework:        Unknown

Packages:
  @assistant-ui/react           0.12.27
  @assistant-ui/core            0.1.16
  @assistant-ui/store           0.2.9
  @assistant-ui/tap             0.5.10
  assistant-stream              0.3.12
  assistant-cloud               0.1.22
  @assistant-ui/react-ai-sdk    1.3.21
  @assistant-ui/react-markdown  0.12.11
  @assistant-ui/react-devtools  1.0.6

Ecosystem:
  react       18.3.1
  react-dom   18.3.1
  ai          6.0.169
  zod         4.3.6
  zustand     5.0.12
  typescript  5.9.3

### Affected area(s)

React / UI Components

### Runtime adapter in use

### Steps to reproduce

1. macOS → System Settings → Keyboard → Input Sources → add German (or U.S. International - PC) and switch to it.
2. Open any page using <ComposerPrimitive.Input asChild>.
3. Click into the composer.
4. Type ` (the dead-key grave). Then try to type any letter.

#### Expected behavior

Typing characters into `<ComposerPrimitive.Input>` always updates the composer text and renders into the input. A missed `compositionend` event (whether from a Chrome quirk on dead-key layouts, an unfocus mid-composition, or any other browser edge case) should self-correct on the next user action — it should not permanently disable the input.

For dead-key layouts specifically: typing `` ` `` should resolve to a literal backtick (or combine with the next letter as an accent), and subsequent keystrokes should continue to register normally.

#### Actual behavior

After a single `compositionstart` without a matching `compositionend`, the input is permanently frozen for the lifetime of the component:

- The textarea keeps focus.
- `keydown` / `keypress` / `input` events still fire on the DOM node.
- The browser-level `value` of the textarea may even update on each keystroke.
- But the React `onChange` handler in `ComposerPrimitive.Input` short-circuits (`if (isComposing) return`), so `aui.composer().setText` is never called.
- As a result: nothing renders to the user, the runtime's composer text stays at whatever it was when composition started, the Send button stays disabled, and the user sees a dead input that they can only recover from by reloading the page or switching threads.

For end users on German / U.S. International / French / Spanish ISO layouts, this happens by accident — typing `` ` `` (which on a German keyboard is right next to the question-mark key) intermittently triggers it and makes the chat unusable.

### Browser (if applicable)

Chrome (Stable) on macOS.

arjunrawal1:
## Summary

Recover `ComposerPrimitive.Input` when a browser drops `compositionend` by treating the next non-composition input as the end of composition and syncing the textarea value back into composer state.

## Changes

- Detect active composition from the native event (`isComposing` or `insertCompositionText`) instead of permanently trusting the internal composition ref in `onChange`.
- Reset the composition ref on the next non-composition input so dead-key/browser edge cases self-correct.
- Add regression coverage for a dropped `compositionend` and for still ignoring active composition text.

## Testing

- `pnpm exec biome check packages/react/src/primitives/composer/ComposerInput.tsx packages/react/src/primitives/composer/ComposerInput.test.tsx`
- `pnpm --filter @assistant-ui/react test src/primitives/composer/ComposerInput.test.tsx`

Fixes assistant-ui/assistant-ui#3923

### DIFF
--- a/packages/react/src/primitives/composer/ComposerInput.test.tsx
+++ b/packages/react/src/primitives/composer/ComposerInput.test.tsx
@@ -1,0 +1,103 @@
+// @vitest-environment jsdom
+
+import { fireEvent, render, screen } from "@testing-library/react";
+import type { FC, PropsWithChildren } from "react";
+import { describe, expect, it } from "vitest";
+import type { AssistantRuntime } from "@assistant-ui/core";
+import { AssistantRuntimeProvider } from "../../legacy-runtime/AssistantRuntimeProvider";
+import { useAssistantRuntime } from "../../legacy-runtime/hooks/AssistantContext";
+import { useLocalRuntime } from "../../legacy-runtime/runtime-cores/local/useLocalRuntime";
+import type { ChatModelAdapter } from "../../index";
+import { ComposerPrimitiveInput } from "./ComposerInput";
+
+const noOpAdapter: ChatModelAdapter = {
+  async *run() {},
+};
+
+const RuntimeProvider: FC<PropsWithChildren> = ({ children }) => {
+  const runtime = useLocalRuntime(noOpAdapter);
+  return (
+    <AssistantRuntimeProvider runtime={runtime}>
+      {children}
+    </AssistantRuntimeProvider>
+  );
+};
+
+const RuntimeCapture: FC<{
+  runtimeRef: { current: AssistantRuntime | null };
+}> = ({ runtimeRef }) => {
+  runtimeRef.current = useAssistantRuntime();
+  return null;
+};
+
+const renderInput = () => {
+  const runtimeRef: { current: AssistantRuntime | null } = { current: null };
+  render(
+    <RuntimeProvider>
+      <ComposerPrimitiveInput aria-label="Message" />
+      <RuntimeCapture runtimeRef={runtimeRef} />
+    </RuntimeProvider>,
+  );
+  return runtimeRef;
+};
+
+const setNativeInputValue = (element: HTMLTextAreaElement, value: string) => {
+  const setter = Object.getOwnPropertyDescriptor(
+    HTMLTextAreaElement.prototype,
+    "value",
+  )?.set;
+  setter?.call(element, value);
+};
+
+const input = (
+  element: HTMLTextAreaElement,
+  value: string,
+  init?: { isComposing?: boolean; inputType?: string },
+) => {
+  setNativeInputValue(element, value);
+  fireEvent(
+    element,
+    new InputEvent("input", {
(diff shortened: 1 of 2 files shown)


okisdev:
thanks for picking this up. the recovery direction is right, and using `inputType === "insertCompositionText"` as a secondary signal alongside `isComposing` is a nice touch.

a few things worth flagging:

**1. the fix is incomplete; it leaves the second half of #3923 unaddressed.**

the proposed change keeps the original "skip `setText` during composition" branch:

```ts
if (isCompositionInput) return;
```

`ComposerPrimitive.Input` is a controlled `<textarea>` whose `value` prop comes from store state. if `setText` is not called while composing, the prop stays stale, and React reconciles the textarea back to the stale value on the next render, wiping in-progress IME characters. this is the canonical React + IME footgun (`facebook/react#8683`) and shows up most reliably under React 19 (which assistant-ui supports per `peerDependencies: react: "^18 || ^19"`). on Korean IME, typing `가` simply shows nothing.

the correct shape is to call `setText(e.target.value)` unconditionally (the DOM value is already what the user just typed, so syncing it back cannot interfere with the IME), and only gate the downstream side effects (plugin cursor tracking) on `isComposing`.

**2. missing changeset.**

the bot already flagged this. per `AGENTS.md`, every PR touching a published package needs a `patch` changeset.

**3. PR title scope.**

the project convention for changes in `packages/react` is `fix(react): ...` (see #2874, #3583, #3870). `fix:` alone reads as cross-cutting in changelogs.

**4. test gaps.**

two specific holes:

- the "recovers" test does not actually exercise the recovery line. the new `onChange` decides composition state purely from the native event; it never reads `compositionRef.current` anymore. removing `compositionRef.current = false` would still leave this test passing. that line is only meaningful for `onSelect` (line 344), which is the handler that still depends on the ref, and `onSelect` is not exercised in the suite.
- no CJK IME coverage. dead keys are one trigger; CJK IMEs (Korean / Japanese / Chinese pinyin) are the larger user-facing surface.

closing in favor of #3924, which already covers all of the above (`act + createRoot` matching `ComposerAttachmentDropzone.test.tsx`, strict per-call `setText` counts, an isolated assertion on the `compositionRef` recovery path, and CJK + dead-key + `isEditing=false` cases). thanks again for the contribution.

PR #3940, opened 2026-05-02
feat(react-ink): add ThreadPrimitive viewport support

ShobhitPatra:
## Summary

  - Add `ThreadPrimitive.Viewport`, `ViewportProvider`, and `ScrollToBottom` for bounded terminal thread history
  - Integrate `ThreadPrimitive.Messages` with viewport state while preserving unbounded behavior outside a viewport
  - Add Ink docs, tests, and a patch changeset for `@assistant-ui/react-ink`

### DIFF
--- a/.changeset/react-ink-thread-viewport.md
+++ b/.changeset/react-ink-thread-viewport.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react-ink": patch
+---
+
+feat(react-ink): add thread viewport primitives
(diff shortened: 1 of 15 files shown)


ShobhitPatra:
Closing. #3966 shipped while this was open and addresses the actual long-thread pain (perf, terminal scrollback) on the same render. The remaining surface area here: bounded viewport, PageUp/PageDown, ScrollToBottom, paused hint  is only useful for apps embedding the thread in a fixed-height pane, which no one has asked for yet.

No concrete plan to re-open. If the embedded-pane use case comes up, the right approach is a smaller PR adding row-level scroll on top of #3966's windowing, not a rebase of this one.

PR #3960, opened 2026-05-05
feat(react-ag-ui): support AG-UI interrupt lifecycle

### LINKED ISSUE #3959: Support AG-UI interrupt-aware run lifecycle in react-ag-ui
### Problem statement

I'm trying to implement human-in-the-loop tool approvals using the AG-UI interrupt spec (`RunFinishedEvent.outcome = { type: "interrupt", interrupts: [...] }` + `RunAgentInput.resume`), but currently `react-ag-ui` doesn't expose interrupt data to the React app.

The event parser in `run-aggregator` strips `outcome` from `RUN_FINISHED` events (only `runId` is extracted), and `CUSTOM` events are dispatched through the subscriber but silently ignored by the aggregator's default case. This means there's no way to detect that a run ended with an interrupt or to render approval UI in response.

The AG-UI core SDKs (TypeScript and Python) now ship native `Interrupt`, `ResumeEntry`, and `RunFinishedOutcome` types, so the protocol layer is ready:

- ag-ui/client version 0.0.53 
- [Related ag-ui commit](https://github.com/ag-ui-protocol/ag-ui/commit/df613e40b857668be9a8986aa7ee388fe65aee31)

### Proposed solution

1. Parse `outcome` from `RUN_FINISHED` events in the event parser
2. When `outcome.type === "interrupt"`, expose the interrupts array through the runtime (e.g. a hook like `useInterrupts()` or a new message status like `{ type: "requires-action", reason: "interrupts", interrupts: [...] }`)
3. Provide a way to submit resume responses (analogous to `addResult` for human tools), which sends the next `RunAgentInput` with the `resume` array populated

This would let apps render approval/confirmation UI when the agent pauses for human input, and resume the run with the user's decisions.

### Area

Runtime / Core

### Alternatives considered

We're currently considering working around this by emitting interrupt data as a `CustomEvent` before the `RunFinishedEvent` and overriding `HttpAgent.runAgent` to wrap the subscriber with a custom `onCustomEvent` callback that captures the data outside of `react-ag-ui`. While this might work it bypasses the framework's event/rendering lifecycle entirely.

### Additional context

- AG-UI interrupt spec: https://docs.ag-ui.com/concepts/interrupts
- The spec supports parallel interrupts (multiple approval requests in one run), `reason` taxonomy (`tool_call`, `input_required`, `confirmation`), and approve-with-edits via `responseSchema`

ShobhitPatra:
Closes https://github.com/assistant-ui/assistant-ui/issues/3959
  ## Summary

  - add local AG-UI interrupt lifecycle contracts until the SDK exports official types
  - preserve `RUN_FINISHED.outcome` and map interrupt outcomes to `requires-action: interrupt`
  - expose `useAgUiInterrupts()` and `useAgUiResumeInterrupts()`
  - send `RunAgentInput.resume` for resolved/cancelled interrupt entries

  ## Test Plan

  - [x] `pnpm --filter @assistant-ui/react-ag-ui test`
  - [x] `pnpm --filter @assistant-ui/react-ag-ui exec tsc --noEmit`
  - [x] `pnpm --filter @assistant-ui/react-ag-ui build`


jnissin:
Thanks for the quick work, this would save us from using an awkward workaround 🙏 🚀

okisdev:
hi @ShobhitPatra, this PR is superseded by #3974, which landed on main for #3959 shortly before review here wrapped up.

the two PRs converged on different shapes:

- #3974 stores interrupts on `metadata.custom.agui.interrupts` and exposes them via `unstable_getPendingInterrupts` / `unstable_submitInterruptResponses` on the runtime.
- this PR uses a `pendingInterrupts` field on the core plus `useAgUiInterrupts` / `useAgUiResumeInterrupts` hooks backed by `@assistant-ui/store` and the `extras` channel.

most of the wire-level coverage (outcome parsing, `requires-action.reason: "interrupt"` state, `RunAgentInput.resume` forwarding, `onRunFinishedEvent` wiring) is now on main via #3974, so the bulk of this PR overlaps. cherry-picking the remaining commits was not viable: every shared file differs at the architecture layer (`pendingInterrupts` class field vs. metadata namespace, hooks via `@assistant-ui/store` vs. `unstable_*` runtime methods).

what this PR genuinely added that #3974 did not:

1. `assertNoPendingInterrupts` on `append` / `reload` / `resume` (AG-UI interrupts spec rule 4 enforced on the client)
2. `AgUiInterruptReason` union (`"tool_call" | "input_required" | "confirmation" | (string & {})`)
3. `onRunFinishedEvent` narrow on the parsed event type, so a misrouted payload no longer suppresses the `onRunFinalized` fallback

those three are now in #3977 with you as `Co-authored-by` on the feat commit. please close this PR in favor of #3977.

PR #3966, opened 2026-05-06
perf(react-ink): virtualize message list + diff-aware reconciler

samdickson22:
## Summary

Two internal changes that drop per-frame work in long ink threads. No public API change — all new knobs are optional and backward-compatible.

- `ThreadPrimitive.Messages` now accepts optional `windowSize` / `windowOverscan`. When set, only the most recent `windowSize + windowOverscan` messages stay mounted and subscribed; older messages remain in the terminal scrollback above the live region but no longer contribute to per-token reconciliation.
- Each rendered message is wrapped in a memoized boundary keyed by `(index, render)`. Re-renders triggered by streaming a single message no longer walk every other message's subtree. With a stable render callback (the common case), this collapses an O(n) reconcile per token into O(1).

## Microbenchmark

`packages/react-ink/benchmarks/long-thread.bench.tsx` — 1000 messages, ~10 s of streaming at ~62 tokens/sec, three runs each:

| metric        | baseline | windowed (50) | delta |
|---------------|----------|---------------|-------|
| frames        | ~600     | ~600          | flat  |
| mean ms/frame | ~17      | ~17           | flat  |
| peak ms/frame | ~78      | ~22           | -71%  |

Peak frame time is the user-visible "stutter" metric; cutting it ~3.5× is what eliminates the flicker reported in adjacent ink-on-terminal projects.

## Test plan

- [x] `pnpm --filter @assistant-ui/react-ink build` — clean
- [x] `pnpm --filter @assistant-ui/react-ink test` — 46/46 passing
- [ ] Manual: render a 1000-message thread with `windowSize={50}` and confirm scrollback contains older messages while the live region only mounts the recent window
- [ ] Manual: existing consumers without `windowSize` see no behavior change

## Backward compatibility

- `windowSize` defaults to undefined → render all messages (legacy behavior)
- `windowOverscan` defaults to 4 (no effect when `windowSize` is unset)
- `MemoMessage` is internal; the only external visible change is per-message memoization, which is invisible when render callbacks are stable

### DIFF
--- a/.changeset/perf-react-ink-virtualization.md
+++ b/.changeset/perf-react-ink-virtualization.md
@@ -1,0 +1,30 @@
+---
+"@assistant-ui/react-ink": patch
+---
+
+perf: virtualized message list and diff-aware reconciler for long ink threads
+
+Two internal changes that drop per-frame work in long ink threads. No public
+API change — all new knobs are optional and backward-compatible.
+
+- `ThreadPrimitive.Messages` now accepts `windowSize` / `windowOverscan`. When
+  set, only the most recent `windowSize + windowOverscan` messages stay
+  mounted and subscribed; older messages remain in the terminal scrollback
+  but no longer contribute to per-token reconciliation. Defaults preserve
+  legacy behavior (render all).
+- Each rendered message is wrapped in a memoized boundary keyed by `(index,
+  render)`. Re-renders triggered by streaming a single message no longer
+  walk every other message's subtree. With a stable render callback, this
+  collapses an O(n) reconcile per token into O(1).
+
+Microbenchmark (`benchmarks/long-thread.bench.tsx`, 1000 messages, ~10 s of
+streaming at 62 tokens/sec, three runs each):
+
+| metric        | baseline | windowed (50) | delta |
+|---------------|----------|---------------|-------|
+| frames        | ~600     | ~600          | flat  |
+| mean ms/frame | ~17      | ~17           | flat  |
+| peak ms/frame | ~78      | ~22           | -71%  |
+
+Peak frame time is the user-visible "stutter" metric; cutting it ~3.5× is
+what eliminates the flicker reported in adjacent ink-on-terminal projects.
--- a/.changeset/perf-react-ink-virtualization.md
+++ b/.changeset/perf-react-ink-virtualization.md
@@ -9,22 +9,29 @@ API change — all new knobs are optional and backward-compatible.
 
 - `ThreadPrimitive.Messages` now accepts `windowSize` / `windowOverscan`. When
   set, only the most recent `windowSize + windowOverscan` messages stay
-  mounted and subscribed; older messages remain in the terminal scrollback
-  but no longer contribute to per-token reconciliation. Defaults preserve
-  legacy behavior (render all).
+  mounted and subscribed; messages above the live region graduate into Ink's
+  `<Static>`, which writes them to stdout exactly once and lets the terminal
+  commit them to scrollback before unmounting the React subtree. Per-token
+  reconciliation work is bounded by the window size regardless of thread
+  length, while scrolled-past output remains in the terminal backbuffer for
+  native scroll. Defaults preserve legacy behavior (render all dynamically).
 - Each rendered message is wrapped in a memoized boundary keyed by `(index,
   render)`. Re-renders triggered by streaming a single message no longer
   walk every other message's subtree. With a stable render callback, this
   collapses an O(n) reconcile per token into O(1).
 
-Microbenchmark (`benchmarks/long-thread.bench.tsx`, 1000 messages, ~10 s of
-streaming at 62 tokens/sec, three runs each):
+Microbenchmark (`benchmarks/long-thread.bench.tsx`, 1000-message thread,
+~5 s of streaming at 62 tokens/sec, append-and-mutate per token, two
+trials each, averaged):
 
-| metric        | baseline | windowed (50) | delta |
-|---------------|----------|---------------|-------|
-| frames        | ~600     | ~600          | flat  |
-| mean ms/frame | ~17      | ~17           | flat  |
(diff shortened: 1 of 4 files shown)


ShobhitPatra (commented):
Hey @samdickson22 , nice work on the memo boundary, that part looks clean .                                                                      
                                                                                                                                                                             
On windowSize though - it's framed as scrollback-preserving, but the code just renders the tail. If someone opens a saved thread with 1000 messages and `windowSize={50}`, the first 950 never get written to the terminal at all, so they can't end up in scrollback. They just disappear on open.                                                   
                                                                                 
Renaming this to something like `tailSize` would be more accurate about what it does, but I’m not sure we want a public option that drops earlier history on initial render. One option could be rendering the prefix once on mount so Ink commits it to scrollback, then trimming after.
          
Worth re-running the benchmark with just the memo change too - curious how much of the peak-frame win is from windowing vs the memo alone.                                 

@okisdev wdyt ?

samdickson22:
> Hey @samdickson22 , nice work on the memo boundary, that part looks clean .
> 
> On windowSize though - it's framed as scrollback-preserving, but the code just renders the tail. If someone opens a saved thread with 1000 messages and `windowSize={50}`, the first 950 never get written to the terminal at all, so they can't end up in scrollback. They just disappear on open.
> 
> Renaming this to something like `tailSize` would be more accurate about what it does, but I’m not sure we want a public option that drops earlier history on initial render. One option could be rendering the prefix once on mount so Ink commits it to scrollback, then trimming after.
> 
> Worth re-running the benchmark with just the memo change too - curious how much of the peak-frame win is from windowing vs the memo alone.
> 
> @okisdev wdyt ?

good shout

samdickson22:
@ShobhitPatra ty for the catch — both fair calls. Pushed `0ffd810c6` addressing them.

**Prefix → terminal scrollback.** `windowSize` no longer drops earlier history. The prefix is now flushed through Ink's `<Static>`: Ink writes those messages to stdout exactly once, the terminal commits them to scrollback, and then the React subtree unmounts. Opening a 1000-message saved thread with `windowSize={50}` writes all 950 prefix messages on mount and only keeps the tail of 50+overscan live. One tradeoff worth flagging: messages above the live region are read-only from the renderer's perspective once committed — edits to scrolled-past messages won't repaint until the thread re-opens. That matches terminal scrollback semantics (committed history is committed) and the doc comment on the prop calls it out.

**Memo vs windowing attribution.** Re-ran the bench (1000 messages, ~5s of streaming at 62 tok/s, append + per-token mutation, two trials averaged):

| mode          | mean ms/frame | peak ms/frame |
|---------------|---------------|---------------|
| legacy        | 231           | 608           |
| memo only     | 189 (-18%)    | 476 (-22%)    |
| windowed (50) | 110 (-53%)    | 252 (-59%)    |

Memo alone trims ~22% off peak; windowing on top of memo adds another ~47%. Both pull their weight.

The numbers are more conservative than the original changeset's -71% peak — the prior bench routed updates through provider-prop re-renders which masked most of the per-mode delta. Bench harness now mounts the readonly thread runtime once and drives `core.setMessages` directly, exercising the production update path.

okisdev (approved):

okisdev:
@samdickson22 — overall this is a solid piece of work. the `<Static>`-based handoff for the windowed prefix is the right architectural call for terminal scrollback (much cleaner than rolling a virtual-list of our own), the 3-mode bench (legacy / memo / windowed) actually makes the perf story verifiable instead of a single number, and gating the new behaviour behind opt-in props with legacy defaults is the right API choice. main things i'd push back on are the components-deps memo regression that cubic and greptile both flagged (which silently degrades the memo for the common inline-`components={{ ... }}` usage) and the lack of test coverage on the new windowing math; both addressed on `pr/3966` on origin.

@ShobhitPatra, good catches. inline notes:

**scrollback semantics:** i think the prefix actually does land in scrollback on initial mount. ink's `<Static>` starts its cursor at 0, so the first render passes the entire prefix slice through `render()` and writes it via `internal_static`; only after the layout effect does the cursor advance to `items.length` and the subtree unmount. on a saved 1000-message thread with `windowSize={50}` you'd get 946 lines emitted at mount, the terminal commits them to its backbuffer as they scroll past, and the react tree only keeps the trailing 54. the new test asserts this split: `staticItemCounts.at(-1)` matches `tailStart`, prefix captures land in `[0..tailStart-1]`, tail captures cover the rest.

**naming** is fair though: `windowSize` reads like "viewport size" in virtualised-list parlance. happy to rename to `tailSize` if you prefer.

**on the bot findings:**

- `components` in the `stableComponents` `useMemo` deps (cubic + greptile P2) is real. inline `components={{ Message: MyMessage }}` literals were busting the memo even though the field-level destructure was supposed to keep it stable. dropped `components` from the deps with a `biome-ignore` covering the suppression; there's a regression test for the inline-literal path now.
- `sample()` frame undercount (greptile P2) is already addressed in the current code: the loop distributes elapsed wall time across `newFrames` entries (`for (i...) frameTimes.push(perFrame)`), so peak ms is per-flush, not per-poll-tick. that comment looks like it was on an earlier diff.

**memo-only attribution** (3-mode bench, 2 trials each, fresh run on the parallel branch):

| mode          | mean  | peak  |
|---------------|-------|-------|
| legacy        | 96ms  | 205ms |
| memo only     | 74ms (-23%) | 170ms (-17%) |
| windowed 50   | 23ms (-76%) | 55ms (-73%)  |

memo alone gets ~20% off both metrics; the bulk of the peak-frame win is windowing. absolute ms are lower than the changeset table because of machine differences, but the shape lines up.

side polish on the parallel branch: clamp negative window dimensions to 0, `displayName` on the memo boundary for devtools, a `pnpm benchmark` runner that picks up every `*.bench.{ts,tsx}` so future bench files just drop in, plus tests covering windowing math, default overscan, clamps, memo short-circuit, and inline-components stability.

PR #3967, opened 2026-05-06
feat(core, react): add MessagePrimitive.GenerativeUI primitive

samdickson22:
## Summary

A new first-class primitive for rendering agent-described React UI from a JSON spec, with a consumer-provided component allowlist as the security boundary.

```tsx
<MessagePrimitive.Parts
  components={{
    generativeUI: { components: { Card, Button, Stat } },
  }}
/>
```

The agent emits a new `generative-ui` message part containing a tree of components by name; `MessagePrimitive.GenerativeUI` walks the spec and resolves each name against the registry you pass in. Unknown names throw a typed `GenerativeUIRenderError` or invoke the optional `Fallback`. Composes with `MessagePrimitive.Parts` via the new `components.generativeUI` option, supports streaming partial specs, and offers an opt-in `sandbox=\"iframe\"` mode in the web distribution for CSS/DOM isolation on top of the allowlist.

## Changes

**Core**
- New `generative-ui` `MessagePart` type, wired through:
  - `packages/core/src/types/message.ts` and `types/index.ts`
  - `runtime/utils/thread-message-like.ts` (coercion)
  - `react/runtimes/cloud/auiV0.ts` (cloud round-trip)
  - `react/types/MessagePartComponentTypes.ts`
- `GenerativeUIRender` + `MessagePrimitive.GenerativeUI` primitives in `packages/core/src/react/primitives/generativeUI/`
- `GenerativeUIRenderError` typed error + optional `Fallback`
- `MessagePrimitive.Parts` accepts `components.generativeUI`

**Web (`@assistant-ui/react`)**
- Re-exports `GenerativeUIRender`, `GenerativeUISpec`, `GenerativeUINode`, `GenerativeUIComponentRegistry`, `GenerativeUIRenderError`
- `GenerativeUISandboxFrame` adds `sandbox=\"iframe\"` mode for CSS/DOM isolation

**Demo**
- `examples/with-generative-ui-primitive/` — self-contained Next.js example using the registry pattern, no provider keys required

**Docs**
- `apps/docs/content/docs/guides/generative-ui.mdx`

## Security model

The allowlist is the security boundary in the same-realm path. Any spec referencing a name outside the registry throws `GenerativeUIRenderError` (or invokes `Fallback`); no eval, no dynamic import. Set `sandbox=\"iframe\"` to additionally run the rendered tree inside an isolated document — defense in depth on top of the allowlist.

## Test plan

- [x] `pnpm --filter @assistant-ui/core --filter @assistant-ui/react build` — clean
- [x] `pnpm --filter @assistant-ui/react test` — 270/270 passing (includes new `generative-ui.test.tsx`)
- [x] `tsc --noEmit` on the example — clean
- [ ] Manual: run example app, confirm both `GenerativeUIRender` (direct) and `MessagePrimitive.Parts` (via parts) paths resolve the allowlist
- [ ] Manual: confirm unknown component name throws `GenerativeUIRenderError`, and that `Fallback` short-circuits it
- [ ] Manual: with `sandbox=\"iframe\"`, the rendered tree is inside an iframe with isolated styles

## Backward compatibility

- All new APIs are additive; no existing API changes
- Adapters that don't emit `generative-ui` parts are unaffected
- `components.generativeUI` is optional on `MessagePrimitive.Parts`


samdickson22:
Pushed `1ac85c02a` addressing CI failures and cubic feedback:

**CI fix.** `react-langchain`, `react-google-adk`, and `react-langgraph` had hard-coded `_exhaustiveCheck` unions on the user-input message-part `type` discriminant. Extending the union with `"generative-ui"` broke their builds. Runtime behavior here is correct — these adapters reject `generative-ui` because the part is agent-emitted, not user-input — so the fix is just to widen the exhaustive-check union so the discriminant covers it.

**Cubic finding (P2, GenerativeUI.tsx:87).** Flipped prop spread order so the explicit React `key` always wins:

```diff
- { key: key ?? path, ...(props ?? {}) }
+ { ...(props ?? {}), key: key ?? path }
```

Verified locally: 5 packages build clean (`core`, `react`, `react-langchain`, `react-google-adk`, `react-langgraph`); `@assistant-ui/react` test suite 270/270 passing.

samdickson22:
Both Greptile P1s addressed in 52f1af4f3:

1. **Selector stability** — moved the `spec` prop merge out of the `useAuiState` selector in both `MessagePrimitive.GenerativeUI` and `GenerativeUISandboxFrame`. Closing over a per-render prop value inside the selector was a recipe for spurious tearing-detection re-renders in `useSyncExternalStore`. Selector now reads store state only; the prop merge happens outside.

2. **Iframe auto-sizing** — the sandbox iframe was inheriting the browser default (~150px) so any taller generative UI was silently clipped. Added a `ResizeObserver` on the mount target that updates an explicit `height` style as the body grows. SSR-safe (skips when `ResizeObserver` is undefined).

okisdev (approved):
heads up before the summary, since the branch moved a fair bit and one fix is load-bearing.

**where it started.** the primitive worked, but it carried more surface than the pattern needs, and one path was quietly broken:
- a `sandbox="iframe"` mode that only scoped CSS/DOM (not real isolation), with comments that contradicted each other, duplicating what `@assistant-ui/safe-content-frame` already does.
- a redundant `$text` node (a bare string already renders as text) and an unused `version` spec field.
- a second example, `with-generative-ui-primitive`, re-scaffolding ~8 boilerplate files and colliding with the existing `with-generative-ui` (we already brand tool UIs as "Generative UI").
- docs that called the allowlist "the security boundary" without saying the agent controls `props`, which get spread onto your allowlisted components.
- a changeset missing the three adapter packages the PR touches, plus a `(part as any)` cast.
- and the real one: the public web `MessagePrimitive.Parts` dropped the `generativeUI` key when it rebuilt its components map, so `<MessagePrimitive.Parts components={{ generativeUI: { components } }} />`, the primary documented API, rendered nothing no matter what allowlist you passed. it hid because the tests only drove `GenerativeUIRender` directly, which skips that wrapper.

**what changed.**
- merged current main (the branch was ~100 commits behind, before the pnpm 11 migration); resolved the `message.ts` / `MessageParts.tsx` conflicts and healed the lockfile.
- cut the sandbox mode, `$text`, and `version`. the surface is now just the `generative-ui` part type, `GenerativeUIRender`, `MessagePrimitive.GenerativeUI`, and `components.generativeUI` on `Parts`.
- folded the demo into `examples/with-generative-ui` as a `/primitive` route and deleted the duplicate example.
- rewrote the docs security section (the allowlist gates which components render, not the props), completed the changeset, typed the cast.
- fixed the dropped-`generativeUI` bug in the web `Parts` wrapper (one line).
- added store-backed tests that exercise the real public API: `Parts` with an allowlist, the no-allowlist warning path, the `Fallback` path, and the standalone primitive reading the part from scope. those are what surfaced the wrapper bug.

**result.** about 670 lines lighter, the surface matches how every other part type is rendered, the primary `Parts` path works, the security story is honest, and the full react suite is green (323 passing) with the new coverage.

**what's left, all your call, none blocking:**
- error-handling default: an unknown component throws during render. the example passes a `Fallback` so it soft-fails, but you may want the library itself to default to graceful degradation instead of throwing. that's a contract decision.
- prop validation: the allowlist gates names, not props. a per-component schema (zod / standard-schema) would turn the documented caveat into an enforced boundary. it's net-new API, so it's yours to shape.
- naming: this and tool UIs both read as "Generative UI". a cross-link between the two guides (or a more specific name here) would make it clear when to reach for which.

approving on this shape. clean primitive, and it lands in a good place.

Issue #3968, opened 2026-05-06
Runtime Error: MessagePartText can only be used inside text or reasoning message parts and tapClientLookup: Index 5 out of bounds (length: 4)


Alireza-Iranpour:
### Checklist

- [x] I have searched existing issues and this is not a duplicate
- [x] This is a reproducible bug, not a usage question (use [Discord](https://discord.gg/S9dwgCNEFs) for questions)

### Environment info

"@ai-sdk/react": "^3.0.103",
"@assistant-ui/react": "^0.12.28",
"@assistant-ui/react-ai-sdk": "^1.3.21",
"@assistant-ui/react-devtools": "^1.0.6",
"@assistant-ui/react-markdown": "^0.12.11",
"ai": "^6.0.101",
"react": "19.2.3",
"shadcn": "^4.0.5",
"next": "16.1.6",

### Affected area(s)

Thread Management, React / UI Components, Content Rendering (Markdown, Syntax Highlighting)

### Runtime adapter in use

useAISDKRuntime / useChatRuntime

### Steps to reproduce

I use toolUI and thread history using drizzle and my own DB
sometimes when I refresh page or switch between threads I face one of the two issues 

## Error Type
Runtime Error

## Error Message
MessagePartText can only be used inside text or reasoning message parts.

    at MarkdownTextImpl (src/components/assistant-ui/markdown-text.tsx:23:5)
    at children (src/components/assistant-ui/thread.tsx:414:26)
    at AssistantMessage (src/components/assistant-ui/thread.tsx:379:11)
    at ThreadMessage (src/components/assistant-ui/thread.tsx:111:10)
    at children (src/components/assistant-ui/thread.tsx:90:22)
    at Thread (src/components/assistant-ui/thread.tsx:89:13)
    at Assistant (src/components/assistant-ui/modules/assistant.tsx:64:17)
    at ThreadPage (src/app/(chat-aui)/chat/[id]/page.tsx:10:10)

## Code Frame
  21 | const MarkdownTextImpl = () => {
  22 |   return (
> 23 |     <MarkdownTextPrimitive
     |     ^
  24 |       remarkPlugins={[remarkGfm]}
  25 |       className="aui-md"
  26 |       components={defaultComponents}

Next.js version: 16.1.6 (Turbopack)

--------------------------------
## Error Type
Runtime Error

## Error Message
tapClientLookup: Index 5 out of bounds (length: 4)

[...]

PR #3976, opened 2026-05-08
fix(store): suppress zombie-child throws in useAuiState

samdickson22:
## Summary

- `useAuiState` now caches the last successful slice and returns it when a selector throws after the initial render.
- Mirrors react-redux's `useSelector` pattern for the same problem.
- First-call throws still propagate so real bugs are not silently swallowed.

## Why

Fixes the recurring `tapClientLookup: Index N out of bounds (length: M)` runtime error users hit on thread switches and history reloads — see #3968 (filed today, React 19 + Next 16) and #3652 (open since March, React 18 legacy mode).

The throw originates in a stale message child whose selector reads through the proxy into `tapClientLookup.get({ index })` after the parent has already shrunk the indexed list — but before React has unmounted the child. React 18 concurrent mode dodges it via root-first re-render; legacy mode and some React 19 paths still surface the throw because `useSyncExternalStore`'s second `getSnapshot` call (post-`forceStoreRerender`) is uncaught.

This is the classic **zombie-child** problem and the fix shape is well-known: cache the last good snapshot, return it on selector throw, let reconciliation finish, and the zombie unmounts cleanly. react-redux applies the same pattern in `useSelector`.

## Scope

Defensive fix in the consumer hook only. The deeper root-cause fix discussed in #3652 (scope-isolated subscriptions, per @ShobhitPatra's analysis) would touch the core subscription path across multiple files and is intentionally out of scope here — this PR exists so users stop hitting the crash today.

## Test plan

- [x] New unit test simulates a notify-time selector throw and asserts the hook returns the last-good slice rather than propagating the error.
- [x] New unit test asserts a first-call throw still propagates (no silent failures on real bugs).
- [x] Recovery path tested — once the selector stops throwing, fresh values flow again.
- [x] Existing `@assistant-ui/store` tests still pass (8/8).

### DIFF
--- a/.changeset/fix-zombie-child-tapclient-lookup.md
+++ b/.changeset/fix-zombie-child-tapclient-lookup.md
@@ -1,0 +1,20 @@
+---
+"@assistant-ui/store": patch
+---
+
+fix(store): suppress zombie-child throws in `useAuiState`
+
+`useAuiState` now caches the last successful slice and returns it when a
+selector throws after the initial render. This is the same pattern
+react-redux uses for the zombie-child problem.
+
+Fixes the recurring `tapClientLookup: Index N out of bounds (length: M)`
+runtime error reported on thread switches and history reloads (e.g.
+issues #3968, #3652). The throw originates in a stale message child
+whose selector reads through the proxy into `tapClientLookup.get` after
+the parent has already shrunk the indexed list, but before React has
+unmounted the child. Returning the previously-observed slice keeps
+reconciliation moving so the zombie unmounts cleanly.
+
+The first selector call is never suppressed — real bugs surfaced on the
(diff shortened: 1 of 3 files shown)


samdickson22:
Pushed `46517e586` addressing the Greptile P1.

`getServerSnapshot` no longer writes to or reads from `lastSliceRef` — it just returns `selector(proxiedState)` directly. The zombie-child guard now only applies to `getSnapshot`, so:

- A real selector failure on the initial client render (or during SSR) still propagates.
- The cache is populated only by the client subscription path, which is the only place the zombie-child race actually exists.

Comment updated to make the scope explicit.

Yonom:
We decided this is not something we'd want to

PR #4002, opened 2026-05-10
fix(core): render thread list items by stable ID to eliminate zombie-child crash

### LINKED ISSUE #3652: React 17: Switching threads crashes when message components are mounted with `ThreadPrimitive.Messages`  ("tapClientLookup: Index 0 out of bounds")

shashank-100:
Fixes #3652

## Summary

- Keys `ThreadListItems` children by thread ID instead of numeric index
- Adds `ThreadListItemByIdProvider` which uses `item({ id })` instead of `item({ index })`

## Root cause

When a thread is deleted, the store notifies all subscribers simultaneously. React needs to both re-render the parent (to drop the deleted child) and re-run `getSnapshot` on existing children. The zombie child at the now-invalid index fires `getSnapshot` before the parent can unmount it, causing `tapClientLookup` to throw `Index N out of bounds` and crash the tree.

With index-based keys (`key={0}`, `key={1}`, `key={2}`), when a thread is deleted the indexes don't change — React sees the same keys and doesn't unmount anything. With ID-based keys (`key="abc"`), when a thread is deleted its ID disappears from `threadIds` — React sees the key is gone and unmounts the child before `getSnapshot` can fire on it.

## Test plan

- [ ] Delete a thread from the list — no crash, item removed cleanly
- [ ] Switch between threads rapidly — no crash
- [ ] Archived thread list behaves the same

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


Yonom:
Thank you for this PR!

**Note 1:**
There is an accessibility regression with this PR: the users' keyboard tab-focus is reset when they switch between different branches of a message (the Action bar lives inside a message and keying the component by id instead of index causes the action bar identity to no longer be stable)

**Note 2:**
I am still unable to reproduce the zombie child problem. This is frustrating because it makes it impossible to validate the correctness of this PR...  On the first look, it is not clear to me why the zombie child problem would not reappear if the ID of a message stops appearing (as opposed to the index).

Furthermore, I would love to fix the underlying problem, there are many other ByIndex queries we do.

shashank-100:
Thanks for the detailed review, Yonom! Let me address both concerns.

## How to reproduce the zombie-child crash

Minimal reproduction:
1. Use React 18 with `ReactDOM.render` (not `createRoot`) — this bypasses `useSyncExternalStore`'s zombie-child protection
2. Render a `ThreadList` with 2+ threads
3. Delete any thread that is not the last one

What happens step by step:
- Thread at index 1 is deleted → the store shrinks `threadIds` from `[A, B]` to `[A]`
- The store notifies all subscribers simultaneously
- The `ThreadListItemByIndexProvider` at `index=1` (the zombie child) fires `getSnapshot` **before** its parent can unmount it
- `getSnapshot` calls `aui.threads().item({ index: 1, archived: false })` → which resolves to `threadIds[1]` → `undefined` → then `threadItems.get({ key: undefined })` → throws `Key "undefined" not found`

In React 18 with `createRoot`, `useSyncExternalStore` prevents this by flushing parent unmounts before child re-renders. With `ReactDOM.render` that guarantee is gone.

## Why ID-based keying fixes it

With `key={index}`:
- React sees keys `0, 1` before and `0` after deletion
- It **reuses** the component at `key=0` (updates it) and tries to unmount `key=1`
- But the zombie at `key=1` fires `getSnapshot` before unmount → crash

With `key={threadId}`:
- React sees keys `"A", "B"` before and `"A"` after deletion
- It sees key `"B"` is gone and **immediately removes** it from the VDOM
- The component at `"A"` is unaffected and doesn't shift
- No stale index access → no crash

The key insight is: ID-based keying makes each component's lifecycle tied to the existence of its specific data, not its position. When `"B"` disappears from `threadIds`, React knows to unmount that subtree before any `getSnapshot` can run on it.

PR #4020, opened 2026-05-13
fix(composer): accept highlighted trigger popover item on Tab

### LINKED ISSUE #4018: ComposerPrimitive.Unstable_TriggerPopover: Tab should select the highlighted item
**Note:** Claude handled the investigation and prose via back-and-forth. Idea, decisions, and everything else are mine. — @Seluj78

## Summary

`ComposerPrimitive.Unstable_TriggerPopover` doesn't bind <kbd>Tab</kbd> for selection. Users expect Tab to accept the highlighted item (autocomplete convention from CLIs, IDEs, command palettes, GitHub's own `@`/`#` pickers, Discord, Slack, etc.). Currently only <kbd>Enter</kbd> selects; Tab falls through to the textarea (focus move or tab character insert depending on container).

## Repro

1. Mount a `Composer.Root` with `Unstable_TriggerPopoverRoot` + a `Unstable_TriggerPopover` (any `char`, e.g. `/`).
2. Type the trigger char in the input → popover opens, first item highlighted.
3. Press <kbd>Tab</kbd>.
4. Expected: highlighted item is selected (same as <kbd>Enter</kbd>).
5. Actual: nothing happens in the popover; focus moves out of the textarea (or a tab character is inserted, depending on host).

## Root cause

`packages/react/src/primitives/composer/trigger/triggerKeyboardResource.ts` — `handleKeyDown` switch handles `ArrowDown` / `ArrowUp` / `Enter` / `Shift+Enter` / `Escape` / `Backspace`. No case for `Tab`, so it falls through to the default `return false` and the input handles it natively.

Reproduced against `@assistant-ui/react@^0.12.28`. Verified by reading `dist/primitives/composer/trigger/triggerKeyboardResource.js:27-82`:

```js
switch (e.key) {
    case "ArrowDown": { /* ... */ }
    case "ArrowUp":   { /* ... */ }
    case "Enter":     { /* ... */ }
    case "Escape":    { /* ... */ }
    case "Backspace": { /* ... */ }
    default: return false;
}
```

## Proposed fix

Add a `case "Tab":` arm mirroring `Enter`:

```ts
case "Tab": {
    if (e.shiftKey) return false;          // let Shift+Tab keep its native focus-back behavior
    e.preventDefault();
    const item = navigableList[highlightedIndex];
    if (!item) return true;
    if (isTriggerItem(item)) selectItem(item);
    else selectCategory(item.id);
    return true;
}
```

Tab and Enter both accept the highlighted entry. Shift+Tab is preserved for users who want to step focus backward out of the popover. Empty popovers (no item to accept) still `return true` to swallow the Tab so focus doesn't escape unexpectedly while the popover is open.

Happy to send a PR if the maintainers are open to it.

## Why this matters

The popover is the canonical autocomplete affordance — `@` mentions, `/` slash commands, command palettes. Across the conventions users carry in (Claude Code's CLI, GitHub's PR composer, Slack/Discord composers, Notion mentions, every shell), <kbd>Tab</kbd> is the universal autocomplete key. Without it, the popover feels half-built and users default to typing the full name.

## Notes

- The `Unstable_TriggerPopover` family is marked experimental; treating this as part of stabilization is fine.
- If there's a reason Tab was intentionally omitted (e.g. accessibility — preserving native focus-move), happy to discuss; my reading of the relevant ARIA combobox patterns is that Tab accepting the selection is the recommended behavior when the popover is open.

— context: hit this in [`agent-of-empires`](https://github.com/njbrake/agent-of-empires) where the cockpit composer uses `Unstable_TriggerPopover` for `@` files and `/` commands.

serhiizghama:
## Problem

`ComposerPrimitive.Unstable_TriggerPopover` only binds <kbd>Enter</kbd> for selection. <kbd>Tab</kbd> falls through to the underlying textarea, so focus moves out of the composer or a tab character is inserted depending on the host, instead of accepting the highlighted entry.

This diverges from the autocomplete convention used in CLIs, IDEs, command palettes, GitHub's `@`/`#` pickers, Slack, Discord, and Notion mentions, where <kbd>Tab</kbd> accepts the highlighted suggestion. Users coming from any of those flows reach for Tab and the popover feels half-built.

Closes #4018.

## Solution

In `triggerKeyboardResource.ts`, the `handleKeyDown` switch now handles `Tab` alongside `Enter`. Both keys accept the highlighted item or drill into the highlighted category. <kbd>Shift+Tab</kbd> still passes through (`return false`) so native focus traversal keeps working, and an empty `navigableList` still swallows Tab so focus doesn't escape while the popover is open.

## Testing

Added `triggerKeyboardResource.test.ts` covering:

- Tab selects the highlighted item
- Tab selects a highlighted category (drill-in)
- Shift+Tab is not consumed
- Empty navigable list still consumes Tab without calling selectors
- Tab is ignored when the popover is closed

```bash
pnpm --filter @assistant-ui/react exec vitest run src/primitives/composer/trigger
# Test Files  2 passed (2)
# Tests       24 passed (24)
```

Includes a patch changeset for `@assistant-ui/react`.

### DIFF
--- a/.changeset/trigger-popover-tab-select.md
+++ b/.changeset/trigger-popover-tab-select.md
@@ -1,0 +1,9 @@
+---
+"@assistant-ui/react": patch
+---
+
+fix(composer): select highlighted item on Tab in trigger popover
+
+`ComposerPrimitive.Unstable_TriggerPopover` only accepted the highlighted entry on Enter. Tab fell through to the underlying textarea, moving focus out of the composer or inserting a tab character. This diverged from the autocomplete convention used in CLIs, IDEs, command palettes, GitHub, Slack, Discord, and Notion, where Tab accepts the highlighted suggestion.
+
+Tab now mirrors Enter and selects the highlighted item or category. Shift+Tab still passes through so native focus traversal keeps working.
--- a/packages/react/src/primitives/composer/trigger/triggerKeyboardResource.test.ts
+++ b/packages/react/src/primitives/composer/trigger/triggerKeyboardResource.test.ts
@@ -1,0 +1,105 @@
+import { describe, it, expect, vi } from "vitest";
+import { createResourceRoot } from "@assistant-ui/tap";
+import type {
+  Unstable_TriggerCategory,
+  Unstable_TriggerItem,
+} from "@assistant-ui/core";
+import { TriggerKeyboardResource } from "./triggerKeyboardResource";
+
+const item = (id: string): Unstable_TriggerItem => ({
+  id,
+  type: "command",
+  label: id,
+});
+
+const category = (id: string): Unstable_TriggerCategory => ({
+  id,
+  label: id,
+});
+
+const makeKeyEvent = (key: string, shiftKey = false) => ({
+  key,
+  shiftKey,
+  preventDefault: vi.fn(),
+});
+
+const render = (
+  overrides: Partial<Parameters<typeof TriggerKeyboardResource>[0]> = {},
+) => {
+  const props = {
+    navigableList: [item("a"), item("b"), item("c")],
+    isSearchMode: false,
+    activeCategoryId: null as string | null,
+    query: "",
+    popoverId: "popover",
+    open: true,
+    selectItem: vi.fn(),
+    selectCategory: vi.fn(),
+    goBack: vi.fn(),
+    close: vi.fn(),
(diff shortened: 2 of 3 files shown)


okisdev (approved):

okisdev:
thanks!

PR #4062, opened 2026-05-20
fix(react): scroll initial thread messages into view

### LINKED ISSUE #4009: scrollToBottomOnInitialize doesn't fire with useLocalRuntime + initialMessages
## Summary
When mounting `<ThreadPrimitive.Viewport>` over a runtime created via `useLocalRuntime(adapter, {initialMessages})`, the viewport stays at `scrollTop = 0` instead of scrolling to bottom on initial render — even though `scrollToBottomOnInitialize` defaults to `true`.

## Versions
- `@assistant-ui/react`: 0.14.0
- `@assistant-ui/core`: 0.2.0
- `@assistant-ui/store`: 0.2.10

## Minimal repro
```tsx
const adapter: ChatModelAdapter = { run: async function* () { yield {content: []} } }
const initialMessages: ThreadMessageLike[] = [
    {role: "user", content: [{type: "text", text: "hello"}]},
    {role: "assistant", content: [{type: "text", text: "world"}]},
    // ... enough messages to overflow the viewport
]

function App() {
    const runtime = useLocalRuntime(adapter, {initialMessages})
    return (
            <ThreadPrimitive.Viewport
                className="h-96 overflow-y-scroll"
                scrollToBottomOnInitialize
            >
                <ThreadPrimitive.Messages components={{
                    UserMessage: () => <MessagePrimitive.Root><MessagePrimitive.Content/></MessagePrimitive.Root>,
                    AssistantMessage: () => <MessagePrimitive.Root><MessagePrimitive.Content/></MessagePrimitive.Root>,
                }}/>
            </ThreadPrimitive.Viewport>
        </AssistantRuntimeProvider>
    )
}
```

**Expected**: viewport mounts already scrolled to bottom (last message visible).
**Actual**: viewport mounts at `scrollTop = 0` (first message visible at top).

theamodhshetty:
## Summary
- make `useThreadViewportAutoScroll` also react to the first non-empty `thread.messages` state, not only the `thread.initialize` event
- keep the existing initialize-event path, but cover the case where `initialMessages` load before the viewport subscribes
- add a jsdom regression test and a patch changeset for `@assistant-ui/react`

Fixes #4009

## Tests
- `pnpm install --frozen-lockfile`
- `pnpm --filter assistant-stream --filter assistant-cloud --filter @assistant-ui/core build`
- `pnpm --filter safe-content-frame build`
- `pnpm exec biome check --write packages/react/src/primitives/thread/useThreadViewportAutoScroll.ts packages/react/src/primitives/thread/useThreadViewportAutoScroll.test.tsx`
- `pnpm --dir packages/react exec vitest run src/primitives/thread/useThreadViewportAutoScroll.test.tsx`

### DIFF
--- a/.changeset/fix-viewport-initial-messages-scroll.md
+++ b/.changeset/fix-viewport-initial-messages-scroll.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react": patch
+---
+
+Scroll thread viewports to bottom when initial messages are already loaded before the viewport subscribes to initialization events.
--- a/packages/react/src/primitives/thread/useThreadViewportAutoScroll.test.tsx
+++ b/packages/react/src/primitives/thread/useThreadViewportAutoScroll.test.tsx
@@ -1,0 +1,132 @@
+// @vitest-environment jsdom
+
+import { render, waitFor } from "@testing-library/react";
+import type { FC, PropsWithChildren } from "react";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { AssistantRuntimeProvider } from "../../context";
+import * as MessagePrimitive from "../message";
+import * as ThreadPrimitive from "../thread";
+import { useLocalRuntime } from "../../legacy-runtime/runtime-cores/local/useLocalRuntime";
+import type { ChatModelAdapter, ThreadMessageLike } from "../../index";
+
+class ResizeObserverMock {
+  observe = vi.fn();
+  disconnect = vi.fn();
+}
+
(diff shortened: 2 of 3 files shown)


AVGVSTVS96:
thanks for the PR! went with #4063 instead, similar fix but removed the deprecated initialize event and refactored a bit

theamodhshetty:
Thanks for the heads up. That makes sense to me — #4063 covers the same state-based path and cleans up the deprecated initialize event too. Appreciate you landing it.

PR #4067, opened 2026-05-20
fix: avoid disposing OpenCode runtime in StrictMode

### LINKED ISSUE #4064: OpenCode runtime fails to stream in StrictMode
### Checklist

### Environment info

```
Environment:
  OS:               macOS 26.3.1 (arm64)
  Node.js:          v24.14.0
  Package Manager:  npm 11.9.0
  Framework:        Vite 6.4.2

Packages:
  @assistant-ui/react           0.14.5
  @assistant-ui/core            0.2.2
  @assistant-ui/store           0.2.10
  @assistant-ui/tap             0.5.11
  assistant-stream              0.3.14
  assistant-cloud               0.1.27
  @assistant-ui/react-markdown  0.14.0

Ecosystem:
  react       19.2.6
  react-dom   19.2.6
  vite        6.4.2
  zod         4.4.3
  zustand     5.0.13
  typescript  5.9.3
```

### Affected area(s)

Streaming (assistant-stream), Runtime / Core

### Runtime adapter in use

Other / Custom

### Steps to reproduce

1. Set up simple React app
2. Make sure to use `<StrictMode>`
3. Add OpenCode runtime for assistant-ui and set up OpenCode

When going to the app, you should see new messages when reloading, but they are not being streamed in.
Disabling `<StrictMode>` fixes the error.

Expected: Messages stream in as they are created
Actual: User needs to manually refresh browser to see new messages

### Browser (if applicable)

_No response_

### Code snippet or error log

```tsx

```

### Additional context

It looks like the issue lies here: https://github.com/assistant-ui/assistant-ui/blob/main/packages/react-opencode/src/useOpenCodeRuntime.ts#L293

```
  const registry = useMemo(() => createRegistry(client), [client]);

  useEffect(() => {
    return () => {
      registry.dispose();
    };
  }, [registry]);
```

A registry is created and memoized. WITHOUT strict mode, `registry.dispose` is called when the component unmounts, but with strict mode enabled, React will rerun the effect on mount and call `registry.dispose()` right away.

I am not familiar enough with the assistant-ui codebase to really suggest how to fix this, but I would think that using a ref for the registry instead of `useMemo` which can be re-created inside of the effect could work. The effect also could be responsible for re-enabling the registry if it was previously disposed.

Herrtian:
Fixes #4064.

Delays OpenCode registry disposal by one tick so StrictMode's remount check does not close the active event stream.

Checks:
- pnpm --filter @assistant-ui/react-opencode exec vitest run src/OpenCodeEventSource.test.ts src/OpenCodeThreadController.test.ts src/openCodeThreadState.test.ts src/useOpenCodeStreamingTiming.test.ts
- pnpm exec biome check packages/react-opencode/src/useOpenCodeRuntime.ts .changeset/quiet-opencode-streams.md
- git diff --check

### DIFF
--- a/.changeset/quiet-opencode-streams.md
+++ b/.changeset/quiet-opencode-streams.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react-opencode": patch
+---
+
+fix(react-opencode): keep the OpenCode runtime registry alive during React StrictMode remount checks.
--- a/packages/react-opencode/src/useOpenCodeRuntime.ts
+++ b/packages/react-opencode/src/useOpenCodeRuntime.ts
@@ -15,6 +15,7 @@ import {
   useEffect,
   useEffectEvent,
   useMemo,
+  useRef,
   useSyncExternalStore,
 } from "react";
 import type {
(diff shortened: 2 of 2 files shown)


okisdev:
thanks a lot for putting this together and for the clear diagnosis. the root cause is right: strictmode replays the effect cleanup, and the current cleanup permanently stops the opencode event stream.

i’m going to supersede this with #4068 because this pr solves it by delaying cleanup by one tick. that works for the strictmode replay case, but it still relies on react effect timing. #4068 moves the fix into the runtime lifecycle instead: the event source can be recreated, controllers can resubscribe after cleanup, session state is preserved, and reconnect behavior stays owned by the adapter.

really appreciate you catching this and sending the fix. it made the underlying issue much easier to verify and turn into a lifecycle-level patch.

PR #4083, opened 2026-05-22
feat(docs): expose example docs pages to docs assistant

### LINKED ISSUE #4082: Expose example docs pages to the docs assistant tools
## Summary

The docs assistant can browse and read `/docs` pages through `listDocs` and `readDoc`, but it does not currently expose the official `/examples` docs pages through those tools.

While evaluating our Xulux coding agent against assistant-ui docs, we found that implementation questions often benefit from example docs pages as much as conceptual docs. Since examples are already available as a separate Fumadocs source, the docs assistant should be able to discover and read them too.

## Current Behavior

`listDocs()` returns root docs sections from `source.pageTree`.

`readDoc({ slugOrUrl })` reads pages from `source.getPage(...)`.

Because `/examples` uses a separate source, these calls are not available today:

```ts
listDocs({ path: "examples" })
readDoc({ slugOrUrl: "examples/ai-sdk" })
```

## Proposed Behavior

Expose example docs pages as part of the docs assistant tool surface.

`listDocs()` should include:

```json
{ "type": "folder", "name": "examples", "url": "/examples" }
```

These should work:

```ts
listDocs({ path: "examples" })
readDoc({ slugOrUrl: "examples/ai-sdk" })
```

## Proposed Implementation

- Import the existing examples source from `@/lib/source`.
- Add `examples` to the root `listDocs()` result.
- Route `listDocs({ path: "examples" })` to `examplesSource.pageTree`.
- Route `readDoc({ slugOrUrl: "examples/..." })` to `examplesSource.getPage(...)`.
- Enable processed markdown for the examples collection for parity with docs/blog generated content.

## Impact

This lets the docs assistant answer implementation questions using official example docs pages, not only reference and guide docs.

vivek100:
## Summary

Closes #4082.

Exposes official `/examples` docs pages through the docs assistant tools.

## Changes

- Adds processed markdown generation for the examples collection.
- Adds `examples` to the root `listDocs()` result.
- Supports browsing example docs pages with `listDocs({ path: "examples" })`.
- Supports reading example docs pages with `readDoc({ slugOrUrl: "examples/..." })`.

## Testing

Tested locally in the docs app on this branch.

Verified the docs assistant can now use the examples tool paths:

```ts
listDocs()
listDocs({ path: "examples" })
readDoc({ slugOrUrl: "examples/ai-sdk" })
```

Also ran Biome on the touched files:

```sh
node_modules\.bin\biome.cmd check apps/docs/app/api/doc/chat/route.ts apps/docs/source.config.ts
```

### DIFF
--- a/apps/docs/app/api/doc/chat/route.ts
+++ b/apps/docs/app/api/doc/chat/route.ts
@@ -6,7 +6,7 @@ import path from "node:path";
 import { injectQuoteContext } from "@assistant-ui/react-ai-sdk";
 import { checkRateLimit } from "@/lib/rate-limit";
 import { validateDocChatInput } from "@/lib/validate-input";
-import { source } from "@/lib/source";
+import { source, examples as examplesSource } from "@/lib/source";
 import { getModel } from "@/lib/ai/provider";
 import { frontendTools } from "@assistant-ui/react-ai-sdk";
 import { createBashTool } from "bash-tool";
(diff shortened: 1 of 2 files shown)


okisdev:
the feature itself is good. pushed a refactor onto the branch to tighten a few things in the implementation, here's what and why:

- the page tree mapping (the `flatMap` over `page`/`folder` nodes) was duplicated between the new `examples` branch and the existing generic branch in `listDocs`. pulled it into one `listChildren` helper.
- `listDocs` and `readDoc` disagreed on how they matched the `examples` namespace: `listDocs` used `path === "examples"` (exact match) while `readDoc` used `slugs[0] === "examples"` (prefix match). so `readDoc` could read `examples/<nested>` but `listDocs` couldn't browse into it. it doesn't bite today because `content/examples` is flat, but it's an inconsistent contract between the two tools. routed the `examples` path through `findFolderByPath` as well, so both behave the same.
- `readDoc` had two near identical blocks (one per source) running the same `getPage` + `getLLMText` flow. collapsed them into a single dispatch that picks the source and segments up front.
- `includeProcessedMarkdown` on the examples collection wasn't doing anything for these tools. `getLLMText` renders `page.data.body` (the compiled MDX) and converts that to markdown, it never reads the processed markdown export, so the flag only generated build output that nothing consumed. dropped it, which also leaves `source.config.ts` identical to `main` and shrinks the PR to a single file.

`biome check` and `tsc --noEmit` on the docs app both pass, no casts needed, and behavior is unchanged for every current examples page.

okisdev (approved):

PR #4089, opened 2026-05-23
example: use rslib as example to build tap

zmzlois:
add rslib build example in preparation for tsgo migration


Yonom:
since rspack is using tsc under the hood for declarations and declaration map generation, we don't get most of the benefits of rspack (performance)

unfortunately, rspack doesn't seem to help out of the box with the .js file extension appends or keeping the reference tags

this is great to benchmark against tsgo, keeping it around until then

PR #4110, opened 2026-05-26
fix(react): use instant scroll in top anchor reserve to eliminate scroll jitter

### LINKED ISSUE #4104: bug: visible scroll jitter on send when `turnAnchor="top"`
## Description

When using `turnAnchor="top"` on `ThreadPrimitive.Viewport`, there is a visible scroll "tick" (a few pixels of jitter) every time the assistant starts responding. The user message briefly shifts down by a few pixels, momentarily revealing the previous message's chat bubble above it.

## ~~Root cause~~

In `mountTopAnchorReserve.ts`, the `apply()` function uses `behavior: "smooth"` for the scroll correction. ~~Because `apply()` is scheduled via `requestAnimationFrame` (through `createFrameScheduler`), there is always at least one frame of delay between the DOM change (assistant message mounting) and the scroll correction. The browser renders one frame at the "wrong" position, then the `smooth` scroll animation makes the correction visible as a tick.~~

> [!IMPORTANT]
> **Maintainer note:** this is an incorrect diagnosis of the issue, see https://github.com/assistant-ui/assistant-ui/issues/4104#issuecomment-4550542184 for the root cause and fix

## Expected behavior

The anchor correction should be invisible. Using `behavior: "instant"` (at least for the **initial** anchor positioning when the run starts) would eliminate the visible tick — the correction would still happen but within the same paint.

## Reproduction

1. Use `<ThreadPrimitive.Viewport turnAnchor="top">`
2. Send a message in a chat that already has previous messages
3. Observe the user message when the assistant response starts — it shifts down by a few pixels before stabilizing

## Environment

- `@assistant-ui/react`: 0.14.8
- Browser: Chrome (Linux), also reproducible on other browsers
- React 19

serhiizghama:
## Problem

When using `turnAnchor="top"` on `ThreadPrimitive.Viewport`, there is a visible scroll "tick" (jitter) every time the assistant starts responding. The user message briefly shifts down a few pixels, momentarily revealing the message above it.

The root cause is in `mountTopAnchorReserve.ts`. The `apply()` function uses `behavior: "smooth"` for the scroll correction:

```ts
if (Math.abs(viewport.scrollTop - targetScrollTop) > 1)
  viewport.scrollTo({ top: targetScrollTop, behavior: "smooth" });
```

Since `apply()` is already deferred by at least one frame via `requestAnimationFrame`, the browser renders one frame at the wrong position first, then the smooth animation makes even a sub-pixel correction visually noticeable as a jitter.

Fixes #4104.

## Solution

Change `behavior: "smooth"` to `behavior: "instant"`. The correction is already happening at RAF time (synchronized with the browser's rendering cycle), so instant behavior eliminates the visible tick while still performing the correction:

```ts
viewport.scrollTo({ top: targetScrollTop, behavior: "instant" });
```

## Testing

1. Use `<ThreadPrimitive.Viewport turnAnchor="top">` in a chat with existing messages
2. Send a new message
3. Observe that the user message no longer shifts/jitters when the assistant response starts

### DIFF
--- a/packages/react/src/primitives/thread/topAnchor/mountTopAnchorReserve.ts
+++ b/packages/react/src/primitives/thread/topAnchor/mountTopAnchorReserve.ts
@@ -106,7 +106,7 @@ export const mountTopAnchorReserve = (store: TopAnchorStore) => {
     );
 
     if (Math.abs(viewport.scrollTop - targetScrollTop) > 1) {
-      viewport.scrollTo({ top: targetScrollTop, behavior: "smooth" });
+      viewport.scrollTo({ top: targetScrollTop, behavior: "instant" });
     }
 
     if (anchorId !== undefined) lastScrolledAnchorId = anchorId;
--- a/.changeset/fix-top-anchor-scroll-jitter.md
+++ b/.changeset/fix-top-anchor-scroll-jitter.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react": patch
+---
+
+fix(react): use instant scroll in top anchor reserve to eliminate visible jitter


packages/react/src/primitives/thread/topAnchor/mountTopAnchorReserve.ts
```diff
@@ -106,7 +106,7 @@ export const mountTopAnchorReserve = (store: TopAnchorStore) => {
     );
 
     if (Math.abs(viewport.scrollTop - targetScrollTop) > 1) {
-      viewport.scrollTo({ top: targetScrollTop, behavior: "smooth" });
+      viewport.scrollTo({ top: targetScrollTop, behavior: "instant" });
```
AVGVSTVS96, line 109: changing `scrollTo` behavior to `instant` is a workaround that may resolve this scroll jitter bug but introduces a UX regression: our expectation is when a new user message is sent, the thread scrolls down smoothly to anchor the new user message at the top -- this change makes viewport scroll to the new user message instantly, which isn't our intent

AVGVSTVS96:
thanks for the PR! as noted in https://github.com/assistant-ui/assistant-ui/pull/4110#discussion_r3301125160, we won't merge this as-is because it introduces a UX regression.

i'll work on finding the root cause for the scroll jitter, will update when i find a fix.

Issue #4101, opened 2026-05-25
makeAssistantTool tools not included in HTTP requests when using useRemoteThreadListRuntime with runtimeHook


oneyoung19:
### Checklist

- [x] I have searched existing issues and this is not a duplicate
- [x] This is a reproducible bug, not a usage question (use [Discord](https://discord.gg/S9dwgCNEFs) for questions)

### Environment info

```
Environment:
  OS:               macOS 26.2 (arm64)
  Node.js:          v22.14.0
  Package Manager:  pnpm 10.14.0
  Framework:        Next.js 16.1.6

Packages:
  @assistant-ui/react           0.14.5
  @assistant-ui/core            0.2.2
  assistant-stream              0.3.14
  @assistant-ui/react-ai-sdk    1.3.26
  @assistant-ui/react-markdown  0.14.0
  @assistant-ui/react-lexical   0.2.0

Ecosystem:
  react       19.2.3
  react-dom   19.2.3
  next        16.1.6
  vite        8.0.10
  ai          6.0.168
  zod         4.3.6
  zustand     5.0.11
  typescript  5.9.3

Warnings:
  ! @assistant-ui/core should not be installed directly — it is an internal dependency
```

### Affected area(s)

Runtime / Core

### Runtime adapter in use

useAISDKRuntime / useChatRuntime

### Steps to reproduce

When using `useRemoteThreadListRuntime` with a `runtimeHook`, tools registered via `makeAssistantTool` are not sent in the HTTP request body — the `tools` field is an empty object `{}`.

The same tools work correctly when the runtime hook is used directly as the top-level runtime (without `useRemoteThreadListRuntime`).

## Reproduction

**`thread-runtime-provider.tsx`**

```tsx
'use client'

import { AssistantRuntimeProvider, useRemoteThreadListRuntime } from '@assistant-ui/react'
import { threadListAdapter } from '../adapters/thread-list-adapter'
import { useThreadRuntime } from '../runtime/use-thread-runtime'
import { GetLocalTimeTool } from '@/components/assistant-ui/tools/time-tool'

export const ThreadRuntimeProvider = ({ children }) => {
  // ❌ tools field in HTTP request body is {}
  const runtime = useRemoteThreadListRuntime({
    runtimeHook: useThreadRuntime,
    adapter: threadListAdapter,
  })

  // ✅ tools field is correctly populated — but loses thread list management
  // const runtime = useThreadRuntime()

  return (
[...]

PR #4113, opened 2026-05-26
test(react, react-ai-sdk): cover tool propagation through useRemoteThreadListRuntime runtimeHook

ShobhitPatra:
## Summary

Adds regression coverage for tool propagation when a runtime is used as the runtimeHook` of useRemoteThreadListRuntime` — the path reported in #4101.

The existing `adapterProvider` test only checks that `modelContext` is *defined* on the hook's adapters. It never checks that a tool registered via makeAssistantTool` on the outer runtime actually shows up in thread.getModelContext().tools` — the data `AssistantChatTransport` puts in the  body. These tests close that gap.

Note: I could not reproduce #4101 on `main`; tools propagate correctly. These tests serve as a regression guard.

  ## Tests

  - `RemoteThreadListRuntime.toolPropagation.test.tsx` — outer runtime + `useLocalRuntime` hook; asserts the tool reaches the inner thread's model context.
  - `useChatRuntime.toolPropagation.test.tsx` — the exact #4101 setup (outer runtime + `useChatRuntime` + `AssistantChatTransport`); asserts the tool is in thread.getModelContext().tools`.

  Test-only; no changeset

### DIFF
--- a/packages/react-ai-sdk/src/ui/use-chat/useChatRuntime.toolPropagation.test.tsx
+++ b/packages/react-ai-sdk/src/ui/use-chat/useChatRuntime.toolPropagation.test.tsx
@@ -45,6 +45,12 @@ const timeTool = {
   toolName: "get_local_time",
   type: "frontend" as const,
   description: "Get current local time",
+  parameters: {
+    type: "object" as const,
+    properties: {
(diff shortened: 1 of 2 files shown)


Yonom:
The issue turned out to be something other than the propagation failing, I'm closing this as the test would not be very useful

PR #4138, opened 2026-05-27
Add `fetchOptions` passthrough for A2A client/runtime requests

sdg9670f:
`A2AClient` and `useA2ARuntime` could not pass `RequestInit` options like `credentials: "include"` into underlying fetch calls. This PR adds first-class `fetchOptions` support in both APIs and documents usage.

- **Client API: add `fetchOptions` to `A2AClientOptions`**
  - Added `fetchOptions?: Omit<RequestInit, "headers" | "body" | "method" | "signal">`.
  - Stored once on the client instance and applied to all request paths.

- **Fetch plumbing: apply across all relevant request types**
  - `fetchJSON`: `...this.fetchOptions` spread before `...options`.
  - `streamMessage`: spread before `method`/`headers`/`body`.
  - `getAgentCard`: spread before `headers`.
  - `subscribeToTask`: spread before `headers`.
  - `deleteTaskPushNotificationConfig`: spread before `method`/`headers`.
  - Ordering preserves internal control of `headers`, `method`, `body`, and `signal`.

- **Runtime API: expose and forward `fetchOptions`**
  - Added `fetchOptions?: A2AClientOptions["fetchOptions"]` to `UseA2ARuntimeOptions`.
  - Forwarded into auto-created `A2AClient` when using `baseUrl`.

- **Docs: surface new option in both option tables**
  - Added `fetchOptions` rows to:
    - A2A **Client options**
    - `useA2ARuntime` **options**
  - Clarifies intended use and internally managed fields.

- **Tests: add focused coverage**
  - Added tests ensuring `fetchOptions` is applied across request categories and that internal request fields still win over external overrides.

```ts
const runtime = useA2ARuntime({
  baseUrl: "http://localhost:3011/workspaces/6/agents/a2a",
  basePath: "/v1",
  headers: () => ({ Authorization: "******" }),
  fetchOptions: { credentials: "include" },
});

const client = new A2AClient({
  baseUrl: "https://my-agent.example.com",
  headers: { Authorization: "******" },
  fetchOptions: { credentials: "include" },
});
```

### DIFF
--- a/.changeset/a2a-fetch-options.md
+++ b/.changeset/a2a-fetch-options.md
@@ -1,0 +1,14 @@
+---
+"@assistant-ui/react-a2a": patch
+---
+
+add `fetchOptions` passthrough on `A2AClient` and `useA2ARuntime` so callers can forward `credentials`, `cache`, `mode`, `keepalive`, etc. to the underlying fetch.
+
+```ts
+const runtime = useA2ARuntime({
+  baseUrl: "https://my-agent.example.com",
+  fetchOptions: { credentials: "include" },
+});
+```
+
+`headers`, `body`, `method`, and `signal` stay internally managed: the four fields are stripped at construction time and cannot leak through even if a caller bypasses the type with an `as RequestInit` cast.
--- a/apps/docs/content/docs/runtimes/a2a/client-and-hooks.mdx
+++ b/apps/docs/content/docs/runtimes/a2a/client-and-hooks.mdx
@@ -37,6 +37,7 @@ const runtime = useA2ARuntime({ client });
 | `headers` | `Record<string, string>` or `() => Record<string, string>` | Static or dynamic headers (e.g. for auth tokens). |
 | `tenant` | `string` | Tenant ID for multi-tenant servers (prepended to URL paths). |
 | `extensions` | `string[]` | Extension URIs to negotiate via `A2A-Extensions` header. |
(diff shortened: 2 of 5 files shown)


okisdev (approved):
thanks for this. spawned a few agents to compare against upstream `a2aproject/a2a-js` and the rest of the runtimes in the repo, then pushed a follow-up commit at `0e8596f`:

- **runtime sanitization in the `A2AClient` constructor**: destructure `headers`, `body`, `method`, and `signal` off `fetchOptions` once so the type-level `Omit` matches runtime behavior. without this, a caller using `as RequestInit` could smuggle a `signal` into every request (because `signalInit(undefined)` returns `{}` at all 5 fetch call sites, nothing in `options` overrides `fetchOptions.signal`), and `method` would leak into the two GET-style sites (`getAgentCard`, `subscribeToTask`).
- **broader negative test**: the previous test only exercised `sendMessage` (POST). added the `getAgentCard` (GET) path and smuggled `body` / `signal` so all four reserved fields are covered.
- **changeset**: added the missing `@assistant-ui/react-a2a` patch.
- **docs alignment**: included `fetchOptions` in the `A2AClient` example to match the other options, and mirrored the "internally managed" note onto the `useA2ARuntime` row.

for context (not for this PR): upstream `a2aproject/a2a-js` chose a different shape. they expose `fetchImpl?: typeof fetch` plus a per-call `RequestOptions { signal, serviceParameters }`, so callers who need a single `AbortSignal` for global cancellation today can't reach it through `fetchOptions` (intentionally, since `signal` is reserved). a follow-up adding `fetchImpl?: typeof fetch` would mirror upstream and close that gap.

merging once CI green.

PR #4141, opened 2026-05-27
fix(react): release auto-stick intent on user scroll-up; fix retina at-bottom detection

### LINKED ISSUE #4140: [Bug] useThreadViewportAutoScroll snaps to bottom on subtree mutations after user scrolls away
### Checklist

(closest: #3964 — same code path, closed because the reporter found a project-side CSS issue; @okisdev's comment there describes the exact mechanism this report demonstrates as a reproducible library bug.)

### Environment info

```
Environment:
  OS:               macOS 26.5 (arm64, devicePixelRatio: 2 — retina)
  Node.js:          v22.22.2
  Package Manager:  bun 1.3.14
  Framework:        Next.js 16.2.6

Packages:
  @assistant-ui/react           0.12.28
  @assistant-ui/react-markdown  0.12.11
```

Same code is present on `main` as of v0.14.11 — see "Suggested fix" for diff against current source.

### Affected area(s)

React / UI Components

### Runtime adapter in use

Not adapter-specific — the bug is inside `useThreadViewportAutoScroll`.

### Steps to reproduce

With default `turnAnchor="bottom"`:

1. Start a thread, send a message, wait for a long streamed response to complete.
2. Scroll up away from the bottom (trackpad/wheel).
3. Trigger any subtree mutation inside the viewport — a markdown re-render, image lazy-load, a CSS transition finishing, even a benign attribute flip on a child.
4. Viewport snaps back to the bottom.

Minimal programmatic repro (from devtools console with thread already populated):

```js
const sc = document.querySelector('[data-aui-thread-viewport], .your-viewport-el');
sc.scrollTop = sc.scrollHeight / 4;                       // park mid-thread
sc.querySelector('button')?.setAttribute('data-x', '1');  // trip MutationObserver
// → sc.scrollTop snaps to bottom within ~16ms
```

**Expected:** after the user scrolls away from bottom, subtree mutations should not pull the viewport back to bottom. Auto-stick should only re-engage when the user explicitly scrolls back to bottom (or clicks the ScrollToBottom button).

**Actual:** every subtree mutation calls `scrollToBottom("auto")` via the resize observer, because `scrollingToBottomBehaviorRef.current` is set on `thread.runStart` and only cleared in `handleScroll` when `newIsAtBottom` becomes true — i.e., the user has to *land at the bottom* before the lock releases.

### Browser (if applicable)

Chrome 131 on macOS 26.5 (arm64, retina).

### Code snippet / stack trace

Relevant code in `packages/react/src/primitives/thread/useThreadViewportAutoScroll.ts`:

```ts
const handleScroll = () => {
  // ...
  } else {
    if (newIsAtBottom) {
      scrollingToBottomBehaviorRef.current = null;
    }
    // no release path when the user scrolls UP
    // ...
  }
};

const resizeRef = useOnResizeContent(() => {
  const scrollBehavior = scrollingToBottomBehaviorRef.current;
  if (scrollBehavior) scrollToBottom(scrollBehavior); // ← snaps even if user moved away
  // ...
});
```

Confirmed via setter spy on `scrollTo`: a single benign attribute toggle on a descendant produces a `scrollTo({top: scrollHeight, behavior: "auto"})` originating from `useOnResizeContent[resizeRef] → useThreadViewportAutoScroll[scrollToBottom]`.

This is the mechanism @okisdev described in #3964 ("the only path where i can make autoScroll keep overriding a user scroll up is when there are continuous DOM mutations inside the viewport") — now seen as a reproducible bug rather than expected behavior, since the snap also fires *after* streaming finishes, on stable post-stream subtree updates (markdown syntax highlighting, image final layout, animations completing).

### Additional context — secondary bug found while debugging

`Math.abs(div.scrollHeight - div.scrollTop - div.clientHeight) < 1` is too strict on devices where Chrome clips `scrollTop` 1px short of `scrollHeight - clientHeight`. On retina (`devicePixelRatio: 2`) macOS Chrome:

- `scrollHeight = 1337`, `clientHeight = 549`, so `scrollHeight - clientHeight = 788`
- After `sc.scrollTo({top: scrollHeight})`, `sc.scrollTop` clips to **787**, not 788
- `Math.abs(1) < 1` is `false` → `newIsAtBottom` never registers → `isAtBottom` never updates → `ScrollToBottom` button never disables, even though the viewport is visually at the bottom.

Non-retina monitors and narrow-window layouts land on different sub-pixel values that happen to satisfy `< 1`, which is why this only manifests on some setups.

### Suggested fix

Three small changes:

```diff
   const lastScrollTop = useRef(0);
+  const lastScrollHeight = useRef(0);
   const scrollingToBottomBehaviorRef = useRef<ScrollBehavior | null>(null);

   const handleScroll = () => {
     const div = divRef.current;
     if (!div) return;
     const isAtBottom = threadViewportStore.getState().isAtBottom;
-    const newIsAtBottom = Math.abs(div.scrollHeight - div.scrollTop - div.clientHeight) < 1
+    const newIsAtBottom = Math.abs(div.scrollHeight - div.scrollTop - div.clientHeight) <= 1
       || div.scrollHeight <= div.clientHeight;
     if (!newIsAtBottom && lastScrollTop.current < div.scrollTop) {
       // ignore scroll down
     } else {
       if (newIsAtBottom) {
         scrollingToBottomBehaviorRef.current = null;
+      } else if (
+        lastScrollTop.current > div.scrollTop &&
+        lastScrollHeight.current === div.scrollHeight
+      ) {
+        // User scrolled UP (not a content-driven scroll adjustment).
+        // Release the auto-stick lock so subsequent subtree mutations
+        // don't snap back via the resize observer.
+        scrollingToBottomBehaviorRef.current = null;
       }
       const shouldUpdate = newIsAtBottom || scrollingToBottomBehaviorRef.current === null;
       if (shouldUpdate && newIsAtBottom !== isAtBottom) {
         writableStore(threadViewportStore).setState({ isAtBottom: newIsAtBottom });
       }
     }
     lastScrollTop.current = div.scrollTop;
+    lastScrollHeight.current = div.scrollHeight;
   };
```

Why each piece:
1. `<= 1`: handles the retina sub-pixel `scrollTop` clip.
2. `lastScrollHeight` ref: lets us distinguish "user scrolled up" from "`scrollTop` got smaller because `scrollHeight` shrank under it".
3. The new `else if`: when the user actually scrolls up *and* content size is unchanged, release the auto-stick lock so the next subtree mutation doesn't yank them back.

Verified via live browser instrumentation:

| Step | distance from bottom | ScrollToBottom button |
|---|---|---|
| post-stream (was broken on retina pre-patch) | 1 | **hidden** ✓ |
| user scrolled to top | 718 | visible ✓ |
| trigger DOM mutation while away | 732 | visible, **no snap** ✓ |
| scroll back to bottom | 1 | **hidden** ✓ |

Filler-content test confirms the desired stick-to-bottom while-at-bottom behavior is preserved (append 600px-tall child → scroll follows to new bottom).

PR with the patch + a `patch`-level changeset coming up.

vaniyokk:
Fixes #4140.

`useThreadViewportAutoScroll` had two related bugs surfaced together on Chrome macOS with a retina display:

1. **Subtree mutations snapped the viewport back to bottom after the user scrolled away.** `scrollingToBottomBehaviorRef` was planted on `thread.runStart` and only cleared once `newIsAtBottom` became true. If the user scrolled up before reaching bottom (or after streaming finished while still on the bottom), the ref stayed set and `useOnResizeContent` kept calling `scrollToBottom(scrollBehavior)` on every subtree mutation — markdown re-render, syntax-highlighting pass, image lazy-load, even an attribute flip on a child. The viewport was effectively locked to the bottom until a full reload.

2. **`isAtBottom` never registered `true` on high-DPR displays.** The strict `< 1` threshold misses Chrome macOS at `devicePixelRatio: 2`, which clips `scrollTop` one pixel short of `scrollHeight - clientHeight` (so `Math.abs(1) < 1 === false`). `ScrollToBottom` never moved into its disabled state even at the visual bottom. Non-retina monitors and narrow-window layouts land on different sub-pixel values that happen to satisfy `< 1`, which is why this only manifests on some setups.

## What this PR does

Track `lastScrollHeight` alongside `lastScrollTop` so `handleScroll` can distinguish a user-initiated scroll-up from a content-driven `scrollTop` adjustment (`scrollHeight` grew/shrank under it). When the user actually scrolls up and content size is unchanged, release the auto-stick intent so the next subtree mutation doesn't yank them back. Loosen the at-bottom threshold to `<= 1` to absorb the 1px sub-pixel clip.

## Why the `scrollHeight === scrollHeight` guard

Without it, content-driven adjustments — e.g. a CSS transition completing and shrinking a child, which causes the browser to silently lower `scrollTop` to the new max — would look like "user scrolled up" and incorrectly release the intent. The equality guard keeps the release path bound to actual input events.

## Behavior preserved

`thread.runStart` → at-bottom auto-follow during streaming is unaffected, since the only new branch fires under `lastScrollTop > scrollTop`, which the in-flight smooth-scroll never satisfies (it's monotonically descending toward the new bottom). Verified by appending a synthetic 600px child while parked at the bottom — viewport follows to new bottom.

## Verification

Live browser instrumentation on a thread with a ~1300px-tall streamed response, `clientHeight=549`, `devicePixelRatio=2`:

| Step | distance from bottom | ScrollToBottom button |
|---|---|---|
| post-stream | 1 | **hidden** ✓ (was visible pre-patch on retina) |
| user scrolled to top | 718 | visible ✓ |
| trigger benign attribute mutation on a child while away | 732 | visible, **no snap** ✓ (was snapping to bottom pre-patch) |
| scroll back to bottom | 1 | **hidden** ✓ |
| idle 2s at bottom | 1 | **hidden** ✓ |

Setter-spy stack traces confirmed the offending pre-patch `scrollTo({top: scrollHeight, behavior: "auto"})` originated from `useOnResizeContent[resizeRef] → useThreadViewportAutoScroll[scrollToBottom]`.

## Changeset

`patch` per CONTRIBUTING.md (single `@assistant-ui/react` package).

### DIFF
--- a/.changeset/fix-scroll-snap-on-mutation.md
+++ b/.changeset/fix-scroll-snap-on-mutation.md
@@ -2,14 +2,14 @@
 "@assistant-ui/react": patch
 ---
 
-fix(react): release auto-stick intent on user scroll-up; fix retina sub-pixel at-bottom detection
+fix(react): stop subtree mutations from snapping a scrolled-up viewport back to bottom; loosen at-bottom threshold for high-DPR displays
 
-`useThreadViewportAutoScroll` had two related bugs surfaced together on Chrome macOS with a retina display (`devicePixelRatio: 2`):
+`useThreadViewportAutoScroll` had two related bugs surfaced on Chrome macOS at `devicePixelRatio: 2`:
 
-1. **Subtree mutations snapped the viewport back to bottom after the user scrolled away.** `scrollingToBottomBehaviorRef` was planted on `thread.runStart` and only cleared in `handleScroll` once `newIsAtBottom` became true. If the user scrolled up before reaching bottom (or after streaming finished while still on the bottom), the ref stayed set and `useOnResizeContent` kept calling `scrollToBottom(scrollBehavior)` on every subtree mutation — markdown re-render, syntax-highlighting pass, image lazy-load, even an attribute flip on a child. The viewport was effectively locked to the bottom until a full reload.
+1. subtree mutations snapped the viewport back to bottom after the user scrolled away. `scrollingToBottomBehaviorRef` was planted on `thread.runStart` / `useOnScrollToBottom` / initialize / thread switch and only cleared in `handleScroll` once `newIsAtBottom` became true. while the ref stayed set, every non-style subtree mutation (a Radix `data-state` flip, a markdown re-render, an image lazy-load, an attribute toggle on a child) re-entered `useOnResizeContent`'s callback and called `scrollToBottom(scrollBehavior)`, locking the viewport to the bottom until reload.
 
-2. **`isAtBottom` never registered as `true` on high-DPR displays.** `Math.abs(scrollHeight - scrollTop - clientHeight) < 1` is strict-less-than. Chrome macOS at `devicePixelRatio: 2` clips `scrollTop` one pixel short of `scrollHeight - clientHeight` (`Math.abs(1) < 1 === false`), so even visually at the bottom the store never updated and `ScrollToBottom` never moved into its disabled state.
+2. `isAtBottom` never registered as `true` on high-DPR displays. `Math.abs(scrollHeight - scrollTop - clientHeight) < 1` is strict-less-than, and Chrome macOS at `devicePixelRatio: 2` clips `scrollTop` one pixel short of `scrollHeight - clientHeight` (`Math.abs(1) < 1 === false`), so the store never updated and `ScrollToBottom` never moved into its disabled state.
 
-The fix tracks `lastScrollHeight` alongside `lastScrollTop` so the scroll handler can distinguish a user-initiated scroll-up from a content-driven scrollTop adjustment (scrollHeight grew/shrank under it). When the user actually scrolls up and content size is unchanged, the auto-stick intent is released. The threshold becomes `<= 1` to absorb the 1px sub-pixel clip.
+the fix combines two layers. `handleScroll` now tracks `lastScrollHeight` alongside `lastScrollTop` and releases the auto-stick intent when the user scrolls up with content size unchanged, ruling out content-driven scrollTop shifts. the resize callback also bails when neither `scrollHeight` nor `clientHeight` has changed since the last fire, so mutations that don't move layout never re-enter the snap path. at-bottom auto-follow during streaming is preserved (verified by appending a synthetic 600px child while scrolled to bottom; viewport follows to new bottom). the threshold becomes `<= 1` to absorb the 1px sub-pixel clip.
 
-`thread.runStart` → at-bottom auto-follow during streaming is preserved (verified by appending a synthetic 600px child while scrolled to bottom — viewport follows to new bottom). User scrolling up no longer fights back, and the `ScrollToBottom` button now correctly hides at the bottom on retina displays.
+closes #4140.
--- a/.changeset/fix-scroll-snap-on-mutation.md
+++ b/.changeset/fix-scroll-snap-on-mutation.md
@@ -1,0 +1,15 @@
+---
+"@assistant-ui/react": patch
+---
+
+fix(react): release auto-stick intent on user scroll-up; fix retina sub-pixel at-bottom detection
+
+`useThreadViewportAutoScroll` had two related bugs surfaced together on Chrome macOS with a retina display (`devicePixelRatio: 2`):
+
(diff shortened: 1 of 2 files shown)


okisdev (approved):
**ship-ready.**

pushed one commit on top:

- added a `scrollHeight` / `clientHeight` gate at the resize-callback entry. complementary to your `handleScroll` release valve: the gate skips the snap path entirely on noise mutations (Radix `data-state` flips, attribute toggles that don't move layout), the valve handles the case where a real layout mutation fires while the user is already scrolled up.
- trimmed the two larger comments to match file convention (short, only for hidden invariants).
- reformatted the changeset (lowercase, no em-dashes); folded both layers into the description.

nice repro + retina instrumentation, made review fast.

PR #4145, opened 2026-05-27
Improve docs chat examples discovery

### LINKED ISSUE #4143: Improve docs agent discovery of examples in listDocs
While running evals for our Xulux coding agent, we found that the agent was not exploring examples correctly.

The root \listDocs\ response exposed \examples\, but the entry did not describe that examples are app-shaped guides for common assistant-ui use cases. This made it easier for the agent to skip relevant examples and instead invent structure from lower-level docs.

We already fixed this discovery hint for the main Xulux coding agent. We should apply the same upstream-facing fix to the docs assistant so the docs agent also understands that \examples\ is a useful section to browse when answering implementation and app-building questions.

vivek100:
Closes #4143

## Summary

- Updates the docs assistant \listDocs\ root response for \examples\
- Gives the examples folder the same discovery hint used by the Xulux coding agent
- Makes examples easier for agents to identify as app-shaped guides with recommended patterns and UI structure

## Context

While running evals for our Xulux coding agent, we found that the agent was not exploring examples correctly. We fixed the root examples discovery hint in the main coding agent, and this applies the same upstream-facing fix to the docs chat agent.

## Testing

- \git diff --check -- apps/docs/app/api/doc/chat/route.ts\ passed
- Pre-commit hook ran \oxlint --fix\ and \oxfmt\ for the staged file during commit

### DIFF
--- a/apps/docs/app/api/doc/chat/route.ts
+++ b/apps/docs/app/api/doc/chat/route.ts
@@ -363,7 +363,12 @@ export async function POST(req: Request): Promise<Response> {
                     (node): node is PageTree.Folder => node.type === "folder",
                   ),
                 ),
-                { type: "folder", name: "examples", url: "/examples" },
+                {
+                  type: "folder",
+                  name: "examples",
+                  description:
+                    "Examples of app types users can build with assistant-ui, showing instructions, recommended patterns, and UI structure.",
+                },
               ];
             }
 


Yonom (approved):
LGTM. Replacing the manual `examples` root entry's `url` with a `description` is consistent with `listChildren`'s folder shape, and navigation still works since examples are browsed via the dedicated `segments[0] === "examples"` branch keyed on path/name rather than the `url`. Small, low-risk docs-infra change; no changeset needed (`@assistant-ui/docs` is private). All checks green.

---

Issue #3723, opened 2026-04-01
@assistant-ui/react-a2a: incompatible with a2a-sdk REST handler (parts vs content, /v1/ path prefix)


surfai:
> **Note:** This issue was drafted with AI assistance (Claude Code) while integrating assistant-ui with Google's [A2UI](https://github.com/google/A2UI) project. We wanted to share our findings in case they're helpful.

First off — thank you for the excellent A2A runtime adapter! The developer experience with `useA2ARuntime` and the agent card auto-discovery is really smooth. We ran into a couple of interop issues when connecting to a Python `a2a-sdk` backend and wanted to flag them.

## Environment

- `@assistant-ui/react-a2a`: 0.2.8
- `@assistant-ui/react`: 0.12.21
- Backend: Python `a2a-sdk` with `A2ARESTFastAPIApplication` (REST transport)
- Agent: [A2UI restaurant finder sample](https://github.com/google/A2UI/tree/main/samples/agent/adk/restaurant_finder)

## Issue 1: `parts` vs `content` field name

`@assistant-ui/react-a2a` exports `A2A_PROTOCOL_VERSION = "1.0"` and sends requests to REST endpoints (`/message:send`, `/message:stream`). However, `A2AThreadRuntimeCore.threadMessageToA2AMessage()` builds message objects using a `parts` field:

```js
const a2aMsg = {
  messageId: message.id,
  role: "user",
  parts, // <-- v0.x field name
};
```

The A2A v1.0 proto definition for `Message` uses `content` instead of `parts`:

```
Available Fields: "['messageId', 'contextId', 'taskId', 'role', 'content', 'metadata', 'extensions']"
```

This causes a parse error on the `a2a-sdk` REST handler:

```
google.protobuf.json_format.ParseError: Failed to parse message field:
Message type "a2a.v1.Message" has no field named "parts" at "SendMessageRequest.message".
```

## Issue 2: Missing `/v1/` path prefix

The `a2a-sdk` REST adapter registers routes at `/v1/message:stream`, `/v1/message:send`, etc. The `A2AClient` sends requests to `/message:stream` (no prefix), resulting in a 404.

It's not entirely clear whether `/v1/` is spec-mandated or an `a2a-sdk` convention. A configurable base path (beyond the existing `tenant` option) would help interop.

## Workaround

We created a Next.js API route proxy that:
1. Translates assistant-ui's REST requests into JSON-RPC format for the backend
[...]

PR #4158, opened 2026-05-28
fix(react-a2a): send 'parts' on wire per A2A v1.0 spec

### LINKED ISSUE #4157: Standard has been violated on Message - parts vs content
### Checklist

### Environment info

This has been introduced with pull request https://github.com/assistant-ui/assistant-ui/issues/3723 

I already added info there in alignment with the standard from A2A protocol definition.
Normalization is required now on backend to change content field to parts back in order to have a protocol based implementation.
   Version:
 "@assistant-ui/react": "^0.14.11",
 "@assistant-ui/react-a2a": "^0.2.16",

### Affected area(s)

Streaming (assistant-stream)

### Runtime adapter in use

useA2ARuntime

### Steps to reproduce

Install above mentioned package.
Make a POC and implement a backend to support this.
Content is sent in the A2A communication instead of Parts

Use the A2A standard

### Browser (if applicable)

_No response_

### Code snippet or error log

```tsx

```

### Additional context

_No response_

ShobhitPatra:
# Summary

- Revert the outbound `parts → content` rename from #3725. The wire now sends `parts`, matching the A2A v1.0 spec (`A2A-Version: 1.0` header was always set; the payload now actually matches it).
- Keep the inbound `content → parts` normalization in `A2AClient.normalizeKeys` as defensive cover for any straggler A2A v0.3 server response.
- Flip the corresponding assertion in `A2AClient.test.ts`; inbound-normalization tests stay as-is.

## Why

A2A spec PR [a2aproject/A2A#1100](https://github.com/a2aproject/A2A/pull/1100) (Oct 2025) renamed the gRPC `Message.content` field to `parts`, aligning gRPC with JSON-RPC. As of A2A v1.0 both transports use `parts` — the JSON Schema is generated from the proto, so they cannot diverge.

Timeline check on `a2a-python` (confirmed against the SDK's `Message` definition at each tag):

- `v0.3.25` (Mar 10, 2026): `Message.content` — pre-#1100 schema.
- `v1.0.0-alpha.0` (Mar 17, 2026): `Message.parts` — already aligned with the post-#1100 spec.
- `v1.0.0` GA (Apr 20, 2026): `Message.parts`.

When #3725 merged (Apr 1, 2026), the v1 alpha already shipped `parts`. The bug report it was based on (#3723) was filed against a v0.3.x backend that legitimately used `content`, so #3725 made our outbound match v0.3 wire — while our client still advertises itself as `A2A-Version: 1.0`. That contradiction stayed invisible until v1.0 GA traffic showed up, and #4157 is the first report of it.

The `compat/v0_3/a2a_v0_3.proto` shim in `a2a-python` is the cleanest confirmation: it preserves `repeated Part content = 5` under an explicit `v0_3` namespace, while the live v1.0 types use `parts`.

Fixes #4157. Re-opens the original surface of #3723 only for users still on a frozen pre-v1.0 server — they should upgrade their SDK; the inbound normalization means their responses still parse.

### DIFF
--- a/.changeset/a2a-revert-content-to-parts.md
+++ b/.changeset/a2a-revert-content-to-parts.md
@@ -1,0 +1,6 @@
+---
+"@assistant-ui/react-a2a": patch
+---
+
+fix(react-a2a): send `parts` on the wire per A2A v1.0 spec.
+
--- a/packages/react-a2a/src/A2AClient.test.ts
+++ b/packages/react-a2a/src/A2AClient.test.ts
@@ -384,7 +384,7 @@ describe("A2AClient", () => {
       expect(body.message.messageId).toBe("msg-1");
     });
 
-    it("sends 'content' not 'parts' per A2A v1.0 proto spec", async () => {
+    it("sends 'parts' per A2A v1.0 spec (gRPC + JSON-RPC unified in a2aproject/A2A#1100)", async () => {
       fetchMock.mockResolvedValue(
         mockFetchResponse({
           task: { id: "t1", status: { state: "completed" } },
@@ -394,8 +394,8 @@ describe("A2AClient", () => {
       await client.sendMessage(userMessage);
 
       const body = JSON.parse(fetchMock.mock.calls[0]![1].body);
-      expect(body.message.content).toEqual([{ text: "Hello" }]);
-      expect(body.message.parts).toBeUndefined();
+      expect(body.message.parts).toEqual([{ text: "Hello" }]);
+      expect(body.message.content).toBeUndefined();
     });
 
     it("unwraps task from SendMessageResponse", async () => {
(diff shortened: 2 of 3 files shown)


Yonom (approved):
Approving.

Correctly reverts the outbound `parts → content` rename from #3725. The client always sends `A2A-Version: 1.0`, and A2A v1.0 unified both gRPC and JSON-RPC on `parts` (a2aproject/A2A#1100), so emitting `content` contradicted the advertised version. The inbound `content → parts` normalization in `normalizeKeys` is retained as backward-compat for frozen v0.3 servers, which is the right call.

Verified `A2AMessage` only contains spec-aligned wire fields (`messageId`, `contextId`, `taskId`, `role`, `parts`, `metadata`, `extensions`, `referenceTaskIds`), so the new `{ ...msg, role }` spread cannot leak internal-only fields onto the wire today.

Tests flipped appropriately, changeset is `patch`, and all CI checks are green. The remaining bot suggestions (stale `"Hi from v1 server"` fixture text, doc comment on passthrough) are non-blocking nits.

---

PR #4173, opened 2026-05-30
fix(core): make Simple attachment adapters work without FileReader

### LINKED ISSUE #4166: react-ink attachment adapters crash in terminal (FileReader is browser-only)
# react-ink attachment adapters crash in terminal (`FileReader` is browser-only)

react-ink re-exports `SimpleText`/`SimpleImage`/`CompositeAttachmentAdapter` (`packages/react-ink/src/index.ts:69-71`), but their `send()` uses `FileReader`, which doesn't exist in Node:

- `getFileText`: `core/src/adapters/attachment.ts:97` (`readAsText`)
- `getFileDataURL`: `:54` (`readAsDataURL`)

So sending an attachment in a terminal app throws `ReferenceError: FileReader is not defined`. No terminal-safe adapter exists.

## Possible fixes

1. **Make the core Simple adapters environment-agnostic.** Replace `FileReader.readAsText` with `await file.text()` and `FileReader.readAsDataURL` with `file.arrayBuffer()` + base64. These work in both browser and Node, fixing react-ink without new APIs.
2. **Ship a terminal-safe adapter in react-ink** (e.g. `createTextAttachmentAdapter`) built on `file.text()`, and stop re-exporting the FileReader-based ones from react-ink.

## Tradeoff

**For option 1:** it's the root-cause fix (the core adapters assume a DOM), so it's correct for any non-browser consumer, not just ink. The text path is a one-liner (`await file.text()`); the only real work is the image path, where `readAsDataURL` becomes `file.arrayBuffer()` then base64 then a rebuilt `data:` URL, with a small cross-env base64 helper (`Buffer` is Node-only, `btoa` is browser-only, so there's likely already a util in core/assistant-stream).

**For option 2:** core changes are best avoided when not strictly necessary. Option 1 touches a path that web (and RN, which has its own `FileReader`) hit on every attachment, so it carries blast radius and extra scrutiny purely to serve the lowest-traffic distribution. ink usage is low, so scoping the fix to react-ink keeps the risk where the problem actually is.

I'm leaning option 2: contain it in react-ink and leave the core adapters alone, given ink's low usage and the cost of changing core. If we go that route, worth a one-line note near `getFileText`/`getFileDataURL` that the Simple adapters are DOM-only by design, so this isn't silently rediscovered.

 Would appreciate a sanity check before I pick it up

ShobhitPatra:
Fixes #4166
  
`SimpleImageAttachmentAdapter` / `SimpleTextAttachmentAdapter` used the browser-only `FileReader`, so sending an attachment in a non-browser environment (e.g. `@assistant-ui/react-ink`, which re-exports them) threw `ReferenceError: FileReader is not defined`.
  
  ## Changes
  - `getFileText` now uses `file.text()`.
  - `getFileDataURL` now uses `file.arrayBuffer()` + base64, rebuilding the `data:<mime>;base64,...` URL.
  - Added a `bytesToBase64` helper that prefers `Buffer` (Node) and falls back to `btoa` (browser). No new deps.
  - Added `attachment-adapters.test.ts` covering both adapters' `send()`. It runs in vitest's `node` env (no `FileReader`), so it reproduces the original crash.

Behavior-preserving for web/RN: `file.text()` and `arrayBuffer()`+base64 produce the same output `FileReader` did. RN keeps its own `FileReader`.

### DIFF
--- a/.changeset/attachment-adapters-no-filereader.md
+++ b/.changeset/attachment-adapters-no-filereader.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/core": patch
+---
+
+fix(adapters): make Simple attachment adapters work outside the browser (no more `FileReader`).
--- a/packages/core/src/adapters/attachment.ts
+++ b/packages/core/src/adapters/attachment.ts
@@ -49,13 +49,22 @@ export class SimpleImageAttachmentAdapter implements AttachmentAdapter {
   }
 }
 
-const getFileDataURL = (file: File) =>
-  new Promise<string>((resolve, reject) => {
-    const reader = new FileReader();
-    reader.onload = () => resolve(reader.result as string);
-    reader.onerror = (error) => reject(error);
-    reader.readAsDataURL(file);
-  });
+const bytesToBase64 = (bytes: Uint8Array): string => {
+  if (typeof Buffer !== "undefined") {
+    return Buffer.from(bytes).toString("base64");
+  }
+
+  let binary = "";
+  for (let i = 0; i < bytes.length; i++) {
+    binary += String.fromCharCode(bytes[i]!);
+  }
+  return btoa(binary);
+};
+
+const getFileDataURL = async (file: File): Promise<string> => {
+  const bytes = new Uint8Array(await file.arrayBuffer());
+  return `data:${file.type};base64,${bytesToBase64(bytes)}`;
+};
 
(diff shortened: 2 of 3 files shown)


ShobhitPatra:
closing this. the core approach breaks react-native: it re-exports these same adapters with no override, and RN's Blob/File polyfill has no .text() or .arrayBuffer() - only FileReader (which is why the old
code worked there). so this swaps an ink crash for an RN one

PR #4188, opened 2026-06-01
feat(react-ag-ui): add STATE_DELTA event support

ShobhitPatra:
## Summary

Previously, `STATE_DELTA` events were logged and silently ignored. Agents emitting incremental state updates per the AG-UI spec now have those updates applied to `stateSnapshot`.

- Adds a focused JSON Patch applier at packages/react-ag-ui/src/runtime/json-patch.ts, scoped to the ops AG-UI agents emit via STATE_DELTA (add, remove, replace, move, copy, test). It is not a general-purpose RFC 6902 library: it implements that subset, validates each op, and throws on anything outside it (existence checks on replace/move/copy targets and sources, unknown ops). Full-spec strict-conformance niceties (e.g. JSON Pointer URI-fragment form) are out of scope.
- Wires it into the `STATE_DELTA` case in `AgUiThreadRuntimeCore.handleEvent` with `try` / `catch` + `logger.error`, so a malformed agent delta cannot crash the active run stream. Mirrors the `importMessagesSnapshot` neighbor's error-isolation pattern.
- Covers edge cases: empty-patch short-circuit (no clone), `from` validation on `move` / `copy` with a descriptive error, splice-insert (not overwrite) when `move` / `copy` target an array index.

## Scope

Salvaged from closed #3790, scoped down to `STATE_DELTA` only. The `onCustomEvent` callback (also from that PR) is intentionally not included and can ship separately.
Because the applier validates and throws, an invalid delta is caught and logged rather than silently corrupting state: the agent patches its own snapshot, so failing loud (in the log) is the safer default.

## Test plan

- [x] `pnpm --filter @assistant-ui/react-ag-ui vitest run test/json-patch.spec.ts test/state-delta.spec.ts`: 23/23 pass (15 `applyJsonPatch` unit tests + 1 integration test driving `STATE_DELTA` through `AgUiThreadRuntimeCore` and verifying `stateSnapshot` is updated)
- [x] `pnpm --filter @assistant-ui/react-ag-ui build`: clean
- [x] `pnpm --filter with-ag-ui build`: clean (TypeScript check passes against the runtime's public surface)

### DIFF
--- a/.changeset/state-delta.md
+++ b/.changeset/state-delta.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react-ag-ui": patch
+---
+
+feat(react-ag-ui): add STATE_DELTA event support
--- a/packages/react-ag-ui/src/runtime/AgUiThreadRuntimeCore.ts
+++ b/packages/react-ag-ui/src/runtime/AgUiThreadRuntimeCore.ts
@@ -26,6 +26,7 @@ import {
   toAgUiTools,
 } from "./adapter/conversions";
 import { createAgUiSubscriber } from "./adapter/subscriber";
+import { applyJsonPatch } from "./json-patch";
 
 const optimisticPrefix = "__optimistic__";
 const generateOptimisticId = () => `${optimisticPrefix}${generateId()}`;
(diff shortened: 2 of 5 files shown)


okisdev:
closing in favor of #4210.

that pr keeps the state_delta support, but replaces the custom patch applier with fast-json-patch to match ag-ui’s own client behavior. it also adds coverage for delta-only state updates, invalid patch isolation, and prototype pollution protection.

thanks for the work here. the narrower follow-up should be easier to review and maintain.

PR #4231, opened 2026-06-03
feat: add message queueing for sends during a run

### LINKED ISSUE #4230: Feature request: cross-runtime queued-message API + optimistic UI for sends while isRunning
### Problem statement

When a user (or programmatic CTA) tries to send a new message while the assistant is mid-response, there is no first-class way to **queue** the submission across runtimes. `ComposerRuntime.send()` is gated by `thread.isRunning`, so callers either drop the send, cancel the active run, or roll their own subscription loop.

This is a common UX expectation today — ChatGPT and Claude.app both let users type/queue a follow-up while the model is still streaming. Replicating that in assistant-ui currently requires app-level code:

```ts
const unsubscribe = aui.subscribe(() => {
  if (aui.thread().getState().isRunning) return
  unsubscribe()
  aui.thread().composer().setText(prompt)
  aui.thread().composer().send()
})
```

Even when wired up, **there is no built-in UI surface for the queued message** on non-`assistant-transport` runtimes. The user sees nothing happen until the previous run finishes, which is indistinguishable from a dropped click.

`assistant-transport` already solves this via `connectionMetadata.pendingCommands` + `useCommandQueue`, which renders optimistic messages. That capability isn't exposed for `useLangGraphRuntime` (or other runtimes), so users of those runtimes have to reinvent both the queueing and the UI.

### Proposed solution

A runtime-level queued-message API available on `ThreadRuntime` / `ComposerRuntime` regardless of the underlying runtime:

```ts
// Submit now if idle, else queue and fire when isRunning flips to false.
// Returns a handle to inspect/cancel the queued submission.
const queued = aui.thread().composer().enqueue(prompt)
queued.cancel()
queued.status // 'queued' | 'sent' | 'cancelled'
```

Paired with a UI primitive (or hook) so apps can render the queued state consistently:

```tsx
const pending = useAuiState(s => s.thread.pendingMessages)
// pending: Array<{ id, content, status }> — same shape used by assistant-transport
```

Ideally, `ComposerPrimitive.Send` itself would gain a `queueWhileRunning` opt-in so the default "send" button just works without app-level glue.

### Area

Runtime / Core

### Alternatives considered

- **Cancel + send**: `aui.thread().cancelRun()` then `composer.send()`. Destructive — yanks the in-progress response, jarring UX.
- **Bail (return false / no-op)**: feels broken to users who clicked or typed.
- **App-level subscription loop** (snippet above): works for sending, but every consumer has to also build their own pending-message UI to make the queue visible. No-op on UX consistency.
- **Switch to `assistant-transport`**: not viable for apps already on `useLangGraphRuntime` / other runtimes; the queueing benefit shouldn't require a runtime migration.

### Additional context

- `assistant-transport` precedent: `connectionMetadata.pendingCommands`, `useCommandQueue`, optimistic UI mapping in the `useConverter` docs.
- Real-world driver: implementing campaign/CTA flows that submit prompts programmatically while a previous response may still be streaming. Users expect the new prompt to be visible and to fire once the run completes — not to silently disappear.
- Related primitives already in the API: `AuiIf condition={s => s.thread.isRunning}`, `ComposerPrimitive.Cancel`. A `pendingMessages` selector + an `enqueue()` method would slot in naturally beside these.

ShobhitPatra:
Closes #4230.

Adds a generic, store level message queue so `composer.send()` buffers a message while the thread is running and flushes it when the run ends. It works across the legacy core runtimes (langgraph, external store, local) with no per runtime adapter, and builds on the queue primitives from #3684 (`ComposerPrimitive.Queue`, `QueueItemPrimitive.*`) which were previously only wired into `ExternalThread`.

## What changed

- **Store bridge owns the queue** (`composer-runtime-client.ts`): `send()` buffers a composed `AppendMessage` while the thread is running and flushes the head via `append` when it goes idle. `send({ steer: true })` cancels the current run and appends immediately. `composer.queue` plus `queueItem({ index }).remove()` / `.steer()` are surfaced.
- **Thread bridge** (`thread-runtime-client.ts`): injects a `queueController` (thread `isRunning` + `append` + `cancelRun`) into the thread composer only. Edit composers are untouched and behave exactly as before.
- **`queueWhileRunning` opt in** on `ComposerPrimitive.Send` (react, react-native, react-ink). The default "Send disabled while running" behavior is unchanged, and `assistant-transport`'s `capabilities.queue` gating is untouched. Programmatic `composer.send()` while running always buffers, which covers the CTA / campaign use case in the issue.

This lights up the existing #3684 queue primitives with zero UI changes.

## Why this approach

`useLangGraphRuntime` / `useExternalStoreRuntime` / `useLocalRuntime` all flow through the same store bridge (`RuntimeAdapter` -> `ThreadListClient` -> `ThreadClient` -> `ComposerClient`). Putting the queue there fixes every legacy core runtime at once, lives in the layer the codebase is migrating toward, and needs no per runtime adapter.

## Testing

- End to end tests on a real local runtime through the actual store path: buffer while running then flush on idle, `queueItem.remove`, `send({ steer })`, and idle send unchanged.
- Send gating matrix for `useComposerSend` across `isRunning` x `capabilities.queue` x `queueWhileRunning`.
- Full repo build, lint, and the core and react test suites all pass.

## Follow ups

These are deliberately left out to keep this PR scoped to the queue mechanism and easy to review.

- **Attachment finalization for queued sends.** Queued messages are composed from composer state (matching `ExternalThread`), so the local runtime's attachment adapter finalization that `composer.send` normally performs is skipped. Doing it properly means running the adapter's finalization at enqueue or flush time, which pulls logic back into the legacy composer core and deserves its own change and tests. The common case in this issue is text prompts, so attachment heavy queueing can land as a separate PR.
- **Server side steer.** On client only runtimes (local, external store, langgraph) `steer` is cancel + append, which is the only meaningful semantics since there is no server run to interrupt. A true server side interrupt only applies to `assistant-transport`, which already has its own mechanism, so unifying that path is a separate effort.

### DIFF
--- a/.changeset/store-level-message-queue.md
+++ b/.changeset/store-level-message-queue.md
@@ -1,0 +1,8 @@
+---
+"@assistant-ui/core": patch
+"@assistant-ui/react": patch
+"@assistant-ui/react-native": patch
+"@assistant-ui/react-ink": patch
+---
+
+feat: add a store level message queue so composer.send() buffers while the thread is running and flushes when it ends, with a queueWhileRunning opt in on ComposerPrimitive.Send (works on langgraph, external store, and local runtimes)
--- a/packages/core/src/react/primitive-hooks/useComposerSend.test.ts
+++ b/packages/core/src/react/primitive-hooks/useComposerSend.test.ts
@@ -1,0 +1,68 @@
+import { describe, it, expect } from "vitest";
+
+// Pure replica of the gating predicate in useComposerSend.ts, asserting the
+// full matrix. Keep this expression identical to the hook's `disabled` selector.
+const isDisabled = (s: {
+  canSend: boolean;
+  isRunning: boolean;
+  capabilitiesQueue: boolean;
+  queueWhileRunning: boolean;
+}) =>
+  !s.canSend || (s.isRunning && !s.capabilitiesQueue && !s.queueWhileRunning);
+
+describe("useComposerSend gating", () => {
+  it("disabled when canSend is false regardless of other flags", () => {
+    expect(
+      isDisabled({
+        canSend: false,
+        isRunning: false,
+        capabilitiesQueue: true,
+        queueWhileRunning: true,
+      }),
+    ).toBe(true);
+  });
+
+  it("enabled when idle and sendable", () => {
+    expect(
+      isDisabled({
+        canSend: true,
+        isRunning: false,
+        capabilitiesQueue: false,
+        queueWhileRunning: false,
+      }),
+    ).toBe(false);
+  });
+
+  it("disabled while running by default (no queue capability, no opt-in)", () => {
+    expect(
+      isDisabled({
+        canSend: true,
+        isRunning: true,
+        capabilitiesQueue: false,
+        queueWhileRunning: false,
+      }),
+    ).toBe(true);
+  });
+
+  it("enabled while running when capabilities.queue is true (assistant-transport)", () => {
+    expect(
+      isDisabled({
+        canSend: true,
+        isRunning: true,
+        capabilitiesQueue: true,
+        queueWhileRunning: false,
+      }),
+    ).toBe(false);
+  });
+
+  it("enabled while running when queueWhileRunning opt-in is set", () => {
+    expect(
+      isDisabled({
+        canSend: true,
+        isRunning: true,
+        capabilitiesQueue: false,
+        queueWhileRunning: true,
+      }),
+    ).toBe(false);
+  });
+});
--- a/packages/core/src/react/primitive-hooks/useComposerSend.ts
+++ b/packages/core/src/react/primitive-hooks/useComposerSend.ts
@@ -2,12 +2,14 @@ import { useCallback } from "react";
 import { useAui, useAuiState } from "@assistant-ui/store";
 import type { ComposerSendOptions } from "../../store/scopes/composer";
 
-export const useComposerSend = () => {
+export const useComposerSend = (queueWhileRunning = false) => {
   const aui = useAui();
   const disabled = useAuiState(
     (s) =>
       !s.composer.canSend ||
-      (s.thread.isRunning && !s.thread.capabilities.queue),
+      (s.thread.isRunning &&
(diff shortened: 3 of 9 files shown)


okisdev:
thanks for tackling #4230 @ShobhitPatra. we're going to land it via #4246 instead, so i'll close this one. you put real work in here, so here's the actual reasoning rather than a one-liner.

the deciding factor is where the queue state lives. the `*-runtime-client.ts` bridges are, by convention, pure projections of runtime state; the only state they own is display-ephemeral (like `isCopied` / `isHovering`). buffered domain state lives in the runtime core or a plain engine it owns: pending attachments, optimistic placeholders, and assistant-transport's `commandQueue` all follow that. this PR keeps the queue (`tapState<QueuedSend[]>` plus the flush effect) in the composer bridge, which is the one layer that by convention shouldn't own it.

that placement is also what drives the two reviewer findings. the flush effect re-fires on a stale `isRunning` snapshot before `append` propagates the new run state back through the resource tree, so a queue of two or more drains in a single idle window (reproduces with a two-item test). and because the controller is injected into every thread, a programmatic `send()` while running on assistant-transport buffers in the bridge instead of going through its own `commandQueue`. both are fixable, but the fix converges on owning the queue in the runtime with an explicit in-flight guard, which is what #4246 does.

#4246 puts the queue in a runtime-agnostic `createMessageQueue` engine, exposes it via `ExternalStoreAdapter.queue` and `unstable_enableMessageQueue`, and flips the existing `capabilities.queue` so the Send button and the Enter key read one gate rather than a separate `queueWhileRunning` prop. it covers external-store, langgraph, and local.

two things from here carried forward: the breadth goal (cover all the legacy runtimes, not just one) is something #4246 matches, and your `useComposerSend` gating-matrix test is good coverage worth keeping. really appreciate the effort, and please do send more.

ShobhitPatra:
@okisdev  thanks for taking the time to write this up Harry, really helpful. the reasoning makes sense and both findings check out. glad parts of it carry forward, will send more

PR #4263, opened 2026-06-04
feat(react-ag-ui): export message conversion helpers

### LINKED ISSUE #4262: Expose fromAgUiMessages so history adapter can use it on page load.
### Problem statement

Hi! I have an agent using adk + https://github.com/trendmicro/adk-agui-middleware, and this middleware supports GET /agents/state to fetch messages from a conversation on page load. Basically get a list of ag-ui events that I can load into the UI when page loads, without having to send an empty message.

But in order to display these messages, I need to convert ag-ui messages > assistant-ui messages, which is exactly the task of fromAgUiMessages. Can this be exported so it can be reused?

### Proposed solution

Export fromAgUiMessages and document it.

### Area

None

### Alternatives considered

Migrate away from /agents/state and do a hacky "empty message" just to get a MESSAGE_SNAPSHOT. 

### Additional context

_No response_

ShobhitPatra:
closes #4262

### DIFF
--- a/.changeset/export-ag-ui-conversions.md
+++ b/.changeset/export-ag-ui-conversions.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react-ag-ui": patch
+---
+
+feat: export fromAgUiMessages, toAgUiMessages and toAgUiTools conversion helpers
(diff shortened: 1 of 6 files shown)


ShobhitPatra:
closing in favor of #4264 same fix, minimal export surface, honest typing instead of a cast

PR #4268, opened 2026-06-04
docs: cross-file toolkit merge, MCP-only export, tool stubs and disabling

Yonom:
Updates the tools authoring docs to match the new generative-compiler behavior (see #4267) and documents two runtime-override patterns.

## Changes (`apps/docs/content/docs/tools/defining-tools.mdx`)

- **Split tools across files** — the example now shows each split file as its own `"use generative"` module that `export default defineToolkit(...)`, with the parent spreading **default imports** (`import weatherTools from "./tools/weather"`). Replaces the old example that spread named imports, which the compiler rejects. Notes that relative paths and `tsconfig` aliases (`@/tools/weather`) both resolve, and that named/opaque imports are rejected.
- **Add MCP server tools** — leads with exporting `defineMcpToolkit({ ... })` directly for an MCP-only toolkit (no longer wrapped in an otherwise-empty `defineToolkit`), and keeps the spread form for combining MCP tools with your own.
- **Tool stubs** — a clearer section on declaring a contract with `execute: stubTool()` and supplying the executor elsewhere via `useAuiToolOverrides` (for executors that close over React state).
- **Disabling a tool** — documents toggling a tool off at runtime with `useAuiToolOverrides({ name: { disabled: true } })`, in addition to the static `disabled: true`.

## Notes

- `@assistant-ui/docs` is private, so no changeset.
- Scoped to be disjoint from #4266 (which removes the Render-only / Keep-schemas sections as part of the generativeTools→AISDKToolkit migration), so the two land without conflict.

### DIFF
--- a/apps/docs/content/docs/tools/defining-tools.mdx
+++ b/apps/docs/content/docs/tools/defining-tools.mdx
@@ -288,9 +288,46 @@ the tool definition from the external system. The client build keeps only
 `type: "backend"` and the renderer (or `renderText`) for matching tool-call
 message parts.
 
-### Dynamic (stateful) tools
+### Tool stubs (supply the executor elsewhere)
 
-When a tool's executor must close over React state, declare its contract with `execute: stubTool()` and supply the real executor at runtime with `useAuiToolOverrides`. See [Dynamic tools](/docs/tools/dynamic-tools).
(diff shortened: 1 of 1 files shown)


Yonom:
Re: merge ordering — this docs PR documents the behavior added in #4267 (cross-file default-import spreads, tsconfig alias resolution, and the bare `defineMcpToolkit` default export). #4267 will be merged first, then this PR.

PR #4292, opened 2026-06-08
feat: add with-cure example project

Morrow7:
## Summary

A new example project `examples/with-cure` demonstrating assistant-cloud thread
history integration with local storage fallback and a thread list sidebar.


okisdev:
hi @Morrow7, thanks for turning the earlier issue into a real PR. i'm closing this one though, and want to be concrete about why so the next one lands.

the core issue is that `with-cure` duplicates an example we already ship. `examples/with-cloud` (and `with-cloud-standalone`) already cover everything this sets out to show: persistent thread history with assistant cloud, a thread list sidebar, the AI SDK integration, and markdown rendering. your `MyRuntimeProvider.tsx` is basically a copy of the one in `with-cloud`, so there's no new ground here for someone to learn from.

a few other things that would block it regardless:

- the name. `with-cure` doesn't describe anything. our examples are named after the integration they demonstrate (`with-langgraph`, `with-mcp`, `with-cloud`); there's no "cure" being shown.
- `app/api/chat/route.ts` uses `result.toDataStreamResponse(...)`, which is the old AI SDK v4 api. nothing in the repo uses it anymore and it won't run against the `ai@^6` you declared in package.json. the current pattern is `convertToModelMessages(messages)` + `result.toUIMessageStreamResponse()` (see `with-cloud`).
- there's a stray `.github/assets/traction.png` change and the branch is `update-traction-chart`, neither of which belongs in a "add an example" PR.
- `.gitignore` carries copy-pasted cruft (`apps/docs/app/api/xulux/...`, `.specify/`, the `.claude` lock) that isn't relevant here.
- the description mentions a "local storage fallback", but there's no local-storage code in the diff; the `baseUrl ? ... : undefined` guard just turns cloud off.

none of this is meant to put you off. like i said on the issue: the contributions worth the most are concrete fixes to something genuinely broken or unclear, not new surface that overlaps what we already have. if you were building on top of the cloud example and hit a rough edge (a confusing doc, an actual bug, a missing setup step), that's the thing to send. if you want, i can point you at a couple of good first issues to start from.

Morrow7:
I understand. However, I really need your recommendation. Thank you.

PR #4302, opened 2026-06-08
feat(react): add reactArtifact single-file React component artifact

okisdev:
## summary

- adds `reactArtifact()`, the flagship artifact kind: the model authors a single-file React component (TSX/JSX, default export) and it renders as a standalone artifact in the shared `SandboxHost`.
- the component is compiled by `@babel/standalone` **loaded inside the sandboxed iframe from a CDN**, so there is no Babel dependency in the `@assistant-ui/react` bundle. `buildReactArtifactHtml()` is pure string templating (import map + Babel script + source stash + blob-import bootstrap + error boundary + success probe).
- unlike `htmlArtifact` (a frontend tool that resolves immediately), this is a **human tool** whose render is the source of the result: the iframe posts an `aui-artifact:status` message and the render calls `addResult({ ok } / { ok: false, error })`, so a compile/runtime error feeds back to the model. human tools render standalone by default.
- ships `minimalImportMap` (default) + `claudeParityImportMap` (opt-in, lucide/recharts/framer-motion/d3/three/react-router, all `?external=react,react-dom`), and exports `buildReactArtifactHtml`.
- compose it: `defineToolkit({ create_react_artifact: reactArtifact() })`.

## design notes

- builds on `SandboxHost` (#4298): the iframe's `aui-artifact:status` / `aui-artifact:size` messages post to `"*"` and SandboxHost's cross-origin guard validates source/origin on receipt, so this does not reinvent #4108's bespoke origin handshake.
- safe-content-frame injects no CSP, so the sandboxed iframe can load Babel / esm.sh / Tailwind from CDNs.

### already verified

- [x] `pnpm --filter @assistant-ui/react exec tsc --noEmit` -> no artifact type errors (only the pre-existing, unrelated `replayBoundaryStream.test.ts`). this is the gate `aui-build` doesn't provide; it caught two real issues pre-push.
- [x] `pnpm turbo test --filter=@assistant-ui/react` -> 362/362 (`reactArtifact` 6, `buildReactArtifactHtml` 3, covering tool shape, compile-into-host, status->addResult ok/error, auto-resize, streaming guard).
- [x] `pnpm --filter @assistant-ui/react build` + oxlint/oxfmt clean.

### reviewer should verify

- [ ] in a real browser: create a React artifact, confirm it compiles + mounts + auto-resizes; force a compile/runtime error and confirm `{ ok: false, error }` feeds back to the model. (jsdom can't run the SCF iframe + CDN Babel, so the actual compile path is browser-only.)

## notes

- scope is the React render kind only. no panel/versioning/update-rewrite yet (a later slice); the artifact renders standalone via the existing `display` routing.
- the `vi.mock("safe-content-frame")` + `fakeRendered`/`flush` helpers are now shared in spirit across three artifact test files; a shared fixture is a reasonable follow-up (low priority).

### DIFF
--- a/.changeset/react-artifact.md
+++ b/.changeset/react-artifact.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react": patch
+---
+
+feat: add reactArtifact, a generative UI tool that compiles a model-authored single-file React component (Babel in a sandboxed iframe) and renders it as a standalone artifact, reporting compile/runtime status back to the model
--- a/examples/with-artifacts/app/api/chat/route.ts
+++ b/examples/with-artifacts/app/api/chat/route.ts
@@ -2,43 +2,50 @@ import { openai } from "@ai-sdk/openai";
 import {
   streamText,
   convertToModelMessages,
-  tool,
   stepCountIs,
-  zodSchema,
+  jsonSchema,
 } from "ai";
 import type { UIMessage } from "ai";
-import { z } from "zod";
 
 export const maxDuration = 30;
 
+type ToolDef = { description?: string; parameters: Record<string, unknown> };
+
+const SYSTEM =
+  "You are a helpful assistant that renders artifacts. Use create_html_artifact " +
+  "for a self-contained HTML document, and create_react_artifact for an " +
+  "interactive single-file React component (TSX with a default export). Prefer " +
+  "an artifact tool over showing code in a code block.";
+
 export async function POST(req: Request) {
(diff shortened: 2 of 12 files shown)


okisdev:
closing: artifacts are being rebuilt as a thin react-side composition over existing primitives (sandbox host, the tool render path, store resources) instead of packaged tools. htmlArtifact reverted in #4303; reactArtifact returns as part of that surface. keeping this branch as the compile-path reference (buildReactArtifactHtml + import map).

PR #4358, opened 2026-06-11
add light-themed assistant UI

Morrow7:
Adds a new light-first assistant example in examples/with-ligh


okisdev:
closing this one as well, and this close is firmer than the last one.

more than half of this diff is the `with-cure` example from #4292, resubmitted byte for byte (the git tree object is identical), with every issue i listed there still in place: the `toDataStreamResponse` call that does not exist in the `ai@6` you depend on, the copy-pasted `.gitignore`, the unrelated `.github/assets/traction.png` change. stacking a new branch on top of a closed PR without addressing any of its feedback tells us the review time we spent was wasted.

the new `with-light` part would not land on its own either. `examples/` demonstrates integrations and runtime features (`with-langgraph`, `with-mcp`, `with-cloud`); a visual theme is not one of those, styling is userland. and it was not run before submitting: the `thread-list`, `thread-list-item` and `thread-list-trigger` classes in `page.tsx` are never defined in the CSS, so the thread list renders unstyled; `globals.css` contains invalid properties like `transition-all-duration`; the summarize route returns Chinese error strings in an English example; and `tsconfig.json` skips the shared `@assistant-ui/x-buildutils/ts/next` preset that every other example extends.

stepping back from the files, the more important thing. a contribution starts with a problem that actually exists. when there is no problem and a PR is invented in order to have something to submit, the result is not neutral; every one of these consumes maintainer attention, which is the scarcest resource a project has, and that attention runs out. if after really using assistant-ui you find nothing broken or confusing, that is a fine outcome, and the honest move is to send nothing here. plenty of other projects have real, labeled problems waiting; browse GitHub trending or the issue trackers of the tools you already use and pick one there. and before contributing anywhere, learn first. reading how mature codebases are structured and tested will do far more for you right now than shipping code nobody asked for.

so the boundary, stated once: no more example PRs here, and no resubmissions of closed ones; from now on they will be closed without review. what stays open is the same thing it has always been: a small fix for a real bug or a confusing doc, found by actually using the library. that kind of PR gets a normal review like anyone else's, no more and no less.

Morrow7:
I just realized I submitted the wrong version; I intended to submit only
the "with-light" version but accidentally submitted "with-cure" instead. Of
course, if you still decide to close my submission, there is nothing I can
do about it—after all, my work simply wasn't good enough.

Yonom:
@Morrow7 Thank you nonetheless for taking the time to open a PR. Our recommendation for external authors is to:

a) open issues first for non-trivial contributions so you can collect maintainer feedback, then work on a PR
(an issue is recommended before adding a new feature, but no issue needed to fix a typo in our docs)

b) focus on smaller PRs for your first contributions

c) contribute examples by publishing them as separate repositories on your own account

I hope this is helpful and I wish you all the best :)

Morrow7:
We apologize for the inconvenience caused and thank you for your
suggestions.

PR #4363, opened 2026-06-11
docs: add commands and monorepo layout to AGENTS.md

wangyue55:
## What changed

AGENTS.md previously covered architecture, changesets, lint rules, and package boundaries, but not how to build, test, or navigate the repo. This adds the missing operational context so coding agents (and new contributors) can be productive without rediscovering it each session:

- **Commands** — install, the required initial `pnpm turbo build` (packages import each other's `dist/` outputs, per CONTRIBUTING.md), lint/autofix, the full test run, and how to scope builds/tests to a single package (turbo filters, single vitest file, `-t` by name).
- **Monorepo layout** — roles of `packages/`, `apps/`, `examples/`, `templates/`, and `python/`, including the constraint that examples resolve `@/components/assistant-ui/*` to `packages/ui` via tsconfig paths and that template components must stay byte-equal with the canonical source (`pnpm sync-templates` checks, `--write` fixes).
- **Architecture** — adds `assistant-stream` (TS/Python protocol pair) and the integration-package layer to the existing diagram, and pins down where the legacy runtime lives (`packages/react/src/legacy-runtime`, `packages/core/src/runtimes`) so the tap-migration note is actionable.

## Why

Documentation-only change; all commands and layout facts were verified against the repo (root `package.json`, `turbo.json`, `pnpm-workspace.yaml`, `scripts/sync-templates.sh`, CONTRIBUTING.md). No changeset needed since no published package is modified.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,3 +1,32 @@
+## Commands
+
+```sh
+pnpm install          # Node >= 24, pnpm 11
+pnpm turbo build      # required once after install — packages import each other's dist/ outputs
+pnpm lint             # oxlint + oxfmt --check
+pnpm lint:fix         # autofix both
+pnpm test             # turbo test (vitest per package) + test:types (tsc --noEmit per package)
+```
+
+Scope work to a single package with turbo filters, or run vitest from inside the package:
+
+```sh
+pnpm turbo build --filter=@assistant-ui/react                # build one package + its deps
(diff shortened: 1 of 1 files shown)


AVGVSTVS96:
AGENTS.md should generally be as concise as possible, we rely on the model knowing how find available commands and search the codebase

PR #4410, opened 2026-06-15
docs(examples): refine gemini landing visuals and animation

ShobhitPatra:
Brings the Gemini demo closer to the current gemini.google.com landing screen.

- Mount choreography: the glow grows out from the composer point (scale plus fade), then the heading slides up and fades in around 0.5s later, so it is staggered instead of everything appearing at once
- Glow reworked as a blurred rounded rectangle so the falloff is uniform instead of an oval, with a deeper indigo blue, centered on the composer
- Darker near black background, narrower composer, white heading
- Send button is now indigo (#1f3b9b) to match the glow, and only shows once the composer has text
- Model picker renamed Fast to Flash with the real descriptions
- Removed the welcome screen disclaimer since the real app only shows it once you are in a conversation

Animations live in a co-located gemini.css, so globals.css is untouched. Docs only, no new deps. Motion respects prefers-reduced-motion.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/4410?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


okisdev (approved):
pushed a few changes on top of this PR. here is the what, why, and what to keep in mind.

## what changed

- removed `gemini.css` and moved the glow and heading animations onto tailwind utilities (`tw-animate-css`, the same `animate-in fade-in slide-in-from-*` set the other docs examples already use). there is no co-located stylesheet at all now.
- lifted the heading above the ambient glow with `relative z-10`.
- gated the empty and conversation layouts on `s.thread.messages.length` instead of `s.thread.isEmpty`.

## why

**the css file.** the animations are fully expressible in tailwind, so a separate stylesheet is not needed. two details worth calling out: tailwind v4 compiles `-translate-x-1/2` to the standalone `translate` property while `tw-animate-css` animates `transform`, so the glow stays centered while it scales; and i added `blur-in-[90px]` so the `enter` keyframe holds the blur at 90px instead of animating it up from 0 (the original kept blur constant). reduced motion is preserved with `motion-safe:` on `animate-in`.

**the washed out heading.** the glow is `position: absolute` (a positioned element) and the `<h1>` was in normal flow (`static`), so by CSS paint order the translucent blurred glow painted on top of the heading and washed the `#1f1f1f` text to gray. the composer already avoided this with `relative z-10`; the heading did not (measured: glow box spans y=427 to 687, heading sits at y=463 to 503, fully inside it). this was already present in the stylesheet version, not introduced by the refactor.

**the composer flashing at the bottom on first load.** `thread.isEmpty` is defined as `messages.length === 0 && !isLoading` (core, `thread-runtime-client.ts`). this example runs on a cloud `useChatRuntime`, which has a roughly 300ms initial loading phase. during that window there are 0 messages but `isLoading` is true, so `isEmpty` is false and the `!isEmpty` branch (composer pinned in the `sticky bottom-0` footer) renders first, then snaps to centered once loading finishes. recorded timeline before the fix: bottom at t=20ms, centered at t=320ms. gating on `messages.length` keeps the count at 0 throughout loading, so the centered state shows immediately; after the fix the timeline goes straight to centered at t=9ms.

## what to watch for next time

- keep docs examples tailwind only. we already ship `tw-animate-css`, so reach for `animate-in fade-in slide-in-from-* zoom-in-* motion-safe:` instead of adding a `.css` file with keyframes. a per example stylesheet is the thing to avoid.
- when you lay text over a decorative `absolute` element (glow, gradient), remember positioned elements paint above static in-flow content, so the decoration can wash out or cover the text. give the text its own stacking context (`relative z-10`) or put the decoration behind it, and judge the rendered result, not just the color value.
- do not gate empty vs conversation layouts on `thread.isEmpty` when the runtime loads async. `isEmpty` folds in `isLoading` and flips mid load, which is what caused the jump. branch on `messages.length`, or handle `isLoading` explicitly if you want a loading state. the general rule: when a layout depends on runtime state, check how it behaves during the loading phase, not only once it settles.
- sanity check examples in the browser across the full load lifecycle (first paint, during load, settled) and in both light and dark. all three issues here were transient or visual and would slip past a static read of the diff.

PR #4447, opened 2026-06-17
feat(react-langchain): resume interrupts via useLangChainRespond

ShobhitPatra:
This PR surfaces v1 `useStream`'s preferred interrupt-resume path as `useLangChainRespond`. It carries the response payload cleanly via `stream.respond` and handles interrupt namespaces, unlike the `useLangChainSendCommand` to `submit(null, { command })` workaround.

Additive only. Mirrors the imperative `useLangChainSubmit` (reads `aui.thread().getState().extras`), exports the hook, and documents it in the Interrupts section and hook-mapping table.

Follow-ups (out of scope here): `respondAll` for parallel interrupts, kept separate to keep this scope minimal and easier to review.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/4447?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/langchain-respond.md
+++ b/.changeset/langchain-respond.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react-langchain": patch
+---
+
+feat(react-langchain): resume interrupts via useLangChainRespond
--- a/apps/docs/content/docs/runtimes/langchain.mdx
+++ b/apps/docs/content/docs/runtimes/langchain.mdx
@@ -353,7 +353,30 @@ Added in v0.0.2 — see issue [#3862](https://github.com/assistant-ui/assistant-
 
 ## Interrupts
 
-LangGraph interrupts pause the graph and wait for client input. `useLangChainInterruptState` exposes the current interrupt; `useLangChainSubmit` resumes the graph with a raw state update.
+LangGraph interrupts pause the graph and wait for client input. `useLangChainInterruptState` exposes the current interrupt; `useLangChainRespond` resumes it with a response payload, while `useLangChainSubmit` resumes the graph with a raw state update.
+
+`useLangChainRespond` is the cleaner resume path — it carries the response payload via `useStream().respond` and handles interrupt namespaces, so you don't have to construct a `Command`.
+
+```tsx
+import {
+  useLangChainInterruptState,
+  useLangChainRespond,
+} from "@assistant-ui/react-langchain";
+
+function InterruptPrompt() {
+  const interrupt = useLangChainInterruptState();
+  const respond = useLangChainRespond();
+  if (!interrupt) return null;
+  return (
+    <div>
+      <pre>{JSON.stringify(interrupt.value, null, 2)}</pre>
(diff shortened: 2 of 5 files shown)


okisdev (approved):
approving. the hook follows the package paradigm and the feature checks out against `@langchain/react@1.0.23`, so this is good to land. i pushed a docs + style polish (3a79dc8d1) on top of your branch; nothing in `useLangChainRespond`'s behavior or signature changed, the commit only touches the reference docs and the hook's doc comment.

**what changed**

- wrapped the new `useLangChainRespond` interrupt example in `<PlatformTabs>` with React, React Native and React Ink variants. it was a bare React-only block before, while the adjacent `useLangChainSubmit` example already shipped all three tabs.
- merged the two hook-mapping rows that both keyed `useLangGraphSendCommand` into one. it now leads with the preferred target (`useLangChainRespond`) and keeps `useLangChainSubmit(null, { command })` as the raw `Command` fallback in the notes. this clears the duplicate-key nit the review bot raised.
- named the resume path in the feature-coverage table ("send raw state update / resume command" now lists `useLangChainRespond` next to `useLangChainSubmit`).
- removed the em-dashes from the new prose and the JSDoc to match the house writing style.

**why**

the hook code itself was already idiomatic, so the runtime was left untouched. it mirrors `useLangChainSubmit` (imperative `useAui()`, extras read deferred to call time, the `asLangChainRuntimeExtras` brand guard, no `useCallback`), `respond` is added symmetrically to both the extras memo and its deps array, and the loose `Record<string, unknown>` options typing matches the existing `submit` field. the only gaps were in the docs: a single-platform example on a page that is otherwise React / RN / Ink everywhere, plus two reference tables that had drifted out of sync (a duplicate key, a missing resume entry).

**suggestions going forward**

- when you add a code example to a runtime page that uses `<PlatformTabs>`, include the React / React Native / React Ink variants from the start. the page treats all three as first-class, so a single-platform block reads as an omission.
- when you add a hook, update the two reference tables together: keep the hook-mapping left-column key unique, and name the new hook in the feature-coverage matrix so it stays accurate.
- the natural next PR is `useLangChainRespondAll`: wire `stream.respondAll` into the extras and expose it the same way as `respond`, so several interrupts pending at one checkpoint can be resolved through the branded path instead of the raw stream object. you already flagged this as a follow-up, which is the right scope to split out.
- style: the repo avoids em-dashes in prose and JSDoc; use a period, semicolon, or parentheses instead.
- the hook code needs no changes; it already follows the package paradigm, so leave it as is.

CI is green on the new commit and `@assistant-ui/react-langchain` tests pass 20/20 locally.

PR #4464, opened 2026-06-18
feat(react-langchain): add useLangChainInterrupts reader for pending interrupts

ShobhitPatra:
This PR adds `useLangChainInterrupts()`, a thin accessor returning every interrupt pending at the current checkpoint, each with `id` and `value`. Mirrors `useLangChainToolCalls`; defaults to an empty array.

Gives `useLangChainRespondAll` callers a typed way to enumerate the pending interrupts and source their ids, which nothing exposed before.

Reworked onto the new adapter layout from #4467: the reader lives in `hooks.ts` reading through `createRuntimeExtras` (`langChainExtras.use`), so no `useStreamRuntime.ts` change is needed. The `interrupts` field is already provided by the runtime; only its type was widened to carry `id`.

### DIFF
--- a/.changeset/langchain-interrupts-reader.md
+++ b/.changeset/langchain-interrupts-reader.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react-langchain": patch
+---
+
+feat(react-langchain): add useLangChainInterrupts reader for pending interrupts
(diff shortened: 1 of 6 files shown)


okisdev (dismissed):

ShobhitPatra:
Closing in favor of a fresh PR off `main` with clean history. This branch went through a rebase across two refactors (#4462, #4467) and several force-pushes, leaving stale review threads from the old layout. The replacement carries the identical final change (the `useLangChainInterrupts` reader on the current adapter layout) as a single commit on `feat/use-langchain-interrupts`.

PR #4480, opened 2026-06-18
style(ui): refine cot/tool disclosure motion, contrast & reduced-motion

ephraimduncan:
Refines the cot/tool disclosure components (`reasoning`, `tool-fallback`, `tool-group`) with a cohesive motion pass: drawer easing, press-scale feedback on triggers/approval buttons, blur-masked reveal, staggered grouped rows, snappier spinner, higher-contrast code, and `motion-reduce` guards.

Scoped to the 3 component files only — no globals.css, deps, or build changes.


okisdev (approved):
Verified across necessity, correctness, conventions, and side effects.

**Necessity:** holds. No linked issue, no competing/duplicate PR. A self-initiated motion + contrast pass on three disclosure components; value is legitimate for a UI component library that ships visual components.

**Conventions:** aligned with existing in-repo patterns.
- `motion-reduce` guards match `dot-matrix`/`number-roll`.
- Press-scale feedback (`active:scale`) matches `assistant-modal`/`tooltip-icon-button`.
- Inlined `cubic-bezier(0.32,0.72,0,1)` is correct for the self-contained copy model — `@assistant-ui/ui` components are copied byte-identical into user projects, so a shared `globals.css` token would not travel with them (the earlier `--ease-drawer` token was removed for exactly this reason).
- `pressable` scale value is now unified with the trigger value (0.98).
- No changeset required — `@assistant-ui/ui` and `assistant-ui-starter-minimal` are both `private: true` (exempt per AGENTS.md). Template sync is byte-identical.

**One non-blocking note:** the `ReasoningFade` render condition was inverted (top fade always on, bottom fade gated on `isPreview`), plus the exit `animate-out`/`fade-out` classes were dropped. This is the only non-cosmetic change. In the non-preview expanded state it reads as an improvement (top fade softens the title→content transition; the old always-on bottom fade could mislead users into thinking more content scrolls below). No code change needed, just flagging it is a behavioral change bundled with the motion work rather than a pure style edit.

CI green on the prior head; the unified `pressable` value is a one-token change with no behavioral impact.

PR #4481, opened 2026-06-18
fix(composition): track open state per accordion item in guide example

ephraimduncan:
## Problem

In the composition guide's Accordion example, `Root` held a single `open`/`setOpen` boolean in context and shared it with every `Item`. With multiple items, opening one would open them all — there was no per-item state. The final "Putting it all together" snippet also hardwired `open={false}` with a no-op setter, so nothing could ever open. On top of that, `Content` destructured `open` but rendered unconditionally, so it never actually hid its content.

## Fix

- **Root** now owns the open state internally (`useState<string | null>`), tracking *which* item's `value` is open — uncontrolled, matching the guide's learning order (controlled/uncontrolled is taught later in `state.mdx`, which lists composition as a prerequisite).
- **Item** takes a `value`, derives its own `open`/`toggle` by comparing against the Root state, and exposes them through a second `AccordionItemContext`.
- **Trigger** / **Content** read their own item's context (`useContext`), so each item toggles independently; `Content` returns `null` when closed and `Trigger` gets `aria-expanded`.
- The final example drops the broken props — each `Item` just gets `value={item.title}`.

Single-open behavior (one item open at a time). The unrelated `asChild?` prop is left untouched (separate `/as-child` concern).

## Verification

Traced the snippets: clicking item 2 sets the Root value to its `value` → only that item's `Content` renders; clicking again closes it; clicking another switches. All three defects resolved.

### DIFF
--- a/composition.mdx
+++ b/composition.mdx
@@ -1,0 +1,260 @@
+---
+title: Composition
+description: The foundation of building modern UI components.
+type: conceptual
+summary: How to break monolithic components into smaller, focused subcomponents using the compound component pattern and naming conventions.
+prerequisites:
+  - /definitions
+related:
+  - /types
+  - /state
+  - /as-child
+  - /polymorphism
+---
+
+Composition, or composability, is the foundation of building modern UI components. It is one of the most powerful techniques for creating flexible, reusable components that can handle complex requirements without sacrificing API clarity.
+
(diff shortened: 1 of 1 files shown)


ephraimduncan:
Closing: wrong approach — this copied the guide markdown into the repo instead of fixing the source in place.

PR #4259, opened 2026-06-04
fix: handle LangGraph reasoning without summary

puneetdixit200:
Fixes #4257.

## Summary
- handle LangChain reasoning content blocks that provide a `reasoning` string without a `summary` array
- keep existing summary-array handling unchanged and fall back to an empty reasoning string if neither field is present
- add a regression test for the summary-less reasoning block shape
- add a patch changeset for `@assistant-ui/react-langgraph`

## Verification
- `pnpm --filter @assistant-ui/react-langgraph test -- convertLangChainMessages.test.ts -t "falls back to reasoning text"` failed before the fix with `TypeError: Cannot read properties of undefined (reading 'map')`\n- `pnpm --filter @assistant-ui/react-langgraph test -- convertLangChainMessages.test.ts -t "falls back to reasoning text"`\n- `pnpm --filter @assistant-ui/react-langgraph build`\n- `pnpm exec oxfmt --check packages/react-langgraph/src/convertLangChainMessages.ts packages/react-langgraph/src/convertLangChainMessages.test.ts packages/react-langgraph/src/types.ts`\n- `pnpm exec oxlint packages/react-langgraph/src/convertLangChainMessages.ts packages/react-langgraph/src/convertLangChainMessages.test.ts packages/react-langgraph/src/types.ts --no-error-on-unmatched-pattern`\n- `git diff --check`\n\nAI assistance was used for implementation and local verification; I reviewed the changes and test output before opening this PR.

PR #4487, opened 2026-06-19
fix(react-langchain): handle reasoning content without a summary

ShobhitPatra:
## What

The `reasoning` content-block converter assumed `summary` is always present and called `part.summary.map(...)`, which throws when a provider sends a plain `reasoning` string instead.

## Change

Make the `reasoning` variant tolerant and fall back: `summary ?? reasoning ?? ""`. Mirrors the handling already in `react-langgraph`, so the two converters reach parity.

## Tests

Added three cases to `convertMessages.test.ts`: summary blocks join to `"a\n\n\nb"`, a bare `reasoning` string passes through, and an empty block yields `""` without throwing.

Internal converter robustness only. No public-API change, no new deps.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/4487?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/react-langchain-reasoning-no-summary.md
+++ b/.changeset/react-langchain-reasoning-no-summary.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react-langchain": patch
+---
+
+fix(react-langchain): handle reasoning content without a summary
--- a/packages/react-langchain/src/convertMessages.test.ts
+++ b/packages/react-langchain/src/convertMessages.test.ts
@@ -8,6 +8,12 @@ const humanMessage = (content: unknown): LangChainBaseMessage => ({
   content,
 });
 
+const aiMessage = (content: unknown): LangChainBaseMessage => ({
+  _getType: () => "ai",
+  id: "msg-1",
+  content,
+});
+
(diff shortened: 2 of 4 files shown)


okisdev:
closing in favor of #4497, which folds this in and extends it. that PR keeps your `summary ?? reasoning ?? ""` guard and the optional-`summary` / `reasoning?` type change, then covers the sibling cases the reasoning-only patch left exposed: the `summary` array element (`s?.text ?? ""`, so `summary: [null]` no longer throws), the `image_url` branch (dropped at the converter when there's no usable url instead of dereferencing `.url`), and the same hardening mirrored into react-langgraph and core's `fromThreadMessageLike` / `sanitizeImageContent`. you're credited via `Co-authored-by` on the squashed commit.

thanks for the original fix and the clean port from #4259.

Issue #4486, opened 2026-06-19
[Bug] Crashes on undefined `summary` / `text` in reasoning content blocks from non-spec SSE streams


FireTable:
## Summary

`@assistant-ui/react-langchain` and `@assistant-ui/core` both crash when an assistant message contains a reasoning content block whose `summary` or `text` field is missing. This happens in practice because some model providers emit reasoning SSE chunks that don't fully populate the spec'd fields (e.g. Anthropic Claude extended thinking sometimes streams a `reasoning` block before `summary` is materialized, or omits `summary` entirely on shorter responses).

The library should defend against these fields being `undefined` instead of throwing.

### 1. `packages/react-langchain/src/convertMessages.ts` — line 43

```ts
case "reasoning":
  return {
    type: "reasoning" as const,
    text: part.summary.map((s) => s.text).join("\n\n\n"),  // ← throws if summary is undefined
  };
```

### 2. `packages/core/src/runtime/utils/thread-message-like.ts` — line 163

```ts
case "text":
case "reasoning":
  if (part.text.trim().length === 0) return null;  // ← throws if text is undefined
  return part;
```

## Stack trace (from a real Next.js + LangChain + Claude-with-extended-thinking app)

```
TypeError: Cannot read properties of undefined (reading 'map')
    at convertMessages.ts:43:32
    at Array.map (<anonymous>)
    at contentToParts (convertMessages.ts:20:6)
    at Object.convertLangChainBaseMessage [as callback] (convertMessages.ts:111:22)
    at useExternalMessageConverter (external-message-converter.ts:426:29)
    at useStreamThreadRuntime (useStreamRuntime.ts:212:26)
    at Object.RuntimeHook [as current] (useStreamRuntime.ts:347:14)
    at useRemoteThreadListRuntime.useCallback[stableRuntimeHook] (useRemoteThreadListRuntime.ts:51:27)
```

## Suggested fix

**react-langchain** — guard `summary`:

```diff
 case "reasoning":
   return {
     type: "reasoning" as const,
-    text: part.summary.map((s) => s.text).join("\n\n\n"),
+    text: (part.summary ?? []).map((s) => s.text).join("\n\n\n"),
   };
```

**core** — guard `text`:

```diff
 case "text":
 case "reasoning":
-  if (part.text.trim().length === 0) return null;
+  if (!part.text?.trim()) return null;
   return part;
```

## Why this matters

[...]

PR #4488, opened 2026-06-19
fix(core): tolerate reasoning parts with undefined text

ShobhitPatra:
This PR fixes part of #4486.

## What

`fromThreadMessageLike` called `part.text.trim()` for `text` and `reasoning` parts, which throws when a provider streams a reasoning block before `text` is populated (Claude extended thinking, OpenAI o-series, etc.).

## Change

Guard with `if (!part.text?.trim()) return null;`. An undefined or whitespace-only reasoning part is dropped instead of crashing the runtime. Behavior is unchanged for populated parts.

## Tests

Added three cases to `thread-message-like.test.ts`: an undefined-text reasoning part is dropped without throwing, a whitespace-only one is dropped, and a part with real text is kept.

Internal robustness only. No public-API change, no new deps.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/4488?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/core-reasoning-undefined-text.md
+++ b/.changeset/core-reasoning-undefined-text.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/core": patch
+---
+
+fix(core): tolerate reasoning parts with undefined text
--- a/packages/core/src/runtime/utils/thread-message-like.ts
+++ b/packages/core/src/runtime/utils/thread-message-like.ts
@@ -160,7 +160,7 @@ export const fromThreadMessageLike = (
             switch (type) {
               case "text":
               case "reasoning":
-                if (part.text.trim().length === 0) return null;
+                if (!part.text?.trim()) return null;
                 return part;
 
               case "file":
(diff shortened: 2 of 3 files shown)


okisdev:
closing in favor of #4497. it keeps your `!part.text?.trim()` guard on the text/reasoning branch and adds the sibling the same malformed-stream class reaches: `sanitizeImageContent` now drops a non-string `image` instead of throwing on `image.match(...)`, so an `image_url` block with a missing url can't crash core either. consolidated with the react-langchain and react-langgraph converter guards so the whole class lands in one patch. you're co-credited on the squashed commit.

thanks for catching the core gap that the langchain report didn't mention.

PR #4492, opened 2026-06-19
build stage

zmzlois:


packages/cli/src/commands/deploy.ts
```diff
@@ -0,0 +1,56 @@
+import { Command } from "commander";
+import chalk from "chalk";
+import * as path from "node:path";
+import { logger } from "../lib/utils/logger";
+import {
+  buildEve,
+  detectProject,
+  type BuildEveOptions,
```
zmzlois, line 8: this should be vercel options

packages/cli/src/commands/deploy.ts
```diff
@@ -0,0 +1,56 @@
...
+          `Detected ${chalk.bold.green("eve")} project${
+            info.eveVersion ? chalk.dim(` (eve ${info.eveVersion})`) : ""
+          }`,
+        );
+        for (const reason of info.reasons) {
+          logger.info(`  ${chalk.dim("•")} ${reason}`);
+        }
+
+        try {
+          const buildOptions: BuildEveOptions = { cwd };
```
zmzlois, line 38: same here

packages/cli/src/lib/utils/logger.ts
```diff
@@ -20,3 +27,9 @@ export const logger = {
     console.log("");
   },
 };
+
+export const debug = (message: string) => {
```
zmzlois, line 31: enable internal debug logs

PR #4495, opened 2026-06-20
fix: pin vulnerable transitive dependencies

ProfTrader:
## Summary
- Add pnpm workspace overrides for vulnerable transitive production dependencies.
- Refresh the lockfile so production audit paths resolve to patched versions.

## Security impact
Clears production advisories for:
- tar path traversal / file-smuggling advisories
- undici TLS/cache/header/WebSocket advisories
- DOMPurify config pollution advisory
- PostCSS stringify XSS advisory
- js-yaml merge-key DoS advisory
- @opentelemetry/core baggage memory allocation advisory
- uuid buffer bounds advisory
- @tootallnate/once control-flow advisory

## Verification
- `pnpm install --lockfile-only`
- `pnpm audit --prod --json` -> 0 vulnerabilities
- `pnpm lint`
- `git diff --check`

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/4495?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/mcp-docs-server/package.json
+++ b/packages/mcp-docs-server/package.json
@@ -41,11 +41,12 @@
   },
   "dependencies": {
     "@modelcontextprotocol/sdk": "^1.29.0",
-    "gray-matter": "^4.0.3",
+    "js-yaml": "^4.2.0",
     "zod": "^4.4.3"
   },
   "devDependencies": {
     "@assistant-ui/x-buildutils": "workspace:*",
+    "@types/js-yaml": "^4.0.9",
     "@types/node": "^25.9.3",
     "tsx": "^4.22.4",
     "vitest": "^4.1.8"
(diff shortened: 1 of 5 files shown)


okisdev:
thanks for the security pass here. the recent dependency update in #4517 regenerated the lockfile and pulled patched versions for most of these advisories, so the manual pnpm overrides are largely redundant now. only a few items remain (mostly transitive through example and build tooling), and we'll handle those separately rather than carry a hand maintained overrides block. closing for that reason, appreciate the contribution.

PR #4511, opened 2026-06-21
fix(cli): ship local components in default create template

Gujiassh:
## Summary
- ship local components with the default create template so `--skip-install` no longer routes through `shadcn add`
- mark the default template as local-components backed in the CLI metadata
- add focused regression coverage for the create path and template sync

## Testing
- `bash scripts/sync-templates.sh`
- `git diff --check`
- `oxfmt --check <touched files>`
- `node <vitest.mjs> run packages/cli/test/lib/create-project.test.ts packages/cli/test/commands/create.test.ts`

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/4511?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


okisdev:
closing as superseded by #4609.

the install-path root cause this PR targets is now fixed with a one-line gate in `transformProject`: `installShadcnRegistry` runs behind `!opts.skipInstall`, matching the existing guard on `installDependencies`. `--skip-install` now skips every install path for all six registry templates (default, cloud, cloud-clerk, langchain, mcp, eve), without copying components into the default template and without reverting the registry-resolve model from #4054.

that makes the main body of this PR (the ~4000 lines of local component copies under `templates/default`) unnecessary: the `hasLocalComponents: true` switch was a template-level workaround for a CLI-level missing gate, and the gate is now in place.

two slices here are independent of the install-path fix and are worth salvaging into a small follow-up PR:

- `packages/ui/src/components/assistant-ui/attachment.tsx`: the `<div role="button" tabIndex={0}>` -> `<button type="button">` change (keyboard activation for the attachment preview dialog trigger).
- `packages/ui/src/components/ui/tooltip.tsx`: the shared `TooltipProvider` restore (cubic flagged the provider-shadowing case this addresses).

both are genuine source fixes that ship to consumers via the registry regardless of the local-components decision, so they stand on their own. happy to open that follow-up if it helps.

PR #4524, opened 2026-06-22
feat(react-ag-ui): infer pending tool-call status in fromAgUiMessages

### LINKED ISSUE #4522: Infer pending tool-call status in fromAgUiMessages
### Problem statement

Problem: During a live AG-UI run, RunAggregator marks the assistant message as { type: 'requires-action', reason: 'tool-calls' } when the run ends with unresolved tool calls. That’s what powers HITL UIs (status === 'requires-action', submit tool result, resume run).

But fromAgUiMessages only reconstructs content (text, reasoning, tool-call parts + merged tool results). It does not restore message status. So when you load an old thread via thread.reset(...), pending ask_user calls look like they’re still running instead of actionable — which is why your ThreadDetailsLogger + unanswered HITL test need the wrapper.

### Proposed solution

Extend FromAgUiMessagesOptions:
```
export type FromAgUiMessagesOptions = {
  ...,
  inferPendingToolCallStatus?: boolean;
};
```
Implementation sketch (end of fromAgUiMessages, after tool-result merging):

```
if (inferPendingToolCallStatus !== false) {
  for (let i = 0; i < converted.length; i++) {
    const msg = converted[i];
    if (msg.role !== "assistant" || !Array.isArray(msg.content)) continue;
    const hasPendingToolCall = msg.content.some(
      (part) =>
        isObject(part) &&
        part.type === "tool-call" &&
        part.result === undefined,
    );
    if (hasPendingToolCall) {
      converted[i] = {
        ...msg,
        status: { type: "requires-action", reason: "tool-calls" },
      };
    }
  }
}
```

### Area

None

### Alternatives considered

_No response_

### Additional context

_No response_

ShobhitPatra:
## What

`fromAgUiMessages` now restores `status: { type: "requires-action", reason: "tool-calls" }` on reconstructed assistant messages that still have an unresolved tool call. A new `inferPendingToolCallStatus` option (default `true`) gates the behavior, with `false` to opt out.

## Why

During a live run, `RunAggregator` ends an unresolved run as `requires-action` / `tool-calls`, which is what powers HITL UIs (status drives "submit tool result" and "resume run").

`fromAgUiMessages` only reconstructed content (text, reasoning, tool-call parts, merged tool results) and never restored status. So loading an old thread via `thread.reset(...)` made pending calls (e.g. `ask_user`) look like they were still running instead of actionable.

This closes that round-trip gap so a cold reload matches the live run, the same principle the existing `showThinking` option already follows.

## How

Status can only be set once every message is converted, because tool results merge into their call parts from later `tool` messages. So this runs as a post-pass at the end of `fromAgUiMessages`, after the merge loop: for each assistant message with array content and no existing status, if any tool-call part has `result === undefined`, it is marked `requires-action` / `tool-calls`.

`inferPendingToolCallStatus` defaults to `true`, mirroring `showThinking`. The opt-out covers the one case the heuristic cannot see: the wire carries no run status, so a run the aggregator force-completed on a success outcome despite unresolved calls would still be inferred as `requires-action`. Setting `false` disables inference for backends where that matters.

## Testing

Added three converter tests: unresolved tool call infers the status, a resolved call does not, and `{ inferPendingToolCallStatus: false }` skips inference. Updated the AG-UI runtime-options docs. Full package test suite (175) passes, lint and format clean, build green.

Closes #4522

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/4524?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/ag-ui-infer-pending-tool-status.md
+++ b/.changeset/ag-ui-infer-pending-tool-status.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react-ag-ui": patch
+---
+
+feat(react-ag-ui): infer requires-action status for unresolved tool calls in fromAgUiMessages
--- a/apps/docs/content/docs/runtimes/ag-ui/runtime-options.mdx
+++ b/apps/docs/content/docs/runtimes/ag-ui/runtime-options.mdx
@@ -66,6 +66,11 @@ messages are gone on the next page load.
 `showThinking: false`, so imported reasoning messages are dropped at conversion
 time, the same way a live run never stores them.
 
+By default `fromAgUiMessages` also restores `requires-action` status on assistant
+messages that still have unresolved tool calls, so a reloaded thread surfaces
+pending human-in-the-loop calls as actionable instead of stuck running, matching
+the live run. Pass `{ inferPendingToolCallStatus: false }` to opt out.
+
 `fromAgUiMessages` reconstructs the text, reasoning, and tool calls of each message. Multimodal user input (`image`, `audio`, `video`, and `document` parts, as well as legacy `binary` parts) is restored as attachments on the user message, so a backend that persists multimodal messages shows them again on reload and re-sends them on the next run. Legacy `binary` parts that only reference a file id are not restored.
 
 ## Thread list (experimental)
--- a/packages/react-ag-ui/src/runtime/adapter/conversions.ts
+++ b/packages/react-ag-ui/src/runtime/adapter/conversions.ts
@@ -479,6 +479,12 @@ export type FromAgUiMessagesOptions = {
    * Matches the `showThinking` option of `useAgUiRuntime`. Defaults to `true`.
    */
   showThinking?: boolean;
+  /**
+   * Whether to restore `requires-action` status on assistant messages that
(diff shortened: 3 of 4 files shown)


okisdev:
thanks for digging into this, the gap you identified is real and the fix ships in #4538. closing this one in favor of that.

the difference is mostly shape: instead of a new `inferPendingToolCallStatus` option, #4538 reuses core's `getAutoStatus`, so AG-UI reload derives status the same way every other external-store runtime already does (react-langchain, react-ai-sdk, etc.) and no new flag is needed. the wire snapshot carries no run outcome, so the success-outcome case can't be distinguished from a genuine pending call on reload for any adapter, which is why #4538 drops the opt-out and documents that limitation rather than describing it as matching the live aggregator.

PR #4555, opened 2026-06-23
feat: add createAISDKRoute helper with escape hatches & options

Kinfe123:
## Summary

Adds `createAISDKRoute` to `@assistant-ui/react-ai-sdk`, a small route helper for the common assistant-ui + AI SDK server route pattern.

Instead of every app hand-writing the same `req.json()` → `convertToModelMessages()` → `aiToolkit.tools({ frontend: tools })` → `streamText()` → `toUIMessageStreamResponse()` flow, users can define the route as:

```ts
export const { POST } = createAISDKRoute({
  toolkit,
  model: openai("gpt-5.4-nano"),
  streamText: {
    stopWhen: stepCountIs(10),
  },
});
```

## Why

The existing manual route setup is flexible, but it asks users to understand several pieces before they can ship a normal chat endpoint:

- how assistant-ui sends frontend tool schemas in the request body
- how `AISDKToolkit.tools({ frontend })` merges frontend tools with server toolkit tools
- how UI messages should be converted with the final AI SDK tool set
- how to return an AI SDK UI message stream response

`createAISDKRoute` keeps that default path short while preserving the same underlying primitives for advanced cases.

## Helper Use Cases

- **Basic chat route**: create a working AI SDK route from a toolkit and model with minimal boilerplate.
- **Toolkit-backed tools**: automatically expose server toolkit tools and merge request-provided frontend tool schemas.
- **Example/template DX**: make examples and starters easier to read by keeping the route focused on configuration.
- **Multiple assistant profiles**: create separate route files for billing, analytics, support, docs, etc. by passing a different `toolkit`, `model`, or route options.
- **Tenant-aware routes**: use `context` and dynamic `model`, `system`, or `streamText` options for auth, org config, provider options, or per-tenant headers.
- **Advanced route customization**: override `messages`, `tools`, `prepareStreamText`, `response`, `toResponse`, or `onError` without leaving the helper.

## What Changed

- Added `createAISDKRoute` and exported it from `@assistant-ui/react-ai-sdk`.
- Added route option types for request body, call settings, response options, and helper configuration.
- Added tests for the default route wiring and advanced escape hatches.
- Updated the AI SDK v6 example route to use `createAISDKRoute`.
- Moved the example weather tool into `app/toolkit.tsx` so the route demonstrates the intended compact helper usage.
- Added a patch changeset for `@assistant-ui/react-ai-sdk`.

## Escape Hatches

The helper intentionally does not hide the AI SDK. Callers can still customize the route with:

- `parseBody` for non-standard request parsing
- `context` for auth/session/tenant lookup
- dynamic `model` and `system`
- custom `messages` conversion
- custom `tools` filtering or augmentation
- `streamText` defaults like `stopWhen`, `providerOptions`, headers, max output tokens, and temperature
- `prepareStreamText` for final full control before calling `streamText`
- `response` for `toUIMessageStreamResponse` options
- `toResponse` for fully custom response handling
- `onError` for route-level error responses

## Validation

- `pnpm exec oxfmt --check ...`
- `pnpm exec oxlint ...`
- `pnpm --filter with-ai-sdk-v6 exec tsc --noEmit`
- `pnpm --filter @assistant-ui/react-ai-sdk exec vitest run src/createAISDKRoute.test.ts`
- Local browser smoke test against `examples/with-ai-sdk-v6`: `POST /api/chat 200` and weather tool response rendered in the chat UI.

Note: local checks showed the existing Node engine warning because this machine is on Node `v23.11.0` and the repo requests `>=24`.

### DIFF
--- a/.changeset/fuzzy-routes-smile.md
+++ b/.changeset/fuzzy-routes-smile.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react-ai-sdk": patch
+---
+
+feat: add createAISDKRoute helper for AI SDK chat routes
--- a/examples/with-ai-sdk-v6/app/api/chat/route.ts
+++ b/examples/with-ai-sdk-v6/app/api/chat/route.ts
@@ -1,50 +1,15 @@
 import { openai } from "@ai-sdk/openai";
-import { frontendTools } from "@assistant-ui/react-ai-sdk";
-import {
-  type JSONSchema7,
-  streamText,
-  convertToModelMessages,
-  type UIMessage,
-  tool,
-  stepCountIs,
-  zodSchema,
-} from "ai";
-import { z } from "zod";
+import { createAISDKRoute } from "@assistant-ui/react-ai-sdk";
+import { stepCountIs } from "ai";
+import toolkit from "../../toolkit";
 
 // Allow streaming responses up to 30 seconds
 export const maxDuration = 30;
 
-export async function POST(req: Request) {
-  const {
-    messages,
-    system,
-    tools,
-  }: {
-    messages: UIMessage[];
-    system?: string;
-    tools?: Record<string, { description?: string; parameters: JSONSchema7 }>;
-  } = await req.json();
-
-  const result = streamText({
-    model: openai("gpt-5.4-nano"),
-    messages: await convertToModelMessages(messages),
-    ...(system ? { system } : {}),
+export const { POST } = createAISDKRoute({
+  toolkit,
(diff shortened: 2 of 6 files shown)


Kinfe123:
I'm closing this in favor of making sure to stay as customizable as possible with out having the escape hatches grows whenever some function stubs inside those abstractions. so to stay less magical and reuse already existed best practices like openai sdk and such it is much better to do with raw way and not confuse anyone with wrapper approach covering only small sets of escape hatches and options

PR #4560, opened 2026-06-23
fix: warn about duplicate generative toolkit spread names

Kinfe123:
## Summary

Adds a compile-time warning for duplicate tool names when a `"use generative"` toolkit composes multiple toolkit fragments with object spread.

Today this pattern is supported:

```tsx
export default defineToolkit({
  ...weatherTools,
  ...databaseTools,
  ...calendarTools,
});
```

But if two fragments define the same tool key, normal JavaScript object spread silently overwrites the earlier tool before `defineToolkit` ever sees the object. That makes the resulting model/tool UI behavior depend on spread order, with no assistant-ui-specific signal.

## What Changed

- Tracks static tool names for compiler-visible toolkit spreads.
- Reads default-exported tool names from imported `"use generative"` toolkit modules without executing user code.
- Checks inline entries, local `defineToolkit(...)` fragments, `defineMcpToolkit(...)` fragments, and imported generative default toolkit spreads for duplicate static names before routing the toolkit.
- Emits a warning rather than failing the build, preserving existing JavaScript object-spread behavior while making the overwrite visible.
- Emits duplicate-name warnings from the client compile pass so split client/server builds log the warning once.
- Treats a toolkit fragment's spread-visible names as its unique public keys, so an internal duplicate is not reported again by every parent spread.
- Scopes imported toolkit-name caching to a single compile pass so rebuilds pick up changed toolkit fragments.
- Renames the shared generative compiler diagnostic prefix from `[assistant-ui/next]` to `[assistant-ui/use-generative]`, since the compiler is used beyond the Next adapter.
- Documents that duplicate names across compiler-visible spreads now warn.

## Warning Behavior

If composition includes a duplicate, the generative compiler now logs a warning and continues compiling:

```txt
[assistant-ui/use-generative] /path/to/src/toolkit.tsx: Duplicate tool name "search" while composing toolkits. JavaScript object spread keeps the last definition.
```

For example, this warns when both `weatherTools` and `calendarTools` expose `search`:

```tsx
export default defineToolkit({
  ...weatherTools,
  ...databaseTools,
  ...calendarTools,
  ...emailTools,
});
```

The output keeps normal JavaScript semantics: the later `search` definition wins.

## Validation

- `pnpm --filter @assistant-ui/x-generative-compiler test`
- `pnpm --filter @assistant-ui/x-generative-compiler build`
- `pnpm lint`

Note: local `pnpm` emitted the existing Node engine warning because this machine is on Node v23.11.0 while the repo requests `>=24`, but all commands exited successfully.

### DIFF
--- a/.changeset/generative-toolkit-duplicate-spreads.md
+++ b/.changeset/generative-toolkit-duplicate-spreads.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/x-generative-compiler": patch
+---
+
+fix: warn about duplicate tool names across generative toolkit spreads
--- a/apps/docs/content/docs/tools/defining-tools.mdx
+++ b/apps/docs/content/docs/tools/defining-tools.mdx
@@ -423,6 +423,10 @@ The compiler splits each file across the client/server boundary on its own, then
 - **Spread a default import** (`import weatherTools from "./tools/weather"`). Relative paths and `tsconfig` path aliases like `@/tools/weather` both resolve. Only the default export crosses the generative-module boundary, so a named import (or any opaque, non-generative import) is rejected.
 - You can also spread a local `defineToolkit(...)` or `defineMcpToolkit(...)` binding declared in the same file.
 
+The compiler checks static tool names across inline entries and compiler-visible
+spreads. If two fragments define the same tool name, the build warns that
+JavaScript object spread will keep the later tool definition.
+
 ### Add MCP server tools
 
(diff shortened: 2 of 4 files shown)


Yonom (approved):

Kinfe123:
this will be augmented with new composable toolkit file structure with much more optimized way of detecting those duplicates and such with helper function to do a safe merge.

PR #4562, opened 2026-06-23
fix(react): disable smooth text streaming under prefers-reduced-motion

ephraimduncan:
## What

`useSmooth` drives a per-character `requestAnimationFrame` typewriter reveal (`TextStreamAnimator`) and is **on by default** for the primary assistant text renderers: `MessagePartText` and `MarkdownText` both default `smooth = true` and route through `useSmooth`. `StreamdownText` defaults `smooth = false` but inherits the same gate when a consumer opts in. The `enabled` gate never consulted the user's reduced-motion preference, so a user who has requested reduced motion still received continuous text-motion on every message.

## Why

This violates **WCAG 2.3.3 (Animation from Interactions)** / `prefers-reduced-motion`. There was previously no reduced-motion handling anywhere in `@assistant-ui/react`.

## How

- Gate the reveal on the existing `useMediaQuery("(prefers-reduced-motion: reduce)")` hook and fold it into `useSmooth`'s `enabled` flag. When reduced motion is requested, the animator never starts and the full text/status commits immediately. Reduced motion takes precedence over an explicit `smooth={true}`, so authors cannot override the user's motion preference. The existing `!enabled` paths already do the right thing (stop the animator, return `state` unchanged), so this is a one-line behavioral gate.
- Because the gate lives in the single shared `useSmooth` hook, all three text renderers (and any external `useSmooth` consumer) inherit the behavior.
- Harden `useMediaQuery` to no-op when `window.matchMedia` is absent, so the new hook usage is safe in environments without it (jsdom tests, edge runtimes). No impact in real browsers, and the sole existing consumer (`ComposerInput`) is unaffected.

## Tests

- Added a `useSmooth` test asserting a running part returns its full text immediately under `prefers-reduced-motion: reduce`.
- `useSmooth` suite: 10/10 pass. `react-streamdown` suite (transitive consumer): 118/118 pass. Lint clean.

Patch changeset included for `@assistant-ui/react`.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/4562?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


okisdev (approved):
lgtm. the reduced-motion gate folds into useSmooth's existing enabled flag, so all three text renderers inherit it from one place and the !enabled short-circuits already commit the full text immediately. reduced motion correctly takes precedence over an explicit smooth=true. the useMediaQuery hardening is strictly safer and leaves ComposerInput unaffected. tests assert the no-reveal contract by referential equality. doc pass adds the missing MarkdownText JSDoc and records the precedence rule.

Yonom (approved):

PR #4570, opened 2026-06-23
feat: add toolkit composition with duplicate warnings

Kinfe123:
## What

Add `mergeToolkits` so apps can split tools by feature and compose them from a single toolkit entrypoint. The helper is left-biased: if two toolkits declare the same model-facing tool name, it keeps the first definition and logs a warning instead of silently overriding it.

This also lets `useAui` accept a plain toolkit for the `tools` scope, so app code can use:

```tsx
const aui = useAui({ tools: appToolkit });
```

instead of wrapping the toolkit manually with `Tools({ toolkit })`.

## Toolkit folder DX

This is the intended shape for larger apps:

```txt
app/toolkits/
  files.tsx
  tasks.tsx
  calendar.tsx
  crm.tsx
  approvals.tsx
  index.tsx
```

Each feature file owns one focused toolkit:

```tsx
// app/toolkits/files.tsx
"use generative";

import { defineToolkit } from "@assistant-ui/react";

export const filesToolkit = defineToolkit({
  search_files: {
    description: "Search files the user has access to.",
    parameters: z.object({ query: z.string() }),
    execute: async ({ query }) => searchFiles(query),
    render: ({ args, result }) => <FileSearchResult query={args.query} result={result} />,
  },
});
```

The folder entrypoint composes those feature toolkits once:

```tsx
// app/toolkits/index.tsx
import { mergeToolkits } from "@assistant-ui/react";
import { approvalsToolkit } from "./approvals";
import { calendarToolkit } from "./calendar";
import { crmToolkit } from "./crm";
import { filesToolkit } from "./files";
import { tasksToolkit } from "./tasks";

export const appToolkit = mergeToolkits(
  filesToolkit,
  tasksToolkit,
  calendarToolkit,
  crmToolkit,
  approvalsToolkit,
);
```

That gives users one import to share across client runtime setup and server route setup, while keeping feature teams from editing one huge `toolkit.tsx` file.

## Client and server usage

Client runtime setup can pass the composed toolkit directly:

```tsx
// app/MyRuntimeProvider.tsx
"use client";

import { AssistantRuntimeProvider, AuiProvider, useAui } from "@assistant-ui/react";
import { appToolkit } from "./toolkits";

export function MyRuntimeProvider({ children }: { children: React.ReactNode }) {
  const aui = useAui({ tools: appToolkit });
  const runtime = useChatRuntime();

  return (
    <AuiProvider value={aui}>
      <AssistantRuntimeProvider runtime={runtime}>{children}</AssistantRuntimeProvider>
    </AuiProvider>
  );
}
```

The server route can import the same composed toolkit and hand it to the AI SDK adapter:

```ts
// app/api/chat/route.ts
import { AISDKToolkit } from "@assistant-ui/react-ai-sdk";
import { appToolkit } from "../../toolkits";

const aiToolkit = new AISDKToolkit({ toolkit: appToolkit });

const result = streamText({
  model,
  messages,
  tools: await aiToolkit.tools({ frontend: tools }),
});
```

On the client, `useAui({ tools: appToolkit })` registers model-visible tool schemas and tool renderers through the existing tools scope. On the server, `AISDKToolkit` receives the same merged toolkit and exposes those tools to the model.

## Non-breaking behavior

This is additive DX, not a breaking change. Existing `useAui()` calls and existing `useAui({ tools: Tools({ toolkit }) })` calls continue to work. The new wrapper only normalizes `tools` when the value is a plain toolkit object; existing resource elements are passed through unchanged.

Duplicate tool names are not an error because that would be disruptive while composing feature folders, but they are visible: `mergeToolkits` warns once per duplicate name and keeps the first definition. That avoids normal object-spread behavior where later tools silently override earlier ones.

## How

- Added `mergeToolkits` and `MergedToolkit` in `@assistant-ui/core/react`.
- Re-exported `mergeToolkits` through web, React Native, and Ink packages.
- Re-exported a core `useAui` wrapper from the public distributions that normalizes a plain toolkit into `Tools({ toolkit })`.
- Registered toolkit model context declaratively so `useAui({ tools: appToolkit })` exposes both model tools and renderer UIs.
- Documented the `app/toolkits/files.tsx`, `tasks.tsx`, `calendar.tsx`, `crm.tsx`, `approvals.tsx`, `index.tsx` pattern.
- Added tests for merging, duplicate warnings, first-definition behavior, type preservation, client `useAui` registration, and server `AISDKToolkit` usage.

## Real-app verification

I verified this in `examples/with-ai-sdk-v6` with temporary `app/toolkits/*` files, a test page, and a route handler, then removed the temporary files before committing. The real app confirmed:

- Server route saw the composed toolkit tools: `create_task`, `lookup_shared`, `request_approval`, `schedule_meeting`, `search_crm`, `search_files`.
- Browser runtime saw the same tools through both `modelContext.toolNames` and `tools.toolUIs`.
- Duplicate `lookup_shared` emitted the expected `[assistant-ui] Duplicate tool name ... Keeping the first definition.` warning.
- No browser error logs were emitted.

## Validation

- `git diff --check`
- `pnpm --filter @assistant-ui/core exec vitest run src/react/model-context/merge-toolkits.test.ts`
- `pnpm --filter @assistant-ui/react exec vitest run src/tests/useAui-toolkit.test.tsx`
- `pnpm --filter @assistant-ui/react-ai-sdk exec vitest run src/generativeTools.test.ts`
- `pnpm --filter @assistant-ui/react-native build`
- `pnpm --filter @assistant-ui/react-ink build`
- `pnpm --filter @assistant-ui/docs exec fumadocs-mdx`
- `pnpm --filter @assistant-ui/docs build`
- `gh pr checks 4570` shows all checks passing or skipped, including changed-package build/test, Oxlint + Oxfmt, Semver Check, Template Sync, Vercel, and code review checks.

Note: local commands emit the existing Node engine warning because this machine is on Node v23.11.0 and the repo requests Node >=24; the commands above still passed.

### DIFF
--- a/.changeset/merge-toolkits-dx.md
+++ b/.changeset/merge-toolkits-dx.md
@@ -1,0 +1,8 @@
+---
+"@assistant-ui/core": patch
+"@assistant-ui/react": patch
+"@assistant-ui/react-native": patch
+"@assistant-ui/react-ink": patch
+---
+
+feat: add mergeToolkits and toolkit shorthand registration
--- a/apps/docs/content/docs/tools/defining-tools.mdx
+++ b/apps/docs/content/docs/tools/defining-tools.mdx
@@ -71,11 +71,11 @@ For a bare React Native app, import `getDefaultConfig` from `@react-native/metro
 </Step>
 <Step>
 
-### Write the toolkit
+### Write a feature toolkit
 
 The file's first line is `"use generative"`, and its default export is `defineToolkit({ ... })`. Each tool is an inline object literal with a `parameters` schema, an `execute`, and a `render` (or `renderText`):
 
-```tsx title="app/toolkit.tsx"
+```tsx title="app/toolkits/files.tsx"
 "use generative";
 
 import { defineToolkit } from "@assistant-ui/react";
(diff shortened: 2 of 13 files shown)


okisdev:
closing as a parallel mechanism to the endorsed composition idiom. toolkit composition already has a canonical answer: spread the default imports into a parent toolkit (defining-tools.mdx:408), and the entire compiler diagnostics series (#4583, #4632, #4617, #4629) exists to statically analyze exactly that shape, including the client/server boundary split. `mergeToolkits` appears nowhere in x-generative-compiler and this PR touches no compiler files, so a composed toolkit built through it is compiler-opaque, which is the very opaque-factory-call class #4617 diagnoses; the folder-DX example in this PR's own description cannot compile through the generative path. the one genuinely novel piece is duplicate-tool-name detection, since a spread is silently right-biased; that belongs in the compiler as a build-time collision diagnostic where the tool-name sets are already statically known, tracked in a follow-up issue. the `useAui({ tools: toolkit })` sugar is a flagship-hook contract change that needs its own design sign-off if wanted, separate from any of this.

PR #4574, opened 2026-06-24
feat(mcp-docs-server): adopt registerTool API with read-only annotations

ShobhitPatra:
## What

Migrate `mcp-docs-server` from the deprecated `server.tool()` registration API to `server.registerTool()`, and advertise both tools as read-only.

## Why

`@modelcontextprotocol/sdk` marks `server.tool()` `@deprecated` in favor of `registerTool()`. The newer API lets the server attach a human-readable `title` and tool `annotations`. Both `assistantUIDocs` and `assistantUIExamples` are pure reads of a baked, local documentation set, so they now advertise `readOnlyHint: true` and `openWorldHint: false`, which are accurate hints MCP clients can surface.

## How

- `src/index.ts`: replace the two `server.tool(...)` calls with `server.registerTool(name, { title, description, inputSchema, annotations }, execute)`. The same input schema and execute callback are reused, so the tool names, input schemas, and response shapes are unchanged. This is additive metadata only, not a contract change.
- Add assertions to the existing `tools/list` protocol test confirming the new `title`/`readOnlyHint` are surfaced.
- No changes to the tool modules or their `zod/v3` schemas (left for a separate PR).

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/4574?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/mcp-docs-server-register-tool.md
+++ b/.changeset/mcp-docs-server-register-tool.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/mcp-docs-server": patch
+---
+
+feat: adopt registerTool API and advertise read-only tool annotations
--- a/packages/mcp-docs-server/src/index.ts
+++ b/packages/mcp-docs-server/src/index.ts
@@ -17,16 +17,24 @@ export const server = new McpServer({
   version: packageJson.version,
 });
 
-server.tool(
+server.registerTool(
   docsTools.name,
-  docsTools.description,
-  docsTools.parameters,
+  {
+    title: "assistant-ui Documentation",
+    description: docsTools.description,
+    inputSchema: docsTools.parameters,
+    annotations: { readOnlyHint: true, openWorldHint: false },
+  },
   docsTools.execute,
 );
-server.tool(
+server.registerTool(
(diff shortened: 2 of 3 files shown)


okisdev (approved):
LGTM. The registerTool migration is correct (verified against the SDK 1.29.0 surface: title is top-level, annotations carries only the hint flags, inputSchema stays a ZodRawShape so JSON-schema conversion is unchanged). Follow-up test refinement on this branch tightens the tools/list assertions: asserts tool.title directly instead of the dead annotations?.title ?? title fallback, and adds the openWorldHint: false assertion the registration advertises. Necessity holds (real @deprecated), no contract change for in-repo consumers (none read title/annotations). One out-of-scope follow-up noted: examples/with-mcp/server still uses the deprecated server.tool() form.

PR #4576, opened 2026-06-24
feat(core): add flat $type generative-ui node with reserved $ namespace

okisdev:
## what

adds the flat `$type` shape generative-ui node to `@assistant-ui/core`, with a reserved `$` namespace for framework keys. first step of the generative-ui IR work.

## why

core's `generative-ui` message part only had the legacy `component` shape (`{ component, props, children }`). we want a flat node shape that is natural for a model to emit, and a reserved-key discipline that keeps the component prop namespace fully free. the `$` prefix convention (`$type`, `$key`, `$action`, `$status`) is assistant-ui's own: one rule covers every framework key, present and future, instead of ad-hoc reserving plain names. a component's own `type` prop (e.g. a button's `type="submit"`) flows through untouched.

## how

- `GenerativeUITypeNode`: flat node with `$type` selector, `$key` list identity, `$action` behavior as data, `children` nesting; every other key is an inline component prop.
- `GenerativeUIAction` plus shared ui token types (`GenerativeUITextSize`, `GenerativeUIColor`, `GenerativeUIWeight`, `GenerativeUIAlign`, `GenerativeUIJustify`, `GenerativeUIButtonStyle`, `GenerativeUIImageSize`).
- `normalizeUINode` / `normalizeSpec`: map both the flat `$type` shape and the legacy `component` shape to one canonical `NormalizedUINode`, stripping the reserved keys from the component prop bag.
- `MessagePrimitive.GenerativeUI` now renders either shape.
- the legacy `component` shape stays as a backward compatible alias; nothing existing is removed.

## not in this pr

- `@assistant-ui/react-generative-ui` already emits `$type`, so it aligns with this standard with no change. folding its parallel `GenerativeUINode` type onto core's shared type is a follow-up with no wire format change.
- action dispatch (the `ActionRegistry` that resolves `$action.type`) and the closed vocabulary plus slack/teams converters are separate later steps. `$action` is carried by the type and normalized, but not yet dispatched.
- the design doc is intentionally kept out of this commit.

## verification

- core tsc clean; 10 new tests pass; the legacy `component` shape `generative-ui.primitive.test.tsx` (4 tests) still passes.
- oxlint and oxfmt clean.

draft while the vocabulary contract and follow-up steps get signed off.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/4576?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


okisdev:
closing this in favor of re-scoping the flat `$type` format into `@assistant-ui/react-generative-ui`. the format from this pr is adopted, but it lands in the wrong home (core). the generative-ui api in `react-generative-ui` gets overridden with the #4576 shape instead, and core keeps only the legacy `generative-ui` part as a backward-compatible carrier. the `present` tool-call unification (a2a/a2ui decoded into a synthetic `present` tool-call part) removes the need for a new core message-part type. the ir design doc and the follow-up implementation land in `react-generative-ui`.

PR #4589, opened 2026-06-24
feat(react-langgraph): expose useLangGraphSetMessages for history pagination

gkkirsch:
## Motivation

`useLangGraphRuntime` owns its LangChain message list internally and surfaces `send`, `interrupt`, `messageMetadata`, and `uiMessages` via hooks — but there's no way for a consumer to **modify** the message list. That blocks two common needs:

- **Windowed history loading** — seed a thread with only the last N messages (instead of the full checkpoint, which scales to seconds on long threads) and prepend older pages on scroll.
- **Optimistic edits** to the rendered list.

## Change

- Add `useLangGraphSetMessages()` (mirrors `useLangGraphSend`) that surfaces the runtime's existing internal `setMessages`.
- Let `setMessages` accept a **functional updater** so consumers can prepend without separately reading state:
  ```ts
  const setMessages = useLangGraphSetMessages();
  setMessages((prev) => [...olderPage, ...prev]);
  ```
  This is additive and backward-compatible (the array form still works).
- Expose it on `LangGraphRuntimeExtras` and provide it in the extras block.

Later stream events continue to merge onto whatever is set, keyed by message id.

## Tests

Added coverage in `useLangGraphMessages.test.ts` for both the array and functional-updater forms. Full package suite passes (153/153); build, oxlint, and oxfmt clean. Changeset included (minor).

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/4589?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


gkkirsch:
Closing — we ended up not needing this. We wanted setMessages to drive history pagination (prepend older pages), but assistant-ui's external store only grows the rendered list at the tail, so prepended messages land on a disconnected branch and don't render. We solved our windowing differently (remount + re-seed). Happy to reopen if a setMessages accessor is independently useful.

PR #4614, opened 2026-06-25
test(react-ai-sdk): cover toolOutputConversion and sliceMessagesUntil

naamanu:
What
----
Adds colocated unit tests for two previously-untested pure helpers in `@assistant-ui/react-ai-sdk`:
- `toolOutputConversion.ts` — `toAISDKContent` and `toAISDKDefaultOutput`
- `ui/utils/sliceMessagesUntil.ts` — `sliceMessagesUntil`

Why
---
The repo convention is to cover converters/pure helpers with colocated tests in both directions, but these two had none. That left the content-mapping branches (text / image-data / file-data, optional filename, part order) and the slice edge cases (null id, not-found throw, trailing-assistant extension, last-message) unverified, so a regression in either would pass CI silently.

How
---
Pure input/output assertions following the existing test style in this package (colocated `*.test.ts`, relative imports, no snapshots). No production source or public API is touched.

This is test-only, so there is intentionally no changeset — the published artifact is unchanged and a version bump would be misleading (addresses the greptile review note).

Verified locally with `pnpm turbo test --filter=@assistant-ui/react-ai-sdk` (11 files / 106 tests pass) and `tsc --noEmit`.


okisdev (approved):
Verified against the source and ran the full suite in a throwaway worktree: 11 files / 106 tests pass (9 new in toolOutputConversion.test.ts, 5 new in sliceMessagesUntil.test.ts), tsc --noEmit clean. Both helpers were previously untested on main, and the new tests match the sibling conventions in usage.test.ts, frontendTools.test.ts, and toCreateMessage.test.ts (colocated *.test.ts, relative imports, no snapshots, msg()/baseMessage fixture builders). Test-only, no public API or changeset, which is correct here. Approving.

PR #4630, opened 2026-06-26
feat: add header brand assets menu

Kinfe123:
## Summary

Adds a Brand Assets context menu to the assistant-ui logo in the docs headers. Right-clicking the logo now lets users copy the logomark SVG, copy the logotype SVG, open the brand guidelines, or download the brand asset kit.

The shared `HeaderBrandLink` is used by the marketing header, docs header, and sub-project layout so the logo behavior stays consistent across the docs surfaces.

## Screenshot
image

## Validation

- `pnpm exec oxlint apps/docs/components/shared/header-brand-link.tsx apps/docs/components/shared/header.tsx apps/docs/components/docs/layout/docs-header.tsx apps/docs/components/shared/sub-project-layout.tsx packages/ui/src/components/ui/context-menu.tsx`
- `pnpm --filter @assistant-ui/docs exec tsc --noEmit --pretty false`

Note: both commands emit the existing Node engine warning because this machine is on Node 23.11.0 while the repo asks for Node >=24.


okisdev (approved):
## review

lgtm; the shipped version now reuses the kit component instead of forking its styling, and the copy action is activation-safe.

- the menu hand-rolled radix-ui primitives and carried verbatim copies of the kit context-menu's content and item classnames, re-implementing a component the docs app can already import via `@/components/ui/context-menu` (the base to radix fallback alias); the copies would drift silently from the kit styling. the retreat from the earlier canonical-kit edit overshot from "don't modify the kit component" to "don't use it at all".
- `copySvg` awaited the fetch before `navigator.clipboard.writeText`, so the write no longer happened inside the user activation window; Safari rejects that, and Chrome only tolerates it while the document stays focused.
- importing the kit component surfaced a latent registry bug (predates this PR): `ContextMenuCheckboxItem` destructured the optional `checked` and re-passed it explicitly, which fails under `exactOptionalPropertyTypes` for any consumer compiling the registry source from a strict tsconfig; the docs app's first import of the file tripped it.

### what we modified

- merges main so the branch sits on the current base/radix kit split, then rebuilds the menu on `@/components/ui/context-menu` (`ContextMenu`/`ContextMenuTrigger`/`ContextMenuContent`/`ContextMenuItem`/`ContextMenuSeparator`), deleting the hand-rolled primitives and both copied classname constants; visuals are unchanged since those classes came from the kit in the first place, and the menu picks up a future base-flavor context-menu automatically through the alias.
- rewrites `copySvg` to construct a `ClipboardItem` synchronously with the pending fetch as its value, so the clipboard write starts inside the user activation window; a comment pins the constraint so it does not get simplified back to fetch-then-write.
- fixes the kit's `ContextMenuCheckboxItem` by letting `checked` ride through `...props` instead of destructure-and-re-pass (behavior identical, and the registry source now compiles under `exactOptionalPropertyTypes`).

### potential further fixes

- the same destructure-and-re-pass pattern remains in `ui/base/dropdown-menu.tsx`, `ui/radix/dropdown-menu.tsx`, and `ui/radix/menubar.tsx`; none is in an in-repo typecheck graph today, but strict registry consumers hit it on `shadcn add`, so a small sweep PR is worth it.
- the copied `logotype.svg` is fixed-black and pastes near-invisible on dark surfaces; deliberate for a canonical brand asset, but the `/brand` page could say so.
- the kit still has no base-flavor context-menu, so this menu resolves to the radix flavor via the fallback alias; adding `ui/base/context-menu.tsx` later flips it to base with zero docs changes.

PR #4637, opened 2026-06-26
chore(style): add view transition theme toggle

Kinfe123:
## Summary

Adds circle-blur top-left theme transition to the existing docs theme toggle.

## Details

- Wraps the existing `next-themes` toggle with `document.startViewTransition` when the browser supports it.
- Adds the circle-blur top-left view-transition CSS in the docs global stylesheet.
- Adds a matching fallback ripple so browsers without View Transitions, including the in-app preview browser, still show a visible top-left blur effect.

## Validation

- `pnpm exec oxfmt --check apps/docs/components/shared/theme-toggle.tsx apps/docs/styles/globals.css`
- `pnpm exec oxlint apps/docs/components/shared/theme-toggle.tsx`
- `pnpm --filter docs exec tsc --noEmit`
- Ran the docs dev server and previewed `/docs` locally.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/4637?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/docs/components/shared/theme-toggle.tsx
+++ b/apps/docs/components/shared/theme-toggle.tsx
@@ -4,6 +4,12 @@ import { useTheme } from "next-themes";
 import { useEffect, useState } from "react";
 import { Moon, Sun } from "lucide-react";
 
+type DocumentWithViewTransition = Document & {
+  startViewTransition?: (updateCallback: () => void) => void;
+};
+
+type Theme = "dark" | "light";
+
 export function ThemeToggle() {
   const { resolvedTheme, setTheme } = useTheme();
   const [mounted, setMounted] = useState(false);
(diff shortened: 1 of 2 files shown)


okisdev:
closing: the reveal origin is hardcoded to the viewport top-left corner (svg mask circle at 0,0 plus `transform-origin: top left`), while the toggle actually renders on the right side of the docs and demo headers and at the bottom of the footer and sidebar layouts, so the circle expands from the opposite corner of the interaction and reads as a wipe glitch rather than a radiating transition. the canonical form of this pattern derives its origin from the click point.

the fallback ripple half of the diff also concentrates most of the open review findings (background literals that drift from the palette, a maximum z-index overlay above dialogs, timer and cleanup machinery on a deprecated ref type) to give browsers without view transitions a decorative degradation of a decorative effect; an instant switch is the platform norm there. two of the remaining threads (the animation shorthand timing reset and the redundant `.dark` arm) are already fixed at the current head but were never resolved.

the view transition half is competently built (the `linear()` expo-out curve and the blurred mask edge are the right tools). the bar for a future version: radiate from the click position via css variables, no fallback animation, and a duration closer to half a second, verified in a real browser against the actual toggle placements.

PR #4650, opened 2026-06-27
fix(react-ai-sdk): make default attachment adapter work without FileReader via shared core helper

ShobhitPatra:
Supersedes #4649. That PR fixed the same `FileReader` bug by patching react-ai-sdk's own copy of the encoder; this one instead reuses core's existing helper and deletes the duplicate, so it fixes the bug and removes the duplication in one change.

## What

`vercelAttachmentAdapter` built its data URL with `new FileReader()`, a browser-only API that is undefined in Node / react-ink / SSR, so sending an attachment there threw `ReferenceError`. It's the default attachments adapter wired unconditionally in `useAISDKRuntime` (including the React Native entry), so the browser dependency leaked into non-browser runtimes.

## How

`@assistant-ui/core` already solved this in `adapters/attachment.ts` (`getFileDataURL`, used by `SimpleImageAttachmentAdapter`), so rather than fix a second copy this exports that helper from `@assistant-ui/core/internal` and has `vercelAttachmentAdapter` consume it.

The shared fallback is hardened so every adapter benefits: it prefers `Buffer` when present and otherwise base64-encodes in `0x8000`-byte windows through `btoa` (a universal global). That fixes core's own gaps, its old fallback assumed `Buffer` exists and threw on bufferless edge runtimes (Workers, Vercel Edge), and an empty MIME type now defaults to `application/octet-stream`. The `FileReader` path stays primary, so browser and React Native stay byte-identical.

## Tests

`core/src/adapters/attachment.test.ts` covers each tier (FileReader, Buffer, chunked `btoa`), the octet-stream default, a 100k-byte regression for the chunked path, and a `SimpleImageAttachmentAdapter.send` case. react-ai-sdk gets a delegation test asserting `send` returns the correct data URL through the shared encoder.

`patch` changeset for both `@assistant-ui/core` and `@assistant-ui/react-ai-sdk`. The empty-MIME default is a small, deliberate behavior change to core.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/4650?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/share-attachment-data-url.md
+++ b/.changeset/share-attachment-data-url.md
@@ -1,0 +1,6 @@
+---
+"@assistant-ui/core": patch
+"@assistant-ui/react-ai-sdk": patch
+---
+
+fix: make the default attachment adapter work without FileReader (Node, react-ink, SSR) by sharing a single getFileDataURL from @assistant-ui/core/internal, whose base64 fallback chunks large inputs and works on runtimes without Buffer
--- a/api-surface/assistant-ui__core.ts
+++ b/api-surface/assistant-ui__core.ts
@@ -733,6 +733,46 @@ type AppendMessage = Omit<ThreadMessage, "id"> & {
   steer?: boolean | undefined;
 };
 
+type AttachmentAdapter = {
+  accept: string;
+  add(state: {
+    file: File;
+  }): Promise<PendingAttachment> | AsyncGenerator<PendingAttachment, void>;
+  remove(attachment: Attachment): Promise<void>;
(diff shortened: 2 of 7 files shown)


Yonom (approved):

PR #4658, opened 2026-06-29
feat(react-ink): add standalone TextInput primitive

ShobhitPatra:
### What

Adds a store-free, controlled `TextInput` primitive to `@assistant-ui/react-ink` and refactors `ComposerInput` to wrap it. `TextInput` owns the editing engine (text buffer, grapheme-aware cursor, eager reducer, full emacs/arrow keymap) behind a plain `value` / `onChange` / `onSubmit` contract, with no runtime or store coupling.

## Why

Ink ships no native text input, so the only editable field today is `ComposerInput`, which hard-wires that editing logic to the composer store (`useAuiState(s => s.composer.text)`, `aui.composer().setText`, `aui.composer().send()`, the `isRunning && !capabilities.queue` guard). Anything needing an editable field outside the thread composer (e.g. editing a value before confirming an action) cannot reuse it.

This is the terminal counterpart to what the other distributions get for free: web's `ComposerInput` wraps the DOM `<textarea>`, React Native's wraps the platform `<TextInput>`, only Ink builds the control by hand. Extracting it gives react-ink that platform-level building block while leaving `ComposerPrimitive.Input` (the primitive consistent across react / react-native / react-ink) unchanged. It sits in the same lane as the existing react-ink-only `DiffView` and `StatusBarPrimitive`.

## How

`TextInput` is the old `ComposerInput` body with store wiring swapped out: `value` feeds `useTextBuffer`, `onChange` replaces `aui.composer().setText`, and `onSubmit` is called directly instead of the composer-send fallback. The echo-dedup map (absorbs the echo when an owner sets `value` to text the buffer just emitted) and its `PENDING_SYNC_CAP` guard move with it, keyed on `value`. `useTextBuffer` is reused as-is; its surface is untouched.

`ComposerInput` keeps its public props and behavior byte-identical, rendering `<TextInput value={storeText} onChange={setText} onSubmit={submit} ... />`. `submit` is the original function unchanged, including caller-supplied `onSubmit` taking precedence over the default `send()` path.

The export is standalone (`export { TextInput, type TextInputProps }`), matching `DiffView` rather than a single-component `*Primitive` namespace.

## Tests

New `src/tests/TextInput.test.tsx` (6 cases): typing fires `onChange` and renders; `submitOnEnter` submits current text; multi-line shift-Enter inserts a newline without submitting; backspace and `Ctrl+U` kill-to-start mutate correctly; an external `value` change reflects into the render with no echo back through `onChange`; placeholder shows only while empty.

The existing `src/tests/ComposerInput.test.tsx` (16 cases: send-on-enter, running/queue guard, `onSubmit` precedence, echo-dedup, full keymap) passes unchanged and guards the refactor.

## Docs

Adds a `## TextInput` section to `apps/docs/content/docs/ink/primitives.mdx` with a controlled-usage example and prop table; keyboard bindings link to the identical `ComposerPrimitive.Input` reference.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/4658?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Yonom (approved):

Kinfe123 (changes requested):
@ShobhitPatra Could single-line TextInput normalize pasted newlines when multiLine is false? The key guards cover Enter/Ctrl-J, but raw pasted input may still include `\n`. can you please check on that and resolve the merge conflict

Kinfe123 (approved):

PR #4660, opened 2026-06-29
feat(react-ink): track hunkIndex on parsed diff lines

ShobhitPatra:
### What

Adds an optional `hunkIndex` field to `ParsedLine` and populates it in both diff parsers. This is the data layer for stageable diffs (hunk include/exclude); it carries no UI and no behavior change to the read-only `DiffView`.

## Why

The diff layer flattens every change into a single `ParsedLine[]`, discarding hunk boundaries. Per-hunk granularity (navigating and toggling whole hunks) needs those boundaries recoverable from the parsed model. Splitting this out first keeps the data change small and independently reviewable ahead of the interaction layer (`StageableDiffView`) that consumes it.

## How

`hunkIndex` is optional, so existing consumers are unaffected.

`parsePatch` tags every line (including context) with a per-file, zero-based counter incremented per `file.chunks` entry. Boundaries come straight from parse-diff, so this path is exact.

`computeDiff` has no native hunks, so it derives them: each contiguous run of change lines (`add`/`del`) is one hunk, and any `normal` gap ends it. Trailing or between-hunk context adopts the preceding hunk; context before the first change stays unindexed (it belongs to no hunk, which avoids dimming the whole file head when the first hunk is later unstaged); a file with no changes gets no hunks. The grouping is deterministic with no context-window threshold, since `computeDiff` does not receive render config. Hunk count is `max(hunkIndex) + 1`.

## Tests

`parsePatch` assigns a per-file index that increments across chunks (multi-hunk patch yields the expected per-line indices). `computeDiff` groups context-separated change regions into distinct hunks, leaves leading context unindexed, and assigns no index when there are no changes. Existing `parsePatch`/`computeDiff` equality assertions are updated for the new field.

## Out of scope

`StageableDiffView`, the selection state and keymap, and the `materializeStaged` reconstruction util all land in follow-up PRs. No change to read-only `DiffView` or its props.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/4660?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Yonom (approved):

okisdev:
This is a clean, additive data-layer change and I recommend merging it as-is. The optional `hunkIndex?: number` field on `ParsedLine` is append-only, the api-surface snapshot is updated in the same PR, a patch changeset is present, and no behavior change reaches the read-only `DiffView` (no consumer of `hunkIndex` exists yet — `grep` confirms `StageableDiffView`/`materializeStaged` are not in this PR). Both producers are implemented correctly and match the description: `parsePatch` (diff-utils.ts:36-73) tags every line of each `file.chunks` entry with a per-file zero-based counter incremented per chunk, so boundaries come straight from parse-diff; `computeDiff` derives hunks via `assignComputedHunkIndices` (diff-utils.ts:124-138), where each contiguous add/del run is one hunk, a normal gap ends the run, trailing/between-hunk context adopts the preceding hunk, leading context stays unindexed (hunkIndex starts at -1 and is only assigned to normal lines once a hunk exists), and a no-change file gets no indices. I traced the multi-hunk `parsePatch` test (expects [0,0,0,0,1,1,1,1]), the context-separated `computeDiff` test, and the leading-context and no-change edges; all hold against the code.

The greptile P2 thread flags inter-hunk context ownership (computeDiff("a\nb\nc\n","A\nb\nC\n") gives shared `b` hunkIndex 0 while the next change gets 1). This is the deliberate design the PR body states outright — context adopts the preceding hunk — and for materialization it is harmless since context is unchanged in both old and new. The only place it could matter is single-hunk rendering/dimming in the follow-up `StageableDiffView`, which is explicitly out of scope here. So it is not a defect in this data-only PR. The one worthwhile follow-up, also noted by the claude review, is that the two producers diverge on leading context (`parsePatch` tags it, `computeDiff` does not); a one-line doc on `hunkIndex` in types.ts stating it may be undefined for context outside any hunk and that producers differ would pin the contract for the consumer PR, but that is polish and does not block this one.

review threads

* refuted: packages/react-ink/src/primitives/diff/diff-utils.ts:131 (greptile-apps P2 thread, unresolved) by @greptile-apps: The factual indexing (shared context b gets hunkIndex 0, next change gets 1) is real, but the PR body states this 'context adopts the preceding hunk' behavior is deliberate, and it is harmless for materialization since context is unchanged in both old and new. The only consequence is single-hunk rendering/dimming in the out-of-scope StageableDiffView consumer, which does not exist yet. Not a defect in this data-only PR; non-blocking.
* confirmed: @claude[bot] review observation on producer divergence (2026-06-29) by @claude[bot]: The observation that parsePatch tags leading context while computeDiff leaves it undefined is accurate and intentional (inherent to the data sources). The suggested one-line doc on hunkIndex is the polish finding above; non-blocking.
* confirmed: @cubic-dev-ai no issues found (2026-06-29) by @cubic-dev-ai: Cubic reported no issues across the 5 files; consistent with this review. Advisory bot, non-blocking.
* confirmed: @Yonom APPROVED (api-surface) by @Yonom: Maintainer approved the api-surface changes; the hunkIndex addition to ParsedLine in api-surface/assistant-ui__react-ink.ts:2529 is consistent with types.ts.

_Review by [Rupic](https://github.com/assistant-ui/rupic). Verified against the diff; runtime behavior was not executed._

PR #4663, opened 2026-06-29
feat(react-ink): add stageable DiffView (hunk include/exclude)

ShobhitPatra:
## What

Adds `StageableDiffView`: an interactive `DiffView` variant for including/excluding individual hunks before approving a change, reporting the staged subset via `onStageChange`. Read-only `DiffView` is unchanged.

## Why

- "Approve this diff, but drop that hunk" is the richest terminal edit interaction, and no primitive offered it.
- First selection state and first keyboard interaction in the diff layer. Builds on the `hunkIndex` boundaries from #4660 (merged).

## How

- State is a pure reducer (`diff-stage.ts`): a `Set` of staged `"{fileIndex}:{hunkIndex}"` keys plus a focused-hunk cursor; actions `toggle` / `stage-all` / `stage-none` / `move-focus`; `toStagedSelection` projection.
- Kept as an internal `useReducer`, not an exported provider. Public surface is just `StageableDiffView` + `StageableDiffViewProps` + `StagedSelection`.
- Renders via the public `DiffRoot` / `DiffContent` with a custom `renderLine`: `[x]` / `[ ]` gutter marker on each hunk's first change line, dims unstaged hunks, highlights the focused hunk.
- Keymap mirrors the `Pressable` `isActive` convention (gated on `isFocused && isActive`): `j`/`k` + arrows move focus, `space` toggles, `a`/`n` all/none.
- `onStageChange` fires synchronously via an eager reducer run (same pattern as `ComposerInput`/`TextInput`), so it does not depend on effect-flush timing.
- Moves the private `getDiffViewFiles` to `diff-utils` as shared `getDiffFiles` (pure relocation, no behavior change); `DiffView` now imports it.

## Tests

- `diff-stage.test.ts`: default all-staged, toggle out/in, stage all/none, focus clamping, no-op move with zero hunks, sorted `toStagedSelection`.
- `StageableDiffView.test.tsx` (`ink-testing-library`): default markers, `space` toggles + emits subset, `j` then `space` toggles hunk 2, `n`/`a` payloads, `isActive={false}` registers no handler.

## Out of scope

- `materializeStaged` (patch reconstruction): follow-up, once agent-inbox needs it.
- In-hunk / intra-line staging, hunk splitting, separate file-level toggle (whole-hunk only).
- v1 limitation: staged rows use plain green/red/dim styling, not the intra-line word highlighting `DiffView` adds (leaves `DiffView`'s renderer untouched; shareable later).
- Staged state does not reset if the `patch` prop is swapped on a mounted component; remount with `key` for a new diff (acceptable for approval flows, follow-up if needed).

### DIFF
--- a/.changeset/stageable-diff-view.md
+++ b/.changeset/stageable-diff-view.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react-ink": patch
+---
+
+feat: add stageable DiffView (hunk include/exclude)
--- a/api-surface/assistant-ui__react-ink.ts
+++ b/api-surface/assistant-ui__react-ink.ts
@@ -2973,6 +2973,20 @@ type SpeechSynthesisAdapter = {
   speak: (text: string) => SpeechSynthesisAdapter.Utterance;
 };
 
+declare const StageableDiffView: (_param48: StageableDiffViewProps) => import("react").JSX.Element;
+
+type StageableDiffViewProps = DiffViewProps & {
+  onStageChange?: ((staged: StagedSelection) => void) | undefined;
+  isActive?: boolean | undefined;
+};
+
+type StagedSelection = {
+  stagedHunks: ReadonlyArray<{
+    fileIndex: number;
+    hunkIndex: number;
+  }>;
+};
+
 type StartRunConfig = {
   parentId: string | null;
   sourceId: string | null;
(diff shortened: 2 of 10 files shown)


ShobhitPatra:
closing this for now. on reflection this is a feature (new interactive primitive, first selection + keyboard state in the diff layer), not a small fix, so it should start with a short issue to align on direction before the code, per how the team works. the specific open question is whether interactive hunk-staging belongs in the library or is app-owned policy, same call that self-closed the earlier HITL primitives

branch is preserved and i'll revive this from an aligned issue when i pick it back up. thanks!

PR #4666, opened 2026-06-29
Add thread list item fork API seam

samdickson22:
## Reviewer Note: Diff Size

This PR looks larger than the implementation because the fork seam changes published API surface across packages that re-export core. Current diff breakdown: one focused fork reconciliation regression test is about 342 lines; generated `api-surface/*` snapshots are about 253 lines across 14 files; the remaining source diff is about 185 lines of core seam plumbing, root exports, and the changeset. The file list is still scoped to the fork API seam, required public-surface snapshots, and regression coverage.

## Summary

Adds the minimal provider-backed thread fork seam:

- `threadListItem().fork({ fromMessageId })` returns the new thread ID
- `RemoteThreadListAdapter.fork` delegates fork creation and message-history copying to the provider
- `forkedFrom` exposes durable source thread/message lineage on thread-list item state
- adapter `list()` and `fetch()` can hydrate persisted lineage after reload

This intentionally does not add a stock fork button, a thread-level capability flag, or generic repository/export-import cloning. Applications decide where to expose the action, and each remote provider owns its native fork semantics.

The OpenCode implementation is split into follow-up PR #4914.

## Why retain lineage

Projects need to distinguish an independently continued fork from an unrelated thread. `forkedFrom` preserves that relationship without making core responsible for copying provider state.

## Verification

- Core focused tests: 3 files, 27 tests passed
- Full package API-surface build: 40 packages passed
- API surface check and TypeScript check passed
- Oxlint and oxfmt passed for changed source files
- Changeset validation passed
- `git diff --check` passed


packages/react-opencode/src/types.ts
```diff
@@ -219,7 +219,6 @@ export type OpenCodeRuntimeExtras = {
   state: OpenCodeThreadState;
   permissions: OpenCodeThreadState["interactions"]["permissions"]["pending"];
   questions: OpenCodeThreadState["interactions"]["questions"]["pending"];
-  fork: (messageId: string) => Promise<string>;
   revert: (messageId: string) => Promise<void>;
```
samdickson22, line 223: **Breaking change to a published public surface.** This PR removes `fork` from `OpenCodeRuntimeExtras` (here), from `OpenCodeThreadControllerLike`/`OpenCodeThreadController` (below), and the `openCodeExtras.fork` helper. `@assistant-ui/react-opencode` is a published, non-private package and all three are part of its exported API surface (`useOpenCodeRuntimeExtras(): OpenCodeRuntimeExtras` is exported), so this is a breaking removal for any consumer calling `extras.fork(messageId)` today.

Per `AGENTS.md`: *"The public surface of any published package is append-only … never remove an export that has shipped … Ship a real behavior change as its own deliberate PR."*

Two concerns:
1. Consider **deprecating** rather than removing — keep `fork(messageId)` as a thin delegate to the new core `threadListItem().fork({ fromMessageId })` seam, marked `@deprecated`, so existing callers don't break.
2. If removal is intentional and maintainer-approved, it shouldn't ride along in the same PR as the additive seam, and the changeset should reflect that it's a breaking change rather than `patch` (minor/major needs maintainer approval per `AGENTS.md`).

---

packages/react-opencode/src/useOpenCodeRuntime.ts
```diff
@@ -155,6 +153,10 @@ const useOpenCodeThreadRuntime = (
 
   return useExternalStoreRuntime<ThreadMessage>({
     ...pickExternalStoreSharedOptions(options),
+    unstable_capabilities: {
+      ...options.unstable_capabilities,
+      fork: true,
```
samdickson22: `fork: true` is spread *after* `...options.unstable_capabilities`, so it unconditionally overrides any caller-supplied value. A consumer who explicitly passes `unstable_capabilities: { fork: false }` can no longer opt out. If the intent is "default on, caller can override", flip the order: `{ fork: true, ...options.unstable_capabilities }`. If fork is always available for OpenCode regardless of caller input, this is fine as-is — but worth a one-line note since it diverges from how `copy` is treated.

---

packages/react-opencode/src/OpenCodeThreadController.test.ts
```diff
@@ -78,7 +78,7 @@ describe("OpenCodeThreadController", () => {
         runConfig: {},
         createdAt: new Date(),
       },
-      { model: "claude" },
+      { model: { providerID: "anthropic", modelID: "claude" } },
```
samdickson22, line 81: This `{ model: "claude" }` → `{ model: { providerID: "anthropic", modelID: "claude" } }` change (and line 95) is unrelated to the fork API seam. Per `AGENTS.md` ("Keep unrelated changes out of the diff"), drive-by edits bury the real change. If it's needed because the test was already red on `main`, that's a separate fix worth its own PR/commit; if it's incidental, please drop it.

---

packages/core/src/runtime/interfaces/thread-runtime-core.ts
```diff
@@ -24,6 +24,7 @@ export type RuntimeCapabilities = {
   readonly edit: boolean;
   readonly reload: boolean;
   readonly delete: boolean;
+  readonly fork: boolean;
```
samdickson22: `fork` is added to `RuntimeCapabilities`, which is a *thread-level* capability, but the actual fork operation lives on `threadListItem().fork()` (thread-list level). That forces every `RuntimeCapabilities` literal across all runtimes to set `fork: false`, yet nothing in this PR appears to read `thread.capabilities.fork` to gate UI/behavior. Is this flag consumed anywhere? If it's purely for future UI gating, a short comment on intent would help; if there's no reader, the thread-level capability may be the wrong seam (the capability arguably belongs alongside the threadListItem fork support, e.g. derived from `RemoteThreadListAdapter.fork`/`onFork` presence).

---

samdickson22 (commented):
## Review: Add thread list item fork API seam

Nice, well-scoped seam. The core `threadListItem().fork({ fromMessageId })` plumbing is clean and consistent — the `forkedFrom` lineage metadata threads correctly through `RemoteThreadData` → `ThreadListItemCoreState` → `ThreadListItemState`, every `RuntimeCapabilities` literal is updated (no missed implementer / type error), the optimistic update in `RemoteThreadListThreadListRuntimeCore.fork` mirrors the existing `detach`/`switchToThread` patterns, unsupported runtimes reject with clear errors, and the OpenCode adapter's `fork` mirrors its `initialize` shape. Tests cover both directions (supported + unsupported) for the remote-thread-list core, the external-store core, the `ThreadListItemRuntime` surface, and the OpenCode adapter. 👍

A few things to address before this comes out of draft:

**1. Breaking removal of public API in `@assistant-ui/react-opencode` (main concern).** `OpenCodeRuntimeExtras.fork`, `OpenCodeThreadControllerLike.fork`, `OpenCodeThreadController.fork()`, and the `openCodeExtras.fork` helper are all removed. That package is published (non-private) and these are part of its exported surface, so this breaks any consumer calling `extras.fork(...)` today. `AGENTS.md` requires the public surface to stay append-only. Suggest deprecating + delegating to the new seam instead of removing, or splitting the removal into its own maintainer-approved PR with a non-`patch` changeset. (inline)

**2. `unstable_capabilities.fork: true` overrides caller input** in `useOpenCodeRuntime.ts` because it's spread last — a caller can't opt out. Flip the spread order if "default-on, overridable" is the intent. (inline)

**3. Unrelated change** — the `model: "claude"` → `model: { providerID, modelID }` edit in `OpenCodeThreadController.test.ts` is unrelated to forking and should be dropped or split out. (inline)

**4. Question on the capability seam** — `RuntimeCapabilities.fork` is thread-level, but fork is a thread-list-item operation. Is `thread.capabilities.fork` actually read anywhere, or is it placeholder for future UI gating? A note on intent (or moving it next to the threadListItem fork support) would help. (inline)

Nothing here is a correctness bug in the happy path — items 1–3 are the actionable ones, with #1 being the blocker. The verification list in the description is thorough; given the noted incomplete `pnpm api-surface` run, double-check the committed api-surface snapshots match before un-drafting.

---

okisdev:
closing. this adds a permanent public API seam across the client API, the RemoteThreadListAdapter contract, and every thread list item state shape (14 api-surface snapshots) without an issue or maintainer direction behind it. per AGENTS.md, a non-trivial feature starts with an issue so a maintainer can confirm the direction before code exists; that matters most for public surface, which is append-only forever. the flagship adapter (Assistant Cloud) also has no fork implementation, so the seam's only real consumer would be OpenCode. implementation quality is not the issue here; the reconciliation race coverage is solid. API decisions of this size start with a design conversation, not a PR.

samdickson22:
claude there was a discussion in person with a maintainer

okisdev:
closing again per the 2026-07-15 rationale, which stands unchanged: a public fork API seam across 14 packages needs maintainer direction on an issue before implementation, and the flagship cloud adapter has no way to implement the seam, which makes the contract unshippable as designed. reopening without new discussion does not change that assessment. if there is new information, an issue describing the fork semantics, and how the cloud thread list would implement them, is the place to make the case.

PR #4670, opened 2026-06-30
feat: add mcp server names to errors

Kinfe123:
## Summary
- Add MCP toolkit entry names to connection and `client.tools()` listing failures.
- Preserve the original failure on `error.cause` while showing a clearer message like `MCP toolkit entry "github" failed to list tools: ...`.
- Document that the `defineMcpToolkit` entry key is a readable debugging label for connection/tool-listing errors; it does not rename the MCP server's tools.

## Testing
- `pnpm exec oxlint packages/react-ai-sdk/src/generativeTools.ts packages/react-ai-sdk/src/generativeTools.test.ts`
- `pnpm --filter @assistant-ui/react-ai-sdk exec tsc --noEmit`
- `pnpm --filter @assistant-ui/react-ai-sdk test -- src/generativeTools.test.ts`
- `git diff --check`

## Follow-up
- #4671 extends this same MCP toolkit entry-name context to `AISDKToolkit.close()` errors. It is stacked on this PR and should be reviewed/merged after this one.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/4670?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


okisdev (approved):
## review

lgtm as `patch`. verified against current main before approving: with this branch's diff applied on top of the AI SDK v7 switch, the full react-ai-sdk suite passes (139/139), and the wrap sites anchor cleanly in the v7-era `generativeTools.ts`.

- the first cut assigned `error.cause` through a manual cast instead of the native `Error` constructor option, which the ES2023 lib already types.
- the branch predated the AI SDK v7 switch and the mcp tool name prefix work on main, so its green CI no longer spoke for the code a merge would actually produce.

### what we modified

- adopts `new Error(message, { cause })` in `toMcpToolkitError`, dropping the cast (addressed during the review round).
- merges main so the branch carries the v7-era `generativeTools.ts` and CI re-runs on the real combination.

### potential further fixes

- `generativeTools.test.ts` carries a pre-existing TS2769 on main (the `execute` call options no longer match the v7 overloads); it is invisible because `test:types` is not CI-gated, unrelated to this diff, and worth a small separate fix.
- #4671 extends the same entry-name context to `close()` errors and is stacked on this branch; once this merges it needs its own main merge and a stale-diff check before review.

PR #4677, opened 2026-06-30
Centralize assistant message construction

piyushzala158:
## Summary

Centralizes construction of full `ThreadAssistantMessage` objects behind a shared internal helper.

This keeps `role: "assistant"` visible where it is the message discriminator or external wire shape, while removing duplicated default object construction for core assistant messages.

## Changes

- Add `createThreadAssistantMessage` as an internal core runtime utility.
- Use the helper in `fromThreadMessageLike`, local runtime assistant run initialization, and external converter error-message creation.
- Add focused tests for helper defaults and metadata preservation.
- Extend `fromThreadMessageLike` coverage to assert assistant defaults still match the previous behavior.

## Why

Several core paths manually constructed assistant messages with the same defaults for `status`, `content`, `createdAt`, and metadata arrays. Keeping those defaults in one place makes future changes less error-prone without changing public APIs or provider message formats.

## Validation

- `NODE_OPTIONS=--max-old-space-size=8192 pnpm api-surface:check`
- `pnpm lint`

Note: `pnpm lint` completed successfully with pre-existing React hooks warnings in unrelated files.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/4677?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/core/src/react/runtimes/external-message-converter.ts
+++ b/packages/core/src/react/runtimes/external-message-converter.ts
@@ -11,6 +11,7 @@ import {
   fromThreadMessageLike,
   type ThreadMessageLike,
 } from "../../runtime/utils/thread-message-like";
+import { createThreadAssistantMessage } from "../../runtime/utils/create-thread-assistant-message";
 import { getAutoStatus, isAutoStatus } from "../../runtime/utils/auto-status";
 import type { ToolExecutionStatus } from "../../runtimes/tool-invocations/ToolInvocationTracker";
 import type { ReadonlyJSONValue } from "assistant-stream/utils";
@@ -322,20 +323,10 @@ const chunkExternalMessages = <T>(
 function createErrorAssistantMessage(
   error: ReadonlyJSONValue,
 ): ThreadAssistantMessage {
-  const msg: ThreadAssistantMessage = {
+  const msg = createThreadAssistantMessage({
     id: generateErrorMessageId(),
-    role: "assistant",
-    content: [],
     status: { type: "incomplete", reason: "error", error },
-    createdAt: new Date(),
-    metadata: {
-      unstable_state: null,
-      unstable_annotations: [],
-      unstable_data: [],
-      custom: {},
-      steps: [],
-    },
-  };
+  });
   bindExternalStoreMessage(msg, []);
   return msg;
 }
--- a/packages/core/src/runtime/utils/create-thread-assistant-message.test.ts
+++ b/packages/core/src/runtime/utils/create-thread-assistant-message.test.ts
@@ -1,0 +1,67 @@
+import { describe, expect, it } from "vitest";
+import { createThreadAssistantMessage } from "./create-thread-assistant-message";
+
+describe("createThreadAssistantMessage", () => {
+  it("fills assistant message defaults", () => {
+    const createdAt = new Date("2026-01-01T00:00:00.000Z");
+
+    expect(createThreadAssistantMessage({ id: "msg-1", createdAt })).toEqual({
+      id: "msg-1",
+      role: "assistant",
+      content: [],
+      status: { type: "running" },
+      createdAt,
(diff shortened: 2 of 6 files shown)


okisdev (changes requested):
direction is fine; this matches how we already centralize shared construction internally (createRuntimeExtras is the closest precedent). but the diff does not compile under the monorepo's strict typecheck and there are two smaller items, so requesting changes.

**1. the Build Changed Packages failure is caused by this diff**

[`with-elevenlabs-scribe#build`](https://github.com/assistant-ui/assistant-ui/actions/runs/28458364637/job/84339237962) compiles core from source and fails at the changed line in `thread-message-like.ts` (trimmed):

```
../../packages/core/src/runtime/utils/thread-message-like.ts:216:9
Type error: Type '{ readonly unstable_state?: ReadonlyJSONValue; ... }' is not assignable to type 'Partial<{ ... }>' with 'exactOptionalPropertyTypes: true'.
  Types of property 'unstable_annotations' are incompatible.
    Type 'readonly ReadonlyJSONValue[] | undefined' is not assignable to type 'readonly ReadonlyJSONValue[]'.
```

the props on `ThreadMessageLike["metadata"]` carry explicit `| undefined`, which is not assignable to `Partial<AssistantMetadata>` under `exactOptionalPropertyTypes`. type the option the same way you typed the scalar options:

```ts
readonly metadata?:
  | { readonly [K in keyof AssistantMetadata]?: AssistantMetadata[K] | undefined }
  | undefined;
```

`pnpm --filter @assistant-ui/core build` does not run a strict tsc pass, which is why your local validation passed; `pnpm --filter @assistant-ui/core exec tsc --noEmit` reproduces the failure and is the right gate before pushing. (the Vercel failure is most likely the same error surfacing through the docs build.)

**2. the helper changes omission semantics for falsy metadata values**

the inline code this replaces omitted `timing` and `submittedFeedback` on truthiness and coerced `isOptimistic` with a truthy check; the helper switched these to `!== undefined` and `=== true`. `fromThreadMessageLike` is a boundary where untyped payloads arrive from external stores, so `metadata: { timing: null }` previously produced a message without a `timing` key and now embeds `timing: null`, which violates the declared `ThreadMessage` shape downstream. keep the three conditional spreads semantically identical to the code they replace, and pin it with a test (a null `timing` input should yield no `timing` key). related: the new `fromThreadMessageLike` coverage asserts with `toMatchObject`, which tolerates added keys and would not catch this; `toEqual` on the full message would.

**3. missing changeset**

`@assistant-ui/core` is published, so this needs a patch changeset even though the change is internal:

```md
---
"@assistant-ui/core": patch
---

refactor: centralize assistant message construction
```

optional, not gating: the voice transcript path in `base-thread-runtime-core.ts` (`_handleVoiceTranscript`) constructs the same shape including `unstable_state: this.state` and is not covered; migrating it would make the centralization complete within core, but it also works as a follow-up.

for scope: I checked the local runtime and error message call sites against the previous inline construction; output is identical for typed consumers and the api surface is untouched, so the items above are the only things gating this.

okisdev:
closing. this extraction doesn't meet the bar we apply to centralization refactors: the three construction sites aren't identical-shape duplication, so the helper ends up an all-optional options bag (every field defaulted, per-field metadata fallbacks, conditional spreads) rather than a shared invariant, and each call site trades locally visible construction for indirection through it. we extract when copies are byte-identical with demonstrated drift cost, and keep divergent shapes local otherwise; there's also no issue or drift bug motivating the change.

PR #4678, opened 2026-06-30
fix(assistant-stream): use delta field for text-delta in UIMessageStreamDecoder

### LINKED ISSUE #4676: Vercel stream protocol mismatch
### Checklist

### Environment info

```
Environment:
  OS:               Windows 10.0.26200 (x64)
  Node.js:          v22.14.0
  Package Manager:  npm unknown
  Framework:        Vite 8.1.0

Packages:
  @assistant-ui/react              0.14.24
  @assistant-ui/core               0.2.19
  @assistant-ui/store              0.2.19
  @assistant-ui/tap                0.9.3
  assistant-stream                 0.3.24
  assistant-cloud                  0.1.34
  @assistant-ui/react-data-stream  0.12.18
  @assistant-ui/react-markdown     0.14.5

Ecosystem:
  react       19.2.7
  react-dom   19.2.7
  vite        8.1.0
  zod         4.4.3
  zustand     5.0.14
  typescript  6.0.3
```

### Affected area(s)

Streaming (assistant-stream)

### Runtime adapter in use

### Steps to reproduce

1. Set up a useDataStreamRuntime
2. Create a custom backend to stream back the following content using the SSE standard of course:
```
{"type":"start-step"} 
{"type":"start","messageId":"7190d813-dafc-4515-a8a7-acdd257e5ec2"} 
{"type":"reasoning-start","id":"4d6757a334c54b38bb292506606df7ed"}  
{"type":"reasoning-delta","id":"4d6757a334c54b38bb292506606df7ed","delta":"The"}  
{"type":"reasoning-delta","id":"4d6757a334c54b38bb292506606df7ed","delta":" user is"} 
{"type":"reasoning-delta","id":"4d6757a334c54b38bb292506606df7ed","delta":" just"}  
{"type":"reasoning-delta","id":"4d6757a334c54b38bb292506606df7ed","delta":" saying \u0022Hello"}  
{"type":"reasoning-delta","id":"4d6757a334c54b38bb292506606df7ed","delta":"\u0022 which is"}  
{"type":"reasoning-delta","id":"4d6757a334c54b38bb292506606df7ed","delta":" a greeting."} 
{"type":"reasoning-delta","id":"4d6757a334c54b38bb292506606df7ed","delta":" I"} 
{"type":"reasoning-delta","id":"4d6757a334c54b38bb292506606df7ed","delta":" should respond with"} 
{"type":"reasoning-delta","id":"4d6757a334c54b38bb292506606df7ed","delta":" a friendly greeting"} 
{"type":"reasoning-delta","id":"4d6757a334c54b38bb292506606df7ed","delta":" and"} 
{"type":"reasoning-delta","id":"4d6757a334c54b38bb292506606df7ed","delta":" offer to help"} 
{"type":"reasoning-delta","id":"4d6757a334c54b38bb292506606df7ed","delta":" them with something"} 
{"type":"reasoning-delta","id":"4d6757a334c54b38bb292506606df7ed","delta":". There\u0027s"} 
{"type":"reasoning-delta","id":"4d6757a334c54b38bb292506606df7ed","delta":" no need"} 
{"type":"reasoning-delta","id":"4d6757a334c54b38bb292506606df7ed","delta":" to use any"}  
{"type":"reasoning-delta","id":"4d6757a334c54b38bb292506606df7ed","delta":" tools for"} 
{"type":"reasoning-delta","id":"4d6757a334c54b38bb292506606df7ed","delta":" this simple greeting"}  
{"type":"reasoning-delta","id":"4d6757a334c54b38bb292506606df7ed","delta":".\n"}  
{"type":"reasoning-end","id":"4d6757a334c54b38bb292506606df7ed"}  
{"type":"text-start","id":"d2f9256a68594a52b481a4fc42cac0b8"} 
{"type":"text-delta","id":"d2f9256a68594a52b481a4fc42cac0b8","delta":"\n\nHello!"}  
{"type":"text-delta","id":"d2f9256a68594a52b481a4fc42cac0b8","delta":" How can I"}  
{"type":"text-delta","id":"d2f9256a68594a52b481a4fc42cac0b8","delta":" help you today"} 
{"type":"text-delta","id":"d2f9256a68594a52b481a4fc42cac0b8","delta":"?"} 
{"type":"text-end","id":"d2f9256a68594a52b481a4fc42cac0b8"} 
{"type":"finish-step"}  
{"type":"finish"} 
```

Expected: to see the reasoning AND the text
Actual: I see only the reasoning but not I got "undefinedundefinedundefinedundefinedundefinedundefined" instead of the text

### Browser (if applicable)

_No response_

### Code snippet or error log

```tsx

```

### Additional context

It seems to me that there is a mismatch between the Vercel stream protocol and the assistant-stream parser in [UIMessageStream](https://github.com/assistant-ui/assistant-ui/blob/main/packages/assistant-stream/src/core/serialization/ui-message-stream/UIMessageStream.ts) line 153 ```controller.appendText(chunk.textDelta);```
In the [vercel documentation](https://ai-sdk.dev/docs/ai-sdk-ui/stream-protocol#text-delta-part) we can see ```data: {"type":"text-delta","id":"msg_68679a454370819ca74c8eb3d04379630dd1afb72306ca5d","delta":"Hello"}```

So it should be `delta `and not `textDelta`

ShobhitPatra:
## Summary

The v6 UIMessageStream decoder read a non-existent `textDelta` field on `text-delta` chunks, while the wire protocol uses `delta`. Spec-compliant backends rendered `undefinedundefined...` for assistant text; reasoning was unaffected because `reasoning-delta` already read `delta`.

## Fix

Normalize the wire format at the SSE parse boundary: legacy `textDelta` payloads are rewritten to `delta` before entering the typed pipeline, so the v5 shape is canonical. `UIMessageStreamChunk` becomes `{ type: "text-delta"; id?: string; delta: string }` and the decoder reads only `chunk.delta`. There is no encoder in this module, so decode is the only boundary. api-surface regenerated.

## Verification

Decoder tests cover both wire shapes: the spec-compliant `delta` test fails against the old source (`expected undefined to be 'Hello'`) and the legacy `textDelta` test proves back-compat. Full assistant-stream suite and full-repo build green.

Closes #4676

## Note
scope note: the tool-call chunk mismatch reported in #4676's comments (`tool-input-start` / `inputTextDelta` vs the decoder's `tool-call-start` / `argsText`) is real but separate from the text fix here, will address in a follow-up pr

### DIFF
--- a/.changeset/fix-ui-message-stream-text-delta.md
+++ b/.changeset/fix-ui-message-stream-text-delta.md
@@ -2,4 +2,4 @@
 "assistant-stream": patch
 ---
 
-fix(assistant-stream): use `delta` field for text-delta in UIMessageStreamDecoder
+fix: accept the AI SDK v6 `delta` wire field for text-delta chunks in UIMessageStreamDecoder
--- a/.changeset/fix-ui-message-stream-text-delta.md
+++ b/.changeset/fix-ui-message-stream-text-delta.md
@@ -1,0 +1,5 @@
+---
+"assistant-stream": patch
+---
+
+fix(assistant-stream): use `delta` field for text-delta in UIMessageStreamDecoder
--- a/api-surface/assistant-stream.ts
+++ b/api-surface/assistant-stream.ts
@@ -11819,8 +11819,7 @@ type UIMessageStreamChunk = {
   id: string;
 } | {
   type: "text-delta";
-  id?: string;
-  delta: string;
+  textDelta: string;
 } | {
   type: "text-end";
 } | {
--- a/api-surface/assistant-stream.ts
+++ b/api-surface/assistant-stream.ts
@@ -11819,7 +11819,8 @@ type UIMessageStreamChunk = {
   id: string;
 } | {
   type: "text-delta";
-  textDelta: string;
+  id?: string;
+  delta: string;
 } | {
   type: "text-end";
 } | {
--- a/packages/assistant-stream/src/core/serialization/ui-message-stream/UIMessageStream.test.ts
+++ b/packages/assistant-stream/src/core/serialization/ui-message-stream/UIMessageStream.test.ts
@@ -38,6 +38,35 @@ describe("UIMessageStreamDecoder", () => {
     const events = [
       JSON.stringify({ type: "start", messageId: "msg_123" }),
       JSON.stringify({ type: "text-start", id: "text_1" }),
+      JSON.stringify({ type: "text-delta", id: "text_1", delta: "Hello" }),
+      JSON.stringify({ type: "text-delta", id: "text_1", delta: " world" }),
+      JSON.stringify({ type: "text-end" }),
+      JSON.stringify({
+        type: "finish",
+        finishReason: "stop",
+        usage: { inputTokens: 10, outputTokens: 5 },
+      }),
+      "[DONE]",
+    ];
+
+    const stream = createUIMessageStream(events);
+    const decodedStream = stream.pipeThrough(new UIMessageStreamDecoder());
+    const chunks = await collectChunks(decodedStream);
+
+    // Find text-delta chunks
+    const textDeltas = chunks.filter(
+      (c): c is AssistantStreamChunk & { type: "text-delta" } =>
+        c.type === "text-delta",
+    );
+    expect(textDeltas).toHaveLength(2);
+    expect(textDeltas[0]?.textDelta).toBe("Hello");
+    expect(textDeltas[1]?.textDelta).toBe(" world");
+  });
+
+  it("should decode legacy textDelta text deltas", async () => {
+    const events = [
+      JSON.stringify({ type: "start", messageId: "msg_123" }),
+      JSON.stringify({ type: "text-start", id: "text_1" }),
       JSON.stringify({ type: "text-delta", textDelta: "Hello" }),
       JSON.stringify({ type: "text-delta", textDelta: " world" }),
       JSON.stringify({ type: "text-end" }),
@@ -53,7 +82,6 @@ describe("UIMessageStreamDecoder", () => {
     const decodedStream = stream.pipeThrough(new UIMessageStreamDecoder());
     const chunks = await collectChunks(decodedStream);
 
-    // Find text-delta chunks
     const textDeltas = chunks.filter(
       (c): c is AssistantStreamChunk & { type: "text-delta" } =>
         c.type === "text-delta",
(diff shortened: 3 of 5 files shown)


Yonom:
The runtime fix is right, but please make the `UIMessageStreamChunk` type change backwards compatible instead of swapping the member shape: keep the existing `textDelta` field name supported alongside `delta`, and make the new `id` optional, e.g.

```ts
| { type: "text-delta"; id?: string; delta?: string; textDelta?: string }
```

with the decoder reading `chunk.delta ?? chunk.textDelta`. The public surface is append-only: backends that conformed to the previous decoder behavior by emitting `textDelta` should keep working on a patch release.

Yonom (changes requested):
Please normalize the wire format instead of widening the chunk type: translate to the right wire shape during encoding/decoding rather than assuming both formats are the same (normalize to the v5 format, `delta`, as the canonical one). As written, making `textDelta` optional is a breaking change to the published `UIMessageStreamChunk` type and requires a version bump of assistant-stream.

ShobhitPatra:
@Yonom  reworked: v5 delta is canonical, legacy textDelta normalized at the sse parse boundary, type no longer widened

okisdev (approved):
## review

LGTM as `patch`. the decode bug is real and externally reported (#4676): `textDelta` never matched any released AI SDK ui-message-stream wire (the field arrived in the v6 upgrade, and the original test encoded the same wrong shape, which is why CI stayed green). the fix now matches the review guidance exactly: the wire's `delta` is accepted at the decode boundary and normalized into the existing published chunk shape, so `UIMessageStreamChunk` keeps `textDelta` required, the api-surface diff is zero, and the patch changeset is legitimate. verified empirically in a throwaway worktree: with the decoder reverted to main, the spec-shape test fails with `expected undefined to be 'Hello'` while the legacy-shape test passes; the ui-message-stream suite passes 12/12 with the fix.

- the first two iterations changed the published type (widened to `{ id?; delta?; textDelta? }`, then renamed to `{ id?; delta }`); both break external consumers of `UIMessageStreamChunk` on a patch release, while the runtime fix itself was right each time.
- `Closes #4676` overpromises slightly: the issue thread also documents the tool-input chunk mismatch, which is tracked separately in #4692.

### what we modified

- reverts `chunk-types.ts` and the api-surface snapshot to main, flips the boundary shim to rewrite the wire's `delta` into the existing `textDelta` shape, and keeps the decoder switch reading `chunk.textDelta`; both wire-shape tests carry over unchanged and pass 12/12.
- rewords the changeset body; the bump stays patch because the published surface is untouched.

### potential further fixes

- the PR description still explains the earlier delta-canonical type approach; worth an edit so it matches the shipped shape.
- `Part of #4676` fits better than `Closes` until #4692 lands, or close the issue deliberately once both halves are in.

okisdev:
consolidating into #4854: both PRs normalize the same decode boundary and were guaranteed to conflict, and #4854 now carries the identical `delta` text-chunk fix in the shape from the review, alongside the `source-url`, flat `file`, and bare lifecycle normalizations. the published surface stays untouched throughout, so patch remains legitimate. re-review moved to #4854.

PR #4681, opened 2026-06-30
fix: ignore stale react-mcp connect results

Kinfe123:
## Summary
- make MCP server connect/auth/disconnect operations latest-action-wins
- ignore stale connect/listTools results so older MCP responses cannot overwrite newer UI state
- add react-mcp Vitest coverage and a package-level test script

## Use case
While testing MCP connections locally, a slow server can finish `connect()` or `listTools()` after the user has already retried, disconnected, or completed a newer connection attempt. Before this change, that stale response could flip the server back to `connected` or `error` and show tools from the wrong attempt. The resource now keeps the visible state aligned with the latest user action.

## Testing
- `pnpm --filter @assistant-ui/react-mcp test -- src/resources/McpServerResource.test.ts`
- `pnpm --filter @assistant-ui/react-mcp exec tsc --noEmit`
- `pnpm --filter @assistant-ui/react-mcp build`


okisdev:
closing as superseded by #5150, which landed the same invariants in a strictly stronger form: `connectionGenerationRef` with `isCurrentConnection` covers every guard point this PR adds (post `buildTransport`, post `connect`, post `listTools`, the auth url path) and additionally checks `mountedRef`, so unmount invalidates a pending attempt, which the attempt counter here does not. transport cleanup is serialized through `pendingTransportRef` and `closeQueuedTransports` rather than closed ad hoc, interrupted OAuth completion rejects explicitly, and StrictMode effect replay is handled. the test coverage and package test script this PR introduced were also delivered independently in the interim, and the conflict state reflects the same connect flow having been rewritten by both changes. this was the right diagnosis and fix shape in June; #5150 landed the fuller version of it.

PR #4707, opened 2026-07-03
chore(examples): thin hover-reveal scrollbars in global styles

ephraimduncan:
Ports the docs scrollbar treatment (#4331) to all web example and template stylesheets: 6px hover-revealed WebKit scrollbars plus a `scrollbar-width: thin` fallback for Firefox, using each app's `--muted-foreground` token. Examples previously shipped default OS scrollbars; this aligns them with assistant-ui.com. Terminal/native examples (with-expo, with-react-ink, with-react-ink-web) excluded. Two examples (with-store, with-heat-graph) use stock Next starter CSS without shadcn tokens, so their blocks substitute `currentColor` for `var(--muted-foreground)`.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/4707?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/examples/waterfall/app/globals.css
+++ b/examples/waterfall/app/globals.css
@@ -35,7 +35,8 @@ body {
 }
 
 @layer base {
-  @supports not selector(::-webkit-scrollbar) {
+  /* Firefox only: standard scrollbar-* props would disable ::-webkit-scrollbar styling in Chromium */
+  @supports (-moz-appearance: none) {
     * {
       scrollbar-width: thin;
       scrollbar-color: transparent transparent;
--- a/examples/waterfall/app/globals.css
+++ b/examples/waterfall/app/globals.css
@@ -33,3 +33,37 @@ body {
   color: var(--foreground);
   font-family: Arial, Helvetica, sans-serif;
 }
+
+@layer base {
+  @supports not selector(::-webkit-scrollbar) {
+    * {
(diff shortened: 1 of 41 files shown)


AVGVSTVS96 (changes requested):
i think we should leave this to users. we should keep templates more minimal and rely on supporting default OS scrollbar settings which this PR would override.

these changes are okay on our docs site because we own it, but templates are for users to create their apps so we shouldn't be opinionated about scrollbar styling there

ephraimduncan:
agreed, closing. templates and examples are user starting points so they should inherit whatever the OS and browser do by default rather than ship opinionated scrollbar styling. the docs site already has its own treatment in apps/docs/styles/globals.css. one thing worth salvaging separately: 8193b11 fixed the @supports not selector(::-webkit-scrollbar) gate that current Firefox no longer enters, and the docs globals still carry that broken gate, so a small docs-only follow-up would fix hover-reveal scrollbars on Firefox there.

PR #4728, opened 2026-07-05
feat(core): surface the upload failure reason on failed attachments

ShobhitPatra:
## What

A failed attachment upload shows a generic error tile with no hint of why it failed. The reason is caught by `CloudFileAttachmentAdapter` but thrown away; only the console sees it. This carries it onto the attachment status and into the UI. 

## How

- `PendingAttachmentStatus` incomplete arm gains `message?: string` (append-only)
- `CloudFileAttachmentAdapter` populates it from the caught error
- composer core forwards it as the `attachmentAddError` event message, generic string stays as fallback
- ui tile tooltip shows it under the file name, "Upload failed" fallback; minimal template synced
- api-surface regenerated: 14 added `message?: string;` lines, nothing removed

Status field instead of throwing because `addAttachment` is un-awaited (throw = unhandled rejection) and only the status persists for the UI, same reasoning as #4714. Plain string instead of `error?: ReadonlyJSONValue` because the consumer is UI text. The event `error` slot stays unset here since it means "the actual thrown object" everywhere else.

## Tests

- adapter: both failure paths assert the populated `status.message`
- composer core: new test for message forwarding (no synthesized `error`), existing test now pins the generic fallback

## Follow-ups 
- react-ink `AttachmentStatus` could append the message next to `x error`
- `CloudAPIError` drops `response.status`; optional `status` field on the SDK error

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/4728?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Kinfe123 (approved):

Kinfe123 (commented):
@ShobhitPatra  nice catch but If an adapter yields an attachment, then throws, the [catch](https://github.com/assistant-ui/assistant-ui/blob/74caf0216b1e969da609f51d8d734b8eb71ccb88/packages/core/src/runtime/base/base-composer-runtime-core.ts#L320-L325) path marks it failed without copying the thrown error message into `status.message`. So the UI still shows generic "Upload failed" instead of "network error". can you cehck on that as follow up pr or commit here.

ShobhitPatra:
@Kinfe123 fixed in 0be29ca, the catch path now copies the thrown message onto status.message so throwing adapters get the real  tooltip too

Kinfe123 (approved):

Yonom (approved):

PR #4755, opened 2026-07-07
feat(react-data-stream): detect stream protocol from response

Kinfe123:
## Summary

Teach `useDataStreamRuntime` to auto-detect legacy data-stream responses from `x-vercel-ai-data-stream: v1` when `protocol` is not set. Explicit `protocol` options still win, and responses without the marker keep the existing `ui-message-stream` fallback.

Docs and READMEs now show the simpler default setup while still mentioning the explicit override for proxies or cross-origin responses that strip or hide the header.

## Why

Users wiring AI SDK v4 `toDataStreamResponse()` or `assistant-stream` backends could accidentally leave out `protocol: "data-stream"` and get the wrong decoder. The response already carries a protocol marker, so the runtime can avoid that whole class of copy/paste mistakes when the header is visible.

## Testing

- `/Users/mac/oss/assistant-ui/node_modules/.pnpm/node_modules/.bin/vitest run packages/react-data-stream/src/protocol.test.ts`
- `./node_modules/.bin/aui-build` in `packages/react-data-stream`
- `./node_modules/.bin/oxlint packages/react-data-stream/src/useDataStreamRuntime.ts packages/react-data-stream/src/protocol.ts packages/react-data-stream/src/protocol.test.ts --no-error-on-unmatched-pattern`
- `git diff --check`

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/4755?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Yonom (approved):

Yonom:
The auto-detection will need to differentiate the v4 and v5 UI message stream formats (the text delta shape differs). AI SDK v5+ `toUIMessageStreamResponse()` sets `x-vercel-ai-ui-message-stream: v1`, and v4's `toDataStreamResponse()` sets `x-vercel-ai-data-stream: v1`, so we can key on the headers explicitly instead of treating every non-data-stream response as UI message stream.

Kinfe123:
@Yonom I opened the follow-up for the UI message stream header differentiation here: #4794. Keeping it separate from this merged PR so any additional changes land in the active follow-up instead of becoming stale on this thread.

PR #4726, opened 2026-07-05
feat(ui): client-side search for the thread list

ShobhitPatra:
Adds a search box to the thread-list registry component that filters loaded threads by title

## What

- Search input below "New Thread", shown once there is at least one thread.
- Filters by title, case-insensitive, and regroups matches by date.
- "No threads found" empty state when nothing matches.

## Scope

Registry component only. No runtime, core, adapter, or cloud API changes. Reuses the existing `useMemo` in `ThreadListItemGroups` that already derives and groups `s.threads.threadIds`. Searches only threads already loaded in the store.

## Follow-up

This is the client-side, loaded-threads-only tier. Searching across unloaded/paginated threads needs a `RemoteThreadListAdapter.search?()` plus a cloud/backend change, which is public-API surface deferred to a separate design discussion.
## Changes

- `thread-list.tsx`: new `ThreadListSearch` (controlled `<Input type="search">`); `ThreadList` owns the query state; `ThreadListItemGroups` filters + empty state.
- `registry.ts`: add `input` to the `thread-list` item's `registryDependencies`.

## Notes

- No changeset (`@assistant-ui/ui` and `@assistant-ui/shadcn-registry` are private).
- No template-sync impact (minimal does not ship `thread-list.tsx`).
- Backwards compatible: default render is unchanged with no query; `ThreadListItems` gains one optional `searchQuery` prop.
- Verified: `with-cloud` builds under strictest; `pnpm lint` clean.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/4726?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #4759, opened 2026-07-07
ci(template-metrics): surface unenforced gate when the main baseline is missing

ShobhitPatra:
When the `install-footprint-main-*` cache is evicted, every template reports `(new)`, the gate has nothing to compare against, and the check passes silently: any LOC increase lands ungated and gets baked into the next baseline. This is exactly how #4726's +64 went green without ever being compared to the 50-line limit.

## What

- `template-metrics.mjs` now writes a third gate verdict, `no-baseline`, when there is no baseline to compare against, and posts the sticky PR comment in that state so the `_No baseline found_` note is visible on the PR.
- The workflow's enforce step turns that verdict into a `::warning::` (check stays green): the gate was not enforced this run, and the next push to main touching `templates/` or `packages/ui/` recreates the baseline.

## Why warn instead of fail

Failing would block every open templates/ui PR from the moment of eviction until a main push refreshes the cache — punishing PRs for an infra event they didn't cause. A loud warning plus a visible PR comment makes the bypass reviewable without turning it into an outage.

## Notes

- Verified locally: the report script writes `no-baseline`/`fail`/`pass` and `post`/`skip` correctly for empty, regressed, and unchanged baselines.
- Workflow-only + repo script; no changeset needed.
- Stacked on #4757 (both edit the enforce step); rebases to a 2-file diff once that merges.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/4759?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/workflows/template-metrics.yaml
+++ b/.github/workflows/template-metrics.yaml
@@ -134,12 +134,19 @@ jobs:
 
       # always(): run even if the comment step failed (e.g. fork 403), so the
       # gate verdict is never lost. Only blocks PRs; pushes refresh the baseline.
+      # A missing baseline (evicted cache) cannot be compared against, so the
+      # gate downgrades to a loud warning instead of passing as if it ran.
       - name: Enforce regression gate
         if: always() && github.event_name == 'pull_request'
         run: |
-          if [ "$(cat metrics-gate.txt 2>/dev/null)" = "fail" ]; then
+          verdict="$(cat metrics-gate.txt 2>/dev/null)"
+          if [ "$verdict" = "fail" ]; then
             echo "::error::Template metrics regressed past the configured limits. See the PR comment."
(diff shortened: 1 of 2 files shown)


okisdev:
closing: #4853 removes the cached baseline entirely and compares against a throwaway worktree at the merge-base, so the no-baseline state this PR surfaces no longer exists as a state class. the silent-pass hole documented here is what motivated that rework.

PR #4777, opened 2026-07-08
chore: bump TypeScript to v7

Kinfe123:
## Summary

- Bump the workspace TypeScript dependency to 7.0.2 and refresh the pnpm lockfile.
- Keep known incompatible compiler-API/tooling workspaces on the TypeScript 6 compatibility alias while still adding a TypeScript 7 alias for the workspace.
- Keep Next.js apps/templates on the TypeScript 6 compatibility alias because Next still loads the old `typescript/lib/typescript.js` entry, which `typescript@7` no longer exposes.
- Fix the small TS7 typecheck fallout in compiler/tests and add the browser-extension CSS module declaration.

## Why the two TypeScript deps?

Packages that support TS7 can use the normal bump:

```json
"typescript": "^7.0.2"
```

Workspaces that still depend on the old compiler package shape use:

```json
"typescript": "npm:@typescript/typescript6@^6.0.2",
"typescript-7": "npm:typescript@^7.0.2"
```

That keeps `node_modules/typescript` compatible for tools like Next.js, while `typescript-7` keeps TS7 installed in the workspace for the actual TS7 migration work.

## Why this helps

- Moves supported parts of the repo onto the new TypeScript 7 compiler package.
- Avoids breaking Next/tooling workspaces that still expect the TypeScript 6 package layout.
- Keeps the migration explicit: compatibility aliases are visible and can be removed later when those tools support TS7 directly.

## CI notes

- Added a scoped `minimumReleaseAgeExclude` for `typescript` and `@typescript/typescript-*` so CI can install the TS7 native platform packages.
- Resolved the merge conflicts with latest `main` and regenerated the lockfile.
- The previous Next build failure was caused by Next resolving `typescript/lib/typescript.js`; the TS6 alias fixes that path.

## Validation

Local checks:

- `pnpm install --lockfile-only --ignore-scripts`
- `pnpm install --frozen-lockfile --ignore-scripts --fetch-timeout=300000`
- `pnpm lint`
- `pnpm --filter with-heat-graph build`
- `pnpm --filter assistant-ui-starter-minimal... build`

GitHub checks are passing for the previously failing lanes: `Oxlint + Oxfmt`, `Build Changed Packages`, `LOC + Bundle Size`, `Template Sync`, `Test Changed Packages`, CodeQL, and Vercel.


Yonom (approved):

Kinfe123:
I also want to callout that i merged with the TS6 alias as a compatibility bridge for Next.js. so as a follow up we will remove the alias once Next.js can consume TypeScript 7 directly.

PR #4779, opened 2026-07-08
feat(ui): thinking-steps trace UI for reasoning, tools, and sources

ephraimduncan:
## What

Reworks the assistant message trace in the shadcn/registry UI into a unified chain-of-thought "thinking steps" presentation:

- **thread.tsx** — groups reasoning, tool-call, and source parts into a collapsible chain-of-thought section with a step timeline and vertical connectors; sources render as a favicon row. `ToolGroup`/`ReasoningGroup` component overrides still dispatch through nested group keys when provided.
- **reasoning.tsx** — each reasoning part becomes a step: first paragraph is the headline, remaining paragraphs expand as a dotted thought timeline; auto-opens while streaming and collapses on completion (scroll-locked via `ReasoningRoot`'s `streaming` mode).
- **tool-fallback.tsx** — tool presets with running/done labels ("Searching web" → "Searched web"), globe/wrench icons, overlapping result favicons with counts, pretty-printed args, and cancelled/error states. Rendering is defensive: missing `toolName` falls back to "Tool", arg serialization never throws.
- **globals.css** (docs + 5 templates) — `working-dot` keyframe animation for the active-step indicator.
- **registry.ts** — thread now pulls in `sources.json` and ships the `working-dot` keyframes via `cssVars`/`css` so registry consumers get the animation.

## Why

Routine tool calls and reasoning are traces of what the model is doing, not standalone content. Folding them into a single chain-of-thought step list reads like the agent run it represents, instead of a flat stack of disconnected collapsibles.

## How

- Built on `MessagePrimitive.GroupedParts` + `groupPartByType` routing reasoning/tool-call/source into `group-chainOfThought`; `standalone-tool-call` parts stay ungrouped.
- `templates/minimal` `tool-fallback.tsx` copy kept byte-equal via `pnpm sync-templates`; minimal's `thread.tsx`/`markdown-text.tsx` remain declared OVERRIDES.
- No published npm package changes, so no changeset (packages/ui and templates are private).

## Screenshots

Thinking-steps trace: chain-of-thought with reasoning steps, web-search tool calls with favicons and result counts, and an expanded tool response panel

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/4779?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/docs/styles/globals.css
+++ b/apps/docs/styles/globals.css
@@ -66,6 +66,16 @@
   }
   --animate-shimmer: shimmer-sweep var(--shimmer-duration, 1000ms) linear
     infinite both;
+  --animate-working-dot: working-dot 1.4s ease-in-out infinite;
+  @keyframes working-dot {
+    0%,
+    100% {
+      opacity: 0.2;
+    }
+    40% {
+      opacity: 1;
+    }
+  }
 }
 
 :root {
--- a/apps/registry/src/registry.ts
+++ b/apps/registry/src/registry.ts
@@ -71,10 +71,26 @@ export const registry: RegistryItem[] = [
       "https://r.assistant-ui.com/follow-up-suggestions.json",
       "https://r.assistant-ui.com/markdown-text.json",
       "https://r.assistant-ui.com/reasoning.json",
+      "https://r.assistant-ui.com/sources.json",
       "https://r.assistant-ui.com/tooltip-icon-button.json",
       "https://r.assistant-ui.com/tool-fallback.json",
       "https://r.assistant-ui.com/tool-group.json",
     ],
+    cssVars: {
+      theme: {
+        "--animate-working-dot": "working-dot 1.4s ease-in-out infinite",
+      },
+    },
+    css: {
+      "@keyframes working-dot": {
+        "0%, 100%": {
+          opacity: "0.2",
+        },
+        "40%": {
+          opacity: "1",
+        },
+      },
+    },
   },
   {
     name: "voice",
@@ -129,8 +145,10 @@ export const registry: RegistryItem[] = [
     ],
     dependencies: [
       "@assistant-ui/react",
+      "@assistant-ui/react-markdown",
       "lucide-react",
       "class-variance-authority",
+      "remark-gfm",
       "tw-shimmer",
     ],
     css: {
--- a/packages/ui/src/components/assistant-ui/reasoning.tsx
+++ b/packages/ui/src/components/assistant-ui/reasoning.tsx
@@ -230,7 +230,7 @@ function ReasoningTrigger({
       <span className="aui-reasoning-trigger-line flex min-h-8 min-w-0 flex-1 items-center gap-1.5 py-1 pe-2 text-start">
         <span
           data-slot="reasoning-trigger-label"
-          className="aui-reasoning-trigger-label-wrapper text-muted-foreground group-hover/trigger:text-foreground relative shrink-0 text-sm leading-6 transition-colors motion-reduce:transition-none"
+          className="aui-reasoning-trigger-label-wrapper text-muted-foreground group-hover/trigger:text-foreground relative min-w-0 truncate text-sm leading-6 transition-colors motion-reduce:transition-none"
         >
           {label}
           {active ? (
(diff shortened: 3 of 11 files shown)


Kinfe123 (commented):
@ephraimduncan this looks so good. but one things i can callout is that it shows website icons for tool results by fetching them from Google. That can leak result domains and break in strict/offline apps, so the default UI should avoid that or may be make it opt-in.

Kinfe123:
@ephraimduncan One small thing: the default `ToolFallback` still fetches favicons from a third-party service, and I don’t think the override is exposed in the default path. Could we make those favicons opt-in/disableable, or expose the override there too or sth

ephraimduncan:
fixed. the favicons now go through the same duckduckgo endpoint sources.tsx already uses (so the kit talks to one favicon service, not two), each icon has an onError letter-chip fallback so blocked or offline environments degrade to initials instead of broken images, and there's a `faviconUrl` prop to point at a self-hosted proxy. favicon fetching for web results is already the kit's default via sources.tsx, so I matched that convention rather than making tool results opt-in on their own — if you'd rather have result favicons (or both surfaces) default-off, happy to flip it, that's a kit-wide call.

AVGVSTVS96:
@ephraimduncan @Kinfe123 @Yonom 

these are significant updates to our default components/templates, and none of it is necessary for achieving the thinking steps trace ui.

assistant-ui is built to make this very easy to implement yourself, we give users all the pieces they need in a minimal template with concise, easy to understand code. users can then customize however they want with the tools we provide them; this is the "idiomatic assistant-ui" way of implementing the same UI:

```tsx
return (
  <MessagePrimitive.Root
    data-slot="aui_assistant-message-root"
    data-role="assistant"
    className="fade-in slide-in-from-bottom-1 animate-in relative -mb-7.5 pb-7.5 duration-150 [contain-intrinsic-size:auto_200px] [content-visibility:auto]"
  >
    <div
      data-slot="aui_assistant-message-content"
      className="text-foreground px-2 leading-relaxed wrap-break-word"
    >
      <MessagePrimitive.GroupedParts
        groupBy={groupPartByType({
          reasoning: ["group-chainOfThought", "group-reasoning"],
          "tool-call": ["group-chainOfThought", "group-tool"],
          source: ["group-chainOfThought", "group-sources"],
        })}
      >
        {({ part, children }) => {
          switch (part.type) {
            case "group-chainOfThought": {
              const active = part.status.type === "running";
              return (
                <ReasoningRoot streaming={active} variant="ghost" className="mb-2">
                  <CollapsibleTrigger className="group text-muted-foreground hover:text-foreground data-[state=open]:text-foreground flex items-center gap-2 py-1 text-sm font-medium transition-colors">
                    <span className="flex size-4 items-center justify-center">
                      <WorkingDot active={active} />
                    </span>
                    Thinking
                    <ChevronDownIcon className="size-3.5 transition-transform group-data-[state=closed]:-rotate-90" />
                  </CollapsibleTrigger>
                  <ReasoningContent aria-busy={active}>
                    <ol className="border-border ms-2 space-y-3 border-s py-2 ps-5">
                      {children}
                    </ol>
                  </ReasoningContent>
                </ReasoningRoot>
              );
            }
            case "group-reasoning":
            case "group-tool":
              return children;
            case "group-sources":
              return (
                <TraceItem icon={GlobeIcon}>
                  <div className="flex flex-wrap items-center gap-1.5">{children}</div>
                </TraceItem>
              );
            case "text":
              return <MarkdownText />;
            case "reasoning":
              return (
                <TraceItem icon={LightbulbIcon}>
                  <Reasoning {...part} />
                </TraceItem>
              );
            case "tool-call":
              return (
                <TraceItem icon={GlobeIcon}>
                  {part.toolUI ?? <ToolFallbackComponent {...part} />}
                </TraceItem>
              );
            case "source":
              return <Sources {...part} />;
            case "data":
              return part.dataRendererUI;
            case "indicator":
              return (
                <span
                  data-slot="aui_assistant-message-indicator"
                  className="animate-pulse font-sans"
                  aria-label="Assistant is working"
                >
                  {"●"}
                </span>
              );
            default:
              return null;
          }
        }}
      </MessagePrimitive.GroupedParts>
      <MessageError />
    </div>

    <div
      data-slot="aui_assistant-message-footer"
      className={cn("ms-2 flex items-center", ACTION_BAR_HEIGHT)}
    >
      <BranchPicker />
      <AssistantActionBar />
    </div>
  </MessagePrimitive.Root>
);
```

---

moving forward please keep additions to registry components (and by extension all our templates) thoughtful and as minimal as possible. assistant-ui's goal is to abstract the complexity and ship a clean starting point for users to build on. always ensure you're not recreating existing APIs, and please be mindful when working on our main product surfaces :)

PR #4783, opened 2026-07-09
perf(ui): memoize diff parsing and contain offscreen diff files

ephraimduncan:
## What
Two perf fixes in the `diff-viewer` registry component for large patches.

## Why
`pairLinesForSplit(file.lines)` was called inline in JSX, re-pairing every line on every render in split view; `parsedFiles` was keyed on `oldFile`/`newFile` object identity, so inline object props re-parsed the whole patch each render. Every file section also paid full layout/paint while offscreen.

## How
- Split pairing folded into the existing `parsedFiles` memo (internal `ParsedFileWithSplitPairs`; exported types/APIs unchanged).
- Memo deps switched to primitive fields (`content`/`name`), preserving the empty-string-is-valid-input behavior.
- `[content-visibility:auto] [contain-intrinsic-size:auto_240px]` on each per-file root — same containment pattern the thread component already uses; the `overflow-x-auto` content element is untouched so horizontal scrolling is unaffected.

## Verification
- `pnpm --filter @assistant-ui/ui exec tsc --noEmit`, `oxlint`, `oxfmt --check` green.
- No `pairLinesForSplit` call remains in JSX; export block unchanged.
- Docs sample checked in browser: unified/split toggle renders identically, containment classes on file roots, horizontal scroll intact; two-file patch renders in order in both modes.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/4783?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/ui/src/components/assistant-ui/diff-viewer.tsx
+++ b/packages/ui/src/components/assistant-ui/diff-viewer.tsx
@@ -29,6 +29,9 @@ interface SplitLinePair {
   left: ParsedLine | null;
   right: ParsedLine | null;
 }
+interface ParsedFileWithSplitPairs extends ParsedFile {
+  splitLinePairs: SplitLinePair[];
+}
 
 function parsePatch(patch: string): ParsedFile[] {
   const files = parseDiff(patch);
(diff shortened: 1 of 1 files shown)


Kinfe123 (approved):

okisdev (approved):
## review

lgtm. two real perf problems fixed with the containment pattern the thread component already ships, and the shipped version addresses the one issue found in review. the nested scroll container concern does not apply here: relevancy accounts for clipping ancestors, and the same classes already work inside the thread viewport's overflow-y-scroll container.

- split pairing was folded into the parse memo (`ParsedFileWithSplitPairs`), so every parse ran `pairLinesForSplit` for every file even in the default unified mode where the pairs are never read; large unified patches paid throwaway work proportional to their total line count.

### what we modified

- replaces the folded pairs with a separate `splitLinePairs` memo gated on `viewMode === "split"` (with `viewMode` in the deps): unified mode short-circuits to an empty array, split mode pairs once per parse or mode change, and the render indexes it per file with a `?? []` fallback that also satisfies `noUncheckedIndexedAccess`.

### potential further fixes

- the assembled `DiffViewer` renders inline `data-slot` divs instead of the exported `DiffViewerFile`/`DiffViewerContent` primitives, so compositions built from the exported pieces do not pick up the containment class; the divergence predates this PR and belongs in a follow-up that makes the assembled component consume its own primitives.
- the PR body's how section still describes the folded `ParsedFileWithSplitPairs` approach rather than the shipped separate memo; a one-line edit keeps the description accurate for future readers.

PR #4786, opened 2026-07-09
perf(ui): contain offscreen message rows with content-visibility

ephraimduncan:
## What
Long threads paid full layout/paint for every offscreen message. This moves CSS containment to the per-message row roots in the canonical thread component and the minimal template.

## Why
`ThreadPrimitive.Messages` renders every message; only the assistant message *content* (not its root) and the user root had containment, and the edit composer had none. Offscreen rows still cost layout/paint at scale.

## How
- `[content-visibility:auto] [contain-intrinsic-size:auto_200px]` on the assistant, user, and edit-composer `MessagePrimitive.Root` elements (the rows the top-anchor system registers). Viewport and message group are untouched.
- Assistant action-bar spacing compensation moved from the footer's negative margin onto the root (`-mb-7.5 pb-7.5`), so paint containment no longer clips the hover action bar (the failure previously tracked for content-level containment).
- Uniform `auto 200px` intrinsic size: smaller role-specific estimates (24/60px) reproduced a bottom-anchoring failure on long-thread load during QA; `auto` keeps the last rendered size thereafter so scroll height stays stable.

## Verification
- `pnpm --filter @assistant-ui/ui exec tsc --noEmit` green.
- Browser QA against a 500-message seeded thread rendering the canonical `Thread`: initial load lands at bottom; scroll up/down stable (no jumps from rows re-entering); scroll-to-bottom button lands correctly; `turnAnchor="top"` send keeps the new user message pinned with no jitter; hover action bar paints inside the contained root; native find-in-page reaches offscreen contained messages; computed styles confirm containment on rows and none on viewport/group.
- #4104: Root-level `[contain-intrinsic-size:auto_200px]` on the anchor-registered `MessagePrimitive.Root` replaces the prior content-div `auto_24px` workaround; the original jitter came from an oversized 300px estimate, and QA confirms `turnAnchor="top"` send stays jitter-free.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/4786?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/ui/src/components/assistant-ui/thread.tsx
+++ b/packages/ui/src/components/assistant-ui/thread.tsx
@@ -336,22 +336,19 @@ const AssistantMessage: FC = () => {
     ReasoningGroup,
   } = useContext(ThreadComponentsContext);
 
-  // reserves space for action bar and compensates with `-mb` for consistent msg spacing
-  // keeps hovered action bar from shifting layout (autohide doesn't support absolute positioning well)
-  // for pt-[n] use -mb-[n + 6] & min-h-[n + 6] to preserve compensation
   const ACTION_BAR_PT = "pt-1.5";
-  const ACTION_BAR_HEIGHT = `-mb-7.5 min-h-7.5 ${ACTION_BAR_PT}`;
+  // Keep the action bar inside the contained root's paint box, then cancel its reserved space in flow.
+  const ACTION_BAR_HEIGHT = `min-h-7.5 ${ACTION_BAR_PT}`;
 
   return (
     <MessagePrimitive.Root
       data-slot="aui_assistant-message-root"
       data-role="assistant"
-      className="fade-in slide-in-from-bottom-1 animate-in relative duration-150"
+      className="fade-in slide-in-from-bottom-1 animate-in relative -mb-7.5 pb-7.5 duration-150 [contain-intrinsic-size:auto_200px] [content-visibility:auto]"
     >
       <div
         data-slot="aui_assistant-message-content"
-        // [contain-intrinsic-size:auto_24px] fixes issue #4104, don't change without checking for regressions
-        className="text-foreground px-2 leading-relaxed wrap-break-word [contain-intrinsic-size:auto_24px] [content-visibility:auto]"
+        className="text-foreground px-2 leading-relaxed wrap-break-word"
       >
         <MessagePrimitive.GroupedParts
           groupBy={groupPartByType({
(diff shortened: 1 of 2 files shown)


okisdev (approved):
This change moves CSS containment from the assistant-message content div onto each per-message `MessagePrimitive.Root` and adds it to the edit-composer wrapper, which is a reasonable placement for `content-visibility:auto` and consistent with the stated goal of skipping layout/paint for offscreen rows. The canonical `packages/ui` source and the `minimal` template are changed in lockstep, `thread.tsx` is already listed in the `OVERRIDES` array in `scripts/sync-templates.sh`, and both touched packages are `private: true`, so no changeset is required (Semver Check is green). The #4104 regression note that was removed inline is addressed in the PR body, so that prior concern is covered.

One thing I could not verify from the diff alone and ask the author to confirm: the spacing compensation does not read as a pure "move." Before, the footer carried `-mb-7.5 min-h-7.5`, where the footer's `-mb-7.5` collapsed through the unpadded `relative` root into the root's own margin-bottom, cancelling the 30px `min-h-7.5` action-bar reservation from the flow. After, the footer keeps `min-h-7.5` with no negative margin, and the root gains `pb-7.5`/`-mb-7.5` (a paint-box extension that is flow-neutral, but does not cancel the footer's reservation). By ordinary CSS margin-collapse against the `flex flex-col gap-y-6` parent, that leaves the footer's ~30px in flow, so each assistant message appears to gain ~30px of bottom footprint versus the prior overlapping behaviour. If that dedicated action-bar space is now intended (no overlap with the next message), please say so explicitly in the body; if net-zero spacing was the intent, the footer's `min-h-7.5` likely needs revisiting. I cannot execute the layout here, so I'm flagging it for visual confirmation rather than asserting a regression.

Two advisory notes from prior review that remain open but do not block: the inline #4104 warning is gone (the body covers it), and the `ACTION_BAR_PT`/`ACTION_BAR_HEIGHT` indirection could collapse to a single constant now that the `-mb-[n+6]` relationship no longer exists — purely readability.

review threads

* fixed: packages/ui/src/components/assistant-ui/thread.tsx:351 (#4104 regression warning) by @claude: The removed inline #4104 warning is addressed in the PR body, which names #4104 and explains the jitter came from an oversized 300px estimate and that turnAnchor=top send stays jitter-free.
* confirmed: packages/ui/src/components/assistant-ui/thread.tsx:341 (collapse ACTION_BAR_PT constant) by @claude: The indirection remains in the current head; ACTION_BAR_PT is still used only to build ACTION_BAR_HEIGHT. Non-blocking readability nit.

_Review by [Rupic](https://github.com/assistant-ui/rupic). Verified against the diff; runtime behavior was not executed._

PR #4793, opened 2026-07-09
fix(templates): surface the real error on failed chat requests

ShobhitPatra:
## What

Adds an `onError` to `toUIMessageStreamResponse` in every AI-SDK template (default, minimal, cloud, cloud-clerk, mcp) so failed requests return the real error text.

```ts
onError: (error) => (error instanceof Error ? error.message : String(error)),
```

## Why

The AI SDK v6 defaults `onError = () => "An error occurred."`, masking server errors from the client. So a user who forgets `OPENAI_API_KEY` gets a generic red bubble while the real `AI_LoadAPIKeyError` shows only in the terminal. Missing/invalid key is the most likely first-run failure; forwarding the message makes it self-diagnosing.

## Impact

- First-run failures now render their real message in the error surface (which already shows `errorText`).
- Happy path unchanged; `onError` only fires on error.
- Terminal logging unchanged (it comes from `streamText`'s separate `onError`).

## Mechanics

- 5 files, one line each. Not touched: `langchain` (already forwards its error) and `eve` (no AI-SDK route).
- No changeset (templates are `private: true`), no new dependency.
- Lint and `tsc --noEmit` clean.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/4793?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/templates/cloud-clerk/app/api/chat/route.ts
+++ b/templates/cloud-clerk/app/api/chat/route.ts
@@ -35,6 +35,8 @@ export async function POST(req: Request) {
 
   return result.toUIMessageStreamResponse({
     sendReasoning: true,
+    onError: (error) =>
+      error instanceof Error ? error.message : String(error),
     messageMetadata: ({ part }) => {
       if (part.type === "finish") {
         return {
--- a/templates/cloud/app/api/chat/route.ts
+++ b/templates/cloud/app/api/chat/route.ts
@@ -35,6 +35,8 @@ export async function POST(req: Request) {
 
   return result.toUIMessageStreamResponse({
     sendReasoning: true,
+    onError: (error) =>
+      error instanceof Error ? error.message : String(error),
     messageMetadata: ({ part }) => {
       if (part.type === "finish") {
         return {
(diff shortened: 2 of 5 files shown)


ShobhitPatra:
on the raw error-forwarding flag: this is a deliberate tradeoff, and worth it for these templates.

- these are getting-started scaffolds. the most likely first-run failure is a missing/invalid OPENAI_API_KEY, and the SDK default `"An error occurred."` gives the user nothing to act on while the real `AI_LoadAPIKeyError` shows only in the terminal.
- that same failure hits the first prod deploy too (env var not set), so a `NODE_ENV` gate would re-mask the most common error exactly where it is hardest to debug (no local terminal). that defeats the fix.
- the templates ship with no auth or rate-limiting and a hardcoded model. anyone graduating to prod revisits this route anyway, and the inline `onError` shows exactly where to sanitize.
- matches the ai sdk error-forwarding cookbook idiom, already used across ~20 files in this repo.

happy to add a one-line readme note that prod deploys should sanitize `onError`, if maintainers prefer that over code.

PR #4798, opened 2026-07-09
fix(docs): remove docs for the dropped useAssistantApi / AssistantIf family

ShobhitPatra:
## What

- Deletes the orphaned `api-reference/hooks/utilities.mdx` page.
- Removes the trailing `## AssistantIf` section from `api-reference/primitives/assistant-if.mdx`.

## Why

Both spots still show imports for names removed by v0.13 (package is now 0.14.26). `hooks/utilities.mdx` is invisible in the sidebar but still a live route advertising `useAssistantApi`, `useAssistantState`, `useAssistantEvent`, and `unstable_useCloudThreadListAdapter`. The `## AssistantIf` section's callout says "removed in v0.13" yet still shows `import { AssistantIf }`.

## Impact

- No coverage lost: every hook the deleted page documented is already on a live, nav-linked page. Nothing links to `hooks/utilities` or the `#assistantif` anchor. The `AssistantIf` -> `AuiIf` rename stays in the migration guides.

## Notes

- A clean `generate:api-reference` run does not recreate the deleted file or touch `assistant-if.mdx`.
- No changeset: `@assistant-ui/docs` is private. Docs-only, no API output change.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/4798?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


okisdev:
closing in favor of #4850. the diagnosis here was right and both deletions were correct, but removing the stale pages by hand does not fix the problem at its source: the generator could neither retire an orphaned hand-authored page nor stop the next one from appearing, so the same rot would grow back. #4850 folds these two deletions in and adds the missing mechanism (unmanaged-page reporting, a --strict mode, and a CI drift gate that fails on any api-reference drift). thanks for surfacing this.

PR #4799, opened 2026-07-09
chore: add build:packages and build:affected root scripts

ShobhitPatra:
## What

Two additive root scripts plus a short CONTRIBUTING section. No existing script or build behavior changes.

- `build:packages` = `turbo build --filter='./packages/*'`:  builds every publishable package, no example or template apps.
- `build:affected` = `turbo build --affected`:  builds only changed packages plus their dependents.

## Why

The only build entrypoint today is `build`, which builds the full 93-task graph (45 packages plus 40 Next.js apps). Neither everyday package check has a memorable script: `build:packages` is the pre-push cross-package export check (cache-friendly, unlike `api-surface:build`'s `--force`), and `build:affected` is the local mirror of CI's `--filter="...[origin/base]"`.

## Impact

- CI: none. No workflow reads these scripts; CI invokes turbo with its own filters.
- Build outputs: none. Additive keys, strict subsets of the existing graph.
- No changeset: root `package.json` is private and CONTRIBUTING.md is unpublished.
- `build:affected` diffs against local `main`; the CONTRIBUTING note says to fetch first if stale.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/4799?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -24,6 +24,12 @@ pnpm turbo build
 
 (some packages rely on build outputs from other packages, even if you want to start the project in development mode)
 
+### Building for a pull request
+
+- `pnpm build:packages` builds every package under `packages/` (no example or template apps). Use it before pushing a change to `packages/*` to catch cross-package export breakage.
+- `pnpm build:affected` builds only the packages you changed plus their dependents, approximating what CI runs on your PR. Fastest check on a feature branch. It compares against your local `main`, so run `git fetch origin main:main` first if yours is stale or missing.
+- `pnpm build` builds everything including the example and template apps. Use it only when you touch `examples/`, `templates/`, or `turbo.json`.
+
 ### Running the project
 
 To run the docs project in development mode:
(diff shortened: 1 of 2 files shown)


Kinfe123 (changes requested):
Good catch from cubic tho. Both notes look valid but low-risk: build:packages should say it builds every package under packages/*, since the filter includes private packages too, and build:affected should say it approximates CI rather than mirrors it because CI uses the fetched remote base ref. I’d apply those wording tweaks, then this looks good to merge.

okisdev:
closing to keep the root script surface minimal. both invocations already work directly (`pnpm turbo build --filter='./packages/*'` and `pnpm turbo build --affected`), and every root alias is one more name to keep in sync with how we invoke turbo. documenting the raw turbo commands in CONTRIBUTING would be welcome as its own small PR. thanks!

PR #4819, opened 2026-07-10
fix: parse CRLF SSE events in A2A client

Kinfe123:
## Summary
- parse A2A SSE responses line-by-line so LF (`\n`), CRLF (`\r\n`), and CR (`\r`) line endings dispatch correctly
- handle CRLF pairs split across stream chunks without treating them as two line endings
- preserve completed-frame behavior by not dispatching an unterminated trailing `data:` frame
- add regression tests for CRLF, CR-only, split-CRLF, and malformed/unterminated stream handling

## Why
A2A streams use Server-Sent Events. The SSE format allows different line endings, and HTTP servers/proxies can emit frames as `\r\n\r\n` or split those delimiters across chunks. The previous parser only searched for `\n\n`, so a compliant server could send valid events while `streamMessage()` kept them buffered and yielded no updates.

This is intended as a compatibility fix: existing LF-delimited streams continue to parse the same way, while CRLF/CR-delimited streams now work too. This helps apps connecting to custom A2A servers, framework adapters, or proxies that normalize line endings differently.

## Validation
- `pnpm --filter @assistant-ui/react-a2a test`
- `git diff --check`


okisdev (approved):
## review

LGTM as `patch` with the parser reshaped on top.

- the premise was verified and is real: `"\r\n\r\n".indexOf("\n\n") === -1`, so a spec-compliant CRLF server starved the old parser entirely; the regression matrix (CRLF, CR, split CRLF, unterminated) is exactly the right coverage, and the revision correctly stopped dispatching unterminated trailing frames
- the remaining issue was idiom, not correctness: the per-code-point state machine with a manual `sawCarriageReturn` flag diverged from how the repo already splits incremental lines (`LineDecoderStream`'s split-plus-tail-buffer shape) and paid per-character cost on the streaming path

### what we modified

- reshaped `processText` to line-oriented splitting on `/\r\n|\r|\n/` with the incomplete tail kept in `lineBuffer`; a chunk-trailing `\r` still terminates its line immediately (same dispatch timing as the per-character version, no event sits buffered waiting for the next chunk) and `pendingLF` swallows the leading `\n` of the next chunk so a split CRLF is not counted twice; empty decoder output (partial multibyte sequences) returns early so the flag survives
- `readEvent`/`processLine` (field parsing, JSON-RPC unwrap, no-dispatch-on-unterminated) are untouched; added one regression test for CR-only delimiters split across chunks
- merged `main` to freshen the base; no behavior interaction

### potential further fixes

- `react-google-adk`'s `AdkClient` has the same hand-rolled `\n\n` scan and most likely the same CRLF starvation; same treatment applies, separate PR
- `assistant-stream`'s internal SSE utilities have their own spec gaps (no CR-only support in `LineDecoderStream`, `SSEEventStream.flush` dispatches unterminated frames) and are not exported; consolidating all three parsers onto one exported, spec-complete utility is the long-term shape

PR #4821, opened 2026-07-10
fix(react-google-adk): parse standard sse frames

Kinfe123:
## Summary
- parse Google ADK SSE responses line-by-line instead of splitting only on `\n\n`
- accept standard SSE `data:` fields with or without a single space after the colon
- support LF, CRLF, and CR line endings, including CRLF delimiters split across stream chunks
- add regressions for the previously failing cases

## Why
This is a focused follow-up to #4819, which fixed the same SSE framing problem in the A2A client. The merged PR's follow-up notes identified Google ADK's hand-rolled parser as having the equivalent CRLF starvation risk, so this PR applies that treatment separately to `AdkClient`.

The ADK stream parser only recognized `data: ` and `\n\n` frame separators. That meant valid SSE frames like `data:{...}\n\n` were ignored, CR-only frames could fail when multiple events were present, and CRLF-delimited streams would not emit updates until the response closed.

The streaming case matters for real ADK/proxy integrations: a server can send a complete `\r\n\r\n` frame while keeping the response open, and the UI should receive that event immediately instead of waiting for connection close.

## Verification
- Added tests that fail on main for `data:` without a space, CR-delimited frames, and CRLF frames emitted before response close
- `pnpm --filter @assistant-ui/react-google-adk test`
- `pnpm --filter @assistant-ui/react-google-adk build`
- `git diff --check`

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/4821?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


okisdev (approved):
## review

LGTM as `patch` with the parser ported to the shared shape.

- the three fixed bugs are real and were pinned by failing-on-main tests: `\n\n`-only splitting starved CRLF streams entirely, `data:` without a trailing space was silently dropped by the old `startsWith("data: ")`/`slice(6)` pair, and CR-only multi-event streams failed
- the controller-driven liveness test (event observed before the response closes, raced against a timeout) is a genuinely stronger assertion than the sibling a2a suite has; worth back-porting there eventually
- the one issue was timing, not correctness: the per-code-point `sawCarriageReturn` state machine was the same shape #4819 shipped and then replaced with a line-oriented split before merging, so landing it here would have reintroduced the divergence

### what we modified

- ported `processText` to the line-oriented parser that merged in #4819 (`aed9ba2a0`): split on `/\r\n|\r|\n/` with the incomplete tail buffered, a chunk-trailing `\r` terminating its line immediately, `pendingLF` swallowing the split CRLF's `\n`, and empty decoder output returning early; the comment is verbatim-shared with the a2a sibling
- kept ADK's deliberate divergence intact: unterminated trailing frames still dispatch at stream end (legacy proxy tolerance, pinned by the existing `handles remaining buffer` test), unlike a2a's spec-strict drop; the consolidation follow-up needs a flush-mode switch for exactly this
- added a CR-only split-across-chunks regression test mirroring the a2a suite

### potential further fixes

- the long-term shape stays the recorded follow-up: one exported CR/CRLF/LF-aware SSE decoder in `assistant-stream` (with a flush-mode option) consumed by a2a, google-adk, and react-pi; with both adapters now byte-identical in parser shape, the extraction is mechanical

PR #4822, opened 2026-07-10
fix(cloud): guard anonymous auth storage

Kinfe123:
## Summary
- treat anonymous Cloud auth localStorage access as best-effort cache reads/writes
- fall back to a new anonymous token when the cached refresh token is malformed or storage is unavailable
- keep the existing valid refresh-token path working even when rotated-token storage fails

## Why
Anonymous Cloud auth stores the refresh token in `localStorage`, but browser storage is not guaranteed to be available or well-formed. A user can have a corrupt `aui:refresh_token` value from an older build/manual testing, or storage can throw in Safari private mode, embedded contexts, strict privacy settings, or when quota is unavailable.

The refresh-token cache should not be a hard dependency for auth. If reading/parsing/removing/writing storage fails, the SDK can still request a fresh anonymous token or use the access token it just received.

## Verification
- Added regressions for malformed cached refresh token JSON
- Added regressions for throwing browser storage access
- Added coverage that the valid cached refresh-token path still uses the refresh endpoint when storing the rotated token fails
- `pnpm --filter assistant-cloud test`
- `pnpm --filter assistant-cloud build`
- `git diff --check`

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/4822?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/cloud-anonymous-storage.md
+++ b/.changeset/cloud-anonymous-storage.md
@@ -1,0 +1,5 @@
+---
+"assistant-cloud": patch
+---
+
+fix: guard anonymous auth refresh token storage
--- a/packages/cloud/src/AssistantCloudAuthStrategy.ts
+++ b/packages/cloud/src/AssistantCloudAuthStrategy.ts
@@ -109,6 +109,67 @@ export class AssistantCloudAPIKeyAuthStrategy implements AssistantCloudAuthStrat
 
 const AUI_REFRESH_TOKEN_NAME = "aui:refresh_token";
 
+type StoredRefreshToken = {
+  token: string;
+  expires_at: string;
+};
+
+const getLocalStorage = (): Storage | null => {
+  try {
+    return typeof localStorage === "undefined" ? null : localStorage;
+  } catch {
+    return null;
+  }
+};
+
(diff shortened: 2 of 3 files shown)


okisdev:
closing as superseded by #4870, which landed the same guard quartet (guarded storage resolution, read, write, remove) in this file three days after this PR opened, with #5157 reworking the surrounding auth flow since; the conflict state reflects that region having been rewritten twice. the one remaining difference, explicit shape validation of the parsed token, is covered behaviorally on main: a malformed value routes through the NaN expiry comparison into removeRefreshToken plus a fresh anonymous token, which also cleans the bad value out of storage. the test file now carries both regression classes this PR added (throwing storage and corrupted JSON). this was the earlier diagnosis of the same real gap; #4870 simply landed it first.

PR #4846, opened 2026-07-11
fix(cloud): normalize base urls across auth strategies

Kinfe123:
## Summary

- normalize custom Cloud base URLs for API-key, JWT, and anonymous authentication
- pass the normalized URL into anonymous token requests as well as regular API requests
- add regression coverage for JWT and anonymous auth while retaining API-key coverage

## Why

A custom base URL ending in a slash was normalized only for API-key auth. JWT and anonymous configurations could therefore request paths such as https://cloud.example.com//v1/threads, which strict proxies and routers may reject.

The normalizer removes trailing slashes from valid URLs while preserving the existing deferred behavior for falsy environment-backed values used during template prerendering.

## Verification

- pnpm --filter assistant-cloud test -- AssistantCloudAPI.test.ts
- pnpm --filter assistant-cloud build
- pnpm --filter assistant-ui-starter-cloud build
- pnpm lint
- pnpm build (86/86 tasks passed)

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/4846?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/cloud-base-url-normalization.md
+++ b/.changeset/cloud-base-url-normalization.md
@@ -1,0 +1,5 @@
+---
+"assistant-cloud": patch
+---
+
+fix: normalize Cloud base URLs across authentication modes
--- a/packages/cloud/src/AssistantCloudAPI.ts
+++ b/packages/cloud/src/AssistantCloudAPI.ts
@@ -65,8 +65,10 @@ type MakeRequestOptions = {
   body?: object | undefined;
 };
 
-const normalizeBaseUrl = (baseUrl: string) =>
-  baseUrl ? baseUrl.replace(/\/+$/, "") : baseUrl;
+const normalizeBaseUrl = (baseUrl: string) => {
+  if (!baseUrl || !baseUrl.endsWith("/")) return baseUrl;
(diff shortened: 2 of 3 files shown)


okisdev (approved):
verified against today's main before merging: the raw baseUrl assignments for JWT and anonymous auth are still live, so the double-slash vector through the /v1 join is real, and the touched file has not changed since the PR was opened. normalizeBaseUrl keeps single-slash parity with the removed regex, the anonymous strategy now receives the normalized url so the token endpoint is fixed too, and the falsy guard preserves the deferred prerender behavior. since the head's CI predates the recent cloud changes, I re-ran the cloud suite on a local merge with current main: 58 passing alongside the auth dedup and history persistence tests. the CodeQL thread resolved itself by construction, the flagged regex is the code this PR deletes.

PR #4861, opened 2026-07-12
fix(react-ai-sdk): retry failed history appends

Kinfe123:
## Summary

- mark an external history message as persisted only after every formatted append succeeds
- stop the current persistence pass after a failed append so descendants are not written ahead of a missing parent
- add a regression test covering a transient parent append failure and the following retry

## Use case

An assistant response can render successfully in the current chat while its history write fails because of a brief network interruption or temporary storage outage. The message remains visible until reload, but the old bookkeeping marked it as persisted anyway. When a later run completed, the missing parent was skipped and a newer descendant could be attempted instead, leaving incomplete Cloud or custom-adapter history after reload. This change keeps the failed parent eligible for the next persistence pass and retries it before continuing with descendants.

## Why

`useExternalHistory` previously added a message ID to its in-memory persisted set before awaiting `formatAdapter.append()`. If that append rejected, later persistence passes treated the message as already stored and skipped it.

The bookkeeping now happens after the append batch succeeds. A failed write is reported and remains eligible for the next persistence pass.

## Test plan

- `pnpm --filter @assistant-ui/react-ai-sdk test` (140 tests)
- `pnpm lint`
- `pnpm turbo build --filter=@assistant-ui/react-ai-sdk`
- verified the new regression test fails against the original ordering and passes with this change

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/4861?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/retry-ai-sdk-history-appends.md
+++ b/.changeset/retry-ai-sdk-history-appends.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react-ai-sdk": patch
+---
+
+fix: retry history messages after a failed persistence append
--- a/packages/react-ai-sdk/src/ui/use-chat/useExternalHistory.test.ts
+++ b/packages/react-ai-sdk/src/ui/use-chat/useExternalHistory.test.ts
@@ -1,7 +1,7 @@
 // @vitest-environment jsdom
 
-import { renderHook } from "@testing-library/react";
-import { describe, expect, it, vi } from "vitest";
+import { act, renderHook, waitFor } from "@testing-library/react";
+import { afterEach, describe, expect, it, vi } from "vitest";
 import type {
   AssistantRuntime,
   MessageFormatAdapter,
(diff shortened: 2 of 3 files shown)


okisdev:
closing as superseded. the diagnosis here was correct against the july 12 file: marking `historyIds` before the awaited appends meant a failed append never retried. since then the file was rewritten three times and the bug is gone as a byproduct: #5087 serialized the persist passes, #5052's changed-message baseline re-arms a message whose pass aborted before the snapshot write, and the per-inner-item bookkeeping from #5051-era work dedupes the replay, so a failed append retries on the next pass without replaying items that already succeeded. verified empirically on current main: the retry regression tests pass unmodified. #5179 lands those tests to pin the behavior, which is the part of this PR that stays valuable; the code half no longer applies to the current structure.

PR #4867, opened 2026-07-13
perf(react-ink-markdown): memoize markdansi render() in MarkdownText

ShobhitPatra:
## What

`MarkdownText` called markdansi's one-shot `render(text, options)` unconditionally in the component body. This PR wraps the component in `React.memo`, so re-renders with unchanged props skip the markdown parse and layout entirely.

## Why

In an Ink app the whole live region re-renders on every composer keystroke and spinner frame. Since `render()` ran in the component body, every visible message's full markdown document was re-parsed and re-laid-out per frame, not just the streaming one. The inputs are pure, so all of that work on unchanged text is redundant. In long threads this is measurable per-frame cost during typing and spinner animation.

React Compiler does not cover this package (aui-build applies it only to packages with a direct `@assistant-ui/tap` dependency), so the memoization has to be manual.

## Impact

- Keystroke and spinner-frame re-renders now skip the markdown parse for every message whose props are unchanged; only the actively streaming message recomputes.
- Rendered output is byte-for-byte identical. `MarkdownTextProps` and the export surface are untouched; the component body is unchanged.
- Streaming behavior is unchanged: while text grows each chunk, the memo misses and re-renders as before.

## Rationale and mechanics

`React.memo` with the default shallow compare is the right mechanism here rather than a `useMemo` inside the body: the component's props are all primitives or stable references, and memo preserves the rest-spread pass-through of any props not yet declared on `MarkdownTextProps`. That forwarding is load-bearing: `examples/with-react-ink-web` passes `color` (not yet a declared prop) through to markdansi so it never falls back to reading `process.stdout` in the browser bundle. An earlier revision of this PR used `useMemo` over an explicit prop list and silently dropped such props; the memo shape fixes that and follows the in-repo precedent in `react-ink`'s internal `MemoMessage`.

- New tests: (1) a black-box memoization check using a stable `highlighter` spy on a single-line fenced code block, asserting exact call counts across a same-props rerender (memo hit), a text change, and an option change; verified red on the unmemoized implementation. (2) A forwarding regression test that smuggles an undeclared `tableEllipsis` option through the spread and asserts it reaches markdansi's output.
- Full suite green (18 tests), lint clean on the touched files, full `turbo build` green (86/86).
- Changeset: patch on `@assistant-ui/react-ink-markdown`.
- No behavior or output change for consumers; perf only.

### DIFF
--- a/.changeset/memoize-ink-markdown-render.md
+++ b/.changeset/memoize-ink-markdown-render.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react-ink-markdown": patch
+---
+
+perf: memoize markdansi render() in MarkdownText
--- a/api-surface/assistant-ui__react-ink-markdown.ts
+++ b/api-surface/assistant-ui__react-ink-markdown.ts
@@ -1,9 +1,9 @@
 import { RenderOptions, Theme, Theme as Theme$1, ThemeName, ThemeName as ThemeName$1 } from "markdansi";
 
-declare const MarkdownText: {
+declare const MarkdownText: import("react").MemoExoticComponent<{
   (_param0: MarkdownTextProps): import("react").JSX.Element;
   displayName: string;
-};
+}>;
 
 declare const MarkdownTextPrimitive: {
   (_param1: MarkdownTextPrimitiveProps): import("react").JSX.Element;
--- a/packages/react-ink-markdown/src/MarkdownText.tsx
+++ b/packages/react-ink-markdown/src/MarkdownText.tsx
@@ -43,15 +43,45 @@ export type MarkdownTextProps = {
   listIndent?: number;
 };
 
+// One shared "resize" listener per stream; per-instance listeners trip Node's
+// ten-listener warning once a thread renders more than ten markdown messages.
+type ResizeStore = {
+  subscribers: Set<() => void>;
+  notify: () => void;
+};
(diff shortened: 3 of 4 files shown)


okisdev (approved):
## review

LGTM as `patch` with the resize fix applied on top.

- the memoization dropped resize reactivity: markdansi reads `process.stdout.columns` at `render()` time only when `width` is omitted and wrapping is on (`dist/render.js:22`), and the app-level re-renders that used to re-run it are exactly what `memo` now blocks, so completed messages kept their stale wrapping after a terminal resize (widening never recovers; narrowing degrades to Ink's character wrap over pre-wrapped lines)
- the premise itself checked out precisely: aui-build gates React Compiler on a direct `@assistant-ui/tap` dependency (`dependsOnTap ? [reactCompiler()] : []` in x-buildutils), so manual memoization here is necessary rather than redundant, and the highlighter call-count test is a solid behavioral pin

### what we modified

- subscribed the memoized impl to the Ink context stdout's `resize` via `useSyncExternalStore`, injecting the live width only under markdansi's own gate (`width` unset, `wrap !== false`), so spinner and keystroke frames still hit the memo while resizes invalidate exactly; the width source also moves from `process.stdout` to Ink's stdout, which is the correct one for embedded stdouts like ink-web
- added a widen-direction resize test (2 lines to 1 line; widening is the clean probe because Ink's own layout wrap masks the narrow direction) and updated the jsdoc

### potential further fixes

- none

PR #4872, opened 2026-07-13
fix(cloud-ai-sdk): retry interrupted history loads

Kinfe123:
## Summary

- release a thread's pending history-load marker when navigation interrupts that request
- prevent a cancelled failed request from clearing a newer replacement load
- add regressions for switching back to a thread and for the replacement-load race

## Why

While playing around with Cloud-backed thread switching locally, I switched from thread A to thread B and back to A before A's history request had finished. I hit an empty thread that would not retry its history load. The effect cleanup cancelled the original request, but left its promise in `meta.loading`, so A continued to look like it was already loading.

The cleanup now releases only the promise it started. This allows A to load again immediately without letting the older cancelled request overwrite or clear the replacement request.

## Validation

- `pnpm --filter @assistant-ui/cloud-ai-sdk test` (58 tests)
- `pnpm --filter @assistant-ui/cloud-ai-sdk build`
- `pnpm lint`

The A → B → A regression reproduced on `main` as one load for A instead of two, and passes with this change.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/4872?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/cloud-ai-sdk-history-load-cancellation.md
+++ b/.changeset/cloud-ai-sdk-history-load-cancellation.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/cloud-ai-sdk": patch
+---
+
+fix: retry interrupted thread history loads
--- a/packages/cloud-ai-sdk/src/chat/useCloudChat.switch.test.ts
+++ b/packages/cloud-ai-sdk/src/chat/useCloudChat.switch.test.ts
@@ -1,8 +1,9 @@
 // @vitest-environment jsdom
 
 import { act, renderHook } from "@testing-library/react";
-import { describe, expect, it, vi } from "vitest";
+import { afterEach, describe, expect, it, vi } from "vitest";
 import type { UseThreadsResult } from "../types";
+import { CloudChatCore } from "../core/CloudChatCore";
 import { useCloudChat } from "./useCloudChat";
 
 function createThreads(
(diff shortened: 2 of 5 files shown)


okisdev (approved):
## review

LGTM as `patch`. verified by transplant: the commit cherry-picks cleanly onto current main (including #5140's async tool callback changes in the same files) and the package passes 62/62 there, and with the two source files reverted exactly the two regressions fail, the switch-back test reproducing the reported one-load-instead-of-two and the core test losing the replacement marker. the fix is a coherent guard pair: the effect cleanup releases `meta.loading` only when it still points at the promise that cleanup owns, the same identity-guard idiom the UIMessageStream args-alias fix used, and the core tail leaves the marker alone on the cancelled path so a late-settling cancelled request cannot clear its replacement. traced the three orderings (cancel then remount, late settle, normal success) and each leaves the marker with its rightful owner.

- none; single commit with both guard halves, regressions for each, spy hygiene via `afterEach(restoreAllMocks)`, and the changeset.

### what we modified

- nothing; single commit.

### potential further fixes

- none. the interrupted-load marker lifecycle is now pinned from both sides; #5134's cloud-change reset is the adjacent open concern and stays separate.

PR #4876, opened 2026-07-13
fix(assistant-stream): support LF, CRLF, and CR line endings

Kinfe123:
## Summary
- parse all SSE-standard line endings in `LineDecoderStream`: LF, CRLF, and CR
- preserve CRLF as a single delimiter when `\r` and `\n` arrive in separate chunks
- add regression coverage for CR-delimited, empty, mixed, and chunk-split lines
- add a patch changeset for `assistant-stream`

## Why
This is another focused follow-up identified during merged PR #4819. `LineDecoderStream` currently splits only on `\n`, which handles LF directly and CRLF by stripping the preceding `\r`, but it cannot recognize standalone CR line endings.

Standalone CR is also valid SSE. A valid stream such as `data: hello\r\r` therefore remains buffered and ends with an incomplete-line error instead of dispatching the event. Fixing the shared line decoder applies consistently to object streams, assistant transport, UI message streams, and data streams.

This complements #4875 without overlapping it: #4875 discards genuinely unterminated events at EOF, while this PR recognizes events terminated by every valid SSE line-ending form.

## Implementation
The decoder scans each chunk character by character: `\n` or `\r` terminates the buffered line immediately, and a `skipNextLineFeed` flag set after `\r` swallows a leading `\n` in the same or the following chunk, so a CRLF pair split across chunks counts as one delimiter instead of two. CR-only lines dispatch as soon as they arrive, partial lines are buffered without rescanning on every chunk, and the strict flush error for a stream ending in an unterminated line is unchanged.

## Verification
- confirmed the CR-only and mixed-line-ending regressions fail on current `main` before the fix
- `pnpm --filter assistant-stream test` (251 passed, 20 skipped)
- `pnpm --filter assistant-stream build`
- scoped `oxlint` and `oxfmt --check`
- `git diff --check`

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/4876?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/assistant-stream-standard-sse-line-endings.md
+++ b/.changeset/assistant-stream-standard-sse-line-endings.md
@@ -1,0 +1,5 @@
+---
+"assistant-stream": patch
+---
+
+fix: parse all standard SSE line endings
--- a/packages/assistant-stream/src/core/utils/stream/LineDecoderStream.test.ts
+++ b/packages/assistant-stream/src/core/utils/stream/LineDecoderStream.test.ts
@@ -40,6 +40,14 @@ describe("LineDecoderStream", () => {
     expect(lines).toEqual(["line1", "line2", "line3"]);
   });
 
+  it("should split lines on CR (\\r)", async () => {
+    const stream = createTextStream(["line1\rline2\rline3\r"]);
+    const lines = await collectLines(
+      stream.pipeThrough(new LineDecoderStream()),
+    );
+    expect(lines).toEqual(["line1", "line2", "line3"]);
+  });
+
   it("should handle CRLF split across chunks", async () => {
     // The \r is at the end of one chunk and \n at the start of the next
     const stream = createTextStream(["line1\r", "\nline2\r\n"]);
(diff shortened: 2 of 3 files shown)


okisdev (approved):
## review

LGTM as `patch`. verified against the SSE spec and by execution: standalone CR is valid SSE and main's `\n`-only split leaves `data: hello\r\r` buffered until the incomplete-line error; the char-scan rewrite handles LF, CRLF, CR, chunk-split CRLF (one delimiter, no phantom empty line), `\r\r` empty lines, and keeps the strict flush error unchanged; the full assistant-stream suite passes on this head (251 passed, 20 skipped) with all four in-package consumers exercised.

- the first commit's regex-split shape rescanned the buffered partial line on every chunk; the follow-up commits moved to a character scan with `skipNextLineFeed`, and the PR body still described the old form. refreshed the body's implementation paragraph at review time so the description matches what ships.

### what we modified

- nothing in code; only the PR description refresh above.

### potential further fixes

- after #4880, assistant-stream carries two line-ending implementations (`SSEEventDecoder`'s regex plus pendingLF, and this char scan). their contracts differ (event-level vs line-level, and #4875 is about to change EOF semantics), so unifying now would be premature; worth revisiting once #4875 lands.

PR #4893, opened 2026-07-13
fix(react-google-adk): normalize session api urls

Kinfe123:
## Summary

- strip trailing slashes from the Google ADK `apiUrl` before constructing session endpoints
- keep existing URL encoding for app, user, session, and artifact identifiers
- add regression coverage for `apiUrl` values ending in one or multiple slashes
- add a patch changeset for `@assistant-ui/react-google-adk`

## Problem

While configuring a local ADK server with:

```ts
apiUrl: "http://localhost:8000/"
```

I found that session requests were constructed with a double slash:

```text
http://localhost:8000//apps/my-app/users/user-1/sessions
```

Permissive development servers may normalize this silently, but strict routers and reverse proxies can treat `//apps/...` as a different path, return a 404, or redirect the request. A redirect is especially problematic for session creation because some redirect responses can change a `POST` into a `GET`.

## Fix

Normalize the configured API URL once before building the shared session base URL. All session and artifact operations then inherit the canonical path, while configurations without trailing slashes remain unchanged.

## Verification

- focused `AdkSessionAdapter` suite: 26 tests passed
- full `@assistant-ui/react-google-adk` suite: 186 tests passed
- package and dependency builds passed
- full repository lint and format checks passed
- changeset validation reports a patch bump
- real local HTTP smoke tests confirmed inputs ending in `/` and `//` both reach `/apps/my%20app/users/user-1/sessions`, not `//apps/...`

### DIFF
--- a/.changeset/google-adk-session-api-urls.md
+++ b/.changeset/google-adk-session-api-urls.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react-google-adk": patch
+---
+
+fix: normalize trailing slashes in ADK session API URLs
--- a/packages/react-google-adk/src/AdkSessionAdapter.test.ts
+++ b/packages/react-google-adk/src/AdkSessionAdapter.test.ts
@@ -342,19 +342,22 @@ describe("createAdkSessionAdapter - load", () => {
 // ── URL construction ──
 
 describe("createAdkSessionAdapter - URL construction", () => {
-  it("normalizes a trailing slash in apiUrl", async () => {
-    mockFetch.mockResolvedValueOnce(
-      new Response(JSON.stringify([]), { status: 200 }),
-    );
+  it.each(["http://localhost:8000/", "http://localhost:8000//"])(
+    "normalizes trailing slashes in apiUrl: %s",
+    async (apiUrl) => {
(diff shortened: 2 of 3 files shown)


okisdev (approved):
## review

LGTM as `patch`.

- the first pass trimmed exactly one trailing slash (`apiUrl.replace(/\/$/, "")` at AdkSessionAdapter.ts:82 in 125e2f6), so `apiUrl` values ending in `//` still produced `//apps/...` in the session base URL
- coverage matched that gap: the initial test exercised only the single-slash input, so the repeated-slash case had no failing test

### what we modified

- replaces the single-slash trim with a trim of the whole trailing slash run, first via `/\/+$/` and finally as a character loop (`trimTrailingSlashes`), which keeps the trim single-pass and makes the regex-engine discussion moot
- extends the coverage to `it.each(["http://localhost:8000/", "http://localhost:8000//"])`, both asserting the canonical base URL
- groups the URL cases under one `URL construction` describe block alongside the existing encoding coverage

### potential further fixes

- `react-pi` carries its own `trimTrailingSlash` (httpClient.ts:72) with a slightly different shape; unifying the helpers is deliberately deferred, matching the repo's pattern of extracting shared utilities only once a third identical consumer appears

PR #4899, opened 2026-07-14
fix(react-ai-sdk): guard toolOutputConversion against non-string mediaType

ShobhitPatra:
## What

`toAISDKContent` in `toolOutputConversion.ts` read `part.mediaType.startsWith("image/")` undefended. Any non-text part whose `mediaType` is missing or not a string threw a `TypeError` inside the user's server route handler. The read is now guarded: `mediaType` falls back to `application/octet-stream` when it is not a string, and the guarded value is emitted in both the `image-data` and `file-data` branches so the provider-bound payload always carries a string MIME type.

## Why

This path is reachable from raw wire data, server side. `frontendTools()` installs `defaultToModelOutput` as the AI SDK `toModelOutput` for every frontend tool, and the `output` it receives comes from the client. `unwrapModelContentEnvelope` only checks `Array.isArray` on the envelope key, so individual parts are never validated: a client sending `{ __aui_modelContent: [{ type: "file", data: "x" }] }` crashed the route with an unhandled TypeError. `generativeTools` / `AISDKToolkit` hit the same line via `toAISDKToModelOutput`, both from the wire envelope and from a user's JS `toModelOutput` returning non-spec parts.

This is the one converter that #4648 (guard converters against missing parts and mediaType) deliberately skipped, completing that merged hardening pattern per the AGENTS.md defensive-converter rule.

## Impact

- Before: a malformed part is an unhandled TypeError (500) inside the user's route handler.
- After: well-formed-enough parts convert normally, and junk without usable data degrades to an SDK-managed validation error instead of an unhandled crash.
- Spec-compliant inputs produce byte-identical output; the fallback only engages on inputs that previously threw.
- No public surface change, no export changes. Patch changeset for `@assistant-ui/react-ai-sdk`.
- Tests: three new cases in the colocated suite (missing `mediaType`, non-string `mediaType`, unknown part type), full react-ai-sdk suite and full turbo build green.

## Rationale

- `typeof` instead of `??`: the input is raw client JSON, so `mediaType: 42` is as reachable as `undefined` and throws the same TypeError through a nullish check.
- Fallback is mandatory: the ai v7 tool-output schema requires `mediaType` to be a string, so passing `undefined` through just moves the crash to `standardizePrompt`.
- `application/octet-stream` matches core's outbound fallback (`packages/core/src/adapters/attachment.ts`) and the AI SDK's own; #4648's `unknown/unknown` is inbound display metadata, not a provider-bound MIME type.
- Null entries inside the parts array stay out of scope, mirroring #4648's array-level guard; the array level here is already covered by `isModelContentEnvelope`.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/4899?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/tool-output-mediatype-guard.md
+++ b/.changeset/tool-output-mediatype-guard.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react-ai-sdk": patch
+---
+
+fix: guard toolOutputConversion against missing mediaType on non-spec tool output
--- a/packages/react-ai-sdk/src/toolOutputConversion.test.ts
+++ b/packages/react-ai-sdk/src/toolOutputConversion.test.ts
@@ -76,6 +76,43 @@ describe("toAISDKContent", () => {
   it("returns an empty value array for empty input", () => {
     expect(toAISDKContent([])).toEqual({ type: "content", value: [] });
   });
+
+  it("falls back to application/octet-stream when mediaType is missing", () => {
+    const parts = [{ type: "file", data: "DDDD" } as any];
+    expect(toAISDKContent(parts)).toEqual({
+      type: "content",
+      value: [
+        {
+          type: "file-data",
+          data: "DDDD",
+          mediaType: "application/octet-stream",
+        },
+      ],
+    });
+  });
+
+  it("falls back to application/octet-stream when mediaType is not a string", () => {
+    const parts = [{ type: "file", data: "EEEE", mediaType: 42 } as any];
+    expect(toAISDKContent(parts)).toEqual({
+      type: "content",
+      value: [
+        {
+          type: "file-data",
+          data: "EEEE",
+          mediaType: "application/octet-stream",
+        },
+      ],
+    });
+  });
+
+  it("does not throw on a part with an unknown type and no mediaType", () => {
+    const parts = [{ type: "widget" } as any];
+    const result = toAISDKContent(parts);
+    expect(result.value[0]).toMatchObject({
+      type: "file-data",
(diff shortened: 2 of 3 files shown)


okisdev (approved):
## review

LGTM as `patch`.

- the unknown-type fallback emits a `file-data` part whose `data` is `undefined` when the junk part carries no data (`toolOutputConversion.ts`, non-text branch); JSON serialization drops the key so providers tolerate it, but it ships a near-empty part where dropping the part would be cleaner
- the unknown-type test pins the fallback with `toMatchObject`, which glosses over that `data: undefined` field instead of asserting the exact emitted shape

### what we modified

- nothing; the original commit ships as submitted (the guard, both branch emissions, the three tests, and the patch changeset all come from the author's single commit)

### potential further fixes

- drop non-text parts that carry no string `data` instead of emitting a placeholder `file-data` part; deferred as out of scope since this PR only removes the crash and the current behavior is not observably wrong through JSON serialization
- `convertMessage.ts:422` guards the same field with `?? "unknown/unknown"`, which misses non-string values and uses a nonstandard MIME placeholder; aligning it to this PR's `typeof` check plus `application/octet-stream` is a small follow-up chore

PR #4906, opened 2026-07-14
feat(react-ink): export attachment Props aliases from the composer and message barrels

ShobhitPatra:
## What

Adds the four missing flat Props aliases to the `composer.ts` and `message.ts` barrels: `AttachmentsProps` and `AttachmentByIndexProps` on both `ComposerPrimitive` and `MessagePrimitive`. Before aliasing, re-points the local flat types at core's namespace Props so the published aliases are accurate, and deletes the local `AttachmentComponents` type that becomes dead as a result.

## Why

Direct sibling of #4888. These four components were the last in the package whose props could not be named at all: their `const` re-exports from `@assistant-ui/core/react` drop the namespace merge (so `.Props` is unreachable) and the barrels never exported the flat types. The existing flat types could not be aliased verbatim: they predate core's `Attachments` props becoming a components/children union, so they miss the `children` render-function arm and admit `components: undefined`, which the real component rejects. Deriving from core's Props fixes that, matching the existing pattern in `message/MessageParts.tsx`.

## Impact

- Type-only change, no runtime output change; the component consts are untouched.
- Append-only: barrels only gain exports. The reshaped flat types and deleted `AttachmentComponents` were never reachable from the published entries.
- `api-surface/assistant-ui__react-ink.ts` regenerated, `api-surface:check` green.
- Patch changeset for `@assistant-ui/react-ink`.
- Gates on raw exit codes: full `turbo run build`, react-ink vitest (231 passed), `pnpm lint`.

## Rationale

Rewriting the types as accurate local unions would recreate the hand-mirror drift this fixes, and skipping them (as #4888 did for `InProgress`/`Messages`) does not apply since those had reachable namespace `.Props` while these did not. `AttachmentComponents` is deleted rather than exported because core keeps its equivalent config type module-private; consumers can reach it as `NonNullable<AttachmentByIndexProps["components"]>`.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/4906?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/ink-attachment-props-aliases.md
+++ b/.changeset/ink-attachment-props-aliases.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react-ink": patch
+---
+
+feat: export attachment Props aliases from the composer and message barrels
--- a/api-surface/assistant-ui__react-ink.ts
+++ b/api-surface/assistant-ui__react-ink.ts
@@ -915,6 +915,8 @@ type ComposerAddAttachmentProps = Omit<PressableProps, "onPress"> & {
 
 declare const ComposerAttachmentByIndex: import("react").FC<ComposerPrimitiveAttachmentByIndex.Props>;
 
+type ComposerAttachmentByIndexProps = ComposerPrimitiveAttachmentByIndex.Props;
+
 declare const ComposerAttachmentByIndexProvider: FC<PropsWithChildren<{
   index: number;
 }>>;
@@ -933,6 +935,8 @@ type ComposerAttachmentsComponentConfig = {
   File?: ComponentType | undefined;
   Attachment?: ComponentType | undefined;
 };
+
+type ComposerAttachmentsProps = ComposerPrimitiveAttachments.Props;
 
 declare const ComposerCancel: (_param17: ComposerCancelProps) => import("react").JSX.Element;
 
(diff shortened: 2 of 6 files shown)


okisdev (approved):
## review

LGTM as `patch`. verified the gap and the fix shape: the `export const` re-bindings in the ink attachment files drop the TypeScript namespace merge, so `.Props` is unreachable and the barrels never exported the flat types; the web package never had this gap because it re-exports the core primitive directly with the merge intact. the four new aliases complete the barrel's existing flat-Props enumeration (AddAttachmentProps, CancelProps, InputProps and friends were already there), and re-pointing the local flat types at core's namespace Props follows the in-package precedent while fixing real drift: the old hand-written shapes predate core's components/children union, miss the children arm, and admit `components: undefined`.

- none in the diff; the deleted `AttachmentComponents` type was never published, so the api-surface change is purely additive.

### what we modified

- nothing.

### potential further fixes

- none; #4907 is the react-native mirror (verified diff-equivalent modulo package names) and merges alongside.

PR #4907, opened 2026-07-14
feat(react-native): export attachment Props aliases from the composerand message barrels

ShobhitPatra:
## What

react-native sibling of #4906. Adds the four missing flat Props aliases to the `composer.ts` and `message.ts` barrels: `AttachmentsProps` and `AttachmentByIndexProps` on both `ComposerPrimitive` and `MessagePrimitive`. Before aliasing, re-points the local flat types at core's namespace Props so the published aliases are accurate, and deletes the local `AttachmentComponents` type that becomes dead as a result.

## Why

The two attachment files were byte-identical to react-ink's pre-#4906 files, so the same gap applies: the `const` re-exports from `@assistant-ui/core/react` drop the namespace merge (so `.Props` is unreachable) and the barrels never exported the flat types, leaving these props unnameable. The existing flat types could not be aliased verbatim: they predate core's `Attachments` props becoming a components/children union, so they miss the `children` render-function arm and admit `components: undefined`, which the real component rejects. Deriving from core's Props fixes that, matching the existing in-package pattern in `message/MessageParts.tsx`. Cross-runtime parity: keeping the runtimes' type surface aligned is the point of shipping this as a mirror of #4906.

## Impact

- Type-only change, no runtime output change; the component consts are untouched.
- Append-only: barrels only gain exports. The reshaped flat types and deleted `AttachmentComponents` were never reachable from the published entries.
- `api-surface/assistant-ui__react-native.ts` regenerated, `api-surface:check` green.
- Patch changeset for `@assistant-ui/react-native`.
- Gates on raw exit codes: full `turbo run build`, react-native vitest (147 passed), `pnpm lint`.

## Rationale

Same calls as #4906: rewriting the types as accurate local unions would recreate the hand-mirror drift this fixes, and `AttachmentComponents` is deleted rather than exported because core keeps its equivalent config type module-private; consumers can reach it as `NonNullable<AttachmentByIndexProps["components"]>`.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/4907?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/native-attachment-props-aliases.md
+++ b/.changeset/native-attachment-props-aliases.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react-native": patch
+---
+
+feat: export attachment Props aliases from the composer and message barrels
--- a/api-surface/assistant-ui__react-native.ts
+++ b/api-surface/assistant-ui__react-native.ts
@@ -868,6 +868,8 @@ type ComposerAddAttachmentProps = Omit<PressableProps, "onPress"> & {
 
 declare const ComposerAttachmentByIndex: import("react").FC<ComposerPrimitiveAttachmentByIndex.Props>;
 
+type ComposerAttachmentByIndexProps = ComposerPrimitiveAttachmentByIndex.Props;
+
 declare const ComposerAttachmentByIndexProvider: FC<PropsWithChildren<{
   index: number;
 }>>;
@@ -886,6 +888,8 @@ type ComposerAttachmentsComponentConfig = {
   File?: ComponentType | undefined;
   Attachment?: ComponentType | undefined;
 };
+
+type ComposerAttachmentsProps = ComposerPrimitiveAttachments.Props;
 
 declare const ComposerCancel: (_param14: ComposerCancelProps) => import("react").JSX.Element;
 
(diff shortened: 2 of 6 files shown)


okisdev (approved):
## review

LGTM as `patch`. react-native mirror of #4906, verified diff-equivalent modulo package names (the only deltas are each barrel's pre-existing member lists). the premise holds identically: the attachment files were byte-identical to ink's before the pair, the `export const` re-bindings drop the namespace merge so `.Props` is unreachable, and the barrels never exported the flat types. deriving the flat types from core's namespace Props follows the in-package `MessageParts.tsx` precedent and fixes the drifted hand-written shapes (missing children arm, `components: undefined` admitted), and the four new barrel aliases complete the existing flat-Props enumeration.

- none in the diff; the deleted `AttachmentComponents` type was never published, api-surface change is purely additive.

### what we modified

- nothing.

### potential further fixes

- none; with #4906 merged this closes the props-nameability gap across all three distributions (the web package never had it, its re-export keeps the namespace merge intact).

PR #4909, opened 2026-07-14
docs: document useChatRuntime runtime options in the ai-sdk v7 guide

ShobhitPatra:
## What

Adds a "Runtime options" section to the AI SDK v7 guide covering `onThreadIdChange`, `joinStrategy`, and `onResume`, and extends the "Advanced: useAISDKRuntime" section with `cancelPendingToolCallsOnSend`.

## Why

These options ship on `useChatRuntime` / `useAISDKRuntime` (`onThreadIdChange` since #4515) but appear nowhere on the AI SDK guide. `joinStrategy` and `onResume` are documented only on the generic external store page, and `onThreadIdChange` is not documented anywhere in guide content, so today they are discoverable only by reading source. AGENTS.md requires docs to stay in lockstep with shipped hooks.

## Impact

- Docs only, single page touched: `apps/docs/content/docs/runtimes/ai-sdk/v7.mdx`
- No code changes, no published package touched, so no changeset (apps/docs is private)
- No change to the generated api-reference tree

## Rationale

Each option is documented where it actually exists. `cancelPendingToolCallsOnSend` is deliberately placed under the `useAISDKRuntime` section rather than the new `useChatRuntime` options list because `UseChatRuntimeOptions` does not accept it; it only exists on `useAISDKRuntime`'s options and `useChatRuntime` never forwards it.

Wording is verified against source rather than copied from the backlog claim:

- `onThreadIdChange` emits only the settled thread ID and `undefined` while a fresh thread is uninitialized (core `remote-thread-list/types.ts`); it works without `cloud` because the in-memory adapter settles `remoteId = threadId`, so the example needs no cloud setup
- `onResume` documents the `resumeRun` bridge and the exact `"Runtime does not support resuming runs."` throw when omitted (`external-store-thread-runtime-core.ts`), and points readers to the resumable streams guide for the built-in transport-level reload path, which does not go through `onResume`
- `joinStrategy` mirrors the external store page's semantics and cross-links it

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/4909?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/docs/content/docs/runtimes/ai-sdk/v7.mdx
+++ b/apps/docs/content/docs/runtimes/ai-sdk/v7.mdx
@@ -595,6 +595,56 @@ const runtime = useChatRuntime({
 
 If you need a transport that does not inherit from `AssistantChatTransport`, you forfeit automatic system-message and frontend-tool forwarding; the transport is then a regular AI SDK `ChatTransport` and you control everything explicitly.
 
+## Runtime options
+
+Beyond transports and adapters, `useChatRuntime` accepts options that control thread routing, message grouping, and run resumption.
+
+### onThreadIdChange
+
+Called whenever the active thread's settled ID changes, e.g. to sync the active thread to a URL query param:
+
+```tsx
+const runtime = useChatRuntime({
+  onThreadIdChange: (threadId) => {
+    const url = new URL(window.location.href);
+    if (threadId) {
+      url.searchParams.set("thread", threadId);
+    } else {
+      url.searchParams.delete("thread");
+    }
+    window.history.replaceState(null, "", url);
+  },
+});
+```
+
+Only the settled ID is emitted: while a freshly created thread has not been initialized yet, the value is `undefined`, and the real ID is emitted once the thread initializes (with [AssistantCloud](/docs/cloud/ai-sdk-assistant-ui), this is the cloud thread ID). The transient local ID is never surfaced.
+
+### joinStrategy
+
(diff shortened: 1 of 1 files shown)


Kinfe123 (approved):

okisdev (approved):
## review

LGTM. verified every documented claim against source: all four options exist where the guide says they do (`onThreadIdChange` at useChatRuntime.ts:41 via #4515, `joinStrategy` and `onResume` on both hooks, `cancelPendingToolCallsOnSend` on `useAISDKRuntime` only, matching the placement rationale in the PR body), the settled-id paragraph matches the core comment on `RemoteThreadListThreadListRuntimeCore` nearly word for word, the `"Runtime does not support resuming runs."` string is exact, the `@default true` and `"concat-content"` defaults are exact, and all three link targets (including the external-store join-strategy anchor) resolve.

- none material; the diff shipped accurate on the first commit.

### what we modified

- nothing; single-commit PR, no follow-up commits needed.

### potential further fixes

- `onResumeToolCall` (useChatRuntime.ts:32) is documented nowhere on the site (resumable-streams covers `onResumeError` only); worth a small follow-up docs PR.

PR #4918, opened 2026-07-14
fix(ui): initialize Radix sidebar skeleton width as state

### LINKED ISSUE #4917: Radix sidebar skeleton uses useMemo as a random-value initializer
## Reproduction

Render SidebarMenuSkeleton from the Radix UI tree and preserve the component across rerenders. Its CSS width is produced by Math.random inside useMemo with an empty dependency list. Recomputing or discarding that memo changes the visible skeleton width without a remount. Server rendering and hydration also evaluate independent random values.

The Base UI sibling already uses the correct lazy useState initializer.

## Expected

A skeleton chooses its width once per mount and keeps that width until unmount.

## Minimal fix

Match the Base UI implementation and initialize width with useState(() => ...).

samdickson22:
Fixes #4917

## What

Use a lazy useState initializer for the Radix SidebarMenuSkeleton random width, matching the Base UI sibling.

## Why

The width is mount state, not a derived calculation. useMemo is discardable, so recalculation can visibly change the skeleton width without a remount. Server and client renders can also independently generate different random values.

## Scope

One component and one hook substitution. No API, styling range, or registry behavior changes. The package is private, so no changeset is required.

## Verification

- pnpm exec oxlint packages/ui/src/components/ui/radix/sidebar.tsx
- pnpm exec oxfmt --check packages/ui/src/components/ui/radix/sidebar.tsx
- git diff --check

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/4918?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/ui/src/components/ui/radix/sidebar.tsx
+++ b/packages/ui/src/components/ui/radix/sidebar.tsx
@@ -605,10 +605,9 @@ function SidebarMenuSkeleton({
 }: React.ComponentProps<"div"> & {
   showIcon?: boolean;
 }) {
-  // Random width between 50 to 90%.
-  const width = React.useMemo(() => {
+  const [width] = React.useState(() => {
     return `${Math.floor(Math.random() * 40) + 50}%`;
-  }, []);
+  });
 
   return (
     <div


okisdev (approved):
## review

LGTM. verified as a parity fix: the Base UI sibling already uses the lazy `useState` initializer (base/sidebar.tsx:598), the Radix flavor kept upstream shadcn's `useMemo(() => random, [])`, and the two trees are meant to stay in lockstep. `useMemo` is a discardable performance hint, so a random per-mount value belongs in state; the diff matches the base implementation byte for byte.

- none; single-commit two-line fix, accurate description, correctly no changeset (packages/ui is private).

### what we modified

- nothing.

### potential further fixes

- the SSR/hydration mismatch mentioned in #4917 is orthogonal and exists in both flavors (server and client each roll their own random width); if it ever matters, both siblings need a deterministic width strategy, not an initializer change.

PR #4919, opened 2026-07-14
fix(tap): keep root lifecycle state out of useMemo

### LINKED ISSUE #4915: tap root lifetime state depends on useMemo retention
## Reproduction

On @assistant-ui/tap 0.9.3, mount a React component that hosts useTapRoot, subscribe to the returned root, and queue a resource update. The scheduler, pending queue, root fiber, and subscriber set are allocated independently with React useMemo. If React discards one of those memo caches without unmounting the component, the root can lose the queued update or subscriber, or split its state machine across old and new instances.

React documents useMemo as a performance optimization rather than lifetime storage. The root's mutable lifecycle state should use React state or refs.

## Expected

The scheduler, queue, fiber, and subscribers remain attached to the mounted root until it unmounts, independent of memo-cache retention.

## Minimal fix

Allocate each lifetime object with a lazy useState initializer. Keep useMemo only for the disposable public facade.

samdickson22:
Fixes #4915

## What

Store the tap root scheduler, pending queue, resource fiber, and subscriber set in lazy React state instead of useMemo. The public root facade remains memoized because its identity is only an optimization.

## Why

These objects form one mutable state machine and must survive for the entire mounted component lifetime. React treats useMemo as a discardable optimization, so correctness cannot rely on those memo cells retaining their allocations.

## Scope

This only changes allocation ownership. Scheduling, commit behavior, subscriptions, Strict Mode handling, and the public API are unchanged.

## Verification

- pnpm --filter @assistant-ui/tap exec vitest run src/__tests__/react/useResource.test.tsx src/__tests__/react/concurrent-pending-updates.test.tsx src/__tests__/react/concurrent-render-phase.test.tsx
- pnpm --filter @assistant-ui/tap build
- pnpm exec oxlint packages/tap/src/hooks/useTapRoot.ts
- pnpm exec oxfmt --check packages/tap/src/hooks/useTapRoot.ts .changeset/tap-root-lifetime-state.md
- pnpm changeset status --since upstream/main
- git diff --check

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/4919?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/tap-root-lifetime-state.md
+++ b/.changeset/tap-root-lifetime-state.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/tap": patch
+---
+
+fix: keep tap root scheduler, fiber, queue, and subscribers in React state for the full component lifetime
--- a/packages/tap/src/hooks/useTapRoot.ts
+++ b/packages/tap/src/hooks/useTapRoot.ts
@@ -12,7 +12,7 @@ import {
   setRootVersion,
 } from "../core/helpers/root";
 import { cloneCurrentTapContext, withTapContextRoot } from "../core/context";
-import { useEffect, useEffectEvent, useMemo, useRef } from "react";
+import { useEffect, useEffectEvent, useMemo, useRef, useState } from "react";
 import { useDevStrictMode } from "./utils/useDevStrictMode";
 
 export namespace useTapRoot {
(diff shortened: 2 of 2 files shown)


okisdev (approved):
## review

LGTM as `patch`. verified the premise independently on main: `fiber` lives in a `useMemo` whose discard would flow through the `[fiber]`-keyed unmount effect and replace the root mid-lifetime, losing all resource state without a component remount, and a discarded `subscribers` memo splits notifications between the old and new set. moving the scheduler, queue, fiber, and subscriber set to lazy `useState` makes them genuine lifetime storage; the facade staying memoized is correct since its identity is the only thing that matters and it closes over state-held storage.

allocation is inert (side effects live in render/commit/effects), so StrictMode double-invoked initializers discard garbage-collectable twins, the same exposure as the previous memos. verified by execution: tap 447/447 on this head, and on the combined state with #4920: tap 447/447 across repeated runs, `check:resource-memo` clean on all 18 resource-hosting files, store 10/10, core 525/525.

- none in the diff; the change is the minimal ownership move with the oxlint suppression scoped to the one deferred effect-event reference.

### what we modified

- nothing.

### potential further fixes

- none; #4920 covers the same class for the keyed-resource layer and merges next.

PR #4920, opened 2026-07-14
fix(tap): preserve hosted resource fiber state

### LINKED ISSUE #4916: React-hosted tap fiber registries depend on useMemo retention
## Reproduction

On @assistant-ui/tap 0.9.3, render keyed resources through useResources and update one child. The keyed Map and the React-hosted resource root are mutable lifecycle stores allocated with React useMemo. If React discards either cache while the component remains mounted, existing fibers, dirty flags, committed values, and cleanup ownership are abandoned.

React documents useMemo as a performance optimization rather than lifetime storage.

## Expected

The host root and keyed child registry live for the full mounted component lifetime. Intentional key or hook changes should continue to use the existing explicit remount path.

## Minimal fix

Use lazy useState initializers for the host root and keyed fiber Map without changing keyed remount behavior.

samdickson22:
Fixes #4916

## What

Move the React-hosted tap root and the keyed useResources fiber registry from useMemo into lazy React state.

## Why

Both allocations hold correctness-bearing mutable lifecycle data: root versions, child fibers, dirty flags, committed values, and cleanup ownership. React may discard memoized values, while state is retained for the mounted component lifetime.

## Scope

The explicit keyed remount path is unchanged. This PR only changes ownership of the host root and registry allocations.

## Verification

- pnpm --filter @assistant-ui/tap exec vitest run src/__tests__/react/useResource.test.tsx src/__tests__/react/concurrent-pending-updates.test.tsx src/__tests__/react/concurrent-render-phase.test.tsx
- pnpm --filter @assistant-ui/tap build
- pnpm exec oxlint packages/tap/src/hooks/useResources.ts packages/tap/src/hooks/utils/useResourceFiberHostUtils.ts
- pnpm exec oxfmt --check packages/tap/src/hooks/useResources.ts packages/tap/src/hooks/utils/useResourceFiberHostUtils.ts .changeset/tap-resource-fiber-lifetime.md
- pnpm changeset status --since upstream/main
- git diff --check

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/4920?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/tap-resource-fiber-lifetime.md
+++ b/.changeset/tap-resource-fiber-lifetime.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/tap": patch
+---
+
+fix: preserve React-hosted tap resource roots and keyed fiber registries for the component lifetime
--- a/packages/tap/src/hooks/useResources.ts
+++ b/packages/tap/src/hooks/useResources.ts
@@ -14,7 +14,7 @@ import {
   hasContextDepsChanged,
 } from "../core/context";
 import { useResourceFiberHost } from "./utils/useResourceFiberHostUtils";
-import { useEffect, useMemo } from "react";
+import { useEffect, useState } from "react";
 import { useRenderMemo } from "./utils/useRenderMemo";
 import { depsShallowEqual } from "./utils/depsShallowEqual";
 
@@ -73,7 +73,7 @@ const hasAnyChildContextDepsChanged = (
 export function useResources<E extends ResourceElement<any, any[]>>(
   elements: readonly E[],
 ): ExtractResourceReturnType<E>[] {
-  const fibers = useMemo(() => new Map<string | number, FiberState>(), []);
+  const [fibers] = useState(() => new Map<string | number, FiberState>());
 
   // Process each element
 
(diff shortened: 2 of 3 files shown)


okisdev (approved):
## review

LGTM as `patch`. same defect class as #4919, applied to the keyed-resource layer: the `useResources` keyed fiber Map and the React-hosted resource root in `useResourceFiberHostUtils` are mutable lifetime stores that lived in `useMemo`; a discarded cache abandons existing fibers, dirty flags, committed values, and cleanup ownership while the component stays mounted. lazy `useState` makes them lifetime storage, and the intentional keyed remount path is untouched.

verified by execution on the combined state with #4919 (which merged first): tap 447/447 across repeated runs, `check:resource-memo` clean on all 18 resource-hosting files, store 10/10, core 525/525, dependency builds green. allocation is inert, so StrictMode initializer double-invocation has the same exposure as the previous memos.

- none in the diff; minimal ownership move, the `useReducer` dispatch reference in the root callback stays TDZ-deferred exactly as before.

### what we modified

- nothing.

### potential further fixes

- none; with #4919 merged this closes out the useMemo-as-lifetime-storage class that #4917 first surfaced in the UI kit.

PR #4947, opened 2026-07-15
fix(react-ai-sdk): wait for history before stream resume

### LINKED ISSUE #4944: Race condition: useExternalHistory's setMessages can silently clobber a resumed stream's recovered reply (unstable_useRemoteThreadListRuntime + resumable)
## Summary

When combining `unstable_useRemoteThreadListRuntime` (with a `ThreadHistoryAdapter`, driving `useExternalHistory`) with `AssistantChatTransport`'s `resumable` option, two independent, unawaited mount-time effects both write to the same AI SDK `Chat` instance's `state.messages`, with no coordination or merge logic between them. Depending on which one resolves last, either the resumed stream's recovered reply is silently dropped, or the resume's append is built on a stale snapshot that gets immediately clobbered by the history load.

## Root cause

1. `useExternalHistory` (`packages/react-ai-sdk/src/ui/use-chat/useExternalHistory.ts:82-120`) fires a `useEffect` on mount (once, guarded by `loadedRef`) that calls the async `ThreadHistoryAdapter.load()` (e.g. localStorage + a server fetch). When it resolves, it calls `runtimeRef.current.thread.import(converted)` and `onSetMessagesRef.current(...)`, which `useAISDKRuntime` wires to `chatHelpers.setMessages(messages)` (`useAISDKRuntime.ts:171-173`) — a **wholesale replace** of `state.messages`, not a merge.
2. `useChatThreadRuntime` (`useChatRuntime.ts:118-133`) has a *separate* `useEffect`, also firing on mount, that checks for a pending resumable stream id and calls `chat.resumeStream()` fire-and-forget.
3. AI SDK core's `resumeStream()` (in the `ai` package's `Chat` class) snapshots `this.lastMessage` (i.e. `state.messages` at invocation time) and appends incoming chunks onto that snapshot, mutating `state` in place as they arrive.
4. Both effects are async and unawaited, with no dependency on each other's completion — whichever finishes last wins:
   - If history load's `setMessages` resolves *after* resume has already appended the recovered reply → the reply is silently wiped (state reverts to whatever persisted history looked like before the reply was saved by the history adapter's own post-completion debounce-save).
   - If resume snapshots `lastMessage` *before* history load's replace lands → resume's append is built on a stale array that history load's later `setMessages` immediately overwrites anyway.

Net effect: after an interrupted stream + reload, the recovered reply is lost roughly 50% of the time depending on relative timing of the two async operations, with no error surfaced anywhere.

## Repro scenario

- Use `unstable_useRemoteThreadListRuntime` with a custom `ThreadHistoryAdapter` whose `load()` is async (e.g. localStorage read + a server round-trip).
- Enable `resumable` on `AssistantChatTransport`.
- Start a generation, interrupt the client mid-stream (e.g. reload the tab) before the reply has been persisted by the history adapter's own post-completion save hook.
- Reload the page. `useExternalHistory`'s history load and `useChatThreadRuntime`'s `resumeStream()` both fire concurrently on mount with no ordering guarantee, and race to write `state.messages` last.

## Versions checked

- `@assistant-ui/react-ai-sdk`: reproduced against 1.3.37; checked the changelog through the latest 1.3.40 — no relevant fix in that range.
- `assistant-stream`: 0.3.23 → 0.3.25, no relevant fix either.
- `@assistant-ui/react`: 0.14.23 → 0.14.26, no relevant fix either.

## Suggested direction

No coordination point is currently exposed between these two systems: `useAISDKRuntime` doesn't surface its history-loading state to `useChatThreadRuntime`, so the resume trigger has no way to gate on it. A fix likely needs one of:

- Delaying `resumeStream()` until the history adapter's `load()` promise has resolved and been applied, or
- Re-deriving the resume's base snapshot from post-history-load state instead of the snapshot taken at effect-mount time, or
- Some explicit "history loaded" signal that `useChatThreadRuntime` can await before firing resume.

Given this, we've left `resumable` disabled for now and fall back to a disk-backed poll + full-page-reload recovery path instead. Happy to provide more repro detail (or a minimal reproduction repo) if that's useful.

Gujiassh:
## Summary

Fixes #4944.

- expose external history loading on the initial render instead of one effect later
- defer automatic resumable stream recovery until history loading has settled
- cover both the asynchronous history state and resume ordering with regression tests

## Validation

- `pnpm --filter @assistant-ui/react-ai-sdk test` (158 tests)
- target `oxlint` and `oxfmt --check`
- `pnpm --filter @assistant-ui/react-ai-sdk build`
- `git diff --check`

The standalone package type check still reports the existing `TS2769` in `src/generativeTools.test.ts:526`; the package declaration build passes and the changed files produce no additional diagnostics.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/4947?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


okisdev (approved):
## review

LGTM as `patch`. the race is proven end to end: the counterfactual ordering test fails on main (resume fires during the history load window), the clobber is mechanical since `onSetMessages` is `chatHelpers.setMessages` doing wholesale replacement (`useAISDKRuntime.ts:268`), and the triggering combination (remote thread list + history adapter + resumable transport) is the promoted cloud resume story. the fix is the right shape on both halves: `isLoading` becomes a synchronous derivation (`formatAdapter != null && !hasLoaded`) so the gate is true from the first render instead of one effect late, and the resume effect consumes it through `useSyncExternalStore` over the runtime's own subscribe, re-firing via the dep array with the pre-existing `resumeFiredRef` latch. no new public surface; the hook-to-runtime `isLoading` wiring predates this PR. verified on the merged state: five use-chat suites pass 65/65 with the parameterized mock coexisting with the #5087 persistence harness and #5179 retry pins, and the failure and no-adapter paths both release or bypass the gate.

- none surviving; the branch predated the #5087 rewrite, and the only merge conflict was the test-file mock, resolved as a union that keeps the parameterized `remoteId` shape.

### what we modified

- the branch-update merge resolves the test-file collision with the #5087/#5179-era tests; autofix then normalized the import formatting.

### potential further fixes

- none. resume now strictly follows history settlement in both success and failure orderings.

PR #4954, opened 2026-07-15
fix: stop isJSONArray from consuming the array index as recursion depth

ShobhitPatra:
## What

`is-json.ts` passed `.every`'s array index as `isJSONValue`'s `currentDepth` (`value.every(isJSONValue)`), so `isJSONArray(Array(102).fill(1))` returned false (element 101 trips the `> 100` depth guard) and each element's nesting budget shrank by its index. Fixed to `value.every((item) => isJSONValue(item))`, matching `isJSONObject`'s per-value semantics. The byte-identical copy in `packages/react/src/utils/json/is-json.ts` gets the same fix so the copies stay byte-equal.

Also adds the missing colocated tests for `is-json.ts` and `is-json-equal.ts` (writing them is what surfaced the bug), and inlines the `hasOwn` alias in `stable-stringify-tool-args.ts`, kept in #4901 only to stay verbatim with the since-deleted adapter copies.

## Why

These modules are load-bearing but untested: `isJSONValue` gates cloud `auiV0` persistence, `isJSONValueEqual` backs `useJSONEqual`, and the react copy feeds the mcp-apps bridge.

## Impact

- `isJSONArray` has no in-repo consumers and the module is on no exports map, so no user-facing behavior changes today; the fix removes a footgun for the first consumer.
- No public surface change; the `hasOwn` inline is behaviorally zero-diff (single call site).

## Notes

- Regression verified both ways: the new `isJSONArray` test fails on the unfixed code, passes with the fix.
- Tests pin surprising-but-current semantics as characterization, not changes: `new Date()` / `new Map()` / `new Set()` pass `isJSONValue` (records with no own enumerable string keys). If that is ever deemed a bug, it is a separate behavior-changing issue.
- The internal `MAX_JSON_DEPTH` guard in `isJSONValueEqualAtDepth` is intentionally uncovered: unreachable through the public API (the `isJSONValue` gate rejects >100-deep input first); kept as defense in depth.
- React copy gets the fix but no duplicated test files; core tests cover the shared logic.
- Single changeset, `@assistant-ui/core` + `@assistant-ui/react`, both patch.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/4954?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved JSON array validation for deeply nested data.
  * Ensured array positions no longer affect validation depth limits.
  * Improved consistency when comparing JSON values, including nested structures and special values.
  * Preserved stable argument serialization while improving property handling.

* **Tests**
  * Added comprehensive coverage for JSON validation and deep equality, including arrays, objects, recursion limits, and edge cases.


okisdev (approved):
## review

LGTM as `patch`.

the bug is real and precisely characterized: `value.every(isJSONValue)` feeds the array index into `currentDepth`, so a valid 102-element array is rejected (index 101 trips the depth guard) and every element's nesting budget shrinks by its position. reproduced both failure modes directly and confirmed the fix restores them, with the boundary math in the tests checking out exactly (depth 100 accepted, 101 rejected, and the `a === b` shortcut in `isJSONValueEqual` correctly gated by the upfront validation). the new colocated tests are the bigger payload: `isJSONValue` gates cloud `auiV0` persistence and `isJSONValueEqual` backs `useJSONEqual`, and both were previously untested; the surprising cases (`new Date()` / `new Map()` passing `isJSONValue`) are pinned honestly as characterization. the `hasOwn` inline is a fair ride-along given #4901's verbatim-copy rationale dissolved with the adapter copies.

### what we modified
- merged main into the branch, resolving the modify/delete conflict by accepting #4971's deletion of the react copy; core now hosts the single implementation, so the fix reaches react via `@assistant-ui/core/internal`
- scoped the changeset to `@assistant-ui/core` only; react no longer carries the code and its dependency floor moves automatically at release time

### potential further fixes
- `isJSONArray`/`isJSONObject` still have no in-repo consumers. the tests now pin them, so keeping them beside the exported `ReadonlyJSONArray`/`ReadonlyJSONObject` types is reasonable; pruning them stays an option if the unused surface bothers anyone later.

PR #4955, opened 2026-07-15
refactor: remove dead is-json copy from assistant-stream

ShobhitPatra:
## What

Deletes `packages/assistant-stream/src/utils/json/is-json.ts`, an unreferenced third copy of the `is-json` guards that diverges from the core/react implementation in two risky ways: it accepts `NaN`/`Infinity` (bare `typeof === "number"` check) and has no recursion-depth guard, so deeply nested input throws `RangeError` (both reproduced).

## Why

The file is dead code: nothing in the package imports its functions and neither entry barrel (`index.ts`, `utils.ts`) re-exports it, so the divergent behavior is unreachable today. Deleting beats aligning it with core: there is no consumer to align for, and a fixed-but-unused copy remains a footgun inviting the next contributor to import the wrong module. Surfaced by the verify pass on #4954, which fixed the core/react copies' depth bug.

## Impact

- No exported name or subpath is removed. The exports map is closed (no wildcard entries), so the per-module dist files this drops (`dist/utils/json/is-json.*`) were never importable; every other dist file is byte-identical before and after (verified via sha256 of both builds).
- Sibling `json-value.ts` (type-only, still imported) and the parse utilities stay untouched.

## Notes

- Zero in-repo references confirmed by grep across the package and the monorepo.
- Changeset: `assistant-stream` patch.
- Full `pnpm turbo run build` green.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/4955?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


okisdev (approved):
## review

LGTM as `patch`.

verified the deletion is safe three ways: grep finds zero references in the package and across the monorepo (every `utils/json/is-json` import resolves to core's or react's own copy), the entry barrels and exports map never mentioned the file at any point in git history (`git log -S "is-json"` on `index.ts`, `utils.ts`, and `package.json` is empty), and the closed exports map means the compiled `dist/utils/json/is-json.*` was never resolvable from outside the package. so the file was born dead in #1459 and no shipped export is removed. the divergence rationale also checks out: the live core and react copies carry the `currentDepth` guard and reject `NaN`/`Infinity`, while this copy had neither, and a fixed but unused copy is exactly the module the next contributor imports by mistake. type consumers are unaffected since `json-value.ts` stays exported through `assistant-stream/utils`.

### potential further fixes
- core and react still maintain duplicate `is-json.ts` copies. consolidation is a separate cross-package decision and this repo extracts shared copies only with real consumers and drift evidence, so leaving them in place is correct for this PR.

PR #4958, opened 2026-07-15
fix(core): preserve optimistic update order

Kinfe123:
## Summary

- preserve overlapping optimistic updates in invocation order rather than request completion order
- keep completed optimistic effects active in the correct order while another request is pending
- retain completed mutations when a stale list response reconciles thread state
- add regression coverage for request completion and list/mutation ordering

## Bug

While exercising rapid consecutive thread renames, I reproduced this sequence:

1. Rename `Untitled` to `Project Alpha`.
2. Before that request settles, rename it again to `Project Beta`.
3. Let the `Alpha` request resolve, followed by the `Beta` request.

The final local title incorrectly became `Project Alpha`. Completed optimistic callbacks were stored and replayed in request completion order, so the older `Alpha` callback was reapplied after the newer `Beta` update settled. The final UI state therefore depended on network timing instead of the order of the user actions.

## Fix

Each optimistic transform now receives a monotonic invocation order before its request starts. Pending and completed optimistic effects are applied in that order. Optimistic-only updates replay already-completed effects that were invoked after them, while `then`-based reconciliation continues to replay every completed mutation so a stale `list()` response cannot erase a delete or archive regardless of which operation started first.

## Verification

- confirmed the new regression test fails on `main` with `Expected: Project Beta`, `Received: Project Alpha`
- `pnpm --filter @assistant-ui/core test` (58 files, 541 tests)
- `pnpm lint`
- `pnpm exec changeset status --since origin/main`

### DIFF
--- a/.changeset/optimistic-update-order.md
+++ b/.changeset/optimistic-update-order.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/core": patch
+---
+
+fix: preserve optimistic updates in invocation order
--- a/api-surface/assistant-ui__core.ts
+++ b/api-surface/assistant-ui__core.ts
@@ -2894,6 +2894,7 @@ type OnSchemaValidationErrorFunction<TResult> = ToolExecuteFunction<unknown, TRe
 declare class OptimisticState<TState> extends BaseSubscribable {
   private readonly _pendingTransforms;
   private readonly _completedOptimistics;
+  private _nextTransformOrder;
   private _baseValue;
   private _cachedValue;
   constructor(initialState: TState);
--- a/packages/core/src/runtimes/remote-thread-list/optimistic-state.test.ts
+++ b/packages/core/src/runtimes/remote-thread-list/optimistic-state.test.ts
@@ -1,0 +1,63 @@
+import { describe, expect, it } from "vitest";
+import { OptimisticState } from "./optimistic-state";
+
+const deferred = () => {
+  let resolve!: () => void;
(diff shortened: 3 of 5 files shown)


okisdev (approved):
## review

lgtm as `patch`. verified both directions at runtime: at the head the two target suites pass 7/7 and the full core suite passes 541/541, and with `optimistic-state.ts` reverted to main exactly the two invocation-order tests fail, reproducing the alpha-over-beta bug in both settle orders.

the ordering design survives an adversarial pass: the monotonic invocation stamp plus the sorted pending-and-completed overlay arbitrates optimistic-versus-optimistic overlap, while `then` reconciliation deliberately replays every completed mutation because it represents a wholesale server snapshot that must not erase local deletes or archives; the asymmetry is design, and the new stale-list test pins it as a guard for the conditional replay (the old unconditional replay covered that case, so it passes on main by construction). the replay-newer rule also heals the settle-order-versus-invocation-order mismatch when an older mutation folds into the base after a newer one. the idempotency requirement on optimistic callbacks predates this change and stays documented; the failure path is unchanged (a rejected mutation never enters the completed list and rolls back in finally). the api-surface delta is a private field declaration only.

- completed optimistic callbacks were replayed unconditionally after every settle, so the final state depended on network completion order instead of user action order

### what we modified

### potential further fixes

- completed effects now live in both the base fold and the display overlay; the class comment covers the double representation, worth keeping in sync if the reducer shape evolves

PR #4968, opened 2026-07-16
fix(react-a2a): preserve history branches on load

Kinfe123:
## Summary

Follow-up to merged #4776, applying the same branch-preserving history fix to the A2A runtime.

- preserve `parentId` and `headId` when loading A2A thread history
- expose the full message repository so persisted sibling responses remain available to the branch picker
- keep sibling branches intact while switching, regenerating, or editing messages
- retain a linear fallback for malformed legacy history

## Why

While testing A2A history with two assistant responses sharing the same user-message parent, reloading flattened the repository into `[user, response A, response B]` because `__internal_load()` discarded every stored `parentId`. The branch picker then had no sibling relationship to navigate.

This follows the repository-backed approach introduced for the same AG-UI root cause in #4776. Editing now creates and selects a sibling branch without reparenting or dropping the original response subtree.

## Follow-up

A2A now preserves the edited source subtree as a sibling branch. AG-UI still removes that source during edits, so aligning its edit behavior can remain a separate follow-up to #4776.

## Test plan

- `pnpm --filter @assistant-ui/react-a2a test` (135 tests)
- `pnpm lint`
- `pnpm build` (86/86 tasks passed)


okisdev (approved):
## review

lgtm as `patch`. verified both directions at runtime: at the head the react-a2a suite passes 135/135, and with the two source files reverted to main exactly the eight discriminating tests fail (branch preservation, the three malformed-history fallbacks, hidden siblings across branch switches, reparenting, regeneration siblings, and edit-as-sibling).

the important architectural call checks out: this mirrors the current ag-ui shape (the single-repository SSOT from #4938) rather than the older #4776 snapshot, and each of the hard-won mapping lessons from that migration is handled. branch-preserving head moves go through `switchToBranch` instead of the destructive `resetHead`, the export-skips-optimistic edge does not apply because this core only uses `addOrUpdateMessage`, duplicate ids, cycles, and orphaned parents are caught by degenerate detection with a topological import and a linear fallback (so `performOp`'s cycle check can never throw on load), and the memoized `export()` payload is invalidated in all four mutation wrappers. the adapter passes `messageRepository` and keeps `setMessages` for the branch-switch capability bit.

- none beyond the reported bug; the nine review threads already hardened edited-branch preservation and the malformed-history coverage

### what we modified

### potential further fixes

- ag-ui still removes the edited source subtree where a2a now keeps it as a sibling; the PR body flags aligning ag-ui as the follow-up to #4776
- the degenerate-input policies differ deliberately (a2a falls back to linear wholesale, ag-ui collapses duplicate ids last-write-wins); both are test-pinned defensive paths, worth unifying only if a shared loader ever emerges

PR #4970, opened 2026-07-16
feat(react-mcp): support `listResources()` pagination

Kinfe123:
## Summary

- allow `listResources()` to accept an optional MCP pagination cursor
- forward the cursor to the connected MCP client and cover it with a regression test
- document how a resource browser requests the next page

## Why

This is a follow-up to merged [PR #4699](https://github.com/assistant-ui/assistant-ui/pull/4699), which first exposed MCP resource listing and identified pagination as unfinished. A server can return `nextCursor`, but the zero-argument public method left resource browsers stuck on the first page because they could not send that cursor back.

After this change, callers can request the next page with:

```ts
type ResourcePage = {
  resources: Array<{ uri: string }>;
  nextCursor?: string;
};

const firstPage = (await server.listResources()) as ResourcePage;
const secondPage = firstPage.nextCursor
  ? await server.listResources({ cursor: firstPage.nextCursor })
  : undefined;
```

Existing `listResources()` calls continue to work unchanged.

## Validation

- `pnpm --filter @assistant-ui/react-mcp test` (22 tests)
- `pnpm --filter @assistant-ui/react-mcp build`
- filtered API surface generation and check for `@assistant-ui/react-mcp`
- `pnpm lint`
- `pnpm --filter @assistant-ui/docs build` (441 static pages)

### DIFF
--- a/.changeset/react-mcp-resource-pagination.md
+++ b/.changeset/react-mcp-resource-pagination.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react-mcp": patch
+---
+
+feat: support cursor pagination when listing MCP resources
--- a/api-surface/assistant-ui__react-mcp.ts
+++ b/api-surface/assistant-ui__react-mcp.ts
@@ -124,7 +124,9 @@ type MCPServerMethods = {
   disconnect: () => Promise<void>;
   remove: () => Promise<void>;
   callTool: (name: string, args: unknown) => Promise<unknown>;
-  listResources: () => Promise<unknown>;
+  listResources: (params?: {
+    cursor?: string | undefined;
+  }) => Promise<unknown>;
   readResource: (uri: string) => Promise<unknown>;
   completeAuth: (callbackUrl: string) => Promise<void>;
 };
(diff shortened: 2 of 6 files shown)


okisdev (approved):
## review

LGTM as `patch`.

verified the gap and the fix: #4699's review flagged pagination as unreachable with the zero-arg signature, and this threads the cursor through exactly where the pinned SDK expects it (`listResources(params?: ListResourcesRequest['params'], ...)`, checked against `@modelcontextprotocol/sdk` 1.29). the zero-arg path is unchanged since `params` stays `undefined`, the regression test pins both the forwarding and the result passthrough, and api-surface, docs, and the changeset all move in the same PR. exposing `{ cursor?: string | undefined }` instead of the SDK's full params type keeps the surface minimal, which fits the raw-result style of the scope.

### potential further fixes
- the `listResources` JSDoc one-liner in `mcp-scope.ts` does not mention the cursor param; a short "pass cursor for the next page" in the same style as the sibling comments would keep the doc comment in step with the signature.

PR #4983, opened 2026-07-16
feat(ui): polish attachment tile, remove button, and preview dialog

ephraimduncan:
## What

Visual polish pass on the attachment components (`attachment.tsx` + `attachment.radix.tsx` in `@assistant-ui/ui`, synced to `templates/minimal`), plus the docs samples that preview them.

- **Tile:** the 1px border is replaced by an inset hairline ring on an `after:` pseudo (`black/10`, `white/10` in dark); hover is a subtle `foreground/10` wash instead of dimming the whole thumbnail with `hover:opacity-75`; the tile finally gets a real `focus-visible` ring (it is keyboard-focusable via `tabIndex={0}` but previously only had the UA default outline); light press scale (`0.96`).
- **Remove button:** white chip with black icon → translucent dark chip (`bg-black/50`, white icon, backdrop blur) that works on any thumbnail in both themes; button grows 14px → 20px with the hit area extended to ~32px via `after:-inset-1.5`.
- **Status overlays:** uploading/error overlays get a soft backdrop-blur treatment; error switches from a red tint + red border to a `ring-destructive/60` ring with the icon on a blurred backdrop.
- **Preview dialog:** image tiles show `cursor-zoom-in`; the dialog image gets rounded corners and a fade-in on load; the close button darkens on hover instead of tinting its icon red.
- **Motion:** composer attachments animate in (fade + zoom), tile/remove/add buttons get press feedback — all new animation and press transitions are gated behind `prefers-reduced-motion` (`motion-reduce:animate-none` / `motion-reduce:transition-none`), following the pattern in `assistant-modal` / `reasoning` / `tool-fallback`.
- **Docs:** `apps/docs/components/docs/samples/attachment{,.radix}.tsx` are updated so the `/docs/ui/attachment` preview matches what `shadcn add attachment` installs.

No new props, no new dependencies, no API changes — class-only changes.

## Why

The attachment tiles were the least polished part of the default composer: no focus treatment on a keyboard-focusable element, a hover state that dimmed the thumbnail itself, a white remove chip that clashes in dark mode, and a 14px remove target that is hard to hit.

## Notes for review (deliberate behavior changes)

- `cursor-zoom-in` comes from the `DialogTrigger` and merges onto the tile; non-image attachments render no dialog (`if (!src) return children`), so they keep the plain pointer.
- The enlarged remove hit area overlaps the tile's top-right corner — clicks there remove instead of opening the preview, by design.
- The X icon's `stroke-[2.5]` now applies in light mode too (previously `dark:`-only).
- Base variant only: the Base UI `Avatar` ships its own `after:` border ring which would double with the new tile ring, so the thumb adds `after:hidden`; the Radix `Avatar` has no such pseudo, hence no counterpart in `attachment.radix.tsx`.
- Error ring is 1px at 60% opacity (was a solid `border-destructive`); the error icon + tooltip message keep the state legible — see screenshots.

## Screenshots

Before/after across states (hover/focus/press are rendered by applying the equivalent utilities statically). Left to right: image, file, hover, focus, pressed, remove-hover, uploading, error, add button.

**Light**

before/after light

**Dark**

before/after dark

## Checks

- `pnpm sync-templates` green — `templates/minimal` copy is byte-equal with `packages/ui`
- oxfmt / oxlint clean on all touched files
- No changeset: `@assistant-ui/ui`, templates, and docs are private packages

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/4983?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/docs/components/docs/samples/attachment.radix.tsx
+++ b/apps/docs/components/docs/samples/attachment.radix.tsx
@@ -13,9 +13,9 @@ function AttachmentTileStatic({ name, isImage }: AttachmentTileStaticProps) {
   const attachmentType = isImage ? "Image" : "Document";
 
   return (
-    <div className="aui-attachment-root relative">
+    <div className="aui-attachment-root animate-in fade-in-0 zoom-in-95 relative duration-200 motion-reduce:animate-none">
       <div
-        className="aui-attachment-tile aui-attachment-tile-composer border-foreground/20 bg-muted size-14 cursor-pointer overflow-hidden rounded-[14px] border transition-opacity hover:opacity-75"
+        className="aui-attachment-tile aui-attachment-tile-composer bg-muted hover:after:bg-foreground/10 focus-visible:ring-ring/50 relative size-14 cursor-pointer overflow-hidden rounded-[14px] transition-transform outline-none after:pointer-events-none after:absolute after:inset-0 after:rounded-[inherit] after:ring-1 after:ring-black/10 after:transition-colors after:ring-inset focus-visible:ring-3 active:scale-[0.96] motion-reduce:transition-none dark:after:ring-white/10"
         role="button"
         tabIndex={0}
         aria-label={`${attachmentType} attachment: ${name}`}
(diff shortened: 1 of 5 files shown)


okisdev (approved):
## review

lgtm. reviewed the before/after captures state by state in both themes and this is a real upgrade in the kit's own direction: hairline inset ring instead of a solid border, hover wash instead of dimming the thumbnail, a proper focus-visible ring replacing the UA outline, the error state stepped down from red-tint-plus-border to a hairline destructive ring over a blurred backdrop, and the remove chip flipped to a translucent dark treatment that works on any thumbnail in both themes.

the details hold up under a close read: the duplicated `dark:after:ring-destructive/60` is load-bearing (it overrides the base `dark:after:ring-white/10` in dark mode), the base-flavor `after:hidden` on the thumb is justified because the base `Avatar` ships its own `after:` border ring that would double with the tile ring (and the radix `Avatar` has none, so no counterpart there), tailwind v4's `transition-transform` covers the `scale` property so the press feedback animates, and every new animation is gated behind `motion-reduce` following the existing kit pattern. template sync is green so the minimal copy stays byte-equal, and the docs samples match what `shadcn add attachment` installs. the enlarged remove hit area overlapping the preview corner is the right priority for a 56px tile.

- none; the three deliberate behavior changes are documented in the PR body with their rationale

### what we modified

### potential further fixes

- none

PR #4994, opened 2026-07-16
fix(docs): restore source-linked package builds

### LINKED ISSUE #4991: [docs] workspace alias breaks react-generative-ui subpaths
## Summary

The docs app cannot build after the Slack converter landed because its broad `@assistant-ui/*` workspace path alias maps the exported `@assistant-ui/react-generative-ui/slack` subpath to the nonexistent `packages/react-generative-ui/slack/src` directory. This blocks local and PR verification for any docs change.

## Type Of Work In Progress

Building the complete docs app while validating a new framework integration and its generated API reference.

## Reproduction Steps

1. Check out `assistant-ui/assistant-ui` at `9a343db87` (current `main` when reproduced).
2. Install dependencies with `pnpm install`.
3. Run `pnpm --filter @assistant-ui/docs build`.

## Expected Behavior

The docs build resolves the package's exported `./slack` entry from `packages/react-generative-ui/src/slack.ts` and completes.

## Actual Behavior

Turbopack fails in `apps/docs/components/gallery/template-code-tabs.tsx`:

```text
Module not found: Can't resolve '@assistant-ui/react-generative-ui/slack'
Import map: aliased to relative '../../packages/react-generative-ui/slack/src'
```

`apps/docs/tsconfig.json` currently has `"@assistant-ui/*": ["../../packages/*/src"]`, so a package subpath is captured as part of the package-name wildcard.

## Context

- Area/package: `apps/docs`, `@assistant-ui/react-generative-ui/slack`
- Root cause: the source file exists at `packages/react-generative-ui/src/slack.ts`, but the workspace alias resolves a different path

## Minimal Reproduction

`pnpm --filter @assistant-ui/docs build` is deterministic on a clean checkout at the commit above. Adding a package-specific source-subpath alias makes the import resolvable.

### LINKED ISSUE #4992: [react-generative-ui] Slack alert level fails TypeScript source check
## Summary

`packages/react-generative-ui/src/slack/constants.ts` applies `as const` to an indexed expression in `buildAlertBlock`. TypeScript 6 rejects that syntax, which prevents consumers that typecheck the source package (including the docs app's workspace aliases) from building.

## Type Of Work In Progress

Building the docs app after correcting its workspace resolution for the newly exported Slack subpath.

## Reproduction Steps

2. Resolve `@assistant-ui/react-generative-ui/slack` to `packages/react-generative-ui/src/slack.ts`, as the docs workspace intends for local packages.

A narrower check also exposes the source error (alongside pre-existing test-only errors):

```bash
pnpm --filter @assistant-ui/react-generative-ui exec tsc --noEmit
```

## Expected Behavior

`buildAlertBlock` typechecks and returns a `SlackAlertLevel` for every supported tone.

## Actual Behavior

```text
packages/react-generative-ui/src/slack/constants.ts:160:17
TS1355: A 'const' assertion can only be applied to references to enum members, or string, number, boolean, array, or object literals.
```

The resulting `level` is also widened to `string`, producing TS2322 at the return value.

## Context

- Area/package: `@assistant-ui/react-generative-ui`, Slack converter
- Impact: source consumers and the docs app cannot typecheck the exported Slack subpath

## Minimal Reproduction

The deterministic docs build above reaches this error immediately after the subpath is resolved. Replacing the indexed-object assertion with a literal-preserving tone mapping removes both source errors without changing runtime behavior.

samdickson22:
## Why

The current docs source build is blocked by a three-step chain of existing workspace defects:

1. the broad `@assistant-ui/*` alias misresolves `react-generative-ui` subpaths
2. once those sources load, TypeScript 6 rejects the Slack alert-level assertion
3. once that typechecks, split `@types/hast` versions make markdown AST types incompatible

Keeping these as stacked PRs leaves the earlier PRs red by construction, so the fixes are combined here as one green docs-build prerequisite.

## What changed

- resolve `@assistant-ui/react-generative-ui/*` inside the package's real `src` directory
- preserve the Slack alert-level literal union with a direct conditional
- pin workspace HAST types to `3.0.5`
- add a patch changeset for the published `@assistant-ui/react-generative-ui` fix

## Verification

- `pnpm --filter @assistant-ui/react-generative-ui test` (24 files, 527 tests)
- `pnpm --filter @assistant-ui/react-generative-ui build`
- `pnpm why @types/hast -r` reports only `@types/hast@3.0.5`
- lockfile contains no `@types/hast@3.0.4` resolution
- `pnpm changeset status --since upstream/main --verbose`
- docs production build succeeds and generates all current pages on Node 24.11.1

Closes #4991
Closes #4992

Note (maintainer): the @types/hast pin originally bundled here was dropped during adoption; interim dependency updates on main already unified the hast types, verified by a clean docs source typecheck without the override. #4997 is closed separately with that rationale.


Kinfe123 (dismissed):
I looked into the remaining TypeScript error. TypeScript 6 rejects as const on the computed { ... }[tone] expression, while removing it widens level to string.
I think the simplest fix is:
```ts
const level = tone === "danger" ? "error" : tone;
```
Another valid option is applying as const to a levelByTone map before indexing it. Both preserve the exact "info" | "success" | "warning" | "error" union without changing runtime behavior.
Since this PR already includes the alias fix, Slack type correction, HAST deduplication, and patch changeset, I think rebasing it onto current main and rerunning the docs build is the cleanest way forward.

okisdev (approved):
adopted from the draft queue: the three-defect chain is real on current main (the broad workspace alias misresolves react-generative-ui subpaths, and the indexed as-const assertion fails the source typecheck once those sources load), the alias fix follows the tsconfig's existing dedicated-subpath pattern, and the conditional keeps the danger-to-error mapping with identical behavior. the bundled @types/hast override was dropped during adoption: interim dependency updates already unified the types, verified by a clean docs source typecheck without it, so #4997 closes separately rather than through this PR. docs tsc, the react-generative-ui suite (527), and its build are green on the merged head; the remaining delta is exactly the changeset, the alias, and the constants fix.

PR #4999, opened 2026-07-16
fix(deps): deduplicate hast types

### LINKED ISSUE #4997: [deps] split @types/hast versions break docs source typecheck
## Summary

Current main resolves `@types/hast` to both 3.0.4 and 3.0.5. The workspace source package `@assistant-ui/react-markdown` uses 3.0.5 while its `react-markdown` dependency resolves the same public AST types through 3.0.4. With `exactOptionalPropertyTypes`, the docs app then rejects the component map because the two `Element` declarations are distinct.

## Type Of Work In Progress

Running the full docs production build after validating new framework-integration docs against current main.

## Reproduction Steps

2. Run `pnpm install` from a clean dependency state.

## Expected Behavior

All workspace packages and markdown dependencies share one `@types/hast` version, and the docs typecheck succeeds.

## Actual Behavior

Next.js fails at `packages/react-markdown/src/primitives/MarkdownText.tsx:137`:

```text
Type ... is not assignable to type 'Components' with 'exactOptionalPropertyTypes: true'.
Type '...@types+hast@3.0.4...Element' is not assignable to type '...@types+hast@3.0.5...Element'.
```

`pnpm why @types/hast -r` reports both 3.0.4 and 3.0.5. `packages/react-markdown` directly requests `^3.0.5`, while transitive unified/markdown packages remain on a lock resolution of 3.0.4.

## Context

- Area/package: root dependency graph, `@assistant-ui/react-markdown`, docs app
- Impact: a clean docs production build fails during TypeScript checking

## Minimal Reproduction

The docs build is deterministic after a clean install. A root pnpm override that deduplicates `@types/hast` to 3.0.5 removes the cross-version type identity and allows the build to proceed.

samdickson22:
Closes #4997

## Root cause

The workspace resolves both `@types/hast 3.0.4` and `3.0.5`. Source-linked assistant-ui Markdown code and `react-markdown` then expose structurally incompatible HAST `Element` types under exact optional-property checking.

## What changed

Pin the workspace graph to `@types/hast 3.0.5` with a root pnpm override and normalize the lockfile so stale `3.0.4` snapshots cannot survive downstream dependency additions.

## Verification

- `pnpm why @types/hast -r` reports only `@types/hast@3.0.5`
- clean lockfile install on Node 24.11.1
- cumulative current-main docs build (448 static pages)
- `pnpm changeset status --since codex/mastra-slack-types --verbose` reports no release

This changes only workspace dependency resolution and does not modify a published package. Depends on #4995.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -4,6 +4,9 @@ settings:
   autoInstallPeers: true
   excludeLinksFromLockfile: false
 
+overrides:
+  '@types/hast': 3.0.5
+
 importers:
 
   .:
@@ -4307,7 +4310,7 @@ importers:
         specifier: workspace:*
         version: link:../x-buildutils
       '@types/hast':
-        specifier: ^3.0.5
+        specifier: 3.0.5
         version: 3.0.5
       '@types/react':
         specifier: ^19.2.17
@@ -4547,7 +4550,7 @@ importers:
         specifier: ^16.3.2
         version: 16.3.2(@testing-library/dom@10.4.1)(@types/react-dom@19.2.3(@types/react@19.2.17))(@types/react@19.2.17)(react-dom@19.2.7(react@19.2.7))(react@19.2.7)
       '@types/hast':
-        specifier: ^3.0.5
+        specifier: 3.0.5
         version: 3.0.5
       '@types/react':
         specifier: ^19.2.17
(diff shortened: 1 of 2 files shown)


samdickson22:
Folded into #4994 because the alias and Slack fixes expose this HAST type split during the same independent docs build. The issue, override, lockfile proof, and verification are preserved in #4994.

PR #5019, opened 2026-07-17
fix(cli): repair voice and interactable scaffolds

### LINKED ISSUE #5018: cli: three hasLocalComponents examples scaffold with unresolvable component imports
`with-elevenlabs-conversational`, `with-livekit`, and `with-interactables` are flagged `hasLocalComponents: true` in `PROJECT_METADATA` (`packages/cli/src/commands/create.ts`), but none of them ships a local `components/` directory; in the monorepo their `@/components/assistant-ui/*` imports resolve through tsconfig aliases into `packages/ui`. after #5013 the scaffold tsconfig is standalone (the dangling `extends` and monorepo aliases are stripped for every scaffold), but the `hasLocalComponents` gate still skips `scanRequiredComponents` and the registry install, so the scaffolded imports point at `./components/assistant-ui/*` which does not exist and `next build` fails. before #5013 these scaffolds were equally broken (the tsconfig extended `@assistant-ui/x-buildutils/ts/base`, which `transformPackageJson` removes from devDependencies), so this predates that change; only `minimal` and `with-expo` actually carry local components.

candidate fixes:

- flip the three entries to `hasLocalComponents: false`, after verifying every component they import (`voice`, `markdown-text`, `tooltip-icon-button`, `thread`, ...) is resolvable as an `@assistant-ui/*` registry item, so the create flow installs them.
- or derive the gate from whether the copied scaffold contains a `components/` directory instead of static metadata, which cannot drift.

either way the fix needs an end to end check: scaffold each of the three and run `next build`.

AkhilTrivediX:
## Summary

- route the ElevenLabs Conversational, LiveKit Voice, and Interactables examples through the CLI component-install flow
- use the registry's generated component layout in those examples
- detect root-level assistant-ui component imports when scaffolding

## Root cause

These examples were marked as shipping local components even though they rely on registry-managed components. The CLI therefore skipped component installation and left imports unresolved in generated projects.

## Verification

- `pnpm --filter assistant-ui exec vitest run test/commands/create.test.ts test/lib/create-project.test.ts --reporter=dot`
- `pnpm --filter assistant-ui build`
- scaffolded `with-elevenlabs-conversational` with the local CLI, then ran `pnpm build` successfully

Fixes #5018

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5019?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/bright-tomatoes-join.md
+++ b/.changeset/bright-tomatoes-join.md
@@ -1,0 +1,5 @@
+---
+"assistant-ui": patch
+---
+
+fix: make the ElevenLabs, LiveKit, and Interactables CLI examples self-contained
--- a/.changeset/create-scaffold-sanitation.md
+++ b/.changeset/create-scaffold-sanitation.md
@@ -1,0 +1,5 @@
+---
+"assistant-ui": patch
+---
+
+fix: sanitize tsconfig for every create scaffold and surface registry install failures
--- a/examples/with-elevenlabs-conversational/app/voice-thread.tsx
+++ b/examples/with-elevenlabs-conversational/app/voice-thread.tsx
@@ -1,14 +1,14 @@
 "use client";
 
-import { MarkdownText } from "@/components/markdown-text";
+import { MarkdownText } from "@/components/assistant-ui/markdown-text";
 import {
   VoiceOrb,
   VoiceConnectButton,
   VoiceMuteButton,
   VoiceDisconnectButton,
   deriveVoiceOrbState,
-} from "@/components/voice";
-import { TooltipIconButton } from "@/components/tooltip-icon-button";
+} from "@/components/assistant-ui/voice";
+import { TooltipIconButton } from "@/components/assistant-ui/tooltip-icon-button";
 import {
   AuiIf,
   MessagePrimitive,
--- a/examples/with-elevenlabs-conversational/app/voice-thread.tsx
+++ b/examples/with-elevenlabs-conversational/app/voice-thread.tsx
@@ -1,14 +1,14 @@
 "use client";
 
-import { MarkdownText } from "@/components/assistant-ui/markdown-text";
+import { MarkdownText } from "@/components/markdown-text";
 import {
   VoiceOrb,
   VoiceConnectButton,
   VoiceMuteButton,
   VoiceDisconnectButton,
   deriveVoiceOrbState,
-} from "@/components/assistant-ui/voice";
-import { TooltipIconButton } from "@/components/assistant-ui/tooltip-icon-button";
+} from "@/components/voice";
+import { TooltipIconButton } from "@/components/tooltip-icon-button";
 import {
   AuiIf,
   MessagePrimitive,
--- a/examples/with-interactables/app/page.tsx
+++ b/examples/with-interactables/app/page.tsx
@@ -1,7 +1,7 @@
 "use client";
 
 import { useCallback } from "react";
-import { Thread } from "@/components/assistant-ui/thread";
+import { Thread } from "@/components/thread";
 import {
   AssistantRuntimeProvider,
   AuiProvider,
--- a/examples/with-interactables/app/page.tsx
+++ b/examples/with-interactables/app/page.tsx
@@ -1,7 +1,7 @@
 "use client";
 
 import { useCallback } from "react";
-import { Thread } from "@/components/thread";
+import { Thread } from "@/components/assistant-ui/thread";
 import {
   AssistantRuntimeProvider,
   AuiProvider,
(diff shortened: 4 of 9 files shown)


AkhilTrivediX:
Good catch. I updated the root-level import scan to check whether the referenced component already exists in the generated scaffold before adding it to the Assistant UI registry install list. The regression test covers a local `SignupForm` alongside a missing registry `thread` component, and confirms that only `@assistant-ui/thread` is installed.\n\nValidated with the targeted CLI tests, `oxfmt --check`, `oxlint`, and the CLI build.

AkhilTrivediX:
I traced the failing Build Changed Apps job to the example source imports introduced by this PR. The examples must keep their monorepo `@/components/assistant-ui/*` aliases to build in-repo; the CLI now rewrites those aliases to registry output paths only after copying a no-local-components project, before scanning and installing the required components.\n\nAdded coverage for the transformation and verified:\n\n- targeted CLI tests\n- `oxfmt --check` and `oxlint`\n- CLI build\n- dependency-aware builds for `with-interactables`, `with-elevenlabs-conversational`, and `with-livekit` (all passed)

okisdev:
thanks for jumping on #5018 so quickly, and welcome! closing this one, but I want to be clear about what you got right, because the hard part was right: the root cause diagnosis (the three examples were mislabeled as shipping local components) was exactly correct, the changeset was correct, and the tests you added showed good instincts, especially the guard against treating existing local components as registry requests.

the fix landed in #5021 (merged as `00fea578f`), and its core is the same three-entry `hasLocalComponents` flip this PR contained. the reason this PR could not merge as written comes down to the two additions around that flip:

**the import rewrite inverts the actual registry layout.** deployed registry items install into `components/assistant-ui/`: `curl https://r.assistant-ui.com/voice.json | jq '.files[0].path'` returns `components/assistant-ui/voice.tsx`, and the installed voice component itself imports `@/components/assistant-ui/tooltip-icon-button`. so the original `@/components/assistant-ui/*` imports were already correct for both worlds: the monorepo (resolved by the example tsconfig aliases) and the generated scaffold (where the registry writes the files). I verified this end to end on all three examples with unchanged imports: scaffold, registry install, and `next build` all pass. after the rewrite, a generated project imports `@/components/voice` while the file lands at `components/assistant-ui/voice.tsx`, which cannot resolve.

**the rewrite also broke the monorepo build.** Build Changed Apps failed on this PR with `Module not found: Can't resolve '@/components/thread'` at `examples/with-interactables/app/page.tsx`, because nothing in the repo maps root-level component names; the aliases only cover the `assistant-ui/` and `ui/` prefixes. example code has to keep compiling in two worlds at once, in-repo through the aliases and post-scaffold through the registry layout, and `@/components/assistant-ui/*` is the one shape that satisfies both.

**the scanner heuristic is a footgun** (greptile flagged this as well): treating any root-level `@/components/x` import without a matching local file as an `@assistant-ui/x` registry item means a typo in a local import becomes a fabricated registry request, and since #5013 a failed install exits nonzero, so the whole scaffold dies on a confusing error about a component that never existed.

a few suggestions that would have caught this before pushing:

- when a fix seems to require changing a convention, first verify where the artifact actually lands; the one line curl above answers the layout question directly
- run the monorepo build for any touched example (`pnpm build`, which is what the failing Build Changed Apps job runs); example changes must keep compiling in-repo, not only in scaffolds
- verify the end to end claim on every affected target and against the final state of the diff; the scaffold build cannot have passed with the rewritten imports, so that run was likely against a stale checkout
- prefer the smallest diff that fixes the root cause; here the metadata flip alone was the entire fix, and the extra surface is where all the failures came from

if you would like a follow-up, a genuinely useful one is open: replace the mirrored name list in `create.test.ts` with a truth check that derives each example's expected `hasLocalComponents` from whether `examples/<name>/components` exists on disk. that test would have caught this drift automatically, and it is small and self-contained. looking forward to the next one.

AkhilTrivediX:
Thanks for the thorough explanation and the welcome. I understand the two-world import layout now: the registry and the monorepo both require the `@/components/assistant-ui/*` form, so the metadata correction was sufficient. I should have verified the registry artifact path and run the affected monorepo build against the final diff before pushing. I appreciate the concrete follow-up suggestion and will carry those checks into future scaffold work.

PR #5020, opened 2026-07-17
fix(docs): sandbox hosted artifact previews

### LINKED ISSUE #5003: [docs] sandbox artifact previews or remove the safe-execution claim
## Summary

The hosted Artifacts page says generated code uses "Safe Execution" and "runs in a sandboxed environment," but the iframe in `apps/docs/components/examples/artifacts-view.tsx` renders model-generated `srcDoc` without a `sandbox` attribute. That leaves the hosted implementation's generated HTML and JavaScript same-origin with the docs page.

Either add an intentionally restrictive iframe sandbox and verify the supported artifact behavior, or remove/correct the safe-execution claim until isolation exists. The preferred fix is to sandbox the preview. PR #4936 already adds `sandbox="allow-scripts"` to the standalone `with-artifacts` work panel, but it does not change the separate hosted docs implementation.

## Type Of Work In Progress

Auditing the two Artifacts implementations while reviewing the chat-to-work-panel example in #4936 and the hosted-example failure in #4856.

## Reproduction Steps

1. Open `apps/docs/content/examples/artifacts.mdx` and note the "Safe Execution" feature claim.
2. Open `apps/docs/components/examples/artifacts-view.tsx`.
3. Inspect the preview iframe that renders `srcDoc={artifact}`.
4. Observe that it has no `sandbox` attribute.
5. Compare it with the iframe introduced in PR #4936, which uses `sandbox="allow-scripts"`.

## Expected Behavior

- A page advertised as safely rendering model-generated HTML isolates that content with an explicit, reviewed iframe sandbox policy.
- Required artifact interactions continue to work inside the sandbox.
- The documentation describes the actual isolation guarantees and does not overstate them.

## Actual Behavior

- The hosted docs preview renders generated HTML and JavaScript in an unsandboxed `srcDoc` iframe.
- The docs describe the execution as sandboxed even though the isolation attribute is absent.
- The standalone fix in #4936 does not protect the separate docs iframe.

## Acceptance Criteria

- [ ] The hosted Artifacts preview uses an explicit restrictive sandbox policy, or the safe-execution claim is removed until it does.
- [ ] Generated JavaScript needed by the demo still runs inside the preview.
- [ ] Browser verification confirms artifact code cannot access or mutate the parent docs document.
- [ ] The documentation states the actual preview isolation guarantees.
- [ ] The result is compatible with the planned deduplication of the hosted and standalone Artifacts implementations.

## Context

- Environment: macOS 15.6.1, Node v22.17.1, pnpm 11.10.0
- Repository state: `main` at `5d888e797`
- Area/package: `apps/docs` and `examples/with-artifacts`
- Related work: #4856, #4931, PR #4936; structural follow-up #5004
- Logs/errors: no runtime error is required; this is visible directly in the iframe markup.

## Repro Script

A separate script is unnecessary because the mismatch is deterministic from the MDX claim and iframe attributes. Browser verification should be added with the fix to assert parent-document isolation.

AkhilTrivediX:
## Summary

- sandbox the hosted Artifacts preview with `allow-scripts` so generated JavaScript continues to run without inheriting the documentation page origin
- align the Artifacts page copy with the actual isolation guarantee

## Root cause

The hosted preview rendered untrusted `srcDoc` content in an iframe without a sandbox attribute. That gave generated artifact code the same origin as the documentation page despite the page claiming safe execution.

## Validation

- `pnpm exec oxfmt --check apps/docs/components/examples/artifacts-view.tsx apps/docs/content/examples/artifacts.mdx`
- `pnpm exec oxlint apps/docs/components/examples/artifacts-view.tsx`

Closes #5003

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5020?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/docs/components/examples/artifacts-view.tsx
+++ b/apps/docs/components/examples/artifacts-view.tsx
@@ -51,6 +51,7 @@ export const ArtifactsView = () => {
               <iframe
                 title="artifact-preview"
                 className="h-full w-full"
+                sandbox="allow-scripts"
                 srcDoc={artifact}
               />
             )}
--- a/apps/docs/components/examples/artifacts-view.tsx
+++ b/apps/docs/components/examples/artifacts-view.tsx
@@ -51,7 +51,7 @@ export const ArtifactsView = () => {
               <iframe
                 title="artifact-preview"
                 className="h-full w-full"
-                sandbox="allow-scripts"
+                sandbox="allow-scripts allow-forms"
                 srcDoc={artifact}
               />
             )}
--- a/apps/docs/content/examples/artifacts.mdx
+++ b/apps/docs/content/examples/artifacts.mdx
@@ -18,18 +18,18 @@ An open-source implementation of Claude Artifacts that allows users to generate 
 
 ## Features
 
-- **Real-time Preview**: Generated artifacts render instantly in a sandboxed iframe
+- **Real-time Preview**: Generated artifacts render instantly in an isolated sandboxed iframe
 - **Code Generation**: Creates complete HTML, CSS, and JavaScript applications
 - **Interactive Elements**: Artifacts can include buttons, forms, and animations
 - **Iteration Support**: Ask the AI to modify and improve generated artifacts
-- **Safe Execution**: Code runs in a sandboxed environment for security
+- **Safe Execution**: Code runs in a sandboxed iframe with scripts enabled but no access to the parent documentation page
 - **Export Options**: Download generated artifacts as standalone files
 
 ## How It Works
 
 1. **Request**: Ask the AI to create something (e.g., "Create a calculator app")
 2. **Generation**: The AI generates the necessary HTML, CSS, and JavaScript
-3. **Rendering**: Code is safely rendered in a preview panel
+3. **Rendering**: Code is safely rendered in an isolated preview panel
 4. **Iteration**: Request changes and see updates in real-time
 
 ## Use Cases


AkhilTrivediX:
Good catch. I added `allow-forms` alongside `allow-scripts`, so artifact forms can submit while the iframe remains sandboxed without `allow-same-origin`. Formatting and lint checks pass.

okisdev (approved):
## review

LGTM.

verified against #5003: the preview iframe at `artifacts-view.tsx:51` currently renders model-generated `srcDoc` with no sandbox, so generated code runs same origin with the docs page while the copy claims safe execution. `sandbox="allow-scripts allow-forms"` without `allow-same-origin` puts the content in an opaque origin (scripts run, no access to the parent document, cookies, or storage), which is exactly the isolation the page now describes, and the component itself has no parent-iframe coupling that sandboxing could break. the updated copy matches the real guarantee. no changeset needed since apps/docs is private.

- the first commit used `allow-scripts` alone, which blocks native form submission inside the sandbox and would have regressed the documented "buttons, forms" interactive artifacts.

### what we modified

- adopted `allow-forms` in a clean follow-up commit, restoring form submission while keeping the opaque-origin isolation intact; the threat model (protect the docs origin) is unaffected since form targets never gain same-origin access.

### potential further fixes

- the sibling preview in `examples/with-artifacts/app/page.tsx` is still unsandboxed on `main`; that surface is scoped to #4936 per the issue split, but if #4936 stalls it deserves the same one-line fix on its own.
- the PR description still says `allow-scripts` only; the commit history tells the full story, but updating the summary would keep the description accurate.
- generated artifacts that call `alert()` or touch `localStorage` degrade silently in an opaque origin; intentional, and consistent with the new copy.

PR #5025, opened 2026-07-17
refactor(react-ag-ui): share interrupt-response validation between submitInterruptResponses and steerAway

ShobhitPatra:
This PR  dedupes the interrupt-response validation in react-ag-ui so both interrupt entry points validate through one shared helper.

## What

`submitInterruptResponses` and `steerAway` each carried a near identical per-entry validation loop (interruptId presence, status validity, unknown id, duplicate), differing only in the error prefix. This extracts the loop into one private helper, `collectInterruptResponses`, on `AgUiThreadRuntimeCore`.

- Per-method error prefixes are preserved: `[agui] submitInterruptResponses:` vs `[agui] steerAway:`.
- Each caller keeps its own policy: submit still requires a response for every open interrupt and runs the expiry checks; steerAway still fills unanswered interrupts with `cancelled`.
- No public surface change; patch changeset for `@assistant-ui/react-ag-ui`.

## Why

The copies had already drifted in wording and check order, and any check added to one method silently misses the other. Sharing the loop keeps the two error surfaces consistent by construction.

## Impact

Accept/reject behavior is unchanged; only the error message selected under multiple simultaneous violations moves, plus two wording unifications:

| Input | Before | After |
| --- | --- | --- |
| submit: unknown id | `unknown interrupt ids: <ids>` | `unknown interrupt id <id>` (matches steerAway) |
| submit: unknown id + unanswered interrupts | missing-responses error | unknown-id error |
| submit: same unknown id twice | duplicate error | unknown-id error |
| steerAway: entry without interruptId | `every response must have an interruptId` | `every entry must have an interruptId` |

- One test regex updated (the plural unknown-ids pin); new tests pin the previously untested checks and the ordering change.
- react-ag-ui suite green (230 tests), full turbo build green.

## Rationale

The two methods' policies differ by design (submit is strict answer-everything, steerAway is cancel-what-I-skipped), so only the per-entry validation is shared and the policies stay at the call sites.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5025?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/ag-ui-shared-interrupt-validation.md
+++ b/.changeset/ag-ui-shared-interrupt-validation.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react-ag-ui": patch
+---
+
+refactor: share interrupt-response validation between submitInterruptResponses and steerAway
--- a/packages/react-ag-ui/src/runtime/AgUiThreadRuntimeCore.ts
+++ b/packages/react-ag-ui/src/runtime/AgUiThreadRuntimeCore.ts
@@ -425,38 +425,17 @@ export class AgUiThreadRuntimeCore {
       );
     }
 
-    const responsesById = new Map<string, AgUiResumeEntry>();
-    for (const entry of responses) {
-      if (!entry || typeof entry.interruptId !== "string") {
-        throw new Error(
-          "[agui] submitInterruptResponses: every entry must have an interruptId",
-        );
-      }
-      if (entry.status !== "resolved" && entry.status !== "cancelled") {
-        throw new Error(
-          `[agui] submitInterruptResponses: invalid status "${entry.status}" for interrupt ${entry.interruptId}`,
-        );
-      }
-      if (responsesById.has(entry.interruptId)) {
-        throw new Error(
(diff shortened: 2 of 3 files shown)


okisdev (approved):
verified as drift normalization rather than plain dedup: the two inline loops had already diverged in wording and check order, the shared helper adopts the per-entry shape, and the design-level policy split (submit answers everything and runs expiry, steerAway fills unanswered with cancelled) stays at the call sites. the three error-selection changes are enumerated honestly in the impact table and pinned by the new tests; accept and reject behavior is unchanged and error text is not a wire contract. suite re-run locally at 230 passing, and no other reference to the old messages exists outside the updated spec.

PR #5027, opened 2026-07-17
refactor(ui): align CommandDialog with upstream composition contract

ephraimduncan:
## What

Realigns `CommandDialog` in the vendored command component (`packages/ui/src/components/ui/{base,radix}/command.tsx`) with shadcn upstream's current composition contract:

- `showCloseButton` default: `true` → `false`
- Removed the internal `<Command className="[&_[cmdk-…]]…">` auto-wrap; `children` now render directly inside `DialogContent`
- Props type: `Omit<React.ComponentProps<typeof Dialog>, "children"> & { …; children: React.ReactNode }` (children now required)

## Why

Registry items declare bare `"command"` registryDependencies, which the shadcn CLI resolves against the **official upstream registry** — not our vendored copies. Upstream changed `CommandDialog` to render children directly (canonical usage passes an explicit `<Command>` child). A user following current shadcn examples against our vendored file gets nested cmdk roots (`<Command>` inside the auto-wrapped `<Command>`).

**Behavior change** (old → new contract): callers must now pass an explicit `<Command>` child and opt into the close button. No in-repo consumer uses `CommandDialog` (only definitions + exports), and `templates/minimal` does not ship `command.tsx`, so nothing needed migration.

## Deliberately deferred drift

Only the `CommandDialog` contract is synced. Upstream's `InputGroup`-based `CommandInput` and the `CommandItem` checked-state indicator are **not** copied — upstream's version depends on `input-group`, which has no Base twin in this repo. Upstream's `cn-*` classes are also skipped (this repo doesn't use that class system); the existing `"overflow-hidden p-0"` is kept.

## Verification

- `grep CommandDialog` across packages/apps/templates/examples → only definitions/exports
- `pnpm sync-templates` ✓, `pnpm lint` ✓, `pnpm -C apps/registry build` ✓, `packages/ui` `tsc --noEmit` ✓
- `packages/ui` is private → no changeset

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5027?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/ui/src/components/ui/base/command.tsx
+++ b/packages/ui/src/components/ui/base/command.tsx
@@ -34,14 +34,14 @@ function CommandDialog({
   description = "Search for a command to run...",
   children,
   className,
-  showCloseButton = true,
+  showCloseButton = false,
   ...props
-}: React.ComponentProps<typeof Dialog> & {
+}: Omit<React.ComponentProps<typeof Dialog>, "children"> & {
   title?: string;
   description?: string;
   className?: string;
   showCloseButton?: boolean;
-  children?: React.ReactNode;
+  children: React.ReactNode;
 }) {
   return (
     <Dialog {...props}>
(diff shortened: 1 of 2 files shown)


okisdev:
closing: the premise does not hold against upstream. three checks against the canonical sources, all taken today:

1. `shadcn-ui/ui` main (`apps/v4/registry/new-york-v4/ui/command.tsx`): `CommandDialog` still defaults `showCloseButton = true`, still auto-wraps children in `<Command className="...cmdk...">`, and keeps `children` optional. that is exactly the contract this PR removes.
2. the upstream file's last change is a 2026-03-02 formatting commit, so there was no composition contract change months before or since this PR was opened.
3. the served registry artifact (`ui.shadcn.com/r/styles/new-york-v4/command.json`), which is what a bare `"command"` registryDependency actually installs, matches main: auto-wrap present, `showCloseButton = true`, children optional.

the motivation chain in the body (vendored copies must stay in sync with what bare registry deps install) is correct, and it is exactly why this cannot merge: the current vendored copy already matches upstream, and applying this diff would make the canonical shadcn usage `<CommandDialog><CommandInput/>...</CommandDialog>` (no explicit `<Command>` child) render cmdk primitives with no cmdk root, breaking the palette and silently dropping the close button. the `InputGroup` based `CommandInput` mentioned in the body also does not exist on upstream main; `CommandInput` there is a plain `CommandPrimitive.Input` wrapper.

PR #5039, opened 2026-07-18
refactor: share parseDataUrl and httpUrlPattern from core internal

ShobhitPatra:
This PR deduplicates a small data-URL parsing helper that had been copy-pasted into four adapter packages, moving it to the shared internal surface in core.

## What

- New `packages/core/src/utils/data-url.ts` with `parseDataUrl` and `httpUrlPattern`, exported from `@assistant-ui/core/internal`, with a colocated test.
- react-ag-ui, react-langchain, and react-langgraph drop their module-private copies and import the shared ones. Call sites unchanged; the bodies were byte-identical (same regex, same return shape), so behavior is bit-for-bit the same.
- Regenerated `api-surface/assistant-ui__core.ts` (the generator snapshots the `./internal` subpath) included.

## Why

Four byte-identical copies of the same pair existed (ag-ui, langchain, langgraph, and a fourth added by #5036 in react-a2a), flagged in review on #5036. AGENTS.md says to reuse the shared core primitive instead of per-adapter copies, and `core/internal` already carries exactly this pattern (`isJSONValue`/`isRecord`, `stableStringifyToolArgs`, `getFileDataURL` all live there "so the X lives in one place").

## Notes

- react-a2a's copy is deliberately not touched: it lives in unmerged #5036 and swaps to the shared helper in the queued converter follow-up after that merges, keeping this PR conflict-free with it (verified: disjoint file sets, including the api-surface snapshots).
- Exporting a bare RegExp (`httpUrlPattern`) is a repo first; core's guard style is functions. Chosen because every call site uses only `.test()`, the regex has no `/g` flag so the shared instance is stateless, and it keeps all migrated call sites byte-identical. Happy to wrap in an `isHttpUrl()` if preferred.
- The three adapters declare core in `dependencies`, and changesets' default `updateInternalDependencies: "patch"` rewrites their core range when all four release together, so no version-floor gap for the new import.
- Existing adapter suites (ag-ui conversions.spec, langchain convertMessages.test, langgraph convertLangChainMessages.test) exercise the data-URL paths and pass unchanged; they are the regression net for the swap.
- One changeset covering all four packages (patch).

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5039?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


okisdev (approved):
verified against the extraction criterion this repo has settled on (shape identity, drift or proliferation evidence, existing dependency edge): all four copies are byte-identical down to the regex, the fourth appeared in #5036 review as live proliferation, and every consumer already declares core in dependencies. this is the parse direction, distinct from the construction-side normalizer that was previously rejected for divergent call-site shapes. during adoption I merged current main (the only conflict was the api-surface snapshot, resolved by regenerating), completed the sweep by migrating the react-a2a copy now that #5036 is on main, added react-a2a to the changeset, and re-ran all five suites green (core 629, ag-ui 230, langchain 108, langgraph 190, a2a 172). the bare RegExp export is fine as argued: no /g flag, every call site uses .test() only.

PR #5040, opened 2026-07-18
fix(react-pi): isolate event listener errors

Kinfe123:
## Summary

- isolate errors thrown by individual React Pi event subscribers
- continue delivering the shared SSE event to remaining subscribers
- keep subscriber failures out of `onStreamError` and the reconnect path
- add regression coverage and a patch changeset

## Problem

React Pi shares one SSE connection per thread and fans each event out to every subscriber. While testing multiple subscribers, I found that one callback throwing stopped the dispatch loop before later callbacks ran. The exception then escaped into `openPiEventStream`, where it was treated as a transport failure: `onStreamError` fired and the stream reconnected even though the network and event were valid.

This can leave another runtime, chat view, or telemetry subscriber stale while also causing an unnecessary reconnect.

## Fix

Catch subscriber exceptions independently, report the faulty callback with React Pi context, and continue dispatching the event. Genuine fetch, response, parsing, and stream failures still use `onStreamError` normally. This matches the isolation behavior already used by Pi's server-side supervisor and the OpenCode event source.

## Testing

- `pnpm --filter @assistant-ui/react-pi exec vitest run src/client/httpClient.test.ts`
- `pnpm --filter @assistant-ui/react-pi test` (147 tests)
- `pnpm --filter @assistant-ui/react-pi build`
- `pnpm exec oxlint packages/react-pi/src/client/httpClient.ts packages/react-pi/src/client/httpClient.test.ts`
- `pnpm exec oxfmt --write packages/react-pi/src/client/httpClient.ts packages/react-pi/src/client/httpClient.test.ts .changeset/tidy-pi-listeners.md`

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5040?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


okisdev (approved):
## review

LGTM as `patch`. verified two ways: the PR's two commits cherry-pick cleanly onto current main and pass 27/27 in `httpClient.test.ts` there, and after merging main into the branch (conflict was confined to the test file; the source hunk auto-merged byte-identical to the verified transplant) the full react-pi suite passes 167/167. with the dispatch loop reverted to main, exactly the new isolation test fails by timeout, which is the bug: one throwing subscriber starves the rest and the exception escapes into `openPiEventStream`'s outer catch, firing `onStreamError` and tearing down a healthy connection for a spurious reconnect (`eventSource.ts` routes `onEvent` through the same try as transport errors).

- the first commit's regression test hand-rolled a fetch mock and restored the `console.error` spy only on the success path, duplicating the file's existing `fakeFetch` helper and leaking the spy on assertion failure.

### what we modified

- the follow-up commit reuses `fakeFetch`, extracts the `sseResponse` helper, and moves spy cleanup to `onTestFinished` so it runs regardless of assertion outcome.
- the merge commit brings the branch onto current main; its only conflict was the test file against the #5111 per-client isolation changes, resolved to keep both suites side by side.

### potential further fixes

- the isolation idiom now exists verbatim in react-opencode and react-pi; if a third adapter grows a listener fan-out, that is the point to consider a shared helper, per the repo's rule of three.

PR #5049, opened 2026-07-19
fix(react-ai-sdk): persist pending tool approvals

### LINKED ISSUE #5046: Thread History adapter - Tool calls responses waiting for approval not triggering save (append) to messages
### Checklist

### Environment info

i will write this later

### Affected area(s)

Thread Management

### Runtime adapter in use

### Steps to reproduce

you give the chat a thread history adapter + tool calls WITH needsApproval:true and do not input your response to the tool call that is waiting for approval

Expected: after i refresh the page the approval should still be pending waiting for me.
Actual: it doesnt trigger save and doesnt save

### Browser (if applicable)

_No response_

### Code snippet or error log

```tsx

```

### Additional context

_No response_

Gujiassh:
## Summary

- persist assistant messages when a run pauses for tool-call approval
- keep other interrupted message statuses on their existing persistence path
- add a regression test and patch changeset for `@assistant-ui/react-ai-sdk`

Fixes #5046

## Root cause

`useExternalHistory` appends messages when a run stops, but its readiness check only accepted messages without a status or with a terminal `complete`/`incomplete` status. Tool calls waiting for approval use `requires-action/tool-calls`, so the persistence pass skipped the assistant message and a reload lost the pending approval state.

This change admits that specific stable approval state into the existing append flow. It does not change the history adapter schema, stored message format, parent mapping, or public API.

## Validation

- `pnpm --filter @assistant-ui/react-ai-sdk exec vitest run src/ui/use-chat/useExternalHistory.test.ts` (7 tests)
- `pnpm --filter @assistant-ui/react-ai-sdk test` (157 tests)
- `pnpm --filter @assistant-ui/react-ai-sdk build`
- `pnpm lint` (existing warnings remain elsewhere in the workspace)
- `git diff --check`

The new regression was the only failing focused test against the unchanged readiness predicate because `append` was never called, then passed with this change. A direct package `tsc --noEmit` still reaches the existing `src/generativeTools.test.ts:526` `TS2769`; the package declaration build passes and the affected files produce no diagnostics.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5049?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/quiet-tools-wait.md
+++ b/.changeset/quiet-tools-wait.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react-ai-sdk": patch
+---
+
+Persist assistant messages while tool calls await approval.
--- a/packages/react-ai-sdk/src/ui/use-chat/useExternalHistory.test.ts
+++ b/packages/react-ai-sdk/src/ui/use-chat/useExternalHistory.test.ts
@@ -1,6 +1,6 @@
 // @vitest-environment jsdom
 
-import { renderHook } from "@testing-library/react";
+import { act, renderHook, waitFor } from "@testing-library/react";
 import { describe, expect, it, vi } from "vitest";
 import type {
   AssistantRuntime,
@@ -9,6 +9,7 @@ import type {
   ThreadHistoryAdapter,
   ThreadMessage,
 } from "@assistant-ui/core";
+import { bindExternalStoreMessage } from "@assistant-ui/core";
 
 vi.mock("@assistant-ui/store", () => ({
   useAui: () => ({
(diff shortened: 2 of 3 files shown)


okisdev:
thanks for taking this on, and for the clean regression test; the diagnosis of the `isReady` guard was exactly right. closing in favor of #5051, which landed the same guard change plus the machinery the one line version was missing: once a paused message is appended, its later content can only reach storage through the optional `update`, so adapters without `update` (including the documented DIY shape) would silently never persist the post-approval tool output and final answer, and new-id continuations were dropped even on Assistant Cloud. #5051 gates early persistence on `update` support, tracks persisted inner ids so continuations are appended instead of silently skipped, and defers run telemetry until completion; the full analysis is in #5050. the fix ships in the next @assistant-ui/react-ai-sdk patch.

Issue #5038, opened 2026-07-18
bug: `mountTopAnchorReserve` tears down the reserve in the between-turns anchor gap, collapsing scroll height under an at-bottom user (composer visibly jumps on send)


JacobCutts13:
### Checklist

- [x] I have searched existing issues and this is not a duplicate
- [x] This is a reproducible bug, not a usage question

(related but distinct: #4104 documents a few-pixel send jitter root-caused to `content-visibility` styles in the registry `thread.tsx` — we had already removed those styles and still hit this. This report is a several-hundred-pixel scroll-height collapse originating inside `mountTopAnchorReserve` itself, present on `main` and in the latest release. Also related: #3648, the same class of rAF-vs-React-commit timing issue.)

### Environment info

```
Environment:
  OS:               macOS (arm64, devicePixelRatio: 2 — retina)
  Node.js:          v22.22.2
  Package Manager:  pnpm 10.33.0
  Framework:        Next.js 16.2.10

Packages:
  @assistant-ui/react           0.14.26  (teardown code unchanged in 0.14.27 and current main)
  @assistant-ui/react-markdown  0.14.5
```

### Affected area(s)

React / UI Components

### Runtime adapter in use

useExternalStoreRuntime

### Steps to reproduce

The key precondition is a **multi-frame gap between turns during which no anchor is active**. With an external-store runtime whose `isRunning` derives from server state (poll/SSE), this gap is just real network latency:

1. `<ThreadPrimitive.Viewport turnAnchor="top" autoScroll>`, external-store runtime, `isRunning` driven by server acknowledgment.
2. Complete at least one turn, so the previous turn's reserve element is mounted with a substantial height (~400px in our traces), and sit at the bottom.
3. Send a message. The message list now ends `[..., assistant, user]` → `getActiveTopAnchorTurn` returns `null` → `apply()` takes the bail-out branch → **the previous turn's reserve is zeroed and removed**.
4. The new anchor can't mount until `isRunning` flips true (server ack) and the assistant placeholder exists — several frames later on a real network.

Local/dev setups with near-instant acks make the gap ~0 frames and mask the bug entirely: in our automated testing (4,800+ instrumented frames, headless Chromium), the reserve was only ever *resized*, never torn down, and the composer never moved a pixel. On a real deployment with real latency it reproduces on every send.

### Expected vs actual behavior

Expected: sending a message from the bottom produces one clean motion (the send reveal, then the pin-to-top scroll once the run starts).

[...]

PR #5059, opened 2026-07-20
fix(react): defer auto-scroll during top-anchored runs

### LINKED ISSUE #5037: bug: `autoScroll`'s instant bottom-snap ignores `hasActiveTopAnchor()`, front-running the top-anchor reserve's smooth scroll
### Checklist

(closest: #4104 — same symptom family, different mechanism; that one was root-caused to `content-visibility` styles in the registry `thread.tsx`. #4110 proposed smooth→instant in the reserve as a workaround, which was correctly rejected. This report is about a missing guard in `useThreadViewportAutoScroll` itself, present on `main` and in the latest release.)

### Environment info

```
Environment:
  OS:               macOS (arm64, devicePixelRatio: 2 — retina)
  Node.js:          v22.22.2
  Package Manager:  pnpm 10.33.0
  Framework:        Next.js 16.2.10

Packages:
  @assistant-ui/react           0.14.26  (bug verified still present in 0.14.27 and current main)
  @assistant-ui/react-markdown  0.14.5
```

### Affected area(s)

React / UI Components

### Runtime adapter in use

useExternalStoreRuntime

### Steps to reproduce

1. `<ThreadPrimitive.Viewport turnAnchor="top" autoScroll>` — note `autoScroll` explicitly enabled (it defaults to `false` under `turnAnchor="top"`; we enable it so poll/SSE arrivals outside a turn get stick-to-bottom behavior, which works well otherwise).
2. Populate a thread with enough history to scroll; sit at the bottom.
3. Send a message and let a run start, so `mountTopAnchorReserve` activates: it inserts/resizes its reserve element and requests a **smooth** scroll to the pin-near-top target.
4. The reserve's own DOM insertion/resize trips `useOnResizeContent`. If `isAtBottom` is still `true` at that moment, the handler's third branch fires `scrollToBottom("instant")`.

Expected: once a top anchor is active, the reserve owns viewport positioning — the resize handler's first branch already encodes exactly this rule ("Let the top-anchor reserve own scrolling while a run starts to avoid a bottom-scroll race").

Actual: only the *pending-intent* branch defers to the anchor. The `autoScroll` branch below it doesn't check `hasActiveTopAnchor()`, so it fires an **instant** snap on the reserve's own mutation. Because the reserve is deliberately sized so that max-scroll coincides with the pin target (`computeTopAnchorReserve`), the instant snap lands almost exactly where the reserve's *smooth* scroll was about to animate to — the intended smooth motion is replaced by a teleport. On the next frame the reserve sees it's within 1px of target and skips its own scroll, so the instant snap "wins" every time the timing lines up.

### Browser (if applicable)

Chrome (macOS, retina); also observed in headless Chromium (Playwright)

### Code snippet or error log

```tsx
// packages/react/src/primitives/thread/useThreadViewportAutoScroll.ts (~L169)
const scrollBehavior = scrollingToBottomBehaviorRef.current;
if (scrollBehavior && hasActiveTopAnchor()) {
  // Let the top-anchor reserve own scrolling while a run starts to avoid a bottom-scroll race.
  scrollingToBottomBehaviorRef.current = null;
} else if (scrollBehavior) {
  scrollToBottom(scrollBehavior);
} else if (autoScroll && threadViewportStore.getState().isAtBottom) {
  scrollToBottom("instant"); // ← fires during an active top anchor; no hasActiveTopAnchor() check
}
```

Suggested one-line fix, verified in production via a pnpm patch:

```diff
-} else if (autoScroll && threadViewportStore.getState().isAtBottom) {
+} else if (autoScroll && !hasActiveTopAnchor() && threadViewportStore.getState().isAtBottom) {
   scrollToBottom("instant");
 }
```

This completes the ownership rule the first branch already documents. All other `autoScroll` behavior (stick-to-bottom for non-turn content arrivals) is unaffected, since no anchor is active in those cases.

One consideration for maintainers: `topAnchorTurn`/the anchor element persist after a run completes (by design). With this guard, an at-bottom stick for content arriving *between* turns is also suppressed while the stale anchor lingers. If that's undesirable, the guard could be narrowed to "anchor active AND run in progress" — happy to adjust the PR either way.

### Additional context

Found while consolidating scroll behavior in a production app (chat with `turnAnchor="top"` + `autoScroll`), against the installed 0.14.26, the published 0.14.27 tarball, and current `main` source.

**How this was verified, precisely:** we instrumented every branch of the resize handler with logging in the *guarded* build. The logs repeatedly show frames with `scrollBehavior: null, hasActiveTopAnchor: true, isAtBottom: true` where the added guard was the only thing preventing `scrollToBottom("instant")` — i.e. the unguarded branch demonstrably fires in exactly the state where the reserve owns positioning. What we did **not** isolate is this branch's standalone visual severity: in our end-to-end reproduction the dominant visible jump came from the reserve teardown (#5038), with this branch firing during that trace's no-anchor window. We report this one as an ownership-rule violation with mechanism-level evidence; how visible it is on its own will vary with timing and hardware.

Filed together with #5038 (`mountTopAnchorReserve` tearing down the reserve in the between-turns anchor gap) — the two were found in the same investigation but have separate mechanisms and separate one-line-scale fixes.

> **Disclosure:** this issue was investigated and written by Claude (Anthropic's AI assistant) during a pair-debugging session, at the direction of @JacobCutts13, who reproduced the behavior in a real browser and verified the fix live. Happy to provide the full frame-level traces on request.

Gujiassh:
## Summary

- defer resize-driven instant auto-scroll while a live top-anchored run owns viewport positioning
- preserve at-bottom auto-scroll after the run completes, even though the previous anchor remains mounted
- cover both sides of the run boundary with a regression test and add a patch changeset

Fixes #5037

## Test plan

- `pnpm --filter @assistant-ui/react exec vitest run src/primitives/thread/useThreadViewportAutoScroll.test.tsx`
- `pnpm lint`

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5059?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/calm-anchors-scroll.md
+++ b/.changeset/calm-anchors-scroll.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react": patch
+---
+
+Prevent instant auto-scroll from interrupting active top-anchored runs.
--- a/.changeset/calm-anchors-scroll.md
+++ b/.changeset/calm-anchors-scroll.md
@@ -2,4 +2,4 @@
 "@assistant-ui/react": patch
 ---
 
-Prevent instant auto-scroll from interrupting active top-anchored runs.
+fix: prevent instant auto-scroll from interrupting active top-anchored runs
--- a/packages/react/src/primitives/thread/useThreadViewportAutoScroll.test.tsx
+++ b/packages/react/src/primitives/thread/useThreadViewportAutoScroll.test.tsx
@@ -11,6 +11,7 @@ import {
   vi,
 } from "vitest";
 import { useEffect, useState, type FC, type PropsWithChildren } from "react";
+import { useAuiState } from "@assistant-ui/store";
 import { AssistantRuntimeProvider } from "../../context";
 import * as MessagePrimitive from "../message";
 import { ThreadPrimitiveMessages } from "./ThreadMessages";
@@ -39,6 +40,7 @@ const getMaxScrollTop = (element: Element) =>
   Math.max(0, element.scrollHeight - element.clientHeight);
 
 let forceShortViewportMeasurement = false;
+let viewportMeasurementOffset = 0;
 const resizeObserverCallbacks = new Set<ResizeObserverCallback>();
 
 class TestResizeObserver {
@@ -111,7 +113,9 @@ beforeAll(() => {
       if (this.getAttribute("data-testid") !== "viewport") return 0;
       if (forceShortViewportMeasurement) return this.clientHeight;
       return (
-        document.querySelectorAll('[data-testid="thread-message"]').length * 80
+        document.querySelectorAll('[data-testid="thread-message"]').length *
+          80 +
+        viewportMeasurementOffset
       );
     },
   });
@@ -127,6 +131,7 @@ beforeAll(() => {
 
 afterEach(() => {
   forceShortViewportMeasurement = false;
+  viewportMeasurementOffset = 0;
   resizeObserverCallbacks.clear();
   cleanup();
 });
@@ -173,6 +178,11 @@ const BottomAnchorThread = () => (
     </ThreadPrimitiveViewport>
   </ThreadPrimitiveRoot>
 );
+
+const RunState = () => {
+  const isRunning = useAuiState((s) => s.thread.isRunning);
+  return <output data-testid="run-state">{String(isRunning)}</output>;
+};
 
 const SyncRuntimeProvider: FC<PropsWithChildren> = ({ children }) => {
   const runtime = useLocalRuntime(adapter, { initialMessages: messages });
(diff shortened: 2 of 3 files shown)


okisdev (approved):
## review

LGTM as `patch`. the narrowed guard (`isRunning && hasActiveTopAnchor()`) is the right variant; a bare `hasActiveTopAnchor()` check would have suppressed the between-turns stick, since the anchor element deliberately stays mounted after a run completes.

- the changeset message ("Prevent instant auto-scroll from interrupting active top-anchored runs.") did not follow the changelog convention for `@assistant-ui/react` entries, which use a lowercase `fix:` prefixed sentence without a trailing period.

### what we modified

- rewords the changeset to `fix: prevent instant auto-scroll from interrupting active top-anchored runs` to match the existing changelog entries.

### potential further fixes

- #5038 (the reserve teardown collapse in the between-turns anchor gap) was found in the same investigation but has a separate mechanism; it stays out of scope here and still needs its own fix.
- the regression test pins the guard at the branch level; the real microtask versus rAF ordering that makes the unguarded snap always win cannot be expressed in jsdom, so browser level coverage stays out of scope.

PR #5081, opened 2026-07-20
fix(core): settle frame tool calls on dispose

Kinfe123:
## Summary

- reject in-flight AssistantFrame tool requests when the host is disposed
- reject calls through stale tool handlers after disposal
- preserve successful result handling and clear request timers during disposal

## Why

When an AssistantFrame unmounts or navigates away during a tool call, `dispose()` currently clears the pending-request map without settling its promises. The timeout later sees no map entry either, so the tool promise remains pending forever and can leave the model run waiting indefinitely.

A minimal reproduction is:

```ts
const result = tool.execute(args, context);
host.dispose();
await result; // never settled before this change
```

Pending calls now reject with `AssistantFrameHost has been disposed`. Requests made through a stale tool reference after disposal reject immediately without posting to the frame.

## Testing

- `pnpm --filter @assistant-ui/core test -- src/model-context/frame/host.test.ts` (63 files, 584 tests)
- `pnpm --filter @assistant-ui/core exec tsc --noEmit`
- `pnpm --filter @assistant-ui/core... build`
- `pnpm lint`

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5081?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/quiet-frames-settle.md
+++ b/.changeset/quiet-frames-settle.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/core": patch
+---
+
+fix: reject pending AssistantFrame tool calls when the host is disposed
--- a/api-surface/assistant-ui__core.ts
+++ b/api-surface/assistant-ui__core.ts
@@ -291,6 +291,7 @@ declare class AssistantFrameHost implements ModelContextProvider {
   private _requestCounter;
   private _iframeWindow;
   private _targetOrigin;
+  private _disposed;
   constructor(iframeWindow: Window, targetOrigin?: string);
   private handleMessage;
   private updateContext;
--- a/api-surface/assistant-ui__react.ts
+++ b/api-surface/assistant-ui__react.ts
@@ -454,6 +454,7 @@ declare class AssistantFrameHost implements ModelContextProvider {
   private _requestCounter;
   private _iframeWindow;
   private _targetOrigin;
+  private _disposed;
   constructor(iframeWindow: Window, targetOrigin?: string);
   private handleMessage;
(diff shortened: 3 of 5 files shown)


okisdev (approved):
## review

LGTM as `patch`. verified in a clean worktree at `47fe13b`: the new suite passes 3/3, and with `host.ts` reverted to main both dispose regressions fail by test timeout, which is the bug in its purest form: the tool promise never settles because `dispose()` cleared the pending map and the 30s timeout callback then finds no entry to reject. the timer-cleanup claim holds through the existing wrapper (the map stores a `clearTimeout`-wrapping reject, so settling on dispose also clears the timer, which the `getTimerCount` assertion pins). no overlap with the recent frame work: #5163 touched the provider side and #5096 lives in safe-content-frame.

- none in the shipped commit; source, tests, and changeset landed complete. the api-surface snapshots (the new private `_disposed` member) were left to the autofix bot.

### what we modified

- the autofix commit regenerates the two api-surface files for the added private member.

### potential further fixes

- `callTool` still ignores the execution context's `abortSignal`, so an aborted run waits out the full 30s timeout instead of cancelling the frame request. pre-existing and orthogonal to this fix, worth an issue if frame tools see heavier use.

PR #5085, opened 2026-07-21
fix(react): preserve top anchor reserve between turns

### LINKED ISSUE #5038: bug: `mountTopAnchorReserve` tears down the reserve in the between-turns anchor gap, collapsing scroll height under an at-bottom user (composer visibly jumps on send)

Gujiassh:
## Summary

* preserve the existing top-anchor reserve while an external runtime is between turns and the next anchor has not mounted yet
* clear sticky top-anchor state when its message IDs leave the current branch, then gate reserve preservation on that valid state
* move the preserved reserve after optimistic content, then let the next active target reposition and resize it
* keep cleanup when top anchoring is disabled, the viewport disappears, the active turn is invalid, or the reserve effect unmounts

Fixes assistant-ui/assistant-ui#5038.

## Validation

* added a current-main red regression that observed the reserve being disconnected during the between-turns gap
* reserve lifecycle and top-anchor state tests cover between-turn preservation, invalid cleanup, and thread switch, new chat, last-turn deletion, and branch swap transitions
* complete `@assistant-ui/react` suite passes (44 files, 466 tests)
* `@assistant-ui/react` production build passes (928 entries)
* root lint and format checks pass with existing warnings

Standalone package `tsc --noEmit` still reports the existing `ComposerRoot.test.tsx` `setMultiline` diagnostics on current main; this change does not touch that test.

Review in cubic

### DIFF
--- a/.changeset/tidy-anchors-wait.md
+++ b/.changeset/tidy-anchors-wait.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react": patch
+---
+
+fix: preserve the top-anchor reserve between turns
--- a/packages/react/src/primitives/thread/ThreadViewport.tsx
+++ b/packages/react/src/primitives/thread/ThreadViewport.tsx
@@ -111,16 +111,17 @@ const useTopAnchorTurn = (enabled: boolean) => {
     if (!enabled) return undefined;
     return getActiveTopAnchorTargetId(s.thread);
   });
-  const threadMessages = useAuiState((s) => s.thread.messages);
   const topAnchorTurn = useThreadViewport((s) => s.topAnchorTurn);
   const activeTurn = useMemo(() => {
     if (!activeAnchorId || !activeTargetId) return null;
     return { anchorId: activeAnchorId, targetId: activeTargetId };
   }, [activeAnchorId, activeTargetId]);
 
-  const topAnchorTurnIsValid = useMemo(
-    () => isTopAnchorTurnValid(topAnchorTurn, threadMessages),
-    [threadMessages, topAnchorTurn],
+  const topAnchorTurnIsValid = useAuiState(
+    (s) =>
+      enabled &&
+      !!topAnchorTurn &&
+      isTopAnchorTurnValid(topAnchorTurn, s.thread.messages),
   );
 
   useLayoutEffect(() => {
--- a/packages/react/src/primitives/thread/ThreadViewport.tsx
+++ b/packages/react/src/primitives/thread/ThreadViewport.tsx
@@ -23,6 +23,7 @@ import { useTopAnchorReserve } from "./topAnchor/useTopAnchorReserve";
 import {
   getActiveTopAnchorAnchorId,
   getActiveTopAnchorTargetId,
+  isTopAnchorTurnValid,
 } from "./topAnchor/topAnchorTurn";
 
 export namespace ThreadPrimitiveViewport {
@@ -110,10 +111,23 @@ const useTopAnchorTurn = (enabled: boolean) => {
     if (!enabled) return undefined;
     return getActiveTopAnchorTargetId(s.thread);
   });
+  const threadMessages = useAuiState((s) => s.thread.messages);
+  const topAnchorTurn = useThreadViewport((s) => s.topAnchorTurn);
   const activeTurn = useMemo(() => {
     if (!activeAnchorId || !activeTargetId) return null;
     return { anchorId: activeAnchorId, targetId: activeTargetId };
   }, [activeAnchorId, activeTargetId]);
+
+  const topAnchorTurnIsValid = useMemo(
+    () => isTopAnchorTurnValid(topAnchorTurn, threadMessages),
+    [threadMessages, topAnchorTurn],
+  );
+
+  useLayoutEffect(() => {
+    if (!topAnchorTurn || topAnchorTurnIsValid) return;
+
+    threadViewportStore.getState().setTopAnchorTurn(null);
+  }, [threadViewportStore, topAnchorTurn, topAnchorTurnIsValid]);
 
   useLayoutEffect(() => {
     if (!activeTurn) return;
--- a/packages/react/src/primitives/thread/topAnchor/mountTopAnchorReserve.test.ts
+++ b/packages/react/src/primitives/thread/topAnchor/mountTopAnchorReserve.test.ts
@@ -122,6 +122,7 @@ describe("mountTopAnchorReserve", () => {
     ) as HTMLElement;
     const optimisticUserMessage = document.createElement("div");
     reserveHost.append(optimisticUserMessage);
+    const appendSpy = vi.spyOn(reserveHost, "append");
 
     setState({
       turnAnchor: "top",
@@ -134,6 +135,17 @@ describe("mountTopAnchorReserve", () => {
     expect(reserve.isConnected).toBe(true);
     expect(reserve.style.height).toBe("60px");
     expect(reserve.previousElementSibling).toBe(optimisticUserMessage);
+    expect(appendSpy).toHaveBeenCalledTimes(1);
+
+    setState({
+      turnAnchor: "top",
+      element: { viewport, anchor: null, target: null },
+      targetConfig: null,
+      topAnchorTurn: activeTopAnchorTurn,
+    });
+    vi.runOnlyPendingTimers();
+
+    expect(appendSpy).toHaveBeenCalledTimes(1);
 
     const nextAnchor = document.createElement("div");
     const nextTarget = document.createElement("div");
--- a/packages/react/src/primitives/thread/topAnchor/mountTopAnchorReserve.test.ts
+++ b/packages/react/src/primitives/thread/topAnchor/mountTopAnchorReserve.test.ts
@@ -150,6 +150,45 @@ describe("mountTopAnchorReserve", () => {
     expect(reserve.isConnected).toBe(false);
   });
 
+  it("removes the reserve when only the registered anchor unmounts", () => {
+    const viewport = document.createElement("div");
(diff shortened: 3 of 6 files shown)


AVGVSTVS96 (changes requested):
Nice find on the gap teardown, and the partial-unmount branch looks right. Blocking on one thing though:

the `!anchor && !target && !clamp` state isn't unique to the between-turns gap. Thread switch, new chat, deleting the last turn, and branch swaps all produce the same all-null state while the viewport (and the reserve's parent) stay mounted. Since the reserve is a manual DOM node React doesn't own, it just... stays. I repro'd this: switch away from a completed turn and the old ~400px reserve is still sitting at the bottom of the new thread. It also keeps `aui_message-group` non-empty, which defeats `empty:hidden` and shoves the centered new-chat layout around in the registry UI.

`main` tears the reserve down in all of these, so as-is this trades the scroll-collapse bug for a stale-space one.

`topAnchorTurn` is close to the right signal but not quite sufficient on its own: it's intentionally sticky and only clears on `thread.initialize`/`threadListItem.switchedTo`, so deletions/branch changes leave it pointing at messages that no longer exist. I think the clean fix is to make `topAnchorTurn` clear whenever its stored ids leave the current branch, then gate the preserve branch on it being non-null. That gives you an exact "the gap is transient, a next turn is coming" discriminator instead of inferring it from element nullness.

Could you also add regression tests for the terminal transitions (thread switch / new chat / last-turn deletion, asserting the reserve is removed and zeroed), and drop a short comment on the preserve branch explaining the invariant? The three-way guard is pretty hard to trust on a cold read without it.

Gujiassh:
Thanks for the detailed review. Pushed 79199c5e.

- `topAnchorTurn` now clears in a layout effect when either ID leaves current branch messages, covering thread switches, new chats, last-turn deletion, and branch swaps while retaining transient between-turn gaps.
- Reserve preservation now requires non-null validated state; otherwise it is removed and zeroed.
- Added regression coverage for valid preservation, invalid cleanup, and all four transitions.

Verification: React tests 44 files/466 passed; package build, lint (existing warnings), formatting, and diff check passed.

AVGVSTVS96:
Pushed 554a7a3 (maintainer edit) to close the last gap found in review, and did a full local verification of the branch.

**The remaining leak:** `isTopAnchorTurnValid` only checked ID presence, so an external store syncing in a turn completed elsewhere (`[..., user-A, assistant-A]` → `[..., user-A, assistant-A, user-B, assistant-B]` with `isRunning` false) kept the stored turn valid while all elements deregistered, preserving a stale reserve indefinitely. The predicate now requires the anchor to immediately precede the target with only pending user messages after it, which also makes the validity scan cheaper. Added regression tests for the synced-turn and pair-separation cases; the between-turns gap and multi-send remain valid.

**Verification on this branch (local):**
- 22/22 topAnchor unit tests pass; typecheck and lint clean.
- Real-browser run (instrumented external-store harness, per-frame scrollHeight/reserve tracing): reserve preserved through an 800ms ack gap with a pixel-stable composer, exactly one smooth pin-to-top scroll, and reserve teardown within ~2 frames on thread switch, new chat, external resync, and last-turn deletion.
- One transient `scrollHeight` dip observed during the gap was root-caused to `content-visibility` intrinsic sizing on freshly inserted message roots: it reproduces identically with `turnAnchor="bottom"` and is unrelated to this PR (same class as #4104).

Thanks @Gujiassh for the careful iterations on this — merging once CI is green.

AVGVSTVS96 (approved):
All requested changes landed (turn-validity discriminator, strengthened to require the trailing pair in 554a7a3). Verified locally: 22/22 unit tests, typecheck/lint clean, and a real-browser run confirming reserve preservation through the gap plus teardown on thread switch, new chat, external resync, and deletion.

PR #5088, opened 2026-07-21
fix: preserve falsy runtime state

Kinfe123:
## Summary

- preserve `false`, `0`, and empty-string runtime state in local runtime updates and external message conversion
- include those values in React data-stream requests while continuing to omit `null`
- add regression coverage for all three paths

## Why

Runtime state accepts JSON values, so boolean, numeric, and string sentinel states are valid. Truthiness checks treated these values as absent, which could silently reset converted messages to `null` or omit state from a backend request.

The core paths now check specifically for `undefined`. The data-stream request uses nullish coalescing so it preserves falsy values without changing the existing behavior that omits `null` from JSON requests.

## Verification

- reproduced all three falsy values being dropped before the fix
- `pnpm --filter @assistant-ui/core test` (587 tests)
- `pnpm --filter @assistant-ui/react-data-stream test` (12 tests)
- `pnpm lint`
- `pnpm --filter @assistant-ui/core --filter @assistant-ui/react-data-stream build`

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5088?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/falsy-runtime-state.md
+++ b/.changeset/falsy-runtime-state.md
@@ -1,0 +1,6 @@
+---
+"@assistant-ui/core": patch
+"@assistant-ui/react-data-stream": patch
+---
+
+fix: preserve falsy runtime state values
--- a/packages/core/src/react/runtimes/external-message-converter.test.tsx
+++ b/packages/core/src/react/runtimes/external-message-converter.test.tsx
@@ -99,4 +99,25 @@ describe("useExternalMessageConverter", () => {
       text: "HELLO",
     });
   });
+
+  it.each([
+    ["false", false],
+    ["zero", 0],
+    ["an empty string", ""],
+  ])("preserves %s assistant state", (_label, state) => {
+    const callback: useExternalMessageConverter.Callback<TestMessage> = (
+      message,
+    ) => ({
+      role: message.role,
+      id: message.id,
+      content: [{ type: "text", text: message.text }],
+      ...(message.role === "assistant"
+        ? { metadata: { unstable_state: state } }
+        : undefined),
+    });
+
+    const { result } = renderConverter({ callback });
+
+    expect(result.current[1]?.metadata.unstable_state).toBe(state);
+  });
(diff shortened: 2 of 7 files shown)


okisdev (approved):
## review

LGTM as `patch`. verified in a clean worktree at `f03c67c`: all three touched suites pass (31 in core, 4 in react-data-stream), and with the three source files reverted to main exactly the nine falsy cases fail (three per path) while the null-omit case passes on both sides, so the #2203 contract of never sending `null` to the backend is preserved intact. the `!== undefined` idiom also matches what react-ag-ui already does for the same field (`AgUiThreadRuntimeCore.ts`), so core is being aligned to the existing sibling pattern rather than inventing a new one.

- the description covers the data-stream `null` behavior but not the converter and local-runtime side of it: an explicit `unstable_state: null` used to be skipped (previous state kept) and now overwrites. traced through the assistant-stream accumulator this is safe (it seeds from the current message state, so it never emits `null` over a live value) and it is the more correct semantics, but it deserved a sentence in the body.
- the new data-stream test invokes the component as a plain function, which only holds while `useLocalRuntime` stays the hook's sole hook call. the pattern has precedent in core, so fine as is, just brittle by design.

### what we modified

- nothing; the single commit landed complete with source, tests, and the two-package changeset.

### potential further fixes

- if `useDataStreamRuntime` ever gains another hook call, the plain-call test needs a renderer or a react mock like the core precedent uses. no action now.
- the explicit-null-clears-state semantics could be stated in the `unstable_state` JSDoc, but the field is `unstable_` surface, so low priority.

PR #5089, opened 2026-07-21
fix: avoid false copy success without a clipboard handler

Kinfe123:
## What changed

- return without setting `isCopied` when no `copyToClipboard` handler is available
- preserve success feedback after a real clipboard write and suppress feedback after rejected writes
- document the platform clipboard writer required by React Native and Ink

## Why

React Native and Ink do not provide a default clipboard implementation. Their official primitive examples omitted `copyToClipboard`, so pressing Copy ran a successful no-op and displayed `Copied!` even though nothing reached the clipboard.

## Verification

- reproduced the bug with a regression test that failed before the source change
- `pnpm --filter @assistant-ui/core test` (584 tests)
- `pnpm lint`
- `pnpm --filter @assistant-ui/docs build`

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5089?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/action-bar-copy-feedback.md
+++ b/.changeset/action-bar-copy-feedback.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/core": patch
+---
+
+fix: avoid reporting copy success without a clipboard handler
--- a/apps/docs/content/docs/ink/primitives.mdx
+++ b/apps/docs/content/docs/ink/primitives.mdx
@@ -768,9 +768,11 @@ import { ActionBarPrimitive } from "@assistant-ui/react-ink";
 Pressable that copies the message content. Pass a platform clipboard writer through `copyToClipboard`. Supports function-as-children for copy state feedback.
 
 ```tsx
+import clipboard from "clipboardy";
+
 <ActionBarPrimitive.Copy
   copiedDuration={3000}
-  copyToClipboard={copyToClipboard}
+  copyToClipboard={clipboard.write}
 >
   {({ isCopied }) => <Text>{isCopied ? "[Copied!]" : "[Copy]"}</Text>}
 </ActionBarPrimitive.Copy>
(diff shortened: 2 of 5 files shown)


okisdev (approved):
## review

LGTM as `patch`. verified in a clean worktree at `aa3a182`: the new suite passes, and with the source file reverted to main the regression test fails exactly as claimed. also confirmed the web primitive is untouched by this change since it always injects its own `navigator.clipboard` writer, so the disabled fold only bites the React Native and Ink paths where the false `Copied!` actually happened.

- the early return in `useActionBarCopy` stopped the false success feedback, but it left the button enabled and silently inert when no writer is configured: the returned `disabled` only reflected message state, while the web primitive already treats a missing writer as disabled (`packages/react/src/primitives/actionBar/ActionBarCopy.tsx`).
- the React Native docs snippet passed a `copyToClipboard` that was never defined, so it was not copy-pasteable, and the Ink page had no concrete writer or install guidance at all.

### what we modified

- folds the missing handler into the hook's disabled state (`disabled: disabled || !copyToClipboard`) so React Native and Ink render the button disabled instead of pressable but dead, matching the web semantics, and pins it with a test assertion.
- defines the writers in the docs: `expo-clipboard` with a `didCopy` check on the React Native page, `clipboardy` plus the install command on the Ink page, and corrects the prop tables to the actual `(text: string) => void | Promise<void>` signature.

### potential further fixes

- react-native and react-ink still ship no default writer, which is deliberate: neither package can carry `expo-clipboard` or `clipboardy` as a dependency, so the documented host-supplied writer is the contract.
- no Ink example exercises `ActionBarCopy`, so the docs snippet is the only executable reference for the clipboardy setup; wiring a copy button into an Ink example would cover it, low priority.

PR #5090, opened 2026-07-21
fix(core): mark a cancelled run incomplete when the adapter returns cleanly

VihaanAgarwal:
### What breaks

Stopping a run in `LocalRuntime` can leave the assistant message marked `complete` instead of `incomplete` / `cancelled`. The partial answer then looks like a finished one: it is persisted to the history adapter as a completed message, reported to cloud telemetry with status `completed`, and any UI branching on the cancelled status never fires.

Repro with a streaming adapter:

```tsx
const adapter: ChatModelAdapter = {
  async *run({ abortSignal }) {
    for await (const chunk of myStream) {
      if (abortSignal.aborted) return; // cooperative stop
      yield { content: [{ type: "text", text }] };
    }
  },
};
```

Send a message, press stop mid-stream, and the message settles as `complete`.

### Root cause

`performRoundtrip` has two ways to notice a cancellation, and both need the adapter to cooperate in a specific way:

- the `abortSignal.aborted` check inside the `for await` loop, which is only reached if the generator yields *again* after the abort
- an `AbortError` propagating out of the adapter, caught below

An adapter that returns cleanly on abort hits neither, so `message.status` is still `running` when the loop ends and the run is finalized as `complete`. That is the shape the docs actually recommend for the surrounding SDK call: *"Swallow `AbortError` (it is the user cancelling)"* ([local-runtime best practices](https://www.assistant-ui.com/docs/runtimes/custom/local-runtime#best-practices)). The non-streaming branch had no abort check at all, so a promise resolving after the abort always overwrote the cancellation.

The shipped adapters are unaffected because they pass `abortSignal` straight to `fetch`, which throws.

### The fix

Finalize against the abort signal rather than assuming the adapter signalled the cancellation itself:

```ts
if (message.status.type === "running") {
  updateMessage({
    status: abortSignal.aborted
      ? { type: "incomplete", reason: "cancelled" }
      : { type: "complete", reason: "unknown" },
  });
}
```

An adapter that sets a terminal status explicitly is still respected, since the branch only runs while the status is `running`. The in-loop check is kept: it stops consuming the stream immediately instead of waiting for the generator to finish.

### Tests

Three cases in `local-thread-runtime-core.test.ts`: a streaming adapter that returns after the abort, a non-streaming adapter that resolves after the abort, and a normal run that must stay `complete`. The first two fail on `main` with `{ type: "complete", reason: "unknown" }`.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5090?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/cancelled-run-status.md
+++ b/.changeset/cancelled-run-status.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/core": patch
+---
+
+fix: mark a cancelled local runtime run incomplete when the adapter returns without throwing
--- a/packages/core/src/runtimes/local/local-thread-runtime-core.test.ts
+++ b/packages/core/src/runtimes/local/local-thread-runtime-core.test.ts
@@ -408,6 +408,74 @@ describe("LocalThreadRuntimeCore tool approvals", () => {
   });
 });
 
+describe("LocalThreadRuntimeCore cancellation", () => {
+  it("marks the message cancelled when a streaming adapter returns after abort", async () => {
+    let released!: () => void;
+    const streaming = new Promise<void>((resolve) => {
+      released = resolve;
+    });
+
+    const thread = createThread({
+      async *run({ abortSignal }) {
+        yield { content: [{ type: "text", text: "partial" }] };
+        await streaming;
+        if (abortSignal.aborted) return;
+        yield { content: [{ type: "text", text: "partial answer" }] };
+      },
+    });
+
+    const appendPromise = thread.append(userMessage("hi"));
+    await flush();
+
+    thread.cancelRun();
+    released();
+    await appendPromise;
+
+    expect(thread.messages.at(-1)?.status).toEqual({
+      type: "incomplete",
+      reason: "cancelled",
+    });
+  });
+
+  it("marks the message cancelled when a non-streaming adapter resolves after abort", async () => {
+    let released!: () => void;
+    const pending = new Promise<void>((resolve) => {
+      released = resolve;
+    });
+
+    const thread = createThread({
+      async run() {
+        await pending;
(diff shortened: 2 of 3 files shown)


okisdev (approved):
## review

LGTM as `patch`.

verified in a worktree: the full core suite passes (584/584 with the 3 new tests), and reverting only `local-thread-runtime-core.ts` makes the two abort tests fail with exactly the reported `{ type: "complete", reason: "unknown" }`, so the regression tests genuinely pin the fix and the bug is live on main without it. the mechanism claim checks out line by line: the in-loop abort check needs the generator to yield again, the non-streaming branch had no check at all, and the catch only helps adapters that throw, which is precisely not what the local-runtime docs recommend (they say to swallow `AbortError` and return cleanly).

- none blocking. finalizing against `abortSignal.aborted` at the single settle site is the right layer: adapter-set terminal statuses are still respected (the branch only runs while `running`), the in-loop check is kept for early stream exit, and the `detach()` path lands on the same cancelled semantics as the thrown-`AbortError` path.

### what we modified

- nothing; single-commit PR, no review iterations were needed.

### potential further fixes

- a microtask-sized race remains where an abort that lands after the final yield but before finalize marks a fully-produced answer cancelled; defensible semantics (completion was never committed) and not worth widening this PR.

PR #5111, opened 2026-07-22
fix(react-pi): isolate shared streams by client

Kinfe123:
## Summary

- scope shared React Pi SSE connections to a single `createPiHttpClient()` instance
- keep connection deduplication for subscribers using the same client
- add a regression test for separate browser auth identities

## Why

The default browser transport cached streams globally by URL and thread. With cookie authentication, a client created after logout/login could therefore reuse an SSE connection opened by the previous user because the existing request keeps the cookies it connected with.

In the reproduced case, user A opened a thread stream, the browser identity changed to user B, and a second client subscribed to the same URL and thread. Only one network connection opened, so user B received user A's existing stream.

Each client now owns its stream cache. Subscribers within that client still share a connection, while a newly created client establishes its own authenticated stream.

## Testing

- `pnpm --filter @assistant-ui/react-pi exec vitest run src/client/httpClient.test.ts`
- `pnpm --filter @assistant-ui/react-pi test`
- `pnpm --filter @assistant-ui/react-pi build`
- `pnpm lint`
- `pnpm exec changeset status --since=origin/main`

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5111?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/quiet-pi-streams.md
+++ b/.changeset/quiet-pi-streams.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react-pi": patch
+---
+
+fix: isolate shared Pi event streams between client instances
--- a/api-surface/assistant-ui__react-pi.ts
+++ b/api-surface/assistant-ui__react-pi.ts
@@ -1766,12 +1766,6 @@ type SendOptions = {
   steer?: boolean;
 };
 
-type SharedStream = {
-  listeners: Set<(event: PiClientEvent) => void>;
-  close: () => void;
-  closeTimer: ReturnType<typeof setTimeout> | undefined;
-};
-
 type SourceMessagePart = {
   readonly type: "source";
   readonly sourceType: "url";
@@ -2369,10 +2363,6 @@ declare const createSseDecoder: () => {
 };
 
 declare const getPiThreadSupervisor: (options?: PiNodeClientOptions) => PiThreadSupervisor;
-
-declare global {
-  var __assistantUiPiHttpStreams: Map<string, SharedStream> | undefined;
-}
 
 declare global {
   interface Window {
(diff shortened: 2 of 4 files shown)


okisdev (approved):
## review

LGTM as `patch`. verified in a clean worktree at `f014ecc`: the full `httpClient.test.ts` suite passes (26/26), and with the source file reverted to main the new regression test fails with exactly the leak this fixes (`expected fetch to be called 2 times, but got 1 times`, so the second client rode the first client's connection). the global cache dates back to the original #4368 package commit rather than any later fix, so removing it does not undo a known repair, and the canonical consumer pattern (`useMemo(() => createPiHttpClient(), [])` in with-pi) keeps full dedup within a client.

- the api-surface snapshots were not regenerated alongside the removal of the ambient `__assistantUiPiHttpStreams` declaration, so the surface check would have drifted without the bot commit. everything else (source, regression test, changeset) landed complete in the first commit.

### what we modified

- the autofix commit regenerates `api-surface/assistant-ui__react-pi.ts`, dropping the leaked `SharedStream` helper type and the global var declaration from the published type surface.

### potential further fixes

- an app that keeps reusing one client instance across logout/login still rides that client's own cached stream (plus the `streamCloseDelayMs` linger), since the cache key carries no identity. creating a new client per identity is the contract, and after this change it actually delivers a fresh authenticated stream, so this is documentation surface at most.
- cross-client dedup for the default transport is gone by design. it was never documented, and it was also wiring later clients to the first client's `onStreamError` and `reconnectDelay`, so losing it removes a second latent bug rather than a feature.

PR #5113, opened 2026-07-22
fix(core): apply pending mute after voice setup

Kinfe123:
## Summary

- apply a mute request made before asynchronous voice setup completes
- preserve a later unmute request as the latest user intent
- add focused regression coverage for both setup races

## Problem

While testing a delayed voice connection, calling `session.mute()` before `setup()` resolved produced `isMuted: true` but zero calls to the provider controls' `mute()` method. Once setup completed, the UI could therefore report a muted session while the underlying microphone remained unmuted.

The session now reconciles its current mute state immediately after controls become available. A mute followed by an unmute during setup remains unmuted.

## Verification

- reproduced before the fix: `{ uiIsMuted: true, underlyingMuteCalls: 0 }`
- verified after the fix: `{ uiIsMuted: true, underlyingMuteCalls: 1 }`
- `vitest run src/adapters/voice.test.ts` (3 tests)
- `aui-build` for `@assistant-ui/core`

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5113?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/quiet-voices-connect.md
+++ b/.changeset/quiet-voices-connect.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/core": patch
+---
+
+fix: apply pending mute state after voice session setup
--- a/packages/core/src/adapters/voice.test.ts
+++ b/packages/core/src/adapters/voice.test.ts
@@ -1,5 +1,9 @@
 import { afterEach, describe, expect, it, vi } from "vitest";
-import { createVoiceSession, type VoiceSessionHelpers } from "./voice";
+import {
+  createVoiceSession,
+  type VoiceSessionControls,
+  type VoiceSessionHelpers,
+} from "./voice";
 
 const createTestSession = () => {
   let helpers: VoiceSessionHelpers | undefined;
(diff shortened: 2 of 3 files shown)


okisdev (approved):
## review

lgtm as `patch`. verified both directions at runtime: with the fix `vitest run src/adapters/voice.test.ts` passes 3/3, and with `voice.ts` reverted to main the new pending-mute test fails with exactly the reported symptom (`controls.mute` called 0 times while `isMuted` reads true). the premise is real against a shipped adapter: the livekit example ends setup with `setMicrophoneEnabled(true)`, so a mute issued during connect left the UI muted with a live microphone.

- none material; the session already tracked `isMuted` correctly and only lacked the reconcile after `controls` resolved

### what we modified

### potential further fixes

- the reconcile is deliberately mute-only: an `isMuted === false` at setup completion does not call `controls.unmute()`, so a mute followed by unmute during setup relies on the provider finishing setup unmuted (livekit does). distinguishing "user asked to unmute" from "default false" would need a pending-intent flag; fine to leave until an adapter that finishes setup muted shows up

PR #5123, opened 2026-07-23
fix: mouse pointer on model selector thinking controls

Nick-Lucas:
Before:

https://github.com/user-attachments/assets/9fc289a2-e50a-4ccd-b7a6-444e8e69d518

After:

https://github.com/user-attachments/assets/6ee01e4a-3560-49ed-86d0-6b8e93fccfe9

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5123?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/ui/src/components/assistant-ui/model-selector.radix.tsx
+++ b/packages/ui/src/components/assistant-ui/model-selector.radix.tsx
@@ -583,7 +583,7 @@ function ModelSelectorEffort({
     <div
       data-slot="model-selector-effort"
       className={cn(
-        "flex items-center justify-between gap-3 border-t px-3 py-2",
+        "flex cursor-default items-center justify-between gap-3 border-t px-3 py-2",
         className,
       )}
       onKeyDown={(e) => {
--- a/packages/ui/src/components/assistant-ui/model-selector.tsx
+++ b/packages/ui/src/components/assistant-ui/model-selector.tsx
@@ -585,7 +585,7 @@ function ModelSelectorEffort({
     <div
       data-slot="model-selector-effort"
       className={cn(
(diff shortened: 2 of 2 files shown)


Kinfe123 (commented):
I think it is much better to make it cursor pointer instead of default.

Nick-Lucas:
Agree completely @Kinfe123, it's not very web-like, but it seems the AUI pattern is for all buttons to use `default` like a desktop app would? At least it seems to be very common including all other buttons in this component, so I wasn't planning to break convention with this fix

Kinfe123:
@okisdev what do you think about this ?

Yonom (approved):

Yonom:
Thanks for flagging out the cursor. I agree we should use the web platform default, not sure why it ended up this way. If/when we update, I prefer we do it as a separate PR to migrate everything over consistently. So this PR is good to go as-is

Kinfe123:
@Yonom yeah we need a follow up PR for consistency sake. make sense.

Kinfe123 (approved):

Kinfe123:
@Nick-Lucas  we might need to apply the same for `packages/ui/src/components/assistant-ui/model-selector.radix.tsx` too? The Radix variant has the same effort row and should keep parity.

Nick-Lucas:
Ah you beat me to it @Kinfe123 - thanks for adding that!

PR #5126, opened 2026-07-23
fix(core): refresh cloud history persistence on client change

Kinfe123:
## Summary

- allow `CloudMessagePersistence` to receive a current-client getter while preserving the existing constructor
- make the core Cloud history adapter use that getter so login or workspace client changes are reflected without a remount
- preserve local-to-remote message ID mappings across client changes
- add regression coverage for the core adapter and parent chaining across two Cloud clients

## Why

While testing a login or workspace client switch without a full app remount, the history adapter could keep the first `AssistantCloud` instance cached. Subsequent history loads or writes could then continue through the old authenticated client.

The persistence object must remain stable because it owns the local-to-remote message ID mapping used for parent chaining, updates, and deduplication. The fix therefore keeps one persistence per thread item and resolves the current Cloud client at the start of each operation. Existing callers that construct `CloudMessagePersistence` with a fixed client continue to work unchanged.

## Verification

- reproduced the old behavior: after swapping the ref, the third formatted history load still called client A
- verified the fix routes that load to client B
- verified a parent mapped through client A is used as the remote parent when its child is written through client B
- `assistant-cloud`: 5 test files passed, 42 tests passed, 2 skipped
- `@assistant-ui/core`: 65 test files passed, 588 tests passed
- both published packages build successfully
- changed files pass oxlint and oxfmt

### DIFF
--- a/.changeset/bright-clouds-refresh.md
+++ b/.changeset/bright-clouds-refresh.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/core": patch
+---
+
+fix: refresh Cloud history persistence when the Cloud client changes
--- a/.changeset/bright-clouds-refresh.md
+++ b/.changeset/bright-clouds-refresh.md
@@ -1,5 +1,6 @@
 ---
 "@assistant-ui/core": patch
+"assistant-cloud": patch
 ---
 
 fix: refresh Cloud history persistence when the Cloud client changes
--- a/api-surface/assistant-cloud.ts
+++ b/api-surface/assistant-cloud.ts
@@ -275,9 +275,10 @@ type CloudMessage = {
 };
 
 declare class CloudMessagePersistence {
-  private cloud;
   private idMapping;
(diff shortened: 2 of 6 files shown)


okisdev (approved):
verified end to end: the stale capture is real (the adapter's telemetry paths already deref cloudRef.current at use time, persistence construction was the one frozen read), and the stable-instance-plus-getter design is the right resolution of the earlier review round, preserving the id mapping the first revision would have dropped. the WeakMap keying by thread list item composes cleanly with it: a same-backend client refresh reuses the persistence through the live getter, while a workspace switch that rebuilds the list gets fresh entries anyway. constructor overload keeps the fixed-client form working for cloud-ai-sdk, api-surface is updated in-PR, and both suites pass locally (42 plus 2 skipped, 588) after rebuilding the cloud dist; the one failure I saw before rebuilding was my environment's stale dist, not this change.

PR #5133, opened 2026-07-24
fix(react-ink): tolerate transient part type mismatches

### LINKED ISSUE #5131: Part hook deprecation notes still recommend throwing selectors
## summary

the deprecation JSDoc on the five part hooks (useMessagePartText, useMessagePartReasoning, useMessagePartSource, useMessagePartFile, useMessagePartImage) still says "or throw inside the selector to preserve the old hook's strict behavior". after #5118 / #5119 / #5127 we know exactly what a throwing selector does during the thread-switch window: it fires inside useSyncExternalStore's getSnapshot and, in an app without a boundary above the part component, unmounts the React root. the library's own hooks moved off throwing for precisely this reason, so the migration guidance should not steer users into it.

## tasks

- [ ] reword the deprecation notes to recommend returning null for optional rendering and explicitly warn that throwing inside a useAuiState selector can unmount the app during thread switches (api-reference regen in the same PR, the notes are covered JSDoc)
- [ ] react-ink parity: packages/react-ink/src/primitives/messagePart/MessagePartText.tsx still throws on part-type mismatch; align it with the web sentinel behavior or record why ink deliberately diverges

## refs

#5118 (closed; both flavors fixed by #5119 and #5127), #5121, packages/react/src/tests/messagePartSwitchRace.test.tsx

Gujiassh:
## Summary

- update the deprecated part-hook guidance to recommend returning `null` and explain why throwing from a `useAuiState` selector can unmount the React root during a thread switch
- apply the same documentation correction to `useMessagePartData`, which carried the identical stale guidance while already returning `null`
- make all six React Ink message-part primitives return an empty-string sentinel for transient type mismatches, matching the web hook behavior
- regenerate the API reference and add a patch changeset for `@assistant-ui/react` and `@assistant-ui/react-ink`
- follow-up commit `e040af0` applies the same selector guard to Image, File, Source, Reasoning, and Data after review identified the sibling mismatch path

Fixes #5131

## Validation

- focused `MessagePartPrimitive` regression passes (13 tests), including red-before-green mismatch coverage for text, image, file, source, reasoning, and data
- complete `@assistant-ui/react-ink` suite passes (21 files, 237 tests)
- `@assistant-ui/react-ink` typecheck and production build pass
- `@assistant-ui/react` production build passes (928 entries)
- API reference generation is idempotent; all six part-hook entries contain the corrected warning
- root lint and format checks pass with existing warnings elsewhere in the workspace
- the full `@assistant-ui/react` suite still reports two existing `messagePartSwitchRace` failures in the unchanged runtime path; the React changes in this PR are documentation-only

### DIFF
--- a/.changeset/part-selector-sentinels.md
+++ b/.changeset/part-selector-sentinels.md
@@ -1,0 +1,6 @@
+---
+"@assistant-ui/react": patch
+"@assistant-ui/react-ink": patch
+---
+
+Warn against throwing from part selectors and keep React Ink text rendering stable during transient part-type mismatches.
--- a/.changeset/part-selector-sentinels.md
+++ b/.changeset/part-selector-sentinels.md
@@ -3,4 +3,4 @@
 "@assistant-ui/react-ink": patch
 ---
 
-Warn against throwing from part selectors and keep React Ink text rendering stable during transient part-type mismatches.
+Warn against throwing from part selectors and keep React Ink part rendering stable during transient part-type mismatches.
--- a/apps/docs/content/docs/(reference)/api-reference/hooks/primitives.mdx
+++ b/apps/docs/content/docs/(reference)/api-reference/hooks/primitives.mdx
@@ -469,8 +469,9 @@ const useMessagePart: { (): MessagePartState; <TSelected>(selector: (state: Mess
 
 <Callout type="warn">
 <strong>Deprecated.</strong> Use [useAuiState](/docs/api-reference/hooks/state#useauistate) to select and narrow `s.part`.
-Return `null` for optional rendering, or throw inside the selector to
-preserve the old hook's strict behavior.
+Return `null` for optional rendering. Do not throw inside the selector:
+selectors run inside `useSyncExternalStore`'s `getSnapshot`, so a transient
+part mismatch during thread switches can unmount the React root.
 </Callout>
 
 ```tsx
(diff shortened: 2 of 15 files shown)


Kinfe123 (approved):

okisdev (approved):
## review

LGTM as `patch`. verified in a clean worktree at `e040af0`: the ink suite passes 13/13, and with the six primitives reverted to main exactly the six new mismatch cases fail on the throwing selectors, so the guard is load-bearing per type. both tasks of #5131 are delivered, and the JSDoc rewording plus same-PR api-reference regen is the correct flow for covered exports.

one correction to the PR body: the "two existing `messagePartSwitchRace` failures" are not real. `packages/react` vitest executes `@assistant-ui/core` and siblings from built dist, so after checkout the react graph must be built first (`pnpm --filter "@assistant-ui/react..." build`); on a fresh build both race suites pass 10/10 on this branch and on main. worth knowing for future PRs, it has bitten this repo before. the `claude-fork-review` failure is the publish step lacking permissions on fork PRs, advisory and unrelated to your changes.

- the first commit guarded only `MessagePartText`; the five sibling primitives (Image, File, Source, Reasoning, Data) kept the identical throwing selector, so the same thread-switch window still crashed an Ink root through any non-text part.

### what we modified

- `e040af0` extends the guard to all six primitives, returning an empty string from the selector on a transient mismatch and moving label assembly into the selector where needed (Data, Image) so matching parts render byte-identical output.
- adds per-type regression coverage that pins the selector contract for all six primitives.

### potential further fixes

- none blocking. the ink render-prop paths and in-progress gating were untouched and the existing seven tests cover them.

PR #5134, opened 2026-07-24
fix(cloud-ai-sdk): reset chat state on cloud change

Kinfe123:
## Summary

- clear the selected thread when the `AssistantCloud` identity changes
- recreate the chat registry and fresh-session state for the new Cloud client
- ignore late thread actions from the previous Cloud client so they cannot overwrite the active client's state
- keep each registry's chat factory isolated so late work from the previous client cannot bind to the new registry

## Why

When an app switches from Cloud A to Cloud B without remounting, `useThreads()` refreshes B's thread list but previously retained A's selected `threadId`. `useChatRegistry()` also retained cached `Chat` instances, whose transport and persistence callbacks were created with Cloud A. A late operation from A could additionally overwrite B's thread state. That could load or persist through the previous account or workspace client after the UI had switched to B.

This is separate from, and complementary to, #5126, which updates the core Cloud history adapter rather than the `cloud-ai-sdk` hooks.

## Testing

- regression tests for selected thread and cached chat reset across Cloud A → B
- regression coverage for late A callbacks and A → B → A switching
- `pnpm turbo test --filter="...[origin/main]"`
- `pnpm --filter @assistant-ui/cloud-ai-sdk build`
- `pnpm lint`

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5134?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/clean-cloud-switches.md
+++ b/.changeset/clean-cloud-switches.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/cloud-ai-sdk": patch
+---
+
+fix: reset selected threads and cached chats when the Cloud client changes
--- a/packages/cloud-ai-sdk/package.json
+++ b/packages/cloud-ai-sdk/package.json
@@ -52,6 +52,7 @@
   "devDependencies": {
     "@ai-sdk/react": "^4.0.30",
     "@ai-sdk/react-v3": "npm:@ai-sdk/react@^3.0.211",
+    "@assistant-ui/react-ai-sdk": "workspace:*",
     "@assistant-ui/x-buildutils": "workspace:*",
     "@testing-library/dom": "^10.4.1",
     "@testing-library/react": "^16.3.2",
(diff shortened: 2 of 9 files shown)


okisdev (approved):
verified as the sdk-hooks half of the cloud identity-switch family, complementary to #5126 on the core adapter: the thread selection is scoped to the cloud identity object with the derived-null-plus-effect-commit pattern, the registry recreation deliberately inverts the old delegate-to-latest factory so late work binds to the registry it was created for, and the regression suite covers A to B, late A callbacks, and A to B to A. the react-ai-sdk workspace devDependency makes an existing relative-path source import from the contract test explicit in the dependency graph. all six review threads were absorbed in earlier rounds, the suite is green locally at 64, and the claude-review row is a 25-second workflow infrastructure failure on an advisory check, not a finding.

PR #5143, opened 2026-07-24
fix(tap): allow Node processes to exit

Kinfe123:
## Summary

- unref Node.js `MessageChannel` ports while the tap scheduler is idle
- temporarily ref the ports while a flush is pending, then unref them after settlement
- retain the existing `MessageChannel` scheduling behavior in browsers and Node.js
- cover both Node-style and browser-style ports and publish the fix as a patch

## Why

`@assistant-ui/tap` creates a module-level `MessageChannel` during import. In Node.js, its active `MessagePort` keeps short-lived scripts, CLIs, SSR workers, and test processes alive after their work is complete.

The ports now keep Node alive only while scheduler work is actually pending. This fixes process shutdown without changing the scheduler primitive or its macrotask ordering.

## Verification

- `@assistant-ui/tap` test suite: 449 tests passed
- downstream `@assistant-ui/react-mcp` scheduler-sensitive tests: 9 passed
- package TypeScript check
- package build
- lint and formatting checks
- built-package subprocess import exits normally
- scheduled subprocess flush runs before exiting

### DIFF
--- a/.changeset/calm-tap-schedulers.md
+++ b/.changeset/calm-tap-schedulers.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/tap": patch
+---
+
+fix: allow Node.js processes to exit after importing the scheduler
--- a/packages/tap/src/__tests__/basic/scheduler.test.ts
+++ b/packages/tap/src/__tests__/basic/scheduler.test.ts
@@ -1,0 +1,58 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+describe("UpdateScheduler", () => {
+  beforeEach(() => {
+    vi.resetModules();
+  });
+
+  afterEach(() => {
+    vi.unstubAllGlobals();
+  });
+
+  it("prefers setImmediate when available", async () => {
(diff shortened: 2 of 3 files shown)


okisdev:
closing as superseded by #5181, verified empirically rather than by reading alone. on current main's built output, both subprocess probes exit cleanly: scheduling one flush prints and exits 0 (the post-use half this PR targeted), and a cascade probe where the task calls markDirty again runs both flushes and still exits 0. the shipped shape covers all three axes — import-only exit via lazy creation, post-use exit via the ref-on-schedule/unref-at-delivery lifecycle, and cascade-then-exit because the unref runs before flushScheduled, so any re-schedule re-refs after it. two notes for the record: the dual-port unref here is harmless but unnecessary (an unstarted sender port does not hold the node loop, per the probe), and the `try { flushScheduled() } finally { unref() }` shape in this diff carries a cascade race the shipped ordering avoids — a task that marks dirty during the flush refs the port, then the unconditional finally unrefs it, so a short-lived script can exit with the cascaded flush pending. worth avoiding that shape if this pattern comes up elsewhere.

PR #5144, opened 2026-07-24
fix(core): isolate thread runtime event listener errors

Kinfe123:
## Summary

- isolate exceptions thrown or asynchronously rejected by individual thread runtime event listeners
- continue notifying later listeners without changing a successful runtime operation into a rejection
- add regression coverage and a patch changeset for `@assistant-ui/core`

## Why

A `runEnd` listener is observer code, commonly used for telemetry or cleanup. If one listener throws today, a successfully completed `append()` rejects and every listener registered after it is skipped, even though the assistant message is already complete. An async listener can similarly produce an unhandled rejection.

The dispatcher now catches each listener independently and reports synchronous throws and rejected Promise-like returns with the event name, matching the isolation behavior used by voice session listeners.

## Reproduction

Before this change, a completed run with a first `runEnd` listener that throws `telemetry failed` produced:

```json
{"rejected":"telemetry failed","laterListenerCalls":0,"finalStatus":{"type":"complete","reason":"unknown"}}
```

After this change, `append()` resolves, the later listener runs once, rejected async listeners are handled, and the completed message remains intact.

## Verification

- `pnpm exec vitest run src/runtimes/local/local-thread-runtime-core.test.ts` (27 tests)
- `pnpm --filter @assistant-ui/core test` (591 tests)
- targeted `oxlint` and `oxfmt --check`

### DIFF
--- a/.changeset/tidy-thread-events-listen.md
+++ b/.changeset/tidy-thread-events-listen.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/core": patch
+---
+
+fix: isolate thread runtime event listener errors
--- a/packages/core/src/runtime/base/base-thread-runtime-core.ts
+++ b/packages/core/src/runtime/base/base-thread-runtime-core.ts
@@ -182,7 +182,16 @@ export abstract class BaseThreadRuntimeCore implements ThreadRuntimeCore {
     const subscribers = this._eventSubscribers.get(event);
     if (!subscribers) return;
 
-    for (const callback of subscribers) callback(payload);
+    for (const callback of subscribers) {
+      try {
+        callback(payload);
+      } catch (error) {
+        console.error(
+          `[assistant-ui] Thread runtime "${event}" listener threw an error`,
+          error,
+        );
+      }
+    }
   }
 
   public subscribe(callback: () => void): Unsubscribe {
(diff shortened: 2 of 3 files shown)


okisdev (approved):
## review

LGTM as `patch`. verified in a clean worktree at `39e03b4`: the suite passes 27/27, and with the dispatcher reverted to main exactly the two new regressions fail, the sync case because a successful `append()` rejects, the async case with vitest surfacing the leaked unhandled rejection. the in-package precedent claim checks out: `adapters/voice.ts` carries the same isolation with the same `[assistant-ui]` prefix and message shape, and this mirrors it faithfully. on the shared-helper question, the five copies of this idiom now in the tree (voice, store's NotificationManager, react-pi, react-opencode, this one) all differ in shape (async handling, message wording, prefix), so per-site copies remain the right call.

- the first commit only caught synchronous throws; an `async` listener's rejection still escaped as an unhandled rejection, which review flagged.

### what we modified

- the follow-up commit detects thenable returns and routes their rejections through the same reporter, with a dedicated async regression pinned by `vi.waitFor`.

### potential further fixes

- the composer dispatcher (`base-composer-runtime-core.ts`) is now the last bare fan-out loop in the tree: a throwing `send` listener can still break `composer.send()` the same way. same fix shape applies; tracked as a follow-up issue.

PR #5147, opened 2026-07-25
feat(ui): scroll follow-up suggestion pills with edge fades

AVGVSTVS96:
## Summary

`ThreadFollowupSuggestions` previously wrapped pill text onto multiple lines when suggestions overflowed. This matches it to the welcome suggestions pill treatment:

- Pills stay on a single line (`whitespace-nowrap`), centered while they fit.
- On overflow the row scrolls horizontally with no scrollbar.
- A mask fade appears only on an edge with clipped content: at the start position there is no fade on the leading edge; mid-scroll both edges fade; at the end only the trailing edge's fade disappears.
- Fade edges are computed from scroll position (RTL-aware) and kept in sync via a `ResizeObserver`, so container resizes and suggestion changes update the mask.

## Docs

- Expanded the follow-up suggestions sample: composer added below the pills, user bubble matched to the canonical `bg-muted` thread style, and enough suggestions to overflow so the scroll + fade behavior is visible on the page.
- Behavior section documents the single-line scroll and fade rules.
- Callout and Related Components now cross-link the Welcome Suggestions component page.

## Screenshots

| Scroll start (no leading fade) | Mid-scroll (both edges fade) | End (no trailing fade) |
| --- | --- | --- |
| scroll start | mid scroll | scroll end |

## Test plan

- `tsc --noEmit` clean in `packages/ui` and `apps/docs`; oxlint/oxfmt clean.
- Verified in the docs dev server: mask is trailing-edge-only at scroll start, both edges mid-scroll, leading-edge-only at the end.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5147?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.3zlkaBCjP-E1Ihs1WNTw6FbRSZAxYdK1v5Ot_E9qxdLpy3BpnJEz2-Pw6r4?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/apps/docs/components/docs/samples/follow-up-suggestions.tsx
+++ b/apps/docs/components/docs/samples/follow-up-suggestions.tsx
@@ -3,9 +3,11 @@
 import type { FC, PropsWithChildren } from "react";
 import {
   AssistantRuntimeProvider,
+  ComposerPrimitive,
   useExternalStoreRuntime,
   type ThreadMessageLike,
 } from "@assistant-ui/react";
+import { ArrowUpIcon } from "lucide-react";
 import { ThreadFollowupSuggestions } from "@/components/assistant-ui/follow-up-suggestions";
 import { SampleFrame } from "@/components/docs/samples/sample-frame";
 
@@ -24,7 +26,7 @@ const userMessage: DemoMessage = {
 const assistantMessage: DemoMessage = {
   id: "assistant-1",
   role: "assistant",
-  text: "Start by mapping the first-run path, then add a few suggested prompts that help users discover the highest-value workflows.",
+  text: "Start by mapping the first-run path, then add a few suggested prompts that help users discover the highest-value workflows. Keep the empty state focused on one clear action.",
 };
 
 const messages = [
@@ -33,15 +35,12 @@ const messages = [
 ] satisfies readonly DemoMessage[];
 
 const suggestions = [
-  {
-    prompt: "Draft three onboarding prompts",
-  },
-  {
-    prompt: "Turn this into a checklist",
-  },
-  {
-    prompt: "Show me a better empty state",
-  },
+  { prompt: "Draft three onboarding prompts" },
+  { prompt: "Turn this into a checklist" },
+  { prompt: "Write copy for the welcome screen" },
+  { prompt: "Show me a better empty state" },
+  { prompt: "Which metrics show onboarding is working?" },
+  { prompt: "Compare a guided tour vs. suggested prompts" },
 ];
 
 const convertMessage = (message: DemoMessage): ThreadMessageLike => ({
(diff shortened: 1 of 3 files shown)


AVGVSTVS96:
Pushed fixes for the verified findings (e35affee2f):

**Fixed**

- **`scrollbar-none` was docs-only** (claude, Copilot — confirmed): the class comes from the `tailwind-scrollbar` plugin loaded only in `apps/docs`, so it was dead in templates and `shadcn add` projects. Replaced with the same self-contained inline utilities `model-selector.tsx` uses: `[-ms-overflow-style:none] [scrollbar-width:none] [&::-webkit-scrollbar]:hidden`.
- **Broken `/docs/ui/welcome-suggestions` links** (claude, rupic, cubic — confirmed for this PR standalone): that page ships with the in-flight welcome-suggestions branch, not main. Reverted the callout to `ThreadPrimitive.Suggestions` and dropped the Related Components entry; the welcome-suggestions PR can add the cross-links once its page exists.
- **Focus-ring clipping from `overflow-x-auto`** (cubic, claude — confirmed; `overflow-y` computes to `auto`): added `py-1 -my-1` on the scroll container so rings render inside the clip box without changing the row's outer height.
- **`getComputedStyle` per scroll event** (claude — valid): direction is now read once and cached in a ref; the scroll handler is down to the two geometry reads.

**Not changed, with reasons**

- **Greptile P1 "Safari 16.4 ignores unprefixed `mask-image`"** — incorrect: Safari has supported unprefixed `mask-image` since 15.4. The actual prefix laggard was Chrome/Edge <120, so I added `WebkitMaskImage` alongside `maskImage` as cheap hardening, but the Safari claim as stated doesn't hold.
- **"This diverges welcome (wrap) vs follow-ups (scroll)"** — the in-flight welcome-suggestions branch moves welcome pills to the same single-line scroll treatment (`fix(ui): keep welcome suggestion pills on one line`). The two components converge once both land; this PR is simply the follow-up half.
- **Docs sample scope** — the composer and bubble-style updates were an explicit request to make the sample present better on the page, not incidental drift.
- **Changeset** — not needed; `@assistant-ui/ui` and the docs app are both private, as claude and rupic noted.

AVGVSTVS96:
Correction to my previous comment: the claim that "Tailwind v4 core ships no scrollbar utilities" (from the reviews, which I repeated) is outdated. **Tailwind v4.3.0 added `scrollbar-{auto,thin,none}` to core** (tailwindlabs/tailwindcss#19981), and all templates plus the docs app pin `^4.3.2` — so `scrollbar-none` was not actually dead there; it compiles to `scrollbar-width: none` with no plugin.

The inline-utility replacement stays, for the narrower reasons that actually hold:

- `shadcn add` copies this file into arbitrary projects, including ones on Tailwind < 4.3 where the class is a silent no-op.
- Core's version emits only `scrollbar-width: none`, which Safari < 18.2 and Chrome ≤ 120 ignore; the `[&::-webkit-scrollbar]:hidden` fallback covers them.
- It matches the existing `model-selector.tsx` pattern in `packages/ui`.

Kinfe123 (approved):

PR #5183, opened 2026-07-26
refactor(eve): migrate convertEveMessage onto fromThreadMessageLike

### LINKED ISSUE #5135: refactor(eve): migrate convertEveMessages onto fromThreadMessageLike
## Context

While working on the eve adapter I noticed it builds `ThreadMessage` through two divergent paths: staged user messages go through core's `fromThreadMessageLike` (`useEveAgentRuntime.ts:136`), while `convertEveMessage` hand-assembles both roles (`convertEveMessages.ts:216-266`), including its own copy of the assistant metadata defaults and two unchecked `as readonly ...[]` casts.

To be clear on motivation:  no live bug. This is consolidation. Hand-built assembly silently drifts as core hardens the shared path: the helper has since been taught to tolerate parts that omit their declared fields and to thread providerMetadata through, and the hand-built path got neither. Concretely, an Eve text part arriving without `text` is dropped by the helper but passes through eve's converter as an invalid part today. Delegating keeps eve on the shared primitive per AGENTS.md.

The timing also looks right: the converter has not needed a single source change while the upstream SDK moved 0.12 -> 0.22 -> 0.24, so the message surface it maps is demonstrably stable. This is not a new abstraction; it is one adapter migrating onto the existing canonical primitive.

## Proposed change

- Converter internals only. `convertEveMessage` keeps its Eve-specific pre-processing (per-part switch, dynamic-tool state machine, approval mapping, status computation, user fallback text part) and delegates final `ThreadMessage` assembly to `fromThreadMessageLike`.
- No public surface change: exports, signatures, and options of `convertEveMessages` / `convertEveMessage` are unchanged.
- Colocated both-way converter tests extended to pin the migration, including the two known behavior deltas below.
- The defensive posture is preserved: eve's part switch filters unknown part types to null before the like is built, so the helper's throw paths stay unreachable from provider payloads.

## Behavior deltas (named, not hidden)

1. The helper drops empty and whitespace-only assistant text/reasoning parts; the hand-built path keeps them. Accepting the helper semantics aligns the converter with the staged path (which already trims). Practical impact is a transiently absent empty part during streaming; a test pins the new behavior.
2. The helper's user branch dropped `metadata.isOptimistic`, which eve sets from Eve's optimistic flag and which `ThreadMessageLike.metadata` accepts role-agnostically. This was a helper gap, not an eve problem: any external-store consumer passing an optimistic user message lost the eviction semantics. Fix landed as a small core patch first (#5122); the eve migration lands on top once it merges. Two sequenced PRs.

## Out of scope

- Any change to eve's status computation, approval mapping, or send-direction converters (`getEveMessageContent`, `toEveInputResponse`).
- Any change to the helper's part vocabulary or its throw-on-unknown behavior.
- Runtime orchestration in `useEveAgentRuntime`.

ShobhitPatra:
Closes #5135.

## What

`convertEveMessage` previously hand-assembled `ThreadMessage` for both roles, carrying its own copy of the assistant metadata defaults and two unchecked `as readonly ...[]` casts. It now keeps all eve-specific pre-processing (part switches, dynamic-tool state machine, approval mapping, status computation, user fallback text part) and delegates final assembly to core's `fromThreadMessageLike`, matching how eve's own staged path already builds messages. No public surface change.

### Behavior deltas (named in #5135, pinned by tests)

- Empty and whitespace-only assistant text/reasoning parts are now dropped, matching the shared helper's assistant semantics. Practical impact is a transiently absent empty part during streaming.
- `metadata.isOptimistic` on user messages survives conversion. Note for reviewers: this test also passes against the old hand-built converter; what it pins is the dependency on #5122, without which the helper's user branch dropped the flag.

## Mechanics

- Status flows through the helper's `fallbackStatus` parameter, same as the staged path in `useEveAgentRuntime`; user messages never carry status, so the helper's throw path stays unreachable.
- User fallback empty text part is preserved as pre-processing, so eve user messages are never content-empty (unchanged behavior).
- Both content casts deleted; the converter now typechecks against `ThreadMessageLike` with no casts.
- Tests: 6 new pins (delta 1, delta 2 both directions, user fallback, non-convertible parts dropped on both roles); existing 6 tests pass unmodified.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5183?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/eve-from-thread-message-like.md
+++ b/.changeset/eve-from-thread-message-like.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/eve": patch
+---
+
+refactor: migrate convertEveMessage onto fromThreadMessageLike
--- a/packages/eve/src/convertEveMessages.test.ts
+++ b/packages/eve/src/convertEveMessages.test.ts
@@ -141,6 +141,124 @@ describe("convertEveMessages", () => {
     });
   });
 
+  it("drops empty and whitespace-only assistant text and reasoning parts", () => {
+    const data = {
+      messages: [
+        {
+          id: "a1",
+          role: "assistant",
+          parts: [
+            { type: "text", text: "" },
+            { type: "reasoning", text: "   " },
+            { type: "text", text: "Hi" },
+          ],
+        },
+      ],
+    } satisfies EveMessageData;
+
+    const [message] = convertEveMessages(data);
+
+    expect(message?.content).toEqual([{ type: "text", text: "Hi" }]);
+  });
+
+  it("preserves isOptimistic on optimistic user messages", () => {
+    const data = {
+      messages: [
+        {
+          id: "u1",
+          role: "user",
+          metadata: { optimistic: true },
+          parts: [{ type: "text", text: "Hello" }],
+        },
+      ],
+    } satisfies EveMessageData;
+
+    const [message] = convertEveMessages(data);
+
+    expect(message?.metadata.isOptimistic).toBe(true);
+  });
+
+  it("omits isOptimistic on confirmed user messages", () => {
+    const data = {
+      messages: [
+        {
+          id: "u1",
+          role: "user",
+          metadata: { status: "submitted" },
+          parts: [{ type: "text", text: "Hello" }],
+        },
+      ],
+    } satisfies EveMessageData;
+
+    const [message] = convertEveMessages(data);
+
+    expect(message?.metadata).not.toHaveProperty("isOptimistic");
+  });
+
+  it("falls back to an empty text part for user messages without convertible parts", () => {
+    const data = {
+      messages: [
+        {
+          id: "u1",
+          role: "user",
+          parts: [{ type: "file", mediaType: "image/png" }],
+        },
+      ],
+    } satisfies EveMessageData;
+
+    const [message] = convertEveMessages(data);
+
+    expect(message?.content).toEqual([{ type: "text", text: "" }]);
+  });
+
+  it("drops non-convertible user parts without triggering the fallback", () => {
+    const data = {
+      messages: [
+        {
+          id: "u1",
+          role: "user",
+          parts: [
+            { type: "file", mediaType: "image/png" },
+            { type: "text", text: "Hello" },
+          ],
+        },
+      ],
+    } satisfies EveMessageData;
+
+    const [message] = convertEveMessages(data);
+
+    expect(message?.content).toEqual([{ type: "text", text: "Hello" }]);
+  });
+
+  it("drops non-convertible part types instead of throwing", () => {
+    const data = {
+      messages: [
+        {
+          id: "a1",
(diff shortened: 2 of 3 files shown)


okisdev (approved):
## review

LGTM as `patch`. verified in a clean worktree at `d2c2881`: the eve suite passes 12/12, and with the converter reverted to main exactly one test fails, the empty and whitespace-only assistant text drop, which is the single true behavior delta and inherent to the shared helper's semantics (`thread-message-like.ts` returns null for untrimmed-empty text and reasoning). the other eleven pins pass on both sides, including the `isOptimistic` test the description explicitly declares as a dependency pin on #5122 rather than a delta, and the preserved-behavior pins for the user fallback part and non-convertible drops. the claim ledger in the description matches the empirical split exactly. the migration itself is the shared-primitive discipline done right: eve pre-processing stays local, both unchecked content casts are gone, status rides the `fallbackStatus` parameter the staged path already uses, and the converter is net smaller.

- none; single commit with the migration, six pins, and the changeset, opened issue-first with the motivation argued against the SDK's 0.12 to 0.24 stability record.

### what we modified

- nothing; single commit.

### potential further fixes

- none. with this landed, eve builds every message through the shared helper on both paths, and future helper hardening (like the providerMetadata threading and field-omission tolerance that motivated #5135) applies to eve for free.

okisdev (approved):
## review

LGTM as `patch`. verified in a clean worktree at `d2c2881`: the eve suite passes 12/12, and with the converter reverted to main exactly one test fails, the empty and whitespace-only assistant text drop, which is the single true behavior delta and inherent to the shared helper's semantics (`thread-message-like.ts` returns null for untrimmed-empty text and reasoning). the other eleven pins pass on both sides, including the `isOptimistic` test the description explicitly declares as a dependency pin on #5122 rather than a delta, and the preserved-behavior pins for the user fallback part and non-convertible drops. the claim ledger in the description matches the empirical split exactly. the migration itself is the shared-primitive discipline done right: eve pre-processing stays local, both unchecked content casts are gone, status rides the `fallbackStatus` parameter the staged path already uses, and the converter is net smaller.

- none; single commit with the migration, six pins, and the changeset, opened issue-first with the motivation argued against the SDK's 0.12 to 0.24 stability record.

### what we modified

- nothing; single commit.

### potential further fixes

- none. with this landed, eve builds every message through the shared helper on both paths, and future helper hardening (like the providerMetadata threading and field-omission tolerance that motivated #5135) applies to eve for free.

PR #5198, opened 2026-07-26
chore: update dependencies

Yonom:
## Summary

Scheduled run of the `update-deps` skill.

- **JS/TS**: `npx taze major -f -w -r` across all 93 workspace projects, `expo install --fix` in `examples/with-expo` to re-pin React Native/Expo SDK-compatible versions, a fully regenerated `pnpm-lock.yaml`, and a generated changeset (`.changeset/tall-rods-pay.md`) covering the 3 published packages whose `package.json` changed: `@assistant-ui/cloud-ai-sdk`, `@assistant-ui/react-lexical`, `@assistant-ui/react-pi`.
- **Fixed a taze regression in `@assistant-ui/cloud-ai-sdk`**: the package's `devDependencies` use `ai-v6: npm:ai@^6.0.209` and `@ai-sdk/react-v3: npm:@ai-sdk/react@^3.0.211` as intentional old-major pins, exercised by the `test:peer-v6` / `test:types:peer-v6` scripts against its `"ai": "^6.0.0 || ^7.0.0"` / `"@ai-sdk/react": "^3.0.0 || ^4.0.0"` peerDependencies union (per `AGENTS.md`'s union-satellite policy: every major in the union must stay exercised in CI). Taze doesn't understand these are fixed-major pins and bumped both aliases to track the current major (`^7.0.37` / `^4.0.40`), which would have silently turned the peer-v6 test legs into duplicates of the v7 legs. Re-pinned both to the latest patch within their intended major instead (`ai@^6.0.235`, `@ai-sdk/react@^3.0.237`) and verified `pnpm test` in that package still exercises both peer legs (72 tests × 2 legs, plus the `tsc --noEmit` type-check leg).
- **Reverted a taze major bump of `@babel/core` to `^8.0.1`** in `@assistant-ui/x-buildutils` and `examples/with-expo` back to `^7.29.7` (flagged independently by Greptile, Copilot, and cubic-dev-ai). Babel 8 stable is excluded by several transitive `babel-plugin-polyfill-*` packages' peer range (`^7.4.0 || ^8.0.0-0 <8.0.0`, a prerelease-only upper bound), and it raises the Node engine floor to `^22.18.0 || >=24.11.0` versus this repo's `>=24`. Since `@assistant-ui/x-buildutils`'s `package.json` ended up byte-identical to the base branch after the revert, its now-empty changeset entry was removed too. A follow-up review (rupic-app) caught that an incremental `pnpm install` after the manifest edits left 178 stale `@babel/core@8.0.1` references embedded in unrelated composite peer-dependency keys (`next`, `react-native`, `@react-native/metro-config`, etc.) in `pnpm-lock.yaml`; a full `node_modules` + lockfile wipe-and-reinstall (as this repo's dependency-update workflow prescribes) cleared all of them — every `@babel/core` resolution in the lockfile is now `7.29.7`.
- **Python**: `uv lock --upgrade` + `uv sync` for the three packages with lockfiles (`assistant-stream`, `assistant-transport-backend`, `assistant-transport-backend-langgraph`); `python/state-test`'s lockfile had nothing to bump (pre-existing wheel-packaging config issue in that package is unrelated and left untouched).
- **GitHub Actions**: skipped. `ratchet`/`pinact` aren't preinstalled; `go install`'ing `ratchet` worked, but resolving other repos' release SHAs (`actions/checkout`, `actions/cache`, etc.) requires GitHub API access outside this session's repo scope (`assistant-ui/assistant-ui` only), so it can't run here.

## Validation

- `pnpm build` (`turbo build --force`) — 84/84 tasks successful after the final lockfile regen.
- `pnpm lint` — clean (only pre-existing `react-hooks` warnings, no errors, formatting OK).
- `pnpm check:resource-memo` — passes (required whenever `@babel/core` changes).
- `pnpm test` (turbo, all packages) — 76/76 tasks successful, including both `@assistant-ui/cloud-ai-sdk` peer legs.
- `uv run pytest` for the two Python packages with tests — `assistant-stream`: 119 passed / 7 skipped (Redis-dependent); `assistant-transport-backend(-langgraph)`: no tests defined.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Sj23bABGG7kx1zrBTVdefY)_

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5198?utm_source=github" rel="nofollow noreferrer noopener" target="_blank">``&lt;img alt="Review in cubic" src="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"&gt;``</a>


Yonom:
Thanks @greptile-apps, @Copilot, and @cubic-dev-ai — all flagged the same real issue, so responding once here for all four threads.

You're right: `taze major` bumped `@babel/core` to `^8.0.1` in `packages/x-buildutils/package.json` and `examples/with-expo/package.json`. Verified two independent problems with that:

1. **Peer range excludes stable Babel 8.** The transitive `babel-plugin-polyfill-corejs2`/`corejs3`/`regenerator` packages declare `peerDependencies: "@babel/core": "^7.4.0 || ^8.0.0-0 <8.0.0"` — that upper bound (`<8.0.0`) only matches Babel 8 *prereleases*, not the stable `8.0.1` we'd pinned to.
2. **Engine floor mismatch.** `@babel/core@8` requires Node `^22.18.0 || >=24.11.0`, but this repo's root `engines.node` is `>=24`, so Node 24.0–24.10 would be a valid install per our manifest but incompatible with Babel 8.

Fixed in 87b8a3e: reverted `@babel/core` to `^7.29.7` in both manifests (nothing else in either package needed Babel 8 — `babel-plugin-react-compiler` stayed at `^1.0.0` throughout). Re-ran `pnpm install`, `pnpm check:resource-memo` (passes — required by AGENTS.md for any `@babel/core` bump), a full `turbo build` (84/84 tasks), `pnpm lint`, and the full `turbo test` suite (76/76 tasks, all green) to confirm no regressions from the revert.

---

Yonom:
Re: the `typescript` field in `examples/with-expo/package.json` — checked this one before deciding whether to revert it.

It's not a stray edit: `expo install --fix` (the mandated step for this repo's dependency-update workflow, run after `taze`) rewrote it deliberately, because Expo SDK 57 pins `typescript@~6.0.3` for compatibility. The alias package used everywhere else in the repo, `@typescript/typescript6`, only has releases up to `6.0.2` on npm — it hasn't published a `6.0.3`. So the alias literally cannot satisfy what Expo requires here; `expo install --fix` correctly fell back to the real `typescript` package, which does publish a stable `6.0.3` (verified: `npm view typescript@^6.0.3 version` resolves cleanly, and this repo already has precedent for bypassing the alias elsewhere — `examples/with-browser-extension`, `examples/with-react-ink`, `examples/with-tanstack`, and the root `package.json` all pin plain `typescript` directly rather than through the alias).

So while it does diverge from the alias convention used by most examples, the divergence is forced by Expo's compatibility matrix, not incidental — and the full `pnpm install` + `turbo build` (84/84) + `turbo test` (76/76) run already validated it resolves and works. Leaving it as-is.

---

Yonom:
Thanks @rupic-app — both findings were valid, fixed in a4b11f5.

**Incomplete Babel revert.** You were right: `87b8a3e` fixed the direct `@babel/core` pins in `x-buildutils` and `with-expo`, but `pnpm-lock.yaml` still had 178 stale references to `@babel/core@8.0.1` embedded in composite peer keys (`next@16.2.11(@babel/core@8.0.1)...`, `@react-native/metro-config@0.86.0(@babel/core@8.0.1)...`). That happened because I only ran an incremental `pnpm install` after editing the two `package.json` files, instead of the full wipe-and-reinstall this repo's own dependency-update workflow calls for (`rm -rf **/node_modules pnpm-lock.yaml && pnpm install`). I confirmed every actual `@babel/core` peerDependencies range in the lockfile is `^7.x`-only — nothing legitimately needs 8.x — so this was purely an orphaned resolution artifact from the incremental install, not a real requirement anywhere in the graph. Did the full wipe+reinstall: `grep -c "@babel/core@8.0.1" pnpm-lock.yaml` now returns **0**, and `7.29.7` is the only version resolved anywhere.

**Empty changeset entry.** Also correct — removed the `"@assistant-ui/x-buildutils": patch` line from `.changeset/tall-rods-pay.md`. That package's `package.json` is now byte-identical to the base branch (`git diff cdee1e2 HEAD -- packages/x-buildutils/package.json` returns empty), so the entry would have published a no-op release.

Re-validated everything after the clean regen: `pnpm check:resource-memo` passes, full `turbo build --force` is 84/84, `pnpm lint` shows only pre-existing warnings, and `turbo test` is 76/76.

---

Yonom:
Good catch — the PR description was written before the `x-buildutils` changeset entry got removed and never updated to match. Fixed the description to say 3 packages (`@assistant-ui/cloud-ai-sdk`, `@assistant-ui/react-lexical`, `@assistant-ui/react-pi`) and to document the Babel revert / lockfile regen from the follow-up commits, since `@assistant-ui/x-buildutils` isn't being released in this PR.

---

PR #5201, opened 2026-07-26
fix(react-native): wrap ThreadListItems rows in ThreadListItemByIndexProvider

ShobhitPatra:
## What
`ThreadListPrimitive.Items` in react-native now wraps each row in `ThreadListItemByIndexProvider`, so thread list item primitives (Trigger, Archive, Delete, Unarchive) work inside `renderItem` instead of crashing on a missing scope. Also removes the manual provider wrap the with-expo example carried as a workaround.

## Why
Same crash class fixed in ink with #4885: the primitive calls `renderItem` bare, so any threadListItem-scoped child throws. The with-expo drawer was already hand-wrapping rows in the provider to survive, which is exactly the workaround this fix removes.

## Impact
- Thread list item primitives become usable inside `Items` rows, matching core and ink behavior.
- Existing apps that hand-wrapped rows keep working: the inner provider just shadows the new outer one with an identical scope.
- No public surface change, props and types untouched.

## Scope & caveats
- Sibling of merged #4885 (ink); FlatList key mechanics differ from ink's `.map()` + `key`, covered by dedicated `keyExtractor` tests.
- `archived={false}` is hardcoded per core's shared impl, same cut as #4885; an `archived` prop is new surface and tracked as the parity follow-up on the ink side.
- Tests: provider wiring, renderItem args, keyExtractor identity, FlatList prop pass-through; no real-store integration test, the package suite mocks the store everywhere.
- with-expo change is only the now-redundant double-wrap removal, no behavior change there.
- Additive only, no exports removed or reshaped.
- Changeset: patch (`@assistant-ui/react-native`).

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5201?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.P-nKYf8XfIdu3YWgFBtuRiN3hjLc-gMoV7NWnNupT0NP8-wMl4H21IJQId8?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/react-native-thread-list-item-scope.md
+++ b/.changeset/react-native-thread-list-item-scope.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react-native": patch
+---
+
+fix: wrap ThreadListItems rows in ThreadListItemByIndexProvider
--- a/examples/with-expo/components/thread-list/ThreadListDrawer.tsx
+++ b/examples/with-expo/components/thread-list/ThreadListDrawer.tsx
@@ -1,10 +1,6 @@
 import { View, Text, Pressable, StyleSheet } from "react-native";
 import { useSafeAreaInsets } from "react-native-safe-area-context";
-import {
-  ThreadListPrimitive,
-  ThreadListItemByIndexProvider,
-  useAui,
-} from "@assistant-ui/react-native";
+import { ThreadListPrimitive, useAui } from "@assistant-ui/react-native";
 import type { DrawerContentComponentProps } from "expo-router/build/react-navigation/drawer/types";
 import { ThreadListItem } from "./ThreadListItem";
 import { Icon } from "@/components/ui/icon";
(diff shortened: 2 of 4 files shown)


okisdev (approved):
thread list item primitives need a by-index scope, and Items is the place that owns it, same cut as ink #4885 and RN ThreadMessages. the new tests fail on main (no provider calls) and pass on this head. the with-expo drawer can drop its hand wrap.

PR #5205, opened 2026-07-26
fix(assistant-stream): harden SSE and data-stream decoders against prototype pollution

### LINKED ISSUE #5204: harden remaining assistant-stream JSON.parse decode boundaries with secure-json-parse
#5152 moved the assistant-transport decoder to secure-json-parse and #5200 did the same for UIMessageStreamDecoder. two untrusted-wire decode boundaries in assistant-stream still parse with bare JSON.parse:

- `packages/assistant-stream/src/core/utils/stream/SSE.ts` line 48 (SSEDecoder, used by GorpStreamResponse)
- `packages/assistant-stream/src/core/serialization/data-stream/serialization.ts` line 27 (data stream decode)

a frame carrying `__proto__` or `constructor.prototype` keys currently decodes with the pollution key as an own property. switching to `sjson.parse` would bring these in line with the transport and UIMS decoders. note that neither site has the try/catch drop path the other two decoders use (SSEDecoder throws on malformed JSON and errors the stream), so the swap needs a decision on error semantics per site rather than a blind replace, plus colocated vitest coverage and a patch changeset.

shashank-100:
## What

Swap the bare `JSON.parse` at the two remaining untrusted-wire decode boundaries in `assistant-stream` to `secure-json-parse` (`sjson.parse`):

- `packages/assistant-stream/src/core/utils/stream/SSE.ts` — `SSEDecoder` (used by `GorpStreamResponse`)
- `packages/assistant-stream/src/core/serialization/data-stream/serialization.ts` — `DataStreamChunkDecoder`

## Why

Both boundaries parse untrusted wire input. With bare `JSON.parse`, a frame carrying `__proto__` or `constructor.prototype` keys decodes the pollution key as an own property. #5152 (assistant-transport) and #5200 (UIMessage stream) already moved to `secure-json-parse`; this brings the last two decoders in line.

`secure-json-parse` is already a declared dependency of the package, so no new dependency is added.

## How

`sjson.parse` is a drop-in for `JSON.parse` that strips prototype-pollution keys. Per the issue, this is deliberately **not** a blind replace: neither site has the try/catch drop path the transport/UIMS decoders use — both intentionally *throw* on malformed JSON to error the stream. Only the parser is swapped; the throw-on-malformed error semantics are preserved at both sites.

## Tests

Colocated vitest coverage added for both decoders:

- decodes a normal JSON frame
- `__proto__` does not become an own property and does not pollute `Object.prototype`
- `constructor.prototype` does not pollute the prototype
- malformed JSON still errors the stream (semantics unchanged)

Closes #5204

> Note: local Node is v20 and the repo's pnpm requires v22.13+, so the tests were not run locally; relying on CI (Node 22+) to execute them.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5205?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/harden-json-parse-decode-boundaries.md
+++ b/.changeset/harden-json-parse-decode-boundaries.md
@@ -1,0 +1,7 @@
+---
+"assistant-stream": patch
+---
+
+fix(assistant-stream): harden SSE and data-stream decoders against prototype pollution
+
+The `SSEDecoder` and `DataStreamChunkDecoder` decoded untrusted wire input with bare `JSON.parse`, so a frame carrying `__proto__` or `constructor.prototype` keys decoded the pollution key as an own property. Both now parse with `secure-json-parse` (already used by the assistant-transport and UIMessage stream decoders), while preserving the existing throw-on-malformed-JSON error semantics.
--- a/packages/assistant-stream/src/core/serialization/data-stream/serialization.test.ts
+++ b/packages/assistant-stream/src/core/serialization/data-stream/serialization.test.ts
@@ -19,7 +19,9 @@ const decode = async (lines: string[]): Promise<DataStreamChunk[]> => {
 
 describe("DataStreamChunkDecoder", () => {
   it("decodes a chunk into type and value", async () => {
-    const [chunk] = await decode([`9:{"toolCallId":"1","toolName":"x","args":{}}`]);
+    const [chunk] = await decode([
+      `9:{"toolCallId":"1","toolName":"x","args":{}}`,
+    ]);
     expect(chunk).toEqual({
       type: "9",
       value: { toolCallId: "1", toolName: "x", args: {} },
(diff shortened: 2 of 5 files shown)


okisdev:
superseded by #5206, which landed the same two decoder swaps with drop-and-warn semantics matching the UIMessageStream and transport decoders (merged as 235c17e22). the tests here assert strip semantics while the implementation keeps secure-json-parse's default throw action, which is why Test Changed Packages fails: the four pollution tests expect a decoded chunk with the key absent, but the default action raises on those keys. thanks for picking up #5204.

PR #5230, opened 2026-07-27
fix(core): suppress controlled thread id echoes

### LINKED ISSUE #5228: onThreadIdChange echoes prop-initiated switches, breaking the controlled pattern it's designed for
### Checklist

### Environment info
```txt
@assistant-ui/react: 0.14.28
@assistant-ui/react-langgraph: 0.14.13
@assistant-ui/core: 0.2.22
React: 19.2
```

### Affected area(s)
- Core runtime (RemoteThreadList, controlled threadId)

### Description

#4482 introduced `onThreadIdChange` as the "out" half of the controlled pattern — *"threadId in, onThreadIdChange out"* — for syncing the active thread to a URL param. But the notification fires for **every** settled main-thread change, including switches that the controlled `threadId` prop itself initiated. That's an echo the consumer has to filter out, and it breaks the contract the API's own docs describe.

React's controlled-component convention: setting `value` does not fire `onChange`; only user/internal-initiated changes do. Here, pushing a new `threadId` in produces an `onThreadIdChange` right back out:

```
useRemoteThreadListRuntime.ts (controlled effect)
  prop changes → switchToThread / switchToNewThread

RemoteThreadListThreadListRuntimeCore.tsx
  _switchToThread settles → _notifyThreadIdChange() → onThreadIdChange(id)   ← echo
```

The existing dedupe (`_lastNotifiedThreadId`) only suppresses *same-value* re-notifications, not confirmations of prop-driven switches.

### Real-world failure

Our app (Next.js) binds `threadId` to a `?thread=` query param and navigates in `onThreadIdChange`:

1. User is on `/?thread=<id>`
2. User navigates to a different page that has no `?thread` param (an inbox)
3. `threadId` prop becomes `undefined` → controlled effect calls `switchToNewThread()` (correct)
4. Switch settles → `onThreadIdChange(undefined)` fires — the echo
5. The app's handler interprets "thread cleared" as a navigation intent and routes to the new-thread page, **hijacking the navigation the user just made** away from the inbox

### Expected behavior

`onThreadIdChange` should only fire for changes the runtime initiated (user clicked a thread, first message settled a new thread's remote id) — not as a confirmation of the value the consumer just passed in. Concretely: the controlled effect in `useRemoteThreadListRuntime` knows which switches it started; those settlements should not notify (or the core could compare against the current `options.threadId` before notifying).

### Current workaround

Consumer-side echo suppression:

```ts
onThreadIdChange: (id) => {
  if ((id ?? null) === (threadIdProp || null)) return; // echo of our own prop change
  onThreadChange(id ?? null);
},
```

Works, but every URL-syncing consumer will hit this and has to rediscover it — the API's primary use case per #4482 is exactly URL syncing.

Kinfe123:
Closes #5228

## What changed

- carry notification intent with each thread-switch operation
- suppress callbacks only for the exact switch initiated by the controlled `threadId` option
- keep notifications for concurrent user/runtime switches, including switches to the same target
- document the controlled callback behavior
- add hook-level coverage for initial IDs, subsequent IDs, clearing the ID, failed initial loads, and concurrent switches

## Why

A URL-driven `threadId` update used the public switch methods, so its settled value was echoed through `onThreadIdChange`. In apps that navigate from `?thread=<id>` to a page without that parameter, the echoed `undefined` could be interpreted as a fresh navigation intent and redirect the user away from the page they selected.

Suppression now travels with the prop-driven operation through the existing switch generation checks. Failed or superseded controlled switches leave no state behind, while a concurrent runtime switch carries its own callback policy.

## Testing

- `pnpm --filter @assistant-ui/core exec vitest run src/tests/useRemoteThreadListRuntime-controlled.test.tsx src/tests/remote-thread-list-reactive-threadId.test.ts`
- `pnpm --filter @assistant-ui/core test` (75 files, 676 tests)

<a href="https://rupic.dev/s/s2.dcKJZEmr9K_hi4g1kOOAoALqLQF-Mgs3B5xoslXCfhkLe9l6EZEwA5XeEQM?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>


okisdev (approved):
## review

LGTM as `patch`; the controlled pattern now behaves the way its own docs promised, and the test-file replacement is an upgrade, not a deletion.

### verification notes
- suppression travels as a parameter through the existing switch-generation chain, which is what makes the concurrency claims hold: a superseding user switch mid-prop-switch still notifies, a failed prop switch leaves no state behind (notification only happens at settle, so the dedupe cursor never moves on failure), and a suppressed settle still advances `_lastNotifiedThreadId`, keeping the pre-existing same-value dedupe semantics intact.
- red-checked against main: exactly the three behavior-change tests fail (echo suppression plus both empty-string cases) while the three invariant tests pass, so the suite discriminates precisely along the fix boundary. full core suite 75 files / 678 tests green on the branch.
- the deleted 92 lines were a `simulateThreadIdEffect` mirror of the hook's effect ("keep in sync" by hand); the effect itself moved into `setOptions`, so the simulation's subject no longer exists, and the real-core `onThreadIdChange` tests in the same file are kept. the new file drives the real hook through a React render harness.
- bonus fix verified: the old effect's truthiness check sent `threadId=""` to `switchToNewThread`; the new `!== undefined` path treats it as a supplied id, pinned by two tests.
- api-surface adds three private members only (append-only holds), docs and the `types.ts` JSDoc are updated in the same PR, `@assistant-ui/core` patch changeset present.

### potential further fixes
- no test covers mounting without `threadId` and supplying one later (the unconditional `_initialThreadLoaded` set is what makes that path work); mechanism-verified, worth a case whenever this file is next touched.

PR #5253, opened 2026-07-27
fix(react-hook-form): submit forms with standard inputs

Kinfe123:
## Summary

- resolve mounted fields with React Hook Form's `get()` helper
- allow `submit_form` to use ordinary single-ref inputs while preserving grouped-input refs
- cover standard, nested, and radio-group fields with regression tests

## Why

While testing `useAssistantForm()` with a normal registered text input, the AI's `submit_form` tool returned `Unable retrieve the form element` and never dispatched the form submission.

```tsx
const form = useAssistantForm<{ name: string }>();

return (
  <form onSubmit={handleSubmit}>
    <input {...form.register("name")} />
  </form>
);
```

React Hook Form stores ordinary inputs in `_f.ref`, but the existing lookup required `_f.refs`, which is used for grouped radio and checkbox fields. It also indexed `_fields` directly, so nested names such as `profile.name` could not be found. Using React Hook Form's own `get()` lookup and accepting either ref shape lets `requestSubmit()` reach the registered form without changing grouped-input behavior.

## Validation

- `react-hook-form` tests: 4 passed
- `@assistant-ui/react-hook-form` package build
- oxlint on the changed source and test files
- patch changeset included

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5253?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.jgA8zeqSdbwE4JfoA_6fmbxEBECANQw8qVjOXVU8orJGTZagMFWRvUTEbY8?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>


okisdev (approved):
approving. reproduced the failure mode and the fix on the merged head: with the old lookup, the new standard-input and nested-input tests fail exactly as described while the grouped-input and baseline tests stay green, and with `get()` plus the relaxed ref check all four pass (plus `tsc --noEmit` clean). the react-hook-form claims check out against both ends of the peer range: at the 7.83.0 test pin, `ref` is required and `refs` is grouped-only in the field types, and `get` has been a public export since v7.0.0 upstream, so the `^7` peer floor holds without a raise. I also updated the branch with main since #5275's accessor migration landed after your fork; the merge is coherent (main's property accessors plus your lookup fix) and CI re-ran green on it. dropping the controlled-forms extension in the last commit was the right scope call.

PR #5262, opened 2026-07-27
fix(core): restore repository duplicate-id throw; dedupe external-store messages at ingestion

Yonom:
MessageRepository's performOp duplicate-id ancestor scan detects internal tree corruption, so it must fail loudly: the `console.error` + skip introduced by d4bdf2c is restored to a hard `throw`. The scan stays hoisted above the cut block (atomic check-before-mutate). This partially reverts d4bdf2c per maintainer decision.

The recoverable case — an app-provided external-store `messages` array containing the same id twice — is now handled at the trust boundary instead: the external-store ingestion loop dedupes by id before linking, keeping the last occurrence (fresher snapshot) and emitting a `console.warn` naming the duplicated id. The optimistic-placeholder and resetHead logic below the loop is unaffected.

Tests: a contract test pinning the repository throw, and one at the external-store seam (duplicate ids warn, last one wins, no throw).

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5262?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.XMvMJWfPpA5k5W3cLVhTxHhV3vdsEPbB_kXp8aG9kXeONImsv4-runHr28U?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/core-repository-duplicate-id-throw.md
+++ b/.changeset/core-repository-duplicate-id-throw.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/core": patch
+---
+
+Restore the MessageRepository duplicate-id throw (it detects internal corruption); duplicate ids in an external-store messages array are now deduped at ingestion with a warning, keeping the last occurrence.
--- a/.changeset/react-external-thread-duplicate-ids.md
+++ b/.changeset/react-external-thread-duplicate-ids.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react": patch
+---
+
+ExternalThread: dedupe duplicate message ids with a warning (last occurrence wins) instead of throwing on duplicate resource keys.
--- a/packages/core/src/runtime/utils/message-repository.ts
+++ b/packages/core/src/runtime/utils/message-repository.ts
@@ -142,12 +142,9 @@ export class MessageRepository {
         current = current.prev
       ) {
         if (current.current.id === child.current.id) {
-          console.error(
-            new Error(
-              "MessageRepository(performOp/link): A message with the same id already exists in the parent tree. This error occurs if the same message id is found multiple times. This is likely an internal bug in assistant-ui.",
-            ),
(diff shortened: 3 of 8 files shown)


Yonom:
Addressed the three review asks in 785e4fecbb:

1. **parentId regluing** — non-issue in the ingestion loop: parenting is purely positional (`messages[i-1]`), so after a mid-array drop the following message reparents onto the preceding one; no path feeds explicit parentIds into this loop (ExportedMessageRepository/auiV0 import go through the separate `store.messageRepository` branch, untouched by the dedupe). Pinned with a contract test asserting the tree shape after a mid-array dedupe.
2. **Athena path** — confirmed on the dedupe path: `useAssistantTransportRuntime` (react/src/legacy-runtime/runtime-cores/assistant-transport/useAssistantTransportRuntime.ts:317) passes converter output as `messages` (ThreadMessage[], no convertMessage) to `useExternalStoreRuntime` (core/src/react/runtimes/useExternalStoreRuntime.ts) → `ExternalStoreRuntimeCore.setAdapter` → `ExternalStoreThreadRuntimeCore.__internal_setAdapter` → the `store.messages` branch containing the dedupe.
3. **ExternalThread** — was affected: it keys message clients by id via `useClientLookup`, and tap's `useResources` hard-throws on duplicate keys. Applied the same last-occurrence-wins dedupe with a warning + contract test.

PR #5280, opened 2026-07-28
fix(core): key cloud persistence cache on resolved threadListItem methods (sw-96)

Yonom:
## Problem

Regression from the property-API migration (#5275): the `globalPersistence` WeakMap in `AssistantCloudThreadHistoryAdapter` used to key on `aui.threadListItem()` — the bound client instance, stable per logical thread list item. Post-migration it keys on the `aui.threadListItem` accessor proxy, which is minted per host mount, so two mounts of the same logical item get separate `CloudMessagePersistence` caches.

## Fix

Key on the resolved bound client again via `unwrapClientAccessor`, newly exported from `@assistant-ui/store` (previously internal, already used by `Derived`). Restoring `apply` to return `read()` was considered instead, but `accessor-properties.test.tsx` pins `Object.is(aui.thread, aui.thread())` as the post-migration contract, so the call form intentionally returns the proxy — hence resolving through the unwrap helper. This adds a store export (api-surface regenerated); flagging for API-surface review.

Written against main independently of #5279 — the pair conflicts trivially in the `_persistence` key-lookup line; whichever merges second needs a one-line rebase.

## Tests

- Added a store contract test: two accessors over the same client unwrap to the same instance; non-accessor values pass through.
- `pnpm turbo test --filter=@assistant-ui/core --filter=@assistant-ui/store`: 761 + 26 passed. Build, lint, and `api-surface:check` clean.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5280?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.gsS0cIMDFGIyqiGFc14G6pk3l2vIZm2nxeJrT2Sn81OrsC-i2hpdnWohgko?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/persistence-weakmap-key.md
+++ b/.changeset/persistence-weakmap-key.md
@@ -1,0 +1,6 @@
+---
+"@assistant-ui/store": minor
+"@assistant-ui/core": patch
+---
+
+fix: key the cloud message persistence cache on the resolved threadListItem client instead of the per-mount accessor proxy; export unwrapClientAccessor from @assistant-ui/store
--- a/.changeset/persistence-weakmap-key.md
+++ b/.changeset/persistence-weakmap-key.md
@@ -1,5 +1,5 @@
 ---
-"@assistant-ui/store": minor
+"@assistant-ui/store": patch
 "@assistant-ui/core": patch
 ---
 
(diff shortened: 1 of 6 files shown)


Yonom:
Superseded by https://github.com/assistant-ui/assistant-ui/pull/5282 — WeakMap re-key now uses getClientId (sw-98)

PR #5298, opened 2026-07-28
fix(core): refresh cloud attachments on client change

Kinfe123:
## Summary

- let `CloudFileAttachmentAdapter` resolve the current Cloud client when an upload starts while preserving its existing fixed-client constructor
- keep one state-owned attachment adapter instance for component-lifetime identity
- add regression coverage that switches clients and uploads immediately, before the thread-list reload settles

## Why

While testing a workspace switch from Cloud client A to B, I found that the thread list correctly reloaded through B but attachment uploads still requested presigned URLs through A. The stable provider introduced around the thread-list adapter keeps the attachment adapter mounted, and that adapter captured the first client at construction time.

This could upload a newly selected file through the previous account or workspace after login, tenant, or Cloud client changes. The fix follows the current-client getter pattern already used by Cloud history persistence, resolving the active client at the start of each new upload.

This is a follow-up to #5224, which stabilized Cloud thread-list adapter identity and intentionally reloads the list when the resolved Cloud client changes.

## Verification

- reproduced the old behavior: after switching A to B, `generatePresignedUploadUrl()` was still called on A
- verified an immediate upload after the switch calls B, leaves A untouched, and reaches `requires-action`
- focused attachment and `useLocalRuntime` tests pass: 2 files, 7 tests
- `@assistant-ui/core` package build passes
- core API-surface check passes
- changed source files pass oxlint and oxfmt

<a href="https://rupic.dev/s/s2.t1xX72-G-8hdEWEWno8kuXRg4bqwJNNyIQbl5beHX01FzM_zS_eSRyCtBtA?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>


okisdev (approved):
the getter is the right fix. i merged main through the pagination conflict in `useCloudThreadListAdapter` and regenerated the constructor overload against the current `#private` api-surface form.

on this head, `pnpm exec vitest run src/react/runtimes/useLocalRuntime.test.tsx src/react/runtimes/cloud/CloudFileAttachmentAdapter.test.ts` is 7/7. the new case still fails on current main (`firstCloud.files.generatePresignedUploadUrl` called once). pending chips surviving a workspace switch is a follow-up; `AssistantCloud` has no workspace identity to invalidate them.

PR #5299, opened 2026-07-28
fix(docs): pass tools client to AssistantRuntimeProvider in Artifacts example

### LINKED ISSUE #4856: fix(docs): Artifacts example does not expose or invoke render_html
## Problem

The hosted Artifacts example presents itself as an agent that can generate and render interactive artifacts, but the active agent does not receive a working `render_html` tool.

The repository contains an artifact toolkit declaration, but the tool is not correctly connected to the runtime used by the example. The model can respond with ordinary text or code instead of producing an artifact tool call, so the preview does not work as documented.

## Steps to reproduce

1. Open `/examples/artifacts`.
2. Start a new thread.
3. Ask the assistant to create a webpage, calculator, or other HTML artifact.
4. Observe that the assistant does not reliably call `render_html`.
5. Inspect the outgoing `/api/chat` request or active model context and confirm the expected artifact tool is absent or disconnected.

## Expected behavior

- The artifact agent receives an explicit artifact tool.
- Requests for webpages or visual HTML invoke the tool.
- Tool arguments stream into the source panel.
- Preview becomes available after tool execution completes.
- Client-tool, backend-tool, runtime-provider, and system-prompt ownership are explicit.

## Reference implementation

`examples/with-artifacts` demonstrates the intended architecture:

- client toolkit using `externalTool()`
- toolkit attached through `AssistantRuntimeProvider aui={aui}`
- dedicated backend `render_html` tool
- artifact-specific system prompt
- preview enabled after the tool result is available

## Follow-up: make artifacts a Base capability

Consider making artifacts a capability of the main Base example instead of maintaining a separate, divergent chat shell. The generalized contract could support:

```ts
render_artifact({
  type: html | markdown,
  code: string,
  title?: string,
})
```

- Render HTML in a sandboxed iframe.
- Render Markdown with Streamdown so streaming Markdown remains visually stable while tool arguments arrive.
- Retain source and preview tabs.
- Define compatibility for existing `render_html` calls.
- Support desktop and mobile artifact layouts.

## Acceptance criteria

- `/examples/artifacts` exposes a working artifact tool to the model.
- The tool is included in the backend request and recognized by the server.
- A webpage request produces a visible HTML artifact.
- Markdown artifacts render through Streamdown.
- Provider wiring has a regression test.
- Tool invocation and artifact extraction have targeted tests.

samdickson22:
## What

Fixes #4856 — the hosted docs Artifacts demo never invokes `render_html`; asking for a webpage yields plain text instead of a rendered preview.

## Why

The bug is in the hosted docs example, not `examples/with-artifacts` (which is wired correctly). In `apps/docs/contexts/ArtifactsRuntimeProvider.tsx`, the artifacts tools client (`useAui` with the artifacts toolkit) was mounted as an `AuiProvider` **around** `AssistantRuntimeProvider`. But the runtime provider never reads ambient context: it passes an explicit `{ parent: aui ?? null }` into `useAui`, which bypasses the ambient-context default. With `parent: null`, the runtime client creates a fresh empty ModelContext, so the toolkit's `render_html` registration never reached the runtime — the model saw no tools.

## How

Pass the tools client via the `aui` prop and drop the now-redundant outer `AuiProvider` (one file). `AssistantProviderBase` re-installs the parent as an outer `AuiProvider` itself, so descendants (children, DevToolsModal) keep seeing the tools scope — now correctly inherited by the runtime client via the scope prototype chain.

Note: `examples/with-artifacts` wires the same relationship in the opposite direction (runtime provider outside, tools client inside with the runtime client as ambient parent). Both directions are supported by the scope-inheritance design; this fix is the minimal change for the existing file structure and matches the documented purpose of the `aui` prop.

## Verification

- Before/after demo videos attached below (before: request for a webpage returns plain text, no artifact panel; after: `render_html` fires and the preview renders).
- Affected package build passes; oxlint/oxfmt clean.
- `apps/docs` is private → no changeset.

<a href="https://rupic.dev/s/s2.nJrLFzQ0d2PSswXpFPmEXkAc8Xg7VwtotIwsumG-ngIR6CSo__1wLb4I5xg?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

## Demo videos

Before (plain text, no artifact panel) / After (render_html fires, preview renders):

https://github.com/user-attachments/assets/e0173122-d00a-4cfe-ba3c-12aca934eba0

https://github.com/user-attachments/assets/c0b34aeb-00f6-4022-9268-10a1834b8447

### DIFF
--- a/apps/docs/contexts/ArtifactsRuntimeProvider.tsx
+++ b/apps/docs/contexts/ArtifactsRuntimeProvider.tsx
@@ -7,7 +7,6 @@ import {
   AssistantCloud,
   useAui,
   Tools,
-  AuiProvider,
   type Toolkit,
 } from "@assistant-ui/react";
 import { useChatRuntime } from "@assistant-ui/react-ai-sdk";
@@ -64,12 +63,10 @@ export function ArtifactsRuntimeProvider({
   });
 
   return (
-    <AuiProvider value={aui}>
-      <AssistantRuntimeProvider runtime={runtime}>
-        {children}
+    <AssistantRuntimeProvider runtime={runtime} aui={aui}>
+      {children}
 
-        <DevToolsModal />
-      </AssistantRuntimeProvider>
-    </AuiProvider>
+      <DevToolsModal />
+    </AssistantRuntimeProvider>
   );
 }


okisdev (approved):
## review

LGTM as is; correct root-cause diagnosis, minimal diff, and the mechanism checks out at the source level.

### verification notes
- `AssistantProviderBase` always calls `useAui` with an explicit `{ parent }` (null when no `aui` prop), and `useAui`'s ambient-context read lives in the default value of the whole second parameter, so it only applies when that parameter is omitted entirely. wrapping the runtime provider in an ambient `AuiProvider` was therefore provably inert, and `parent: null` resolves to a fresh root client with an empty ModelContext, which is exactly the #4856 symptom.
- the fix uses the `aui` prop for its documented purpose, and the base provider's two-layer structure (parent as the outer `AuiProvider`, runtime client inner) keeps descendants on the tools scope through the prototype chain, so children and `DevToolsModal` are unaffected.
- `examples/with-artifacts` is confirmed wired the opposite, also-valid direction (runtime outside, tools client inheriting ambiently with no second argument), so no change is needed there.
- apps/docs is private, so correctly no changeset. before/after videos plus the source trace cover verification for a demo page with no test infrastructure.

PR #5300, opened 2026-07-28
fix(ui): keep ModelSelector popover side stable while filtering

### LINKED ISSUE #5124: Model Selector jumps around when filtering
### Checklist

### Environment info

Chrome MacOS, reproduced on docs website

```
Environment:
  OS:               macOS 15.6.1 (arm64)
  Node.js:          v24.8.0
  Package Manager:  pnpm 11.10.0
  Framework:        Unknown
  Monorepo:         yes

Packages:
  @assistant-ui/x-changelog  0.0.0

Ecosystem:
  typescript  7.0.2
```

### Affected area(s)

_No response_

### Runtime adapter in use

_No response_

### Steps to reproduce

https://github.com/user-attachments/assets/9a2fdadc-c3dd-4fd7-b8c7-7d3a44f6cf9b

Position of combobox is stable after first opening

### Browser (if applicable)

Chrome

### Code snippet or error log

```tsx

```

### Additional context

Have done a little investigation on this but the fix looks time consuming enough it made sense to make an issue

The BaseUI's PopoverPrimitive.Positioner used by the ModelSelector has an issue where it will choose the right popover location for its initial size, but then when filtering to a smaller size it may jump to the preferred side unnecessarily.

https://github.com/user-attachments/assets/9a2fdadc-c3dd-4fd7-b8c7-7d3a44f6cf9b

BaseUI has a fix for this already called `lazyFlip`, which will retain the initial position even if the preferred one becomes possible, but it's only implemented/exposed in the BaseUI Combobox component, so would either need exposing upstream or a migration of this component to the Combobox to be considered. I've confirmed with a vibed up fix that Combobox does behave correctly, just not a merge-ready output

samdickson22:
## What

Fixes #5124 — the ModelSelector popover jumps between opening upward and downward on every keystroke while typing in its filter when near a viewport edge.

## Why

Base UI Popover's flip collision-avoidance is eager and memoryless: opened near the viewport bottom it flips the popup to top; typing shrinks the filtered list until it fits below again, flipping it back — so each keystroke can bounce the side. Base UI fixed this exact case upstream with an internal `lazyFlip` flag, but in `@base-ui/react` 1.6.0 it is only wired into the Combobox positioner and is not exposed on Popover.

## How

A `useLazyFlipSide` hook in `ModelSelectorContent` observes the popup's `data-side` attribute and feeds the rendered side back as the preferred side — replicating upstream lazyFlip semantics:

- The side stays put while the list grows/shrinks during filtering.
- Flipping remains enabled: if the locked side genuinely stops fitting, floating-ui flips, `data-side` changes, and the new side becomes preferred. No UX regression.
- Once open, a consumer-passed `side` prop acts as the initial preference and the rendered side wins until close (side resets on unmount). This is intentional and matches Base UI's own lazyFlip behavior.

Applied to both flavors — Radix's Popper repositions on popup resize the same way, just less dramatically — keeping base/radix parity. The hook is duplicated per file to match the kit's standalone-file convention for registry copying.

A migration to Base UI Combobox was considered and rejected: it would rewrite the publicly-composable cmdk-based `ModelSelector.*` surface for what is a positioning bug.

## Verification

- Before/after demo videos attached below (before: popover bounces up/down per keystroke; after: side stays stable while filtering, still flips when it truly no longer fits).
- `tsc --noEmit` on packages/ui exit 0; oxlint `--deny-warnings` and oxfmt clean on both files; no template copies of model-selector exist, so Template Sync is unaffected.
- `packages/ui` is private → no changeset.
- No unit test: jsdom performs no layout; the mechanism was verified against Base UI's own lazyFlip implementation.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5300?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.RuOliaOELPByEycITQUoo1mUraW4f-qbMEGgFcpdA0qtDggVhuskMWoPf3o?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

## Demo videos

Before (popover bounces between sides per keystroke) / After (side stays stable while filtering):

https://github.com/user-attachments/assets/59682412-929d-4288-87ad-be4a956d0b8f

https://github.com/user-attachments/assets/2883ce82-2f54-4900-a723-59d6cd9d0d44

### DIFF
--- a/packages/ui/src/components/assistant-ui/model-selector.radix.tsx
+++ b/packages/ui/src/components/assistant-ui/model-selector.radix.tsx
@@ -374,6 +374,38 @@ export type ModelSelectorContentProps = ComponentPropsWithoutRef<
   searchable?: boolean;
 };
 
+// The popover re-evaluates collision flipping whenever the popup resizes, so
+// filtering the list down flips the popup back to the preferred side
+// mid-interaction. Feed the rendered side back as the preferred side, making
+// the popup keep its side until it no longer fits.
+function useLazyFlipSide(): {
+  side: ModelSelectorContentProps["side"];
+  popupRef: (node: HTMLDivElement | null) => void;
+} {
+  const [side, setSide] = useState<ModelSelectorContentProps["side"]>();
+  const observerRef = useRef<MutationObserver | null>(null);
+  const popupRef = useCallback((node: HTMLDivElement | null) => {
+    observerRef.current?.disconnect();
+    observerRef.current = null;
+    if (!node) {
+      setSide(undefined);
+      return;
+    }
+    const sync = () => {
+      const rendered = node.getAttribute("data-side");
+      if (rendered) setSide(rendered as ModelSelectorContentProps["side"]);
+    };
+    sync();
+    const observer = new MutationObserver(sync);
+    observer.observe(node, {
+      attributes: true,
+      attributeFilter: ["data-side"],
+    });
+    observerRef.current = observer;
+  }, []);
+  return { side, popupRef };
+}
+
 /**
  * Hidden input that anchors cmdk's keyboard navigation, keeping the list
  * keyboard-operable without a visible search box. ModelSelectorContent renders
@@ -390,19 +422,23 @@ function ModelSelectorFocusAnchor() {
 function ModelSelectorContent({
   className,
   align = "start",
+  side,
   sideOffset = 6,
   searchable,
   children,
   ...props
 }: ModelSelectorContentProps) {
   const { value } = useModelSelectorContext();
+  const { side: renderedSide, popupRef } = useLazyFlipSide();
   const unfiltered =
     searchable === false || (!searchable && children === undefined);
 
   return (
     <PopoverContent
+      ref={popupRef}
       data-slot="model-selector-content"
       align={align}
+      side={renderedSide ?? side ?? "bottom"}
       sideOffset={sideOffset}
       className={cn(
         "bg-popover/95 w-72 min-w-(--radix-popover-trigger-width) overflow-hidden rounded-xl p-0 shadow-lg backdrop-blur-sm",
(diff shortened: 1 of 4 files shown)


okisdev (approved):
## review

LGTM as is; clean diagnosis-to-mechanism chain and the review loop was already fully absorbed before I got here.

### verification notes
- the load-bearing upstream fact checks out at the pinned version: `lazyFlip` lives in base-ui's internal `useAnchorPositioning` and is wired only into `ComboboxPositioner` at v1.6.0, with zero references in `PopoverPositioner`, so a local mirror was the right call over waiting on upstream exposure or rewriting the composable cmdk surface onto Combobox.
- the mechanism is a fixed point: feeding the rendered `data-side` back as the preferred side cannot oscillate (feeding back the current value is stable), and a genuinely non-fitting locked side still flips once and re-locks. close resets via the null callback ref. both flavors' `data-side` placement is corroborated by this file's own `data-[side=…]` selectors.
- the three review threads were each addressed by follow-up commits before this review (`69e254881` forwards the content refs both flavors need under the `^18 || ^19` peer range; `b1a3dc115` lands the side-semantics JSDoc in both files); resolved accordingly.
- verified in a clean worktree: `tsc --noEmit` on packages/ui exits 0 with built deps, oxlint clean on all four files, no template copies of model-selector (Template Sync unaffected), private package so correctly no changeset.
- residual, not this PR's defect: a registry consumer holding an older `popover.tsx` copy alongside a fresh `model-selector.tsx` loses the ref path on React 18 silently; inherent to copy-based distribution with declared registryDependencies.

PR #5308, opened 2026-07-29
fix(tap): drain large update batches across flush passes instead of dropping them

tiann:
## Problem

`UpdateScheduler` batches dirty resources into macrotask flushes capped at `MAX_FLUSH_LIMIT` (50) tasks per pass. When a single flush contains more dirty schedulers than that, the pass threw `Maximum update depth exceeded` — and the `finally` block cleared the scheduler set, **permanently discarding every dirty scheduler that hadn't run yet**.

50 is easy to exceed with legitimate bulk updates, not just infinite loops. Concrete case from a production app (HAPI, using `useExternalStoreRuntime`): prepending one page of older messages (200 items) into the external store mounts ~600 message/part/composer resources in one tap flush. The flush aborted, the thread never rendered the merged page even though the runtime core had already applied it, and the UI appeared stuck (or kept re-triggering loads). The same hazard exists for any large one-shot update (bulk thread switch, importing a big transcript, etc.).

The initial mount path doesn't hit this because resource creation there is driven by React's render, not by a dirty-scheduler cascade inside an already-running flush.

## Fix

A flush pass now runs at most `MAX_FLUSH_LIMIT` tasks and **defers the rest to a follow-up pass** (same macrotask scheduling), so large batches drain across several passes instead of being dropped:

- queue fully drained → reset, done (unchanged behavior for the common case)
- queue still non-empty → schedule another pass
- queue saturated for `MAX_SATURATED_FLUSH_PASSES` (100) **consecutive** passes → a resource is re-dirtying itself on every run, i.e. a true infinite update loop → keep throwing `Maximum update depth exceeded` (same message as before)

The loop guard therefore still fires, just without false-positiving on large finite batches. Error handling for tasks that throw is unchanged: the batch is discarded and the error propagates.

## Tests

Added `scheduler.batched-drain.test.ts` with a manually-pumped MessageChannel so passes can be stepped synchronously:

- 120 dirty schedulers all run across passes, no throw (previously threw + dropped)
- a self-re-dirtying scheduler still throws `Maximum update depth exceeded`
- the saturated-pass counter resets after a fully drained pass (back-to-back 75-scheduler bursts don't accumulate)

Full `@assistant-ui/tap` suite: 457 tests passing.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5308?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.0KS9rIkoIcaXg2G_NnNw-sfGZON2fp1X0NWvd_YvCe4SNdHj-DwbDQw8JIk?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/scheduler-batched-drain.md
+++ b/.changeset/scheduler-batched-drain.md
@@ -8,13 +8,14 @@ When more than 50 schedulers were dirty in one flush (e.g. hundreds of
 resources mounting from a single bulk state update), the scheduler aborted
 the flush and discarded the remaining dirty schedulers, leaving resources
 permanently stale. A flush now drains the queue completely, chunking
-silently across macrotasks when a batch exceeds one pass, so bulk updates
-of any size land without errors. Infinite-loop detection no longer depends
-on batch shape: a per-scheduler re-run guard (> 50 runs in one flush)
-drops just the offending scheduler, so self-re-dirtying resources and
-mutually re-dirtying rings throw "Maximum update depth exceeded" while the
-rest of the queue keeps flushing; and a documented last-resort bound of
-500000 tasks per burst terminates cascades of freshly minted schedulers
-without ever dropping queued work. `createTapRoot` now holds one
-UpdateScheduler per root (mirroring `useTapRoot`) so the guard covers
-every root type.
+silently across macrotasks when a batch exceeds one pass (1000 tasks), so
+bulk updates of any size land without errors. Infinite-loop detection no
+longer depends on batch size: a per-scheduler re-run guard (> 50 runs in
+one flush) skips just the offending scheduler and retries it next pass,
+so self-re-dirtying resources and mutually re-dirtying rings report
+"Maximum update depth exceeded" while the rest of the queue — and any
+deep-but-finite cascade — keeps flushing; a documented last-resort bound
+of 20000 tasks per burst (~33x a real 600-resource history-page load)
+terminates cascades of freshly minted schedulers without dropping queued
+work. `createTapRoot` now holds one UpdateScheduler per root (mirroring
+`useTapRoot`) so the guard covers every root type.
--- a/.changeset/scheduler-batched-drain.md
+++ b/.changeset/scheduler-batched-drain.md
@@ -9,6 +9,8 @@ resources mounting from a single bulk state update), the scheduler aborted
 the flush and discarded the remaining dirty schedulers, leaving resources
 permanently stale. A flush now drains the queue completely; bulk updates of
 any size land in one batch. Infinite-loop detection no longer depends on
-batch size: it counts per-scheduler re-runs within a burst (> 50), so true
-update loops throw "Maximum update depth exceeded" immediately while finite
-batches of any size drain successfully.
+batch size: per-scheduler re-runs within a burst (> 50) catch resources
+that re-dirty themselves or each other, and a burst-wide task cap (10000)
+backstops unbounded cascades of fresh schedulers — so true update loops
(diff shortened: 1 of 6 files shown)


tiann:
Thanks for the careful reviews — both blocking points are addressed in the new commit:

**flushTapSync no longer drops overflow.** It now loops runFlushPass synchronously until the queue is empty (or the loop guard fires), so all notifications land before it returns — the useAui "structural
rebinds land before the notification returns" contract holds for batches of any size. The per-burst loop counter lives on flushState, so nested flushTapSync calls get a fresh counter instead of
sharing/resetting the outer one.

**The residual 5000-task ceiling is gone.** Loop detection is no longer "N consecutive saturated passes" but per-scheduler re-runs within a single burst (> 50). This is a truer definition of an infinite loop:
- finite batches of any size drain — pinned by a 6000-scheduler test (120 passes, no throw)
- chains of distinct schedulers never false-positive (each runs exactly once)
- a self- or mutually-re-dirtying loop throws on its first saturated pass (~50 runs), so the guard fails as fast as the original code instead of 100x slower

New tests cover both previously-missing cases: flushTapSync above the limit (synchronous drain, nothing deferred to a macrotask) and a burst larger than the old ceiling. Full @assistant-ui/tap suite: 460
passing.

On the width-cap trade-off: keeping the 50-task pass budget is intentional — Set iteration visits schedulers re-added mid-pass, so an uncapped pass is exactly where an infinite loop would hang the main
thread. The width cap is what makes in-pass looping bounded (and countable) at all; the cost is that a 600-resource batch lands across ~12 macrotasks/React commits rather than one. Documented in the code.

Kinfe123 (changes requested):
There is already an unresolved scheduler finding where large finite batches can remain pending after reaching the burst limit. let's resolve that.

Yonom:
Thank you for this. I believe you have identified three real issues:
- UpdateScheduler's bailout should never trigger. (🔄)
  - We should only have 1 UpdateScheduler instance in a typical react app(🔄)
    - https://github.com/assistant-ui/assistant-ui/pull/5357
    - useAui with only Derived scopes should not mount its own scheduler (🔄)
  - Independent UpdateSchedulers should not share the same limit
    - https://github.com/assistant-ui/assistant-ui/pull/5370
- UpdateScheduler's decision to bail leaves unrelated tap roots in a dirty state. The blast radius should be smaller. This is a minor bug, because the bailout should never happen. ✅
  - https://github.com/assistant-ui/assistant-ui/pull/5370
- UpdateScheduler should allow large numbers of tap roots to be mounted in the same pass. It might be helpful to yield to the browser every ~5ms to not block page interactivity. ❌
  - WONTFIX: Given the goal to have 1 UpdateScheduler / assistant-ui app, this wouldn't help us much.

Thank you so much for pointing out the source of this problem. Your PR description is extremely helpful. Thank you also for proposing this PR as the solution.

I suspect we have implemented MAX_FLUSH_LIMIT wrong. Its intent is to prevent infinite loops, in particular across multiple tap-root ping-pongs.

I will be investigating some more, to understand the problem space better and see if I can find a more elegant solution.

Issue #5330, opened 2026-07-29
tap: React-hosted reducer replays duplicate already-committed updates (no cross-lane rebase fidelity)


okisdev:
## Summary

the React-hosted bridge (`useResourceFiberHostUtilsReact`) cannot rebase tap reducer cells the way React rebases its own hook state. when React replays the version reducer chain from a base below the last commit (pending transition + `flushSync` interleave), every replayed `applyUpdate` closure applies onto the cell's already-committed `current`, so updates that were part of the earlier commit are applied twice.

## Repro

`packages/tap/src/__tests__/react/concurrent-version-replay.test.tsx` (added alongside the assistant-ui/assistant-ui#5317 fix): a reducer cell accumulating `(s, c) => [...s, c]`, `startTransition(() => push("t1"))` followed by `flushSync(() => push("s1"))` renders `s1,t1,s1`. the React oracle for the same sequence is `t1,s1`. the behavior is identical before and after the assistant-ui/assistant-ui#5317 fix (the pre-fix throw path double-applied the same way after React's recoverable-error retry), so this is not a regression of that change; the test deliberately asserts delivery containment, not exact equality, to avoid pinning the duplication as contract.

## Mechanism

React keeps per-hook `baseState` + `baseQueue` and rewinds to the base before replaying. tap cells only keep `current` (committed) and `workInProgress`; there is no pre-commit base snapshot to rewind to, and the version counter is a proxy that cannot restore cell state. a related hazard sits in `useReducer.ts` (~150-166): the drain reuses a memoized `eagerState` computed against a different predecessor chain, so rebased cumulative reducers (`v => v + 1`) can land wrong values even when applied once.

## Scope

last-write-wins reducers (the dominant streaming pattern, `useState`-style setters included) are unaffected by duplication, which is why this stays latent in production. accumulating reducers across mixed lanes are affected. fixing it properly needs per-cell base snapshots or replay-aware apply deduplication, which is a design change rather than a patch.

[...]

PR #5331, opened 2026-07-29
fix(tap): accept reducer replays below the committed version instead of throwing

### LINKED ISSUE #5317: tap: "Version is less than committed version" thrown from useReducer replay under React concurrent rendering (RemoteThreadList + streaming)
## Bug description

`@assistant-ui/tap` throws `Version is less than committed version` (`packages/tap/src/core/helpers/root.ts`, `setRootVersion`) from inside the React-hosted reducer bridge. React 19 reports it as a recoverable error and re-renders synchronously, so the app keeps working, but the error fires repeatedly during agent streaming.

## Environment

- `@assistant-ui/react` 0.15.1, `@assistant-ui/core` 0.3.1, `@assistant-ui/store` 0.3.1, `@assistant-ui/tap` 0.9.7, `@assistant-ui/react-langgraph` 0.14.16
- Next.js 16.2.6 (Turbopack, App Router), React 19.2.4
- Single instance of core/store/tap in node_modules (verified — no duplicate-package issue)

## Setup

`useRemoteThreadListRuntime` whose `runtimeHook` is `useLangGraphRuntime`, streaming SSE agent responses. Each alive thread mounts a tap host via `ThreadListItemRuntimeProvider → useAui`. High-frequency dispatches during streaming.

## Stack trace

```
Error: Version is less than committed version
    at setRootVersion (tap/dist/core/helpers/root.js)
    at useResourceFiberHostUtilsReact.useReducer (tap/dist/hooks/utils/useResourceFiberHostUtils.js)
    at updateReducerImpl (react-dom)
    at updateReducer (react-dom)
    at useReducer (tap)
    at useResourceFiberHostUtilsReact (tap)
    at useResourceFiberHost (tap)
    at useTapHost (tap)
    at useHostedAssistantClient (store/useAui)
    at useAui (store)
    at ThreadListItemRuntimeProvider (core)
    at renderWithHooks / workLoopConcurrentByScheduler (react-dom, concurrent render)
```

## Analysis

In `useResourceFiberHostUtilsReact`, the React `useReducer`'s reducer calls `setRootVersion(root, v)` with the React-side version counter, while tap's `committedVersion` advances in `commitRoot` during effects.

Under concurrent rendering, when a render lane is interrupted and React later **replays the reducer from an earlier base state** (update rebase), the replayed `v` can be lower than `root.committedVersion` — a commit already happened for a version the rebased lane hasn't seen. `setRootVersion` hits the `root.committedVersion > version` guard and throws.

The stack shows exactly this: the throw happens inside `updateReducerImpl` during `workLoopConcurrentByScheduler`, i.e. during React's reducer replay, not during a normal dispatch.

## Expected

This looks like the same class of recoverable inconsistency that #5250 downgraded from throw to warn. Suggested handling in `setRootVersion`: clamp to the committed state instead of throwing —

```ts
if (root.committedVersion > version) {
  root.version = root.committedVersion;
  root.changelog.length = 0;
  // warn in development
  return;
}
```

The next committed render restores consistency (which is effectively what React's recoverable-error path already does today, minus the noise).

<a href="https://rupic.dev/s/s2.lVCom6j4ig5IAznhi3U-IsrJAmA0TckVeYQvfDHeZoeMVyZ-_fWBX6uHndg?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

okisdev:
## Problem

`setRootVersion` threw `Version is less than committed version` whenever React 19 concurrent rendering replayed the reducer bridge from a base below the last commit (pending transition rebased across a sync commit, the exact interleave agent streaming produces). React treats the throw as a recoverable error and retries, so the app kept working, but the error fired on every replay and flooded the console/overlay. #5317 reports it.

## Why not the clamp (#5321)

the rollback branch already lands on a consistent snapshot: `root.version = version` runs before the guard, rollback callbacks reset every dirty cell, and clearing the changelog is exactly what the `version === committedVersion` branch does. the throw added nothing on top of that except aborting the render React was going to retry anyway. #5321 proposed clamping `root.version` up to `committedVersion` instead; empirically (same bridge test, deficit >= 2 traces) the upward clamp makes later reducer calls in the same replayed chain re-trigger the rollback and wipe the records their predecessors just applied, and leaves React's counter stranded below the committed version where subsequently dispatched updates are applied and then rolled back before render. keeping the low version is the loss-free shape: the replayed chain ascends from its own base and re-supplies every update, which is how the accidental throw-and-retry path already behaved, minus the error.

## Approach

widen the `version === root.committedVersion` branch to `version <= root.committedVersion` and delete the throw. the widened branch also re-bases `committedVersion` to the replayed version: the between-versions branch derives the changelog base from `committedVersion`, and a stale-high value there over-drops records on a later partial rollback (review-traced arithmetic, pinned by a unit test). `committedVersion` is read nowhere outside tap and the next `commitRoot` overwrites it. no logging: after the change a below-committed version is an expected concurrent replay state, not an anomaly, and erroring on it would recreate the console noise half of #5317. `commitRoot` may subsequently move `committedVersion` downward on a partial-lane commit, which commits the replayed chain's snapshot and stays internally consistent. blast radius: only the React-hosted root can hold `version < committedVersion`; `useTapRoot`'s scheduler calls `setRootVersion` with `committedVersion` or `committedVersion + changelog.length`, never below, so the `committedVersion + changelog.length` consumers at `useTapRoot.ts:49` and `:92` never observe the widened state.

## Verification

- new root-level unit tests pin the contract: below-committed input keeps the given version, clears the changelog, runs rollback callbacks, re-bases `committedVersion` to the given version, does not throw.
- new bridge-level test (`concurrent-version-replay.test.tsx`) drives the real `useResource` React host through `startTransition` + `flushSync` with an `onRecoverableError` probe: on current main it fails (the version error surfaces as a recoverable error), with this change it passes with zero recoverable errors and both updates delivered.
- full tap suite: 33 files / 458 tests green. `oxfmt` clean; package `tsc` shows only the pre-existing bench-file dist-resolution errors of an unbuilt worktree.

## Scope

the bridge test asserts delivery containment rather than exact ordering equality: cross-lane replays duplicate already-committed updates because tap cells have no per-hook base snapshot to rewind to. that rebase-fidelity gap predates this change (the throw path double-applies identically after React's retry) and is tracked in #5330 together with the related `eagerState` memoization hazard.

fixes #5317. supersedes #5321.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5331?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.F3uvGhIrg9n5sXO7pInIbCV1HYFIdmSOZeif4ZMMGtUACodJtwJ3dQJudWQ?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/tap-version-replay.md
+++ b/.changeset/tap-version-replay.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/tap": patch
+---
+
+fix: accept reducer replays below the committed version instead of throwing
--- a/packages/tap/src/__tests__/react/concurrent-version-replay.test.tsx
+++ b/packages/tap/src/__tests__/react/concurrent-version-replay.test.tsx
@@ -3,11 +3,26 @@ import { StrictMode, startTransition } from "react";
 import { flushSync } from "react-dom";
 import { createRoot, type Root } from "react-dom/client";
 import { act } from "@testing-library/react";
+import type { TapRoot } from "../../core/types";
 import { resource } from "../../core/resource";
 import { useResource } from "../../index";
 import { useReducer as useResourceReducer } from "../../react-hooks/useReducer";
 import { useMemo as useResourceMemo } from "../../react-hooks/useMemo";
 import { cleanupAllResources } from "../test-utils";
+
+const probes = vi.hoisted(() => ({ belowCommitted: 0 }));
+
+vi.mock("../../core/helpers/root", async (importOriginal) => {
+  const original =
+    await importOriginal<typeof import("../../core/helpers/root")>();
+  return {
+    ...original,
+    setRootVersion: (root: TapRoot, version: number) => {
+      if (version < root.committedVersion) probes.belowCommitted++;
+      return original.setRootVersion(root, version);
+    },
+  };
+});
 
 let reactRoot: Root | undefined;
 let container: HTMLElement | undefined;
@@ -77,5 +92,6 @@ describe("React-hosted reducer replay below the committed version", () => {
 
     expect(container.textContent).toContain("after");
     expect(recoverable).not.toHaveBeenCalled();
+    expect(probes.belowCommitted).toBeGreaterThan(0);
   });
 });
--- a/packages/tap/src/__tests__/react/concurrent-version-replay.test.tsx
+++ b/packages/tap/src/__tests__/react/concurrent-version-replay.test.tsx
@@ -1,0 +1,81 @@
+import { describe, it, expect, afterEach, vi } from "vitest";
+import { StrictMode, startTransition } from "react";
+import { flushSync } from "react-dom";
+import { createRoot, type Root } from "react-dom/client";
+import { act } from "@testing-library/react";
+import { resource } from "../../core/resource";
+import { useResource } from "../../index";
+import { useReducer as useResourceReducer } from "../../react-hooks/useReducer";
+import { useMemo as useResourceMemo } from "../../react-hooks/useMemo";
+import { cleanupAllResources } from "../test-utils";
+
+let reactRoot: Root | undefined;
+let container: HTMLElement | undefined;
+
+afterEach(async () => {
+  await act(async () => reactRoot?.unmount());
+  container?.remove();
+  reactRoot = undefined;
+  container = undefined;
+  cleanupAllResources();
+});
+
+const useStream = () => {
+  const [chunks, push] = useResourceReducer(
+    (s: readonly string[], c: string) => [...s, c],
+    [] as readonly string[],
+  );
+  return useResourceMemo(() => ({ chunks, push }), [chunks]);
+};
+const Stream = resource(useStream);
+
+/**
+ * A flushSync commit against a pending transition makes React replay the
+ * reducer chain from a base below the committed version, which previously
+ * threw "Version is less than committed version" as a recoverable error on
+ * every replay. The assertions check delivery (every update present) rather
+ * than exact equality: the React-hosted bridge applies replayed updates onto
+ * already-committed cell state, so cross-lane replays duplicate committed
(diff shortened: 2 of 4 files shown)


okisdev:
measurement for the a/b ask, from the branch-vs-main matrix that produced this shape (same bridge test, `startTransition(() => push("t1")); flushSync(() => push("s1"))`, tap output read from the DOM):

- main (throw): `s1,t1,s1`, plus the recoverable "Version is less than committed version" error on the replay
- this branch: `s1,t1,s1`, zero recoverable errors
- #5321's clamp variant under the same test: also `s1,t1,s1` at deficit 1; the clamp's wipe only diverges at deficit >= 2, per the traces in the #5321 close-out

tap's committed output is byte-identical on main and this branch, so the containment assertions measure delivery only and the react-oracle divergence (`t1,s1`) is squarely #5330's. the vacuity concern is closed in 88105f620: a `vi.mock` pass-through probe counts `version < committedVersion` entries and the test asserts the count is positive, and the changelog assertions now observe a real record. the blast-radius argument is now in the body.

PR #5371, opened 2026-07-29
fix(core): preserve messages sent after cancellation

Kinfe123:
## Summary

- reconcile deferred cancellation resyncs with messages delivered afterward
- preserve adapter-owned optimistic messages while excluding only the placeholder generated by core
- retain cancellation deletions and branch restoration for queued, failed, or overlapping sends

## Problem

`cancelRun()` captures the current message snapshot and writes it through `setMessages()` on the next task so branch state can resync after cancellation. If another message arrives before that task runs, replaying the captured snapshot can overwrite the newer message. Simply skipping the replay also loses cancellation changes, including removal of a cancelled trailing user message.

## Fix

The deferred resync now records the messages known before cancellation and the exact IDs cancellation removes. If no new message arrives, it restores the captured branch as before. If a new message arrives, it writes the latest repository state after removing only the cancelled IDs and the exact optimistic placeholder generated by core. Adapter-owned optimistic messages remain intact.

## Verification

- reproduced the original overwrite on fresh `main`
- added regressions for successful, queued, failed, and adapter-optimistic sends
- added coverage for optimistic settlement and cancellation deletion during replacement sends
- external-store runtime tests: 81 passed
- changed-file lint and formatting: passed
- `@assistant-ui/core` build: passed

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5371?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.qGrHBdJQpEji99nXo6JMNwJulkoTDWy0xUF8LkpSEYpvCTV_qevfGGgWb5A?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/external-store-cancel-resync.md
+++ b/.changeset/external-store-cancel-resync.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/core": patch
+---
+
+fix: preserve messages appended immediately after cancelling a run
--- a/api-surface/assistant-ui__core.ts
+++ b/api-surface/assistant-ui__core.ts
@@ -1777,6 +1777,7 @@ declare class ExternalStoreThreadRuntimeCore extends BaseThreadRuntimeCore imple
   extras: unknown;
   private _converter;
   private _store;
+  private _cancelResyncRevision;
   private _toolInvocations;
   beginEdit(messageId: string): void;
   constructor(contextProvider: ModelContextProvider, store: ExternalStoreAdapter<any>);
--- a/api-surface/assistant-ui__core.ts
+++ b/api-surface/assistant-ui__core.ts
@@ -1777,7 +1777,6 @@ declare class ExternalStoreThreadRuntimeCore extends BaseThreadRuntimeCore imple
   extras: unknown;
   private _converter;
   private _store;
-  private _cancelResyncRevision;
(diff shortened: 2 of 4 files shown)


okisdev:
superseded by #5826, which just merged; closing. this was the earliest report of the stale-snapshot resync, and the cancelled-id accounting here pointed in the right direction, credit for that.

the branch predates the cancel path rework (#5798, #5799, #5800 and the trailing-leaf restore), so the diff no longer applies to the current shape. the merged fix reads the repository at flush time and re-applies the cancel's rollbacks there: an empty optimistic head is dropped again, a moved user leaf still sitting as the unanswered tail is re-deleted, and a turn the store kept stays in the thread while the untouched composer draft is taken back.

PR #5372, opened 2026-07-29
refactor(store): evaluate Derived scopes at React level, tap root only for build scopes

Yonom:
Restructures `useAui` so Derived-only mounts (per-message providers) no longer pay for a tap root's scheduler, per-scope fibers, and root re-renders on every store update.

## What changed

- **Derived scopes** now run at the React level: `useAui` calls `useDerived` directly (internals unchanged, including its per-scope `useSyncExternalStore`) inside a `useBuildingClientProvider` wrapper on the host fiber, instead of mounting each Derived scope as a fiber inside a tap root. Per-scope hooks are fine because scope maps are static per call site; entries are iterated in a stable order.
- **Build scopes** keep the single `useTapRoot`, still always called (the hook sequence does not depend on scope count/kind), but it now hosts only the non-Derived scopes. In a Derived-only mount nothing subscribes to or dirties this root, so it renders once at mount and stays inert; its output is the shallow-stable build-accessor list, read through one `useSyncExternalStore`.
- The client envelope (accessors + committed identity) is composed at the React level from both groups in original scope order via `useShallowStable`/`useCommittedClient` (rebased on top of #5367's fixed hook count and #5368's `useShallowStable`/mutable-cell machinery, both preserved).
- Semantics preserved: same envelope identity behavior (value-only updates keep `aui` identity, structural swaps produce a new one), same notification routing, same missing-scope error accessors, and a Derived `get()` that throws on re-selection still keeps the committed value while a throw at mount still propagates.

## Benchmark

500 Derived-only `useAui` mounts (MessageByIndexProvider-shaped, 2 Derived scopes each) under one thread provider; jsdom, `performance.now()` around `act()`, median of 5 iterations, two runs shown as a range. Baseline is current main (ecd7c87, includes #5367/#5368). Benchmark file is a temporary "repro" test on this branch (`packages/store/src/__tests__/useAui.derived-split.repro.bench.test.tsx`), to be deleted at merge.

| Scenario | main | this PR | delta |
| --- | ---: | ---: | ---: |
| Mount 500-message thread | 23.7–25.6 ms | 20.2–21.0 ms | ~1.2x faster |
| Delete first message (bulk identity shift) | 92.3–106.2 ms | 42.5–46.5 ms | ~2.2x faster |
| Unmount + remount (thread switch) | 30.2–31.0 ms | 16.2–17.2 ms | ~1.8x faster |

## Verification

- `@assistant-ui/store`: all pre-existing tests pass unmodified (84 incl. the repro bench)
- `@assistant-ui/core`: 767/767 tests pass unmodified
- `pnpm -F @assistant-ui/store build`, `pnpm -F @assistant-ui/core build` pass
- `node scripts/generate-api-surface.mjs`: no api-surface changes
- `pnpm check:resource-memo`: clean

Refs: sw-120

### DIFF
--- a/.changeset/derived-scopes-react-level.md
+++ b/.changeset/derived-scopes-react-level.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/store": patch
+---
+
+refactor: evaluate Derived scopes at React level; useAui's tap root hosts only build scopes, so Derived-only mounts no longer re-render a tap root per update
--- a/packages/store/src/__tests__/useAui.derived-split.repro.bench.test.tsx
+++ b/packages/store/src/__tests__/useAui.derived-split.repro.bench.test.tsx
@@ -1,0 +1,130 @@
+// @vitest-environment jsdom
+// repro: temporary benchmark for sw-120 (Derived scopes at React level).
+// Logs wall-clock medians; no timing assertions.
+
+import type { FC, ReactNode } from "react";
+import { useState } from "react";
+import { act, cleanup, render } from "@testing-library/react";
+import { afterEach, describe, expect, it } from "vitest";
+import { resource, withKey } from "@assistant-ui/tap";
+import { AuiProvider } from "../utils/react-assistant-context";
+import { useAui } from "../useAui";
+import { useAuiState } from "../useAuiState";
+import { useClientLookup } from "../useClientLookup";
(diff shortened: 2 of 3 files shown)


Yonom:
Superseded by #5373 (shape 3).

PR #5380, opened 2026-07-30
refactor(store): inline useAui helper hooks

Yonom:
Follow-up simplifications to `useAui` spotted during #5373, landed post-merge with sign-off:

1. Inline `createAccessor` — it was a trivial pass-through over `createClientAccessor`; both call sites now call it directly.
2. Merge `useAssistantClient` into `useHostedAssistantClient` — the tap-hosted body is now the `useTapHost` closure instead of a separate hook. One targeted `exhaustive-deps` disable where the linter misreads a prop captured by the host closure.

Not landed: folding `useScopeValue` into `useScopeMount` — the building-client mutation there makes the React Compiler bail, leaving the resource element unmemoized (caught by `check:resource-memo`). The helper now carries a comment saying it is load-bearing.

Internal restructure only: all hooks touched are module-local, no public exports change.

## Verification

- store: 82/82, core: 767/767, react: 372/372 — all unmodified
- `tsc --noEmit` clean, oxlint warning set identical to main's baseline
- api-surface regenerated: no changes

<a href="https://rupic.dev/s/s2.LTeB_GNsVhNGC87WzpkxlQyzmysI87VwoyFDuyt2rTKVvSQbiDMBtkUtYgc?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/store-inline-useaui-helpers.md
+++ b/.changeset/store-inline-useaui-helpers.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/store": patch
+---
+
+refactor: inline single-call-site useAui helper hooks
--- a/packages/store/src/useAui.ts
+++ b/packages/store/src/useAui.ts
@@ -96,13 +96,6 @@ const metaOf = (element: ScopeElement): ScopeMeta => {
   return { source: props.source, query: props.query ?? {} };
 };
 
-const createAccessor = <K extends ClientNames>(
-  name: K,
-  meta: ScopeMeta,
-  read: () => ClientMethods,
-): AssistantClientAccessor<K> =>
-  createClientAccessor<K>({ name, ...meta }, read);
-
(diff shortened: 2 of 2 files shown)


Yonom:
Bot triage: added the missing changeset (patch `@assistant-ui/store`). On Copilot's exhaustive-deps finding: the suppression is not redundant — removing it yields `warning react-hooks(exhaustive-deps): unnecessary dependency: parent ... Outer scope values aren't valid dependencies` because the linter misreads the capitalized host closure; `parent` is a reactive prop of the outer hook and must stay in the array, so the disable stays. The useScopeValue-inline finding was already addressed in 854eedf (fold reverted, comment added).

PR #5388, opened 2026-07-30
docs(store): fix drift against current store source and document s.optional

Yonom:
Drift audit of `apps/docs/content/tap-docs/store/` against the current `packages/store` source.

## Fixes

**api-reference**
- `attachTransformScopes`: signature corrected — first param is the hook function passed to `resource()`, transform mutates `scopes` and returns `void` (was shown returning `ScopesConfig`)
- `AssistantClientAccessor`: type now shows methods exposed as properties with the deprecated call signature, matching `types/client.ts`
- `AssistantState`: adds the `optional` view (`s.optional.<scope>` resolves to `undefined` instead of throwing)
- `useClientList.Props.resource`: `Resource<TMethods, [ResourceProps<TData>]>` (was the removed `ContravariantResource`)
- `useAui()`: removed the stale "does not re-render when scopes change" claim; documented the immutable-client/structural-change behavior and the rooted vs derived-only mode frozen at mount
- Added missing exports: `forwardTransformScopes`, `getClientId`, `ScopesConfig`, `DerivedElement`, `ClientMethods`, `ClientSchema`, `ClientEvents`

**state**
- "Checking if a scope exists" now leads with `useAuiState((s) => s.optional.counter?.count)` (shipped in store 0.3.1, previously undocumented)

**sibling-scopes**
- `attachTransformScopes` examples attach to the hook function, not the resource wrapper (transforms are keyed by `element.hook`; wrapper attachment silently never runs)
- Sibling access uses property style (`clientRef.current.modelContext.register(...)`) instead of the deprecated call style
- Callout mentions `forwardTransformScopes` for composing transforms

**child-scopes**
- Registry sample imports `ClientOutput` so it compiles

## Verification
- All 11 pages compile and serve 200 on the docs dev server with no error overlay
- All updated code samples typecheck against `packages/store/src` + `packages/tap/src` (strict tsc with a scratch `ScopeRegistry` augmentation)
- `meta.json` references only existing pages (unchanged)

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5388?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.xiQX4_9kDzWwOGGqVCEqB51YDykK_veOez_7gJfDCuUemQLedBmIaPkjnT0?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/apps/docs/content/tap-docs/store/api-reference.mdx
+++ b/apps/docs/content/tap-docs/store/api-reference.mdx
@@ -386,7 +386,7 @@ Base constraint for a scope's methods.
 ```ts
 type ClientSchema<
   TMethods extends ClientMethods = ClientMethods,
-  TMeta extends ClientMetaType = never,
+  TMeta extends { source: ClientNames; query: Record<string, unknown> } = never,
   TEvents extends Record<string, unknown> = never,
 > = {
   methods: TMethods;
--- a/apps/docs/content/tap-docs/store/api-reference.mdx
+++ b/apps/docs/content/tap-docs/store/api-reference.mdx
@@ -11,7 +11,7 @@ description: All exports from @assistant-ui/store.
 function useAui(): AssistantClient;
 ```
 
-Returns the store from the nearest `AuiProvider`. Does not re-render when scopes change — returns a stable reference.
+Returns the store from the nearest `AuiProvider`. The client is immutable: state updates keep its identity, while a structural change (a scope resolving to a different instance) produces a new client and re-renders consumers.
 
 ### useAui(scopes)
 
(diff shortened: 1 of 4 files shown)


apps/docs/content/tap-docs/store/api-reference.mdx
```diff
@@ -21,6 +21,8 @@ function useAui(scopes: useAui.Props): AssistantClient;
 
 Takes the store from the nearest `AuiProvider` and extends it by filling the provided scopes. Returns a new `AssistantClient` that includes both the parent's scopes and the newly provided ones.
 
+The mode is fixed at mount: if any scope is a root (non-`Derived`) element, the scopes run in a tap host; an all-`Derived` call site resolves each scope as a plain hook call, so its scope count is fixed. Changing a scope's kind requires a remount (new key).
```
Yonom: drop this

Yonom:
Addressed review: dropped the two-mode paragraph from useAui (Yonom), ClientSchema now shows the real defaulted/constrained signature (greptile/cubic), ClientEvents shows the real conditional validation (greptile/cubic), DerivedElement inlines the non-exported DerivedInstance as ReturnType<AssistantClientAccessor<K>> (claude/cubic).

Issue #5130, opened 2026-07-24
Divergent argsText close without a backend result still auto-submits a parse-error tool result (#5102 directions 2 and 3)


okisdev:
## context

assistant-ui/assistant-ui#5102 tracked the tracker-layer conversion of argsText divergence into auto-submitted error tool results and listed three directions. direction 1 (emit the backend result before closing the args stream, plus a backend-result suppression in ToolExecutionStream) ships in [#5107](<https://github.com/assistant-ui/assistant-ui/issues/5107>), which closes that issue. this issue carries the remaining scope so it does not get lost in the auto-close.

## residual mechanism

when no backend result exists at close time (frontend-executed tools, or HITL and interrupt flows where the result arrives later or never; the production shape from assistant-ui/assistant-ui#5098), the tracker still closes the args stream while it holds a stale divergent prefix. ToolExecutionStream parses the stale prefix, throws `Function parameter parsing failed`, wraps it into `ToolResponse({ isError: true })`, and hosts auto-submit it through onAddToolResult. on useLangGraphRuntime this resumes the graph and can abandon a pending interrupt (Anthropic 400, tool_use without tool_result). [#5107](<https://github.com/assistant-ui/assistant-ui/issues/5107>) does not change this path because setResponse is never called in it.

## directions (carried over from assistant-ui/assistant-ui#5102)

1. divergent-complete snapshot adoption: when a close follows a known-divergent snapshot whose argsText is complete valid JSON, let the tracker hand that snapshot text to the execution path instead of the stale stream prefix. needs a small assistant-stream API addition; the argsText controller is append-only (no replace), so divergent text cannot be reconciled by appending.
2. stop auto-submitting an error result that was produced by internal bookkeeping rather than by the tool or the model. hosts cannot distinguish the parse-error ToolResponse from a genuine tool failure today; a marker on the response (or suppressing the auto-submit for this class) would let interrupt flows survive the divergence instead of resuming with a bogus error.

## also carried

values-mode residual from the same family: LangGraphMessageAccumulator.reconcileMessages and replaceMessages bypass appendMessage, so a values-mode final reconcile can reintroduce a compact re-serialization (and therefore divergence) under extreme render batching. mechanism verified, not yet reproduced end to end.

## prior art

[...]

Issue #5098, opened 2026-07-22
Python LangGraph + Anthropic: final `updates` message replaces streamed tool args with a re-stringified copy — argsText freezes mid-prefix and throws "Function parameter parsing failed" (same root cause as #5074/#3779)


rudokotrla:
### Checklist

- [x] I have searched existing issues and this is not a duplicate (#5074 / #3779 / #3458 are the same root-cause family via different triggers — see below)
- [x] This is a reproducible bug, not a usage question

### Environment info

```
Environment:
  OS:               macOS 15.x (arm64)
  Node.js:          v22.22.2
  Package Manager:  bun 1.3.13
  Framework:        Vite + React 18.3.1

Packages:
  @assistant-ui/react            0.14.27
  @assistant-ui/react-langgraph  0.14.12
  @assistant-ui/core             0.2.21 (transitive)
  assistant-stream               0.3.26 (transitive)

Backend: Python LangGraph + langchain-anthropic (Claude), custom SSE endpoint
forwarding `astream(stream_mode=["messages", "updates"], subgraphs=True)` items
verbatim (the shape `useLangGraphRuntime`'s message accumulator consumes).
```

### Affected area(s)

Provider Adapters, Runtime / Core, Streaming (assistant-stream)

### Runtime adapter in use

useLangGraphRuntime

### Steps to reproduce

This is a third trigger for the root cause behind #3779 and #5074: whenever `resolveToolCallArgs` derives `argsText` from a **different source** than the streamed `partial_json`, the result is not a byte-prefix of what already streamed, and `ToolInvocationTracker` freezes the args stream (`EDGE_CASES.md` A.2). #3458 fixed the *key-order* dimension of that mismatch, #5080 fixes the *argless-first-chunk* dimension; this report is the **whitespace/formatting dimension**, triggered at stream **end** by the LangGraph `updates` event.

1. Python LangGraph backend with an Anthropic model; the agent emits a tool call. Anthropic's streamed `input_json_delta` fragments contain its own formatting, e.g. `{"question": "What…", "options": […], "allow_multiple": false, "allow_custom_answer": true}` (spaces after top-level colons/commas).
2. The `messages` stream-mode chunks accumulate normally: each `AIMessageChunk`'s `tool_call_chunks[].args` is concatenated into `partial_json` by `appendLangChainChunk`, so mid-stream snapshots hand `ToolInvocationTracker` a growing byte-prefix. So far so good.
[...]

PR #5395, opened 2026-07-30
fix: let the tracker rebase a divergent argsText instead of carrying stale partial_json

### LINKED ISSUE #5391: tool args: let the tracker rebase a divergent argsText instead of carrying stale partial_json
supersedes the original framing of this issue. the mechanism below is verified; the fix direction first proposed here is refuted and must not be revived.

## the mechanism

`assistant-message-accumulator.ts` builds argsText by pure append (`part.argsText + chunk.textDelta`), and the chunk protocol has only `text-delta` and `tool-call-args-text-finish`. there is no way to replace text already streamed into a tool call's args controller.

so when `ToolInvocationTracker` meets a snapshot whose argsText is not an extension of what it already streamed, its A.2 branch can only keep the stale prefix. `isEquivalentCompleteArgsText` rescues the case where both texts are complete and parse equal, but not the case where the tracked prefix is still incomplete.

#5099's `mergeStreamedToolCallArgs` is the workaround for that: rather than let the converter emit a re-serialized text the tracker cannot adopt, carry the already-streamed `partial_json` forward onto the incoming message. #5390 extended the same carry to the values path for consistency.

the cost of the workaround is that a carried text wins over the server's own parsed args. `resolveToolCallArgs` (`convertLangChainMessages.ts:71-81`) gives `partial_json` top precedence and derives `args` from that same text, so `chunk.args` is discarded whenever a carry fires.

## why the guard first proposed here is refuted

the original proposal was to carry only when `parsePartialJsonObject(partial_json)` deep-equals the incoming `args`. that regresses #5099. the three cases:

| case | tracked prefix | with carry | with the proposed guard |
|---|---|---|---|
| A: #5099's target, stream still incomplete, server sends the same args complete | incomplete | prefix preserved, stream continues normally | `parse(truncated) != complete args`, so the guard declines the carry and #5098 comes back |
| B: a node rewrites the args in place after the stream completed | complete | part.args stale but consistent with what already executed | part.args new while the tracker still holds (and already executed) the old text, so strictly less coherent |
| C: stream truncated, server sends complete args | incomplete | part.args is a partial parse | part.args complete, execution unchanged |

A and C are indistinguishable at carry time (both are "incomplete prefix plus complete server args"; they differ only in whether more deltas arrive later). a guard that saves C necessarily breaks A, and A is the reason the carry exists.

also worth recording: B is unfixable at this layer whatever we do. the stream completed, so the tool already executed before the rewrite existed.

## the actual fix

give the args controller a replace capability and let the tracker use it, then delete the workaround:

1. **assistant-stream**: a replace chunk (path-addressed, alongside `text-delta`), `TextStreamController.replace(text)`, and the accumulator branch that sets rather than appends. deliberately client-internal: it is produced by the tracker to correct its own controller, never by a backend, so it stays out of `AssistantTransport`'s `KNOWN_CHUNK_TYPES` and the Python mirror needs no change.
2. **core**: the A.2 branch adopts a complete divergent snapshot via replace and closes, instead of keeping the stale prefix. this is also what #5130 direction 1 asked for.
3. **react-langgraph**: delete `mergeStreamedToolCallArgs` and the carry. the tracker now handles divergence natively, so the converter no longer has to emit stale text to accommodate it.

after step 2, A and C both resolve by adopting the authoritative snapshot and executing with correct args, and the hazard class disappears with the carry rather than being guarded.

## scope

one PR. the three parts are one concern (make the tracker able to rebase, then remove the workaround for its inability), nothing else consumes the new primitive so the revert unit is the whole change, and the intermediate states are never exercised.

the one part that needs visible scrutiny is deleting a fix that shipped for a production report. #5099's regression tests move to the tracker level in the same PR rather than disappearing, with a red/green counterfactual: revert the tracker half and the migrated #5098 test must fail.

split into two PRs only if the replace chunk turns out to need wire representation after all (for example if a tracker-produced stream is re-encoded anywhere), in which case the protocol and the Python mirror go first.

<a href="https://rupic.dev/s/s2.Moq9Iqvk6uOzcnu04iBAz9Eh3zEoTO589LWpdovYPf9z7TWvr7dnA_0JbPE?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

okisdev:
closes #5391, and resolves direction 1 of #5130

## summary

`ToolInvocationTracker` could not adopt a snapshot whose `argsText` was not an extension of what it had already streamed: the args controller is append-only, so the A.2 branch could only keep the stale prefix. `mergeStreamedToolCallArgs` in react-langgraph existed to work around that, carrying already-streamed `partial_json` forward so the converter never emitted a text the tracker could not take. that workaround costs correctness, because `resolveToolCallArgs` gives `partial_json` top precedence and derives `args` from it, so a carried text discards the server's own parsed args.

this replaces the workaround with the missing capability:

- **assistant-stream**: a `text-replace` chunk and `TextStreamController.replace(text)`, with accumulator, `ToolCallReader`, and `ToolExecutionStream` support. deliberately client-internal: it is produced locally to correct a controller, never by a backend, so it is excluded from `AssistantTransport`'s `KNOWN_CHUNK_TYPES` via `Exclude<AssistantStreamChunk["type"], "text-replace">` and a frame carrying it is rejected as `unknown-type`. the Python mirror is untouched
- **core**: a divergent snapshot whose `argsText` is complete is authoritative, so the tracker replaces the controller's text and closes under the normal rules. an incomplete divergent snapshot is not authoritative-final and keeps today's behavior (retain the prefix, warn). `EDGE_CASES.md` A.2 updated
- **react-langgraph**: `mergeStreamedToolCallArgs` and `findMatchingToolCall` deleted

replace propagates through the private args event stream, so the accumulated text, `reader.args`, its handles, and the terminal value handed to `handle.end` all converge on the replacement. only intermediate values a consumer already read remain stale, since a consumed stream cannot be un-written.

## why one PR

the parts are strictly sequential (the deletion needs the tracker change, which needs the primitive) and the intermediate states are never exercised. more importantly `TextStreamController` is exported and tracked in `api-surface/assistant-stream.ts`, and this repo's public surface is append-only: shipping the primitive in its own PR would risk permanently adding an unremovable method whose only consumer never lands.

### already verified

- [x] `assistant-stream` 364 passed / 20 skipped, `core` 768 passed, `react-langgraph` 210 passed; repo lint clean
- [x] #5099's four converter tests pinned a mechanism this PR deletes, so the scenario they protected (#5098: an updates event delivers a complete re-serialized args text over an incomplete streamed prefix) is re-pinned at the tracker level as `executes with a complete divergent argsText snapshot`. red/green counterfactual: with only the tracker change reverted, that test fails on zero `execute` calls
- [x] the two values-path tests added in #5390 were rewritten as `appendMessage` spy contracts. their previous form asserted a carry that no longer exists, and the naive replacement (`expect(result[0]).toBe(full)`) passes with or without the routing, so they now assert the hook is invoked once with the pre-replacement entry, which is red without #5390's fix
- [x] `tsc --noEmit` on `packages/core` reports 5 errors in unrelated test files, confirmed pre-existing by re-running against the HEAD versions of both changed files

### reviewer should verify

- [ ] against a live LangGraph backend that streams tool args and then emits an `updates` event, the tool executes with the server's args and no `Function parameter parsing failed` result appears

## notes

- the rewrite case (a node rewriting tool args in place after the stream completed) stays unfixed and is unfixable at this layer: the stream closed and the tool already executed before the rewrite existed
- the guard originally proposed on #5391 (carry only when the parsed text deep-equals the incoming args) is refuted in that issue: it declines exactly the incomplete-prefix case #5099 needs, so it would regress #5098

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5395?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.igRAJuMbkaY_3-rGaRbt5N51v7xBZF6xrrjCQq0gzIEX8IvtqvPtjODQIIQ?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/assistant-stream-text-replace.md
+++ b/.changeset/assistant-stream-text-replace.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/assistant-stream": patch
+---
+
+feat: add a client-internal text-replace chunk and TextStreamController.replace
--- a/.changeset/assistant-stream-text-replace.md
+++ b/.changeset/assistant-stream-text-replace.md
@@ -1,5 +1,5 @@
 ---
-"@assistant-ui/assistant-stream": patch
+"assistant-stream": patch
 ---
 
 feat: add a client-internal text-replace chunk and TextStreamController.replace
--- a/.changeset/react-langgraph-drop-partial-json-carry.md
+++ b/.changeset/react-langgraph-drop-partial-json-carry.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react-langgraph": patch
+---
+
+fix: drop the partial_json carry now that the tracker rebases a divergent argsText
--- a/.changeset/tracker-adopt-divergent-snapshot.md
+++ b/.changeset/tracker-adopt-divergent-snapshot.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/core": patch
+---
+
+fix: adopt a divergent complete argsText snapshot instead of keeping the stale prefix
--- a/api-surface/assistant-cloud.ts
+++ b/api-surface/assistant-cloud.ts
@@ -244,6 +244,9 @@ type AssistantStreamChunk = {
   readonly type: "text-delta";
   readonly textDelta: string;
 } | {
+  readonly type: "text-replace";
+  readonly text: string;
+} | {
   readonly type: "annotations";
   readonly annotations: ReadonlyJSONValue[];
 } | {
--- a/api-surface/assistant-stream.ts
+++ b/api-surface/assistant-stream.ts
@@ -117,7 +117,7 @@ type AssistantMessageTiming = {
 };
 
 type AssistantMetaStreamChunk = (AssistantStreamChunk & {
-  type: "part-finish" | "text-delta";
+  type: "part-finish" | "text-delta" | "text-replace";
   meta: PartInit;
 }) | (AssistantStreamChunk & {
   type: "result" | "tool-call-args-text-finish";
@@ -125,7 +125,7 @@ type AssistantMetaStreamChunk = (AssistantStreamChunk & {
     type: "tool-call";
   };
 }) | (AssistantStreamChunk & {
-  type: Exclude<AssistantStreamChunk["type"], "part-finish" | "result" | "text-delta" | "tool-call-args-text-finish">;
+  type: Exclude<AssistantStreamChunk["type"], "part-finish" | "result" | "text-delta" | "text-replace" | "tool-call-args-text-finish">;
 });
 
 declare class AssistantMetaTransformStream extends TransformStream<AssistantStreamChunk, AssistantMetaStreamChunk> {
@@ -153,6 +153,9 @@ type AssistantStreamChunk = {
 } | {
   readonly type: "text-delta";
   readonly textDelta: string;
+} | {
+  readonly type: "text-replace";
+  readonly text: string;
 } | {
   readonly type: "annotations";
   readonly annotations: ReadonlyJSONValue[];
@@ -11594,6 +11597,7 @@ type TextStatus = {
 
 type TextStreamController = {
   append(textDelta: string): void;
+  replace(text: string): void;
   close(): void;
 };
 
--- a/api-surface/assistant-ui__cloud-ai-sdk.ts
+++ b/api-surface/assistant-ui__cloud-ai-sdk.ts
@@ -248,6 +248,9 @@ type AssistantStreamChunk = {
   readonly type: "text-delta";
   readonly textDelta: string;
 } | {
+  readonly type: "text-replace";
+  readonly text: string;
+} | {
   readonly type: "annotations";
   readonly annotations: ReadonlyJSONValue[];
 } | {
--- a/api-surface/assistant-ui__core.ts
+++ b/api-surface/assistant-ui__core.ts
@@ -442,6 +442,9 @@ type AssistantStreamChunk = {
   readonly type: "text-delta";
   readonly textDelta: string;
 } | {
+  readonly type: "text-replace";
+  readonly text: string;
+} | {
   readonly type: "annotations";
   readonly annotations: ReadonlyJSONValue[];
 } | {
--- a/api-surface/assistant-ui__react-ai-sdk.ts
+++ b/api-surface/assistant-ui__react-ai-sdk.ts
@@ -323,6 +323,9 @@ type AssistantStreamChunk = {
   readonly type: "text-delta";
   readonly textDelta: string;
 } | {
+  readonly type: "text-replace";
+  readonly text: string;
+} | {
   readonly type: "annotations";
   readonly annotations: ReadonlyJSONValue[];
 } | {
--- a/api-surface/assistant-ui__react-data-stream.ts
+++ b/api-surface/assistant-ui__react-data-stream.ts
@@ -262,6 +262,9 @@ type AssistantStreamChunk = {
   readonly type: "text-delta";
   readonly textDelta: string;
 } | {
+  readonly type: "text-replace";
+  readonly text: string;
+} | {
   readonly type: "annotations";
   readonly annotations: ReadonlyJSONValue[];
 } | {
--- a/api-surface/assistant-ui__react-google-adk.ts
+++ b/api-surface/assistant-ui__react-google-adk.ts
@@ -638,6 +638,9 @@ type AssistantStreamChunk = {
   readonly type: "text-delta";
   readonly textDelta: string;
 } | {
+  readonly type: "text-replace";
+  readonly text: string;
+} | {
   readonly type: "annotations";
   readonly annotations: ReadonlyJSONValue[];
 } | {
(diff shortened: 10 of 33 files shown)


okisdev:
converted to draft: this reintroduces the failure mode #5132 was written to prevent, and the branch hides that by inverting the test that guarded it.

on main, `does not auto-submit a parse-error result when divergent argsText closes without a backend result` asserts `expect(onResult).not.toHaveBeenCalled()` and `expect(execute).not.toHaveBeenCalled()` for a human-in-the-loop tool whose args diverge mid-stream. on this branch the same test, under the same name, now asserts the opposite: `execute` ran and a result was auto-submitted, while the `human({request:"approve"})` interrupt raised at the top of the test is still pending.

that is the #5098 damage exactly. the auto-submitted result resumes the graph and abandons the pending interrupt; the only thing that changed is that the submitted result is legitimate rather than a parse error, and the interrupt abandonment was the actual production damage, not the error payload.

so adopting a complete divergent snapshot cannot also mean closing it. correcting the text and deciding to execute have to be separated, and the close decision still has to respect a pending interrupt. that needs design work rather than another patch on this branch.

two smaller confirmed defects on the way to that:

- `.changeset/assistant-stream-text-replace.md` names `@assistant-ui/assistant-stream`; the package is `assistant-stream`, so the changeset targets a package that is not in the workspace
- `replaceArgsText` updates the authoritative args text but never reaches `ToolCallArgsReaderImpl`, which accumulates exclusively from the private delta stream. a `streamCall` tool therefore keeps the stale value including the terminal one handed to `handle.end`. the stated tradeoff was losing preview fidelity, but the terminal value is not preview

okisdev:
both substantive findings in this round are confirmed at source, and together they say something about the mechanism rather than about three missing patches.

`AssistantTransportEncoder`'s transform is an unconditional `controller.enqueue(\`data: ${JSON.stringify(chunk)}\n\n\`)`, so it serializes every chunk including this one. the JSDoc I added claiming the wire transports do not carry it was verified only against the decoder's `KNOWN_CHUNK_TYPES` and is false for the encoder.

`StreamTextHandle.update` derives its delta as `value.substring(this.lastValue?.length || 0)`. that is an append-only assumption, so after a replace it emits an empty string when the replacement is shorter and a mid-string fragment when it is not. `streamText()` consumers get corrupted output, and the `GetHandle` case raised separately is the same shape: a handle that already resolved keeps the pre-replacement value.

the pattern across both review rounds is that each pass finds another consumer that assumes text only grows: the accumulator, `ToolCallArgsReaderImpl`, `StreamTextHandle`, `GetHandle`, the two encoders. adding a member to the shared public chunk union does not announce the broken invariant to any of them, it just makes them silently wrong, and `TextStreamController` is append-only public surface so whatever ships here is permanent.

so this is not ready. the goal stands (the tracker genuinely cannot rebase today, which is what forces the langgraph carry), but "introduce a replace chunk into the shared streaming protocol" propagates an invariant violation across a stack that has append-only baked into at least four independent consumers.

worth evaluating before continuing: confine the correction to the tool-call args path without touching the shared union, for example by having the tracker abandon a diverged part and re-create it so every controller's text still only grows. that trades part identity churn against the invariant, and the "exactly once" contract in EDGE_CASES is where that tradeoff has to be settled.

okisdev:
closing in favour of a smaller fix. #5394 targets the same defect (the unguarded `partial_json` carry masking the server's args) with a guard inside `mergeStreamedToolCallArgs`, and does not need a new chunk type.

what this branch got wrong is the mechanism, not the goal. introducing `text-replace` into the shared `AssistantStreamChunk` union breaks an invariant the whole streaming stack is built on, that text only ever grows, and it does so silently: `StreamTextHandle` derives its delta as `value.substring(lastValue.length)`, a `GetHandle` that already resolved keeps its pre-replacement value, and `AssistantTransportEncoder` serializes every chunk unconditionally so an internal chunk leaks onto the wire. two review rounds each surfaced another consumer with the same assumption, and `TextStreamController` is append-only public surface, so whatever shipped here would be permanent.

the underlying gap is still real and still unowned: the tracker cannot rebase a divergent argsText, which is the reason the langgraph carry exists at all, and it is the same missing capability as #5130 direction 1. if it gets picked up again it should start from a mechanism that keeps every controller's text monotonic (for example abandoning and re-creating a diverged part) rather than teaching the shared protocol to rewind, and the "exactly once" contract in EDGE_CASES is where that tradeoff has to be settled.

PR #5411, opened 2026-07-30
fix(store): preserve notification state in StrictMode

### LINKED ISSUE #5422: StrictMode: Composer drops input and cannot submit in with-external-store
## Summary

The official `examples/with-external-store` Composer drops typed text and cannot submit when rendered under React StrictMode. This reproduces on the current `@assistant-ui/react` 0.15 / `@assistant-ui/store` 0.3 line with React 19.2.8.

## Cloneable reproduction

The reproduction is the existing official example in this repository; no custom runtime or UI is required.

```bash
cd assistant-ui
git checkout 052ce90d8
pnpm exec turbo build --filter=with-external-store... --force
pnpm --filter with-external-store exec next dev
```

Open the printed URL, type any text into **Message input**, and press Enter.

Five fresh Chromium pages on that build all produced the same result: the textarea reset to `""`; neither the user message nor the example's `Hello, world!` response appeared. Removing StrictMode makes the path work, but that is only a workaround.

## Expected behavior

The Composer retains the entered text and Enter submits it through the external-store runtime.

## Root cause

`@assistant-ui/store` is built with React Compiler. The source `useNotificationManager` used `useMemo`, which React Compiler replaced with a `c()` memo cache in the published build.

The hook runs inside a Tap host. During the initial StrictMode double render, Tap recreates the uncommitted compiler memo cache, so the two renders receive different stateful `NotificationManager` instances even though the surrounding `clientRef` and Tap fibers survive. The committed client can retain notification methods from the first manager while live host effects notify the second. Composer state changes, but the UI subscriber attached to the first manager is never notified.

A deterministic compiled-artifact regression observes two manager identities with the original implementation.

## Proposed fix

`NotificationManager` owns mutable listeners and subscribers, so its lifetime must be hook state rather than a disposable memo cache. PR #5411 changes it to lazy `useState` and adds a compiler-specific StrictMode regression test.

After the fix:

- the compiled regression observes exactly one manager identity;
- the freshly rebuilt official example retains input, submits, and renders `Hello, world!` in 5/5 Chromium runs;
- the earlier downstream `fields` dependency workaround is not used.

<a href="https://rupic.dev/s/s2.f-u4TRKljTjS4vCpn1OKAv7TozJPU4hRCIIvC5NhdIQHx5S6v5KZvqlHm_Y?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

nyl199310:
Fixes #5422.

## What

Hold `NotificationManager` in lazy `useState` rather than `useMemo`, so its listener registry remains the same state owner across React StrictMode render replays.

This supersedes the PR's earlier `fields` dependency workaround. The committed-client dependency list is no longer changed.

## Root cause

The source-level lifecycle is not the published lifecycle:

1. `@assistant-ui/store` is built with React Compiler.
2. React Compiler replaces `useNotificationManager`'s `useMemo` with a `c()` compiler memo cache.
3. The hook runs inside a Tap host. During the initial StrictMode double render, Tap starts the second uncommitted render from an empty current compiler cache, so the two renders receive different `NotificationManager` instances.
4. The `clientRef`, Tap fibers, and accessors survive that replay. The committed client can therefore retain notification methods from the first manager while the live host effects notify the second manager.
5. Composer state changes, but the UI subscriber attached to the first manager is never notified, so the controlled input resets to the stale empty value.

Instrumentation of the official example showed the same `clientRef` with notification manager identity changing from `1` to `13` between the StrictMode renders. This is the lifecycle detail missed by the prior source-only analysis.

`NotificationManager` owns mutable listener/subscriber state, so its lifetime must be represented as hook state. A memo cache is intentionally disposable and is not an appropriate state owner. Lazy `useState` expresses that invariant directly and compiles to a real hook rather than `c()` cache entries.

## Deterministic regression test

The new store compiler test imports the actual built `dist/utils/NotificationManager.js`, executes it inside `useTapHost` under `<StrictMode>`, and asserts that every replay observes the same manager.

The exact test was run against both implementations after a fresh store build:

- original `useMemo`: fails deterministically, `Set` size is `2`;
- fixed `useState`: passes, `Set` size is `1`.

The test is wired into the repository's existing `test:react-compiler` phase, alongside the core compiler test.

## End-to-end verification

The red/green browser comparison was established on upstream `main` at `052ce90d8` using the official `examples/with-external-store` app in fresh Chromium pages:

- before: 5/5 runs reset the textarea to `""`; no user message and no response;
- after: 5/5 runs retained the typed value, submitted the user message, and rendered `Hello, world!`;
- no console or page errors in either comparison.

After rebasing the fixed branch onto current upstream `main` (`1df2939a7`), the official example was force-built again and the passing Chromium probe was repeated 5/5 with zero console or page errors.

Additional checks on commit `09dbfb7bd`:

- forced `with-external-store...` build: 17/17 tasks passed;
- `@assistant-ui/store`: 16 files, 84 tests passed;
- compiled StrictMode regression: 1/1 passed;
- formatting, oxlint, and `git diff --check` passed.

### DIFF
--- a/.changeset/fresh-composers-listen.md
+++ b/.changeset/fresh-composers-listen.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/store": patch
+---
+
+fix: prevent Composer updates from being lost under React StrictMode
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "format:fix": "pnpm exec oxfmt",
     "sync-templates": "bash scripts/sync-templates.sh",
     "test": "turbo test && pnpm test:react-compiler && pnpm test:types",
-    "test:react-compiler": "turbo test:react-compiler --filter=@assistant-ui/core",
+    "test:react-compiler": "turbo test:react-compiler --filter=@assistant-ui/core --filter=@assistant-ui/store",
     "test:types": "pnpm -r --workspace-concurrency=4 --no-bail exec sh -c '! test -f tsconfig.json || tsc --noEmit'",
     "prepare": "husky && pnpm --filter @assistant-ui/react-devtools run build:css",
     "deps:check": "npx taze major -f -r",
--- a/packages/store/package.json
+++ b/packages/store/package.json
@@ -29,6 +29,7 @@
   "scripts": {
     "build": "aui-build",
     "test": "vitest run",
+    "test:react-compiler": "vitest run --config vitest.react-compiler.config.ts",
     "test:watch": "vitest"
   },
   "peerDependencies": {
--- a/packages/store/src/utils/NotificationManager.react-compiler.test.tsx
+++ b/packages/store/src/utils/NotificationManager.react-compiler.test.tsx
@@ -1,0 +1,37 @@
+// @vitest-environment jsdom
+
+import { StrictMode } from "react";
+import { cleanup, render } from "@testing-library/react";
+import { afterEach, describe, expect, it } from "vitest";
+import { useTapHost } from "@assistant-ui/tap";
+import {
+  useNotificationManager,
+  type NotificationManager,
+} from "./NotificationManager";
+
+afterEach(() => {
+  cleanup();
+});
+
+describe("compiled useNotificationManager", () => {
+  it("keeps one state owner across a StrictMode render replay", () => {
+    const instances: NotificationManager[] = [];
+
+    function Harness() {
+      useTapHost(function NotificationHost() {
+        instances.push(useNotificationManager());
+        return null;
+      });
+      return null;
+    }
+
+    render(
+      <StrictMode>
+        <Harness />
+      </StrictMode>,
+    );
+
+    expect(instances.length).toBeGreaterThanOrEqual(2);
+    expect(new Set(instances)).toHaveLength(1);
+  });
+});
--- a/packages/store/src/utils/NotificationManager.ts
+++ b/packages/store/src/utils/NotificationManager.ts
@@ -1,4 +1,4 @@
-import { useMemo } from "react";
+import { useState } from "react";
 import type { ClientStack } from "./tap-client-stack-context";
 import type {
   AssistantEventName,
@@ -26,7 +26,7 @@ export type NotificationManager = {
 };
 
 export const useNotificationManager = (): NotificationManager => {
-  return useMemo(() => {
+  return useState<NotificationManager>(() => {
     const listeners = new Map<string, Set<InternalCallback>>();
     const wildcardListeners = new Set<InternalCallback>();
     const subscribers = new Set<() => void>();
@@ -109,5 +109,5 @@ export const useNotificationManager = (): NotificationManager => {
         }
       },
     };
-  }, []);
+  })[0];
 };
--- a/packages/store/vitest.react-compiler.config.ts
+++ b/packages/store/vitest.react-compiler.config.ts
@@ -1,0 +1,22 @@
+import { fileURLToPath } from "node:url";
+import { defineConfig } from "vitest/config";
+
+const compiledNotificationManager = fileURLToPath(
+  new URL("./dist/utils/NotificationManager.js", import.meta.url),
+);
+
+export default defineConfig({
+  resolve: {
+    alias: [
+      {
+        find: /^\.\/NotificationManager$/,
+        replacement: compiledNotificationManager,
+      },
+    ],
+  },
+  test: {
+    environment: "jsdom",
+    globals: true,
+    include: ["src/utils/NotificationManager.react-compiler.test.tsx"],
+  },
+});


nyl199310:
I reworked the evidence in response to the blocking review:

- opened #5422 with a cloneable reproduction using the repository’s existing official `examples/with-external-store` app;
- verified the exact same Chromium interaction on base `6e5c450d7` (`value=""`, `submitted=false`) and with this one-line change (`value="strict mode repro"`, `submitted=true`, `Hello, world!` rendered);
- repeated the before/after comparison in an independent published-package Vite consumer (3/3 browser failures -> 3/3 passing);
- removed the original jsdom test because the review was correct: it also passed on `main` and was not valid regression evidence;
- rewrote the PR body without generated badges and clarified why the sibling derived-only call site is different.

The branch is now only the one-line store fix plus its patch changeset.

Kinfe123 (dismissed):
The `fields` appears stable within the same Tap fiber. so i guess this dependency may be inert. Could we add a deterministic test that fails on main and passes with this change?

nyl199310:
Thanks for blocking on deterministic evidence — the objection was useful. I rebuilt both sides identically and traced the missing lifecycle detail.

The source-level claim that `useNotificationManager` is a stable `useMemo([])` cell does not hold in the shipped build: React Compiler turns it into `c()` memo-cache entries. Inside Tap's initial StrictMode replay, those uncommitted compiler-cache entries are recreated, while the surrounding `clientRef` and Tap fiber survive. Instrumentation showed the same `clientRef` paired with two manager identities.

I replaced the earlier dependency workaround with the root fix: the stateful notification registry now uses lazy `useState`. I also added a compiled-artifact regression that fails on the original implementation with two manager instances and passes with one after the fix. The branch is rebased onto current main, and the official browser example passes 5/5 after failing 5/5 before.

@Kinfe123, could you please re-review the new commit when convenient?

nyl199310:
Updated the branch after the latest review:

- rebased onto current `main` (`1df2939a7`);
- removed the redundant `useEffect(effects)` from the compiled-artifact harness;
- rewrote the changeset as `fix: prevent Composer updates from being lost under React StrictMode`;
- reran `@assistant-ui/store` build, all 84 default store tests, the compiled React Compiler regression, oxfmt, oxlint, and `git diff --check`;
- force-built the official `with-external-store` dependency graph: 17/17 tasks passed;
- reran the official example in Chromium 5 times: input stayed visible before submit, user and assistant messages rendered after submit, and all 5 runs had zero console/page errors.

The identity-bearing `useClientResource` proxy remains out of scope because there is still no concrete failing behavior; this PR stays limited to the reproduced notification-manager defect.

@Kinfe123, the deterministic compiled-output regression and browser verification are now on commit `09dbfb7bd`. Could you please re-review when convenient?

okisdev:
verified this independently, including the deterministic fail-on-main test @Kinfe123 asked for. it now exists in the PR, and i confirmed it discriminates.

**mechanism, confirmed from the compiled artifact rather than the source.** `useMemo` does not survive the build at all. on `f434ebd0c` the compiled `dist/utils/NotificationManager.js` starts:

```js
import { c } from "@assistant-ui/tap/react-shim/compiler-runtime";
const useNotificationManager = () => { const $ = c(3); ... }
```

that `c` is tap's own shim, which routes to `useMemoCache` whenever `peekResourceFiber() !== null`. `nextFiberMemoCache` seeds `workInProgress` from `memoCache.current`, and `current` is still `null` before the first commit, so StrictMode's uncommitted replay starts from an empty cache and constructs a second manager. subscribers then sit on the first while live effects notify the second. with the fix the same file compiles to `useState(_temp)[0]` and `c()` is gone entirely.

**the discriminating test.** rendering the built `useNotificationManager` inside `useTapHost` under `<StrictMode>`:

| build | renders observed | distinct managers |
| --- | --- | --- |
| `f434ebd0c` (useMemo) | 2 | 2 |
| this PR (lazy useState) | 2 | 1 |

it has to import from `dist`, not `src`, because the compiler is the entire mechanism; a source-level test cannot observe this.

**end to end.** on `examples/with-external-store` under `next dev` (StrictMode on by default), swapping only `dist/utils/NotificationManager.js` and changing nothing else: before, typing left `textarea.value === ""` with Send disabled; after, the text is retained, Send enables, Enter renders both the user message and `Hello, world!`. clicking a suggestion (the pure external `setMessages` path) also renders, which is the mechanism behind #5502, so this fixes that issue too rather than only the composer surface.

**on whether this belongs in `useState` at all**, since that was the open design question: yes, and it matches what the repo already does. core holds lifetime-owned mutable state in lazy `useState` in about ten places (`useState(() => new ExternalStoreRuntimeCore(...))`, `LocalRuntimeCore`, `RemoteThreadListRuntimeCore`), and `external-message-converter.ts:412` carries a comment saying memoization is droppable for exactly this reason. `useNotificationManager` was the outlier. i scanned every `useMemo` in `packages/store/src` and `packages/core/src`: it is the only one owning mutable state, so the fix is complete for the current surface. worth noting the asymmetry that makes this subtle: tap's own `useMemo` is cell-based and safe; only React's `useMemo` reaching the compiler inside a tap fiber is affected, and this file imported it from `react`.

merged with current main, `packages/store` is green at 84 plus the compiler test. `turbo.json` already gives `test:react-compiler` `dependsOn: ["build"]`, so the dist-based test is not vacuous in CI.

the earlier `fields` dependency concern is moot: that approach was dropped and the committed-client dependency list is unchanged in the current head.

okisdev (approved):
approving on the evidence in the comment above: the compiled-artifact test discriminates (2 managers on main, 1 with this change), and swapping only the built `NotificationManager.js` into the external-store example restores the composer under StrictMode end to end, including the `setMessages` path behind #5502.

the approach is also the one the repo already uses for lifetime-owned mutable state, and a scan of `packages/store/src` and `packages/core/src` shows `useNotificationManager` was the only `useMemo` owning a mutable registry.

this still needs @Kinfe123 to re-review, since the changes-requested review predates the rewrite: the `fields` dependency it questioned is gone, and the deterministic fail-on-main test it asked for is now in the PR.

PR #5414, opened 2026-07-30
fix(react-ai-sdk): wire replacement chat transports

Kinfe123:
## Summary

- keep the default `AssistantChatTransport` stable across rerenders
- configure the concrete transport from the current render instead of routing setup through the stable proxy
- resolve resumable adapters from the current transport when options change

## Why

`useDynamicChatTransport` intentionally keeps one proxy so AI SDK chat state survives transport option updates. Runtime setup was using that proxy for `instanceof` checks and method calls, even though the proxy can later delegate to a different concrete transport.

This left replacement/default transports without runtime context and thread initialization. Switching from `AssistantChatTransport` to a custom transport could also make a later render call `setRuntime` on a transport that does not implement it, while switching in the opposite direction skipped setup entirely.

The fix keeps the proxy contract but performs setup and resumable-adapter discovery against the current concrete transport.

## Verification

- confirmed the regression tests fail on the previous implementation: unstable default instance, built-in to custom crash, and custom to built-in missing setup
- `pnpm --filter @assistant-ui/react-ai-sdk test` (173 tests)
- `pnpm --filter @assistant-ui/react-ai-sdk... build`
- patch changeset included

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5414?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.bc4DMdZPIJ9Ao8LHqC-gLu0lBbCE0vUtRP37voillOsLvDAlbS8Za-HTljY?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/fix-dynamic-chat-transport-wiring.md
+++ b/.changeset/fix-dynamic-chat-transport-wiring.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react-ai-sdk": patch
+---
+
+fix: keep replacement chat transports connected to the current runtime
--- a/packages/react-ai-sdk/src/ui/use-chat/useChatRuntime.test.ts
+++ b/packages/react-ai-sdk/src/ui/use-chat/useChatRuntime.test.ts
@@ -1,6 +1,7 @@
 // @vitest-environment jsdom
 
 import { act, renderHook, waitFor } from "@testing-library/react";
+import type { UIMessage } from "ai";
 import { beforeEach, describe, expect, it, vi } from "vitest";
 
 const mocks = vi.hoisted(() => {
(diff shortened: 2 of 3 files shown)


okisdev:
the diagnosis here was right: the proxy freezes its prototype on the initial target, so `instanceof` kept passing after a switch while `setRuntime` forwarded to a transport that does not implement it, and the bare `?? new AssistantChatTransport()` default churned per render. but your own #5336 landed the same design one day after this was filed (memoized `defaultTransport`, setup against the concrete `sourceTransport`, resumable discovery keyed on the current transport), and the #5923 migration then moved the whole file to `@assistant-ui/ai-sdk`, which is why this shows as conflicting.

the one piece that did not carry over is test coverage: the migrated suite pins resumable replacement (`observes resumable storage from a replacement transport`) but has no named coverage for default instance stability or the built-in to custom switch. if you want to port those two assertions to `packages/ai-sdk/src/runtime/useChatRuntime.test.ts`, a fresh small PR against current main would be welcome; nothing from this branch applies as-is.

closing as superseded by #5336 and the #5923 migration.

PR #5415, opened 2026-07-30
fix(react): render custom children and support asChild in AttachmentPrimitive.Thumb

ephraimduncan:
## What

`AttachmentPrimitive.Thumb` had two bugs:

1. **Custom children were silently swallowed.** The JSX body `.{ext}` always overrode any `children` passed via the `{...props}` spread, so consumers could never customize the thumb content.
2. **`asChild` threw at runtime.** With `asChild`, Radix `Primitive.div` renders a `Slot`, which requires exactly one React element child (`React.Children.only`). The hardcoded text nodes `[".", ext]` made every `asChild` usage throw.

## How

Render `{props.children ?? `.${ext}`}` instead of hardcoding the extension, matching the established fallback convention used by the sibling primitives (`QueueItemPrimitive.Text`, `SuggestionPrimitive.Title`, `ErrorPrimitive.Message`, `ComposerPrimitive.QuoteText`). The default rendering (`.pdf` etc.) is unchanged when no children are provided.

## Testing

Added colocated `AttachmentThumb.test.tsx` (same harness as `ThreadListItemTitle.test.tsx`) covering:
- default extension rendering
- children overriding the extension fallback
- `asChild` rendering the child element without throwing

Against the pre-fix code, the last two tests fail, so they guard the exact regression. Includes a patch changeset.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5415?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.-H4hWuwWavuSWbqt22pt9UEUukm68GTOZ1ryrdVQK43K9j44rjxX-9-u9-A?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>


okisdev (approved):
approving at `f75d6ce37`. i proved both bugs against current main rather than reading for them: `<Thumb>custom</Thumb>` rendered `<div>.pdf</div>` with the child gone, and `asChild` threw `Primitive.div failed to slot onto its children. Expected a single React element child or \`Slottable\``, because `Primitive` wraps `@radix-ui/react-primitive` and a text node cannot be slotted. `children ?? fallback` is also the shape three siblings already use (`QueueItemText`, `SuggestionTitle`, `ErrorMessage`), so this reads as conforming rather than inventing.

i merged main in rather than sending it back, since #5421 landed on the same lines. one resolution was load bearing and worth recording. the react suite conflicted add/add, and this branch's version contained `it("renders an empty extension for names without a dot")` asserting `<div>.</div>`, which is exactly the defect #5421 removed. taking that side would have restored the `split(".")` selector together with a green test pinning it, so the regression would have been invisible. the merged suite keeps the six semantics cases from main and adds the two new ones for children and `asChild`.

the react-native hunk was the same trap in source form, and all three runtimes now read `children ?? label`. `split(".")` appears nowhere in the tree, and the attachment suites pass at 8, 14 and 17. i also retuned the react changeset, which still said the default was the file extension; after #5421 it is the extension or the attachment type, and changesets become release notes.

PR #5416, opened 2026-07-30
fix(react): ignore Escape during IME composition in ComposerPrimitiveInput

ephraimduncan:
## What

Pressing Escape to dismiss an IME composition candidate (Japanese/Chinese/Korean input) was also cancelling the composer — discarding an in-progress message edit the user never meant to abandon.

## Why

The `useEscapeKeydown` handler in `ComposerPrimitiveInput` had no IME guard. Radix's `useEscapeKeydown` fires on any `key === "Escape"` keydown at the document capture phase, and browsers fire that keydown with `isComposing: true` when Escape cancels a composition. The handler fell through to `composer.cancel()`.

The sibling handlers already guard against this: `handleKeyPress` checks `e.nativeEvent.isComposing`, and `onChange` tracks composition state with a ref. The Escape handler was the one gap.

## How

- Add `if (e.isComposing) return;` at the top of the Escape handler, before plugin dispatch — same ordering as `handleKeyPress`.
- Tests: the `useEscapeKeydown` mock now captures the handler so the real handler body is exercised with dispatched `KeyboardEvent`s; two new tests cover Escape cancelling when cancellable and Escape being ignored mid-composition. Mutation-checked: removing the guard fails the new IME test.
- Changeset: patch for `@assistant-ui/react`.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5416?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.8Qodq9j8S4z7cr9F5Ee-HLkqGCz4M62LjmgJ2gtCutANhGr78EGO5buW1-E?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>


okisdev (approved):
approving at `26723228e`. the part that makes this easy to accept is that it is not a new idea: `handleKeyPress` in the same file already guards on composition in exactly this position, after the scope check and before plugin dispatch, with the same comment. the escape path was the one keydown handler that skipped it.

i read `@radix-ui/react-use-escape-keydown@1.1.5` rather than assuming: it listens at document capture and fires on `event.key === "Escape"` with no composition awareness at all, so anything the browser delivers with that key reaches `composer.cancel()`. the native-versus-synthetic distinction is right too, `e.isComposing` here against the sibling's `e.nativeEvent.isComposing`.

i checked whether it should also consult `compositionRef.current`, since `onChange` uses `nativeIsComposing || compositionRef.current`. it should not: that fallback exists to clear a ref stuck true when the browser drops `compositionend`, which is about avoiding false positives, not about `isComposing` being unreliable on keydown. the enter guard uses the event flag alone, and matching that is the consistent choice.

skipping plugin dispatch during composition is also correct rather than incidental. the first escape belongs to the IME candidate and the second, no longer composing, reaches the popover.

the test is a real guard: removing the three lines fails exactly the IME case, and the file stays at 25 with `src/primitives` at 145.

what i could not verify is the browser half. driving a real CJK composition through CDP is not something i can do here, so whether chrome delivers `key: "Escape"` with `isComposing: true` mid-composition is read from the spec and from this file's own enter guard, which only makes sense under the same behavior. the guard is inert if that combination never occurs, so the asymmetry favors shipping it.

i merged main in rather than leaving the red row. `Semver Check` was failing on `@assistant-ui/react-mcp@minor` and `@assistant-ui/mcp-docs-server@minor`, neither of which this PR touches; its `BASE_SHA..HEAD_SHA` changeset diff had swept in changesets from main against a base recorded five days ago. with the base refreshed the diff contains only `escape-ime-composition.md` and the check passes.

PR #5431, opened 2026-07-30
fix(react-opencode): avoid repeated child session scans

### LINKED ISSUE #5424: react-opencode: syncChildControllers rescans every part on every part event
follow-up from #5412.

`OpenCodeThreadController.syncChildControllers` walks every part of every message to build the set of live Task child session ids, and `shouldSyncChildControllers` includes `part.updated`, which fires for every tool state transition, step boundary, snapshot and patch during a run. so the scan is O(total parts in the transcript) per part event, i.e. quadratic over the length of a session, on the streaming hot path, for a set that only changes when a `task` part appears or disappears.

the constant is small (the walk early-returns on `part.type !== "tool" || part.tool !== "task"`, so it is a type comparison per part rather than a projection), which is why it did not block #5412.

the obvious shape does not work on its own. discovery can be answered from the event's own part via `getOpenCodeTaskSessionId(event.part)`, but removal cannot: `handleServerEvent` commits `this.state = nextState` before calling the sync, so a `part.removed` arrives after the part is already gone from state and there is nothing left to map back to a session id. answering both from the event needs a part id to session id index maintained alongside the state.

so the work is: keep an index of Task part id to child session id, update it on `part.updated` and `part.removed`, and reduce the full walk to `history.loaded` and `message.removed` only.

<a href="https://rupic.dev/s/s2.3TejlG2eUePc21Boa4BUP3Uu2Lu_MEaaTMGyRy0YM7AyAeIztYLCryhFKW8?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

Kinfe123:
## Summary

- index Task part IDs to their child session IDs inside `OpenCodeThreadController`
- update the index incrementally for live part updates and removals
- rebuild from the transcript only after history loads or whole-message removals

## Why

Task child-session synchronization previously walked every part in the transcript after every `part.updated` event. Since those events fire throughout streaming, long sessions accumulated quadratic work even when the updated part could not change the child-session set.

The index also keeps removal behavior correct after the reducer has already removed the part from state, and deduplicates multiple Task parts that reference the same child session.

Closes #5424

## Validation

- `@assistant-ui/react-opencode` tests: 85 passed across 11 files
- `@assistant-ui/react-opencode` build passed
- targeted oxlint and oxfmt checks passed
- regression test verifies a live update in a 50-part transcript performs one Task-session inspection instead of rescanning all 50 parts

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5431?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.2GjKKXJeRl28NMOGQRQjoPLZABb38OB27TneqO9iPZfUaM-8RrpwbNtzIQg?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/calm-tools-index.md
+++ b/.changeset/calm-tools-index.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react-opencode": patch
+---
+
+fix: avoid repeated child session scans during streaming
--- a/api-surface/assistant-ui__react-opencode.ts
+++ b/api-surface/assistant-ui__react-opencode.ts
@@ -1069,6 +1069,7 @@ declare class OpenCodeThreadController implements OpenCodeThreadControllerLike {
   private loadPromise;
   private reconnectSyncToken;
   private readonly childControllersById;
+  private readonly childSessionIdByPartId;
   private ancestorSessionIds;
   private isChildSession;
   private readonly stagedMessages;
@@ -1080,7 +1081,11 @@ declare class OpenCodeThreadController implements OpenCodeThreadControllerLike {
   private attachChildController;
   private detachChildControllers;
   private discard;
+  private rebuildChildSessionIndex;
+  private updateChildSessionIndex;
+  private removeFromChildSessionIndex;
   private syncChildControllers;
+  private syncChildSessionIndex;
   private ensureEventSubscription;
   private handleStreamReconnect;
   dispose(): void;
--- a/packages/react-opencode/src/OpenCodeThreadController.test.ts
+++ b/packages/react-opencode/src/OpenCodeThreadController.test.ts
@@ -4,6 +4,23 @@ import { STREAM_RECONNECTED_EVENT_TYPE } from "./OpenCodeEventSource";
 import { rejectWhenThrowing } from "./testUtils";
 import type { OpenCodeServerEvent } from "./types";
 
+const getOpenCodeTaskSessionIdSpy = vi.hoisted(() => vi.fn());
+
+vi.mock("./openCodeTaskSession", async (importOriginal) => {
+  const original =
+    await importOriginal<typeof import("./openCodeTaskSession")>();
+
+  return {
+    ...original,
+    getOpenCodeTaskSessionId: (
+      part: Parameters<typeof original.getOpenCodeTaskSessionId>[0],
+    ) => {
+      getOpenCodeTaskSessionIdSpy(part);
+      return original.getOpenCodeTaskSessionId(part);
+    },
+  };
+});
+
 const createDeferred = <T>() => {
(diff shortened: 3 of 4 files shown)


okisdev (approved):
approving. this closes #5424, which I filed, and it implements the shape that issue specified rather than the one-liner the original review thread suggested.

verified by measurement rather than by reading: swapping the controller back to main's implementation and running this PR's own test shows a single `part.updated` on a 50-part transcript producing 50 extra `getOpenCodeTaskSessionId` calls (100 against an expected 51), and exactly 1 with the index in place. that is the O(parts)-per-part-event scan #5424 described.

the part that mattered is handled correctly. #5424 noted that discovery can be answered from the event's own part but removal cannot, because `handleServerEvent` commits `this.state = nextState` before the sync runs, so by then a removed part is already gone from state. keying the index on part id and driving removal from `event.partId` answers both without a state lookup. the extra win is that `updateChildSessionIndex` early-returns when the mapping is unchanged, so the overwhelming majority of `part.updated` events (tool state transitions, step boundaries, snapshots, patches) never reach `syncChildControllers` at all.

ordering is right: the index sync runs after the state commit, so `rebuildChildSessionIndex` walks committed state. the ancestor filter survives the move from collection time to derive time. `discard()` clears the index alongside the controllers. the four fixes from #5412 are all still intact (`isChildSession`, the attach-loop break, the skip-reload guard, and the projection's ready gate).

on the tests, the two are not equally load-bearing. `inspects only the updated part during live child-session sync` is the discriminating one and fails against main as above; its mock uses `vi.hoisted` with a spread `importOriginal`, matching the repo convention. `keeps a child controller while another Task part references it` passes against main too, since the old full walk derived the same result from state, so it is a preservation test for the dedup and removal semantics across the refactor rather than proof of the fix. worth having, worth not mistaking for the former. it does incidentally prove the `partID` to `partId` event translation is wired, which the index now depends on.

api-surface carries only the five new private member declarations, so no public surface moved. 85 tests green.

PR #5435, opened 2026-07-31
fix(core): isolate message queue subscriber errors

Kinfe123:
## Summary

- isolate message queue subscriber failures with the shared event-listener helper
- continue notifying later subscribers and advancing an idle queue after a listener throws
- add regression coverage for the interrupted enqueue path

## Why

A message queue subscriber could throw after `enqueue()` stored the item but before `advance()` ran. That left the item queued, prevented `driver.run()` from starting, and skipped every later subscriber. This can turn an unrelated telemetry or UI listener failure into a stalled message send.

The shared listener helper reports the subscriber error while allowing queue processing to continue.

## Testing

- reproduced before the fix: one queued item, zero driver runs, and the later subscriber was not called
- `vitest run packages/core/src/tests/message-queue.test.ts`
- `oxlint packages/core/src/runtime/queue/message-queue.ts packages/core/src/tests/message-queue.test.ts`
- `aui-build` in `packages/core`

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5435?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.ilXHcEcP7MYaqN51wvl-kkyHExuB2y-MJ90JBs3g0OFIxaNZKX_6KS1tKV4?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/silent-queues-listen.md
+++ b/.changeset/silent-queues-listen.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/core": patch
+---
+
+fix: isolate message queue subscriber errors
--- a/packages/core/src/runtime/queue/message-queue.ts
+++ b/packages/core/src/runtime/queue/message-queue.ts
@@ -4,6 +4,7 @@ import {
   type QueueItemState,
 } from "../../store/scopes/queue-item";
 import { generateId } from "../../utils/id";
+import { notifyEventListeners } from "../../utils/notify-event-listeners";
 import { getThreadMessageText } from "../../utils/text";
 import type { ExternalThreadQueueAdapter } from "./external-thread-queue-adapter";
 
@@ -34,7 +35,7 @@ export const createMessageQueue = (
   let suppressIdle = 0;
 
   const notify = () => {
-    for (const callback of subscribers) callback();
+    notifyEventListeners(subscribers, undefined, "Message queue");
   };
 
   const setItems = (next: readonly QueueItemState[]) => {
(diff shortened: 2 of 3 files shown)


okisdev (approved):
approving. the wedge is real and precisely located: `push` assigns the lanes, `setLanes` notifies mid-transition, and `advance()` only runs afterwards, so a throwing subscriber leaves a user's message stored with the driver never started. i verified both directions in a worktree, 10/10 at your head and exactly the new test failing on the merge base with zero driver runs and the item stranded.

one classification note for the record, since this PR predates the model-context notify adjudication and superficially collides with it: the queue's subscriber set is its own set with a driver-internal contract, not a caller-facing state-mutation notify, and the rethrow tier would not fix the defect here — all subscribers would be notified but the rethrow would still abort `advance()`, stranding the queued send and reporting a failure for a message that stays queued. the throwing party is ultimately a runtime state subscriber, and it still rethrows loudly at every direct-mutation site; only the queue-transition hop demotes to a reported error so the driver invariant holds. that is a per-set tier assignment by contract, which is exactly what the one-semantic rule permits, not a second policy inside one set.

Issue #4792, opened 2026-07-09
[Bug]: file attachment source_type hardcoded to "base64" — no way to use URL/file_id for multimodal models


FireTable:
## Summary

In `convertLangChainMessages.ts:343-211` (`getMessageContent`), the SDK hardcodes `source_type: "base64"` for `type: "file"` content parts. There's no way for the adapter to opt into `"url"` or `"id"` source types, even though several providers support PDF/file inputs from URL or file_id.

## Reproduction

1. Implement an `AttachmentAdapter` whose `send()` returns a file content part with a URL in `data` (e.g. R2 public URL):
   ```ts
   return [{
     type: "file",
     data: "https://r2.example/u/abc/file.pdf",
     mimeType: "application/pdf",
     filename: "file.pdf",
   }];
   ```
2. Send the message through `useLangGraphRuntime`.
3. Inspect the outgoing LangChain HumanMessage content.

**Expected:** Either
- `source_type: "url"` is emitted (so LangChain's ChatOpenAI converter wraps our URL as a data URL for providers that need it), OR
- The adapter can opt in to URL passthrough for providers that accept PDF URLs directly (Anthropic, Gemini, etc.).

**Actual:** `source_type` is hardcoded to `"base64"` regardless of what the adapter provided:
```js
case "file": return {
  type: "file",
  data: part.data,        // our URL is passed verbatim
  mime_type: part.mimeType,
  metadata: { filename: part.filename ?? "file" },
  source_type: "base64"   // ← hardcoded; LangChain then prepends
};                          //   `data:${mimeType};base64,` to our URL,
                           //   producing a malformed data URL
```

## Impact

1. **Providers that need base64**: works only if the adapter pre-encodes raw base64 (no `data:` prefix). Adapters that want to keep files in object storage (R2/S3) and pass URLs pay the cost of base64-encoding the full file inline — 10 MiB PDF becomes ~13 MiB of base64 in every message.

2. **Providers that accept PDF URLs** (Anthropic Claude, Gemini, etc.) **can't be used** at all through this SDK — the hardcoded `"base64"` source_type forces LangChain's ChatOpenAI converter path, which then either mangles our URL or rejects it.

3. **File-id support** (OpenAI Assistants `file_id`, Anthropic `document` blocks) is unreachable for the same reason.

## Suggested fix

Two reasonable options:

**Option A — additive field on the content part**
```ts
type FileAttachmentContent = {
  type: "file";
  data: string;
[...]

PR #5440, opened 2026-07-31
fix(react-langgraph): preserve custom fields on file content parts

### LINKED ISSUE #5425: [Bug]: react-langgraph getMessageContent drops custom fields on file content parts — no way to pass an opaque file reference (file id) to the backend
## Summary

In `packages/react-langgraph/src/convertLangChainMessages.ts` (`getMessageContent`), file content parts returned by an `AttachmentAdapter` are rebuilt into a fresh object that keeps only `data` / `mime_type` / `metadata.filename` / `source_type`. Any custom field the adapter puts on the part (e.g. an internal file-storage id) is silently dropped before the message is handed to the LangGraph backend.

#4792 / #4795 / #4800 added `url` and `id` source types, but only for parts that already use the built-in `data` conventions — there is still no way to pass through an adapter-defined field, and no option on `useLangGraphRuntime` to override the append-message → LangChain conversion.

## Use case

Our `AttachmentAdapter.send()` uploads the file to internal storage and wants to emit:

```ts
return {
  ...attachment,
  status: { type: "complete" },
  content: [{
    type: "file",
    source_type: "flexport_file_object",   // dropped
    fileObjectFid: "flx::storage:file_object:…", // dropped
    filename: "invoice.pdf",
  }],
};
```

The backend (a LangGraph server) resolves the id to bytes itself — the browser never holds the file content, so `base64` data is not an option, and the storage URL is signed/short-lived, so `url` is not a stable reference either.

## Current workaround

We smuggle the id through the `data` string with a magic prefix and decode it server-side:

```ts
{ type: "file", data: `flexport_file_object:${fileObjectFid}`, mimeType: "application/octet-stream", filename }
```

This survives conversion (falls into the `base64` branch) but is fragile: it depends on `parseDataUrl` not matching, on `httpUrlPattern` not matching, and every backend consumer must know the prefix. It also means `source_type: "base64"` is a lie on the wire.

## Proposal (either would solve it)

1. Pass through unknown fields on file parts, e.g. `return { ...part, type: "file", mime_type: …, metadata, source_type: … }` so adapter-defined fields survive; or
2. Expose the conversion as an option, e.g. `useLangGraphRuntime({ convertAppendMessage })`, so adapters with custom wire formats can own the mapping.

Happy to send a PR for either direction if maintainers indicate a preference.

## Environment

- `@assistant-ui/react-langgraph` 0.14.16 (also verified on current `main`)

<a href="https://rupic.dev/s/s2.JulZW2_Bymepc4vUUFv2SFJwWxCjBdKeclAivq5ar0x0jlFkAh-CL9JgPwo?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

Solaris-star:
## Summary

`getMessageContent()` rebuilt file content parts with only known keys (`data`, `mime_type`, `metadata`, `source_type`), silently dropping any adapter-defined custom fields (e.g. internal storage ids like `fileObjectFid`).

This change destructures the known keys and spreads the remainder so custom fields survive the AppendMessage → LangChain conversion.

## Before

```ts
// adapter returns
{ type: 'file', source_type: 'flexport_file_object', fileObjectFid: 'flx::...', filename: 'invoice.pdf' }
// after getMessageContent — fileObjectFid is gone
{ type: 'file', data: '...', mime_type: '...', metadata: { filename: 'invoice.pdf' }, source_type: 'base64' }
```

## After

```ts
// custom fields preserved
{ type: 'file', fileObjectFid: 'flx::...', data: '...', mime_type: '...', metadata: { filename: 'invoice.pdf' }, source_type: 'base64' }
```

Closes #5425

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5440?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.mIGaYYm1Kn7axLhNwgK6zog-Am-SLXeVRY_YzuNDvfyD3b4Z63c1o4_KkRA?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/packages/react-langgraph/src/convertLangChainMessages.ts
+++ b/packages/react-langgraph/src/convertLangChainMessages.ts
@@ -325,9 +325,19 @@ export const getMessageContent = (msg: AppendMessage) => {
       case "image":
         return { type: "image_url" as const, image_url: { url: part.image } };
       case "file": {
+        // Preserve adapter-defined custom fields (e.g. storage ids) by
+        // spreading the remainder after extracting known keys (#5425).
+        const {
+          type: _type,
+          data: _data,
+          mimeType: _mimeType,
+          filename: _filename,
+          ...customFields
+        } = part;
         const metadata = { filename: part.filename ?? "file" };
         if (httpUrlPattern.test(part.data)) {
           return {
+            ...customFields,
             type: "file" as const,
             url: part.data,
             mime_type: part.mimeType,
@@ -337,6 +347,7 @@ export const getMessageContent = (msg: AppendMessage) => {
         }
         const parsed = parseDataUrl(part.data);
         return {
+          ...customFields,
           type: "file" as const,
           data: parsed?.data ?? part.data,
           mime_type: parsed?.mimeType ?? part.mimeType,


okisdev:
closing in favor of #5439.

this ground was already covered by #4792 → #4795 → #4800. #4795 was closed because a file block that carries the reference in a non spec key still fails at the provider conversion layer (python's `is_data_content_block` rejects the shape, `@langchain/openai`'s `fromStandardFileBlock` reads the spec keys), and #4800 landed url sniffing while deferring an explicit source type override until there was a real request for it. #5425 is that request, and #5439 implements it in the spec conformant shape (`id` plus `source_type: "id"`, matching the wire union in `packages/react-langchain/src/types.ts`).

three problems specific to this diff:

1. the spread is first, so `type`, `url`/`data`, `mime_type`, `metadata` and `source_type` all overwrite it. the "After" block in the description shows `source_type: 'flexport_file_object'` surviving, but `source_type` is hardcoded after the spread, so it cannot. a custom `metadata` is replaced wholesale for the same reason.
2. `FileMessagePart` is a closed type (`type | filename? | data | mimeType | parentId?`), so after the destructure `customFields` is typed as `{ parentId?: string }`. the only key the type system says this forwards is assistant-ui's internal nested-part `parentId`, onto a standard LangChain file block. `fileObjectFid` is invisible to TS and needs a cast to construct at all.
3. arbitrary top level keys are not in the file block wire union, so the emitted block is non spec. that is the failure mode #4795 was closed for.

separately, react-langchain carries the same converter and is not updated here (cross-runtime parity), and the PR has no changeset and no test.

#5439 covers the reported use case.

Solaris-star:
Fair points across the board — the top-level spread is the wrong shape. Closing this PR to settle the design in #5425 first.

The `metadata` widening approach makes the most sense to me: it's the declared extension point on `MessageContentFile`, it round-trips through `contentToParts` (which already reads `part.metadata`), and it keeps the block spec-shaped. Will follow up in #5425 with a concrete proposal.

PR #5455, opened 2026-07-31
fix(store): preserve notification subscriptions across render replays

ponyfly6:
## Summary

- keep `NotificationManager` as mount-lifetime state with lazy `useState`
- add a StrictMode regression test against the React Compiler output
- run the store compiler regression suite from the root test command

## Problem

On `main` at `932bd90f9c46fd343176791eae3299fc514d2fa4`, the default template's composer accepted DOM input events but immediately rendered an empty controlled value again, leaving the send button disabled.

The store state changed correctly. During a pre-commit render replay, however, the compiled `useMemo` implementation returned a different `NotificationManager`. The committed client could retain the first manager while the active store effect notified the later manager, disconnecting `useSyncExternalStore` subscribers from updates.

## Fix

The manager owns mutable listener and subscriber sets, so its identity is runtime state rather than a droppable computation cache. Initializing it with lazy `useState` preserves one manager for the full mount across render replays.

The new test runs against `packages/store/dist`, subscribes through the manager returned by the first StrictMode render, and notifies through the manager returned by the last render. It fails on the original implementation with zero subscriber calls and passes with one call after this change.

## Reproduction

1. Check out `932bd90f9c46fd343176791eae3299fc514d2fa4`.
2. Run `pnpm install` and `pnpm -C templates/default dev`.
3. Open the template and type in the composer.
4. Observe that the controlled input remains empty and the send button stays disabled.

## Testing

- `pnpm --filter @assistant-ui/store test`
- `pnpm test:react-compiler`
- `pnpm --filter @assistant-ui/store exec tsc --noEmit`
- `pnpm check:resource-memo`

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5455?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/fuzzy-composers-listen.md
+++ b/.changeset/fuzzy-composers-listen.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/store": patch
+---
+
+fix: keep assistant state subscriptions connected across pre-commit render replays
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "format:fix": "pnpm exec oxfmt",
     "sync-templates": "bash scripts/sync-templates.sh",
     "test": "turbo test && pnpm test:react-compiler && pnpm test:types",
-    "test:react-compiler": "turbo test:react-compiler --filter=@assistant-ui/core",
+    "test:react-compiler": "turbo test:react-compiler --filter=@assistant-ui/core --filter=@assistant-ui/store",
     "test:types": "pnpm -r --workspace-concurrency=4 --no-bail exec sh -c '! test -f tsconfig.json || tsc --noEmit'",
     "prepare": "husky && pnpm --filter @assistant-ui/react-devtools run build:css",
     "deps:check": "npx taze major -f -r",
--- a/packages/store/package.json
+++ b/packages/store/package.json
@@ -29,6 +29,7 @@
   "scripts": {
     "build": "aui-build",
     "test": "vitest run",
+    "test:react-compiler": "vitest run --config vitest.react-compiler.config.ts",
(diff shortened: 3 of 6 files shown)


ponyfly6:
Confirmed the requested end-to-end behavior in the real `templates/default` UI at `d27feb1`, using the compiled `@assistant-ui/store` package and clean Next.js dev caches.

- With `useNotificationManager` backed by lazy `useState`, browser-level keyboard input remained in the controlled composer after 1 second and the Send button became enabled.
- I temporarily reverted only the three-line hook change to the original compiled `useMemo`, rebuilt `@assistant-ui/store`, and restarted `templates/default`. The same input was reset to an empty string and Send remained disabled.
- I restored `useState`, rebuilt, and restarted from a clean cache. The composer retained the complete input and Send was enabled again; the restored result was repeated successfully.

This confirms that the compiled manager identity split was the cause of the frozen composer, rather than only an isolated internal invariant. The source is restored exactly to the PR head; no additional PR changes were needed.

okisdev:
closing as superseded: #5411 merged on 2026-08-01 (90b3003b9) with the same lazy useState fix plus a compiled-dist StrictMode regression test, and #5873 has since fixed the underlying replay mechanism in tap itself.

PR #5481, opened 2026-07-31
fix: reload keyed history adapters

Kinfe123:
## Summary

- add an optional `key` to the shared `ThreadHistoryAdapter` contract
- reload keyed history in both local and AI SDK runtimes when the backing account or workspace changes
- have the built-in Cloud history adapter key itself by the current Cloud client
- treat `undefined` as a distinct keyless scope and clear/reload when switching to a concrete key
- ignore stale loads and isolate persistence bookkeeping across scopes

## Why

A runtime can remain mounted while its history source changes after login or a workspace switch. Previously, both core local history and AI SDK external history latched the first load for the runtime's lifetime, so later requests could continue showing or persisting the previous scope.

The adapter now declares the identity of its backing scope:

```tsx
const history = useMemo(
  () => ({
    key: workspaceId,
    load,
    append,
    withFormat,
  }),
  [workspaceId, load, append, withFormat],
);

useChatRuntime({ adapters: { history } });
```

Keep `key` stable for the same persistence scope and change it when the backing history changes. Omitting it preserves the existing load-once behavior. `undefined` is itself the keyless scope, so wait to provide the adapter or mount the runtime until an asynchronously loaded scope key is known if a clear/reload is not intended.

## Validation

- 56 focused core runtime tests
- 37 `useExternalHistory` tests
- 27 `useAISDKRuntime` tests
- `@assistant-ui/core` and `@assistant-ui/react-ai-sdk` builds
- docs, changesets, and public API surfaces updated

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5481?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.uuAWM30arRQOAZHN8Gj4UfxcB_-Rao00EdMdTHUxffD_BFcFrNccbnbOBXo?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/fresh-history-adapters.md
+++ b/.changeset/fresh-history-adapters.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react-ai-sdk": patch
+---
+
+fix: reload external history when the history adapter changes
--- a/.changeset/fresh-history-adapters.md
+++ b/.changeset/fresh-history-adapters.md
@@ -2,4 +2,4 @@
 "@assistant-ui/react-ai-sdk": patch
 ---
 
-fix: reload external history when the history adapter changes
+fix: reload external history when the history adapter key changes
--- a/.changeset/fresh-history-adapters.md
+++ b/.changeset/fresh-history-adapters.md
@@ -1,5 +1,6 @@
 ---
 "@assistant-ui/react-ai-sdk": patch
+"@assistant-ui/core": patch
 ---
 
-fix: reload external history when the history adapter key changes
+fix: reload history when the history adapter key changes
--- a/api-surface/assistant-ui__core.ts
+++ b/api-surface/assistant-ui__core.ts
@@ -2242,6 +2242,7 @@ declare class LocalThreadRuntimeCore extends BaseThreadRuntimeCore implements Th
   private abortController;
   private _queue;
   private _queueRunInFlight;
+  private _historyScopeGeneration;
   private _historyWrites;
   private _chainHistoryWrite;
   private _persistPausedMessage;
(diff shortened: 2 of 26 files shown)


okisdev:
the observation is accurate: history loads once per mounted runtime. but that is the designed invariant rather than a defect; the history scope is fixed for the life of the mount. an account or workspace switch changes the identity of the whole runtime, not just history: the thread list, in flight runs, composer and queue state all belong to the old scope too. remounting the provider with a React key (`<AssistantRuntimeProvider key={workspaceId}>`) handles all of that at once with zero API, while a history-only reload leaves a half-switched runtime (old scope's thread list next to the new scope's history), which is a worse state than the one being fixed.

on top of that, no issue or field report backs the mounted-switch scenario, and `key` on `ThreadHistoryAdapter` is permanent public surface (it ripples through the api-surface of every package that re-exports the type, and every other history path inherits the parity expectation). the react-ai-sdk leg also targets files that #5923 has since moved to `@assistant-ui/ai-sdk`, so the branch cannot be revived as-is.

if there is a concrete product scenario that genuinely needs in-place scope switching, an issue making that case is the right next step. a small docs PR adding the remount idiom to the persistence guide would also be welcome.

closing as unjustified against the remount idiom.

PR #5498, opened 2026-08-01
fix(react-langgraph): preserve run config across tool resumes

Kinfe123:
## Summary

- associate each LangGraph send and streamed message with a small per-run context
- resume frontend tool results with the `runConfig` from the run that emitted them
- keep queued tool-result batches separated when multiple configured turns remain pending
- retain completed sibling tool results across a top-level graph error so a later result can finish the batch
- clear ownership on thread switches and prune ownership for messages that leave the current history
- add focused regression coverage and a patch changeset

## Why

A configured turn selected the expected model on its initial request, but its automatic frontend tool-result continuation called `handleSendMessage(batch, {})`. That discarded the turn's `runConfig`, so LangGraph could resume with default configurable values.

A single "latest config" is not sufficient because `autoCancelPendingToolCalls: false` allows tools from multiple configured turns to remain pending. The runtime now assigns each send a small run context, tags its input and streamed messages with that context, and finds a tool call's owner from the assistant message containing it. Resume batches are keyed by that owner, so each continuation uses the correct configuration without mixing results from separate turns.

Existing messages with stable IDs are assigned an unconfigured context before a new run begins, preventing identified history from inheriting a later turn's settings. Ownership is cleared when switching threads and pruned when messages leave history. A top-level graph error still drops queued resumes, but keeps an incomplete parallel result buffer so a sibling tool result arriving afterward does not strand the whole turn.

## Testing

- `pnpm --filter @assistant-ui/react-langgraph test` (237 tests)
- `pnpm --filter @assistant-ui/react-langgraph build`
- focused `oxlint` and `oxfmt` checks

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5498?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.zVsDJGjxmb1qWwMK6altjQQ1wCt6Vwgm0pEAg5R9nwOrOhAWIOFRFHVIdO8?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>


okisdev:
superseded by #6024, which is on main as `2b4d40be2` and closes #6018.

#6024 records ownership at the decode boundary and injects it on automatic tool resumes and `useLangGraphSendCommand`. this head covered the direct tool resume but not the command-then-tool path, and it is still conflicting with main.

the extra here, splitting the tool-result batch when `autoCancelPendingToolCalls` is false and two turns both have pending tools, is tracked as #6033.

PR #5499, opened 2026-08-01
fix(react): keep escape stop speaking active

Kinfe123:
## Summary

- move the Escape-to-stop-speaking handler from the conditional message action bar to the persistent thread root
- query the current thread speech state when Escape is pressed, leaving the key untouched when nothing is speaking
- add focused keyboard regression coverage and a patch changeset

## Why

I reproduced speech continuing after an older message action bar auto-hid. `ActionBarPrimitive.Root` returns `null` for hidden bars, which unmounted `ActionBarPrimitive.StopSpeaking` and its document-level Escape listener even though the speech session remained active. The source already had a TODO documenting this failure.

Keeping the handler at `ThreadPrimitive.Root` preserves the existing Escape UX for the lifetime of the thread, including when the originating action bar is hidden or unmounted. The visible stop button keeps its existing click behavior. No existing issue or open PR covered this path.

## Testing

- `pnpm --filter @assistant-ui/react test` (381 tests)
- `pnpm lint`

<a href="https://rupic.dev/s/s2.GXI1Ea0tRgs3fF5S1cBixYE525X9ak5Pes_Jr7k7CF4po-1dLlAy0zVbi58?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5499?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/stop-speaking-escape-scope.md
+++ b/.changeset/stop-speaking-escape-scope.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react": patch
+---
+
+fix: keep Escape-to-stop-speaking active when message action bars are hidden
--- a/.changeset/stop-speaking-escape-scope.md
+++ b/.changeset/stop-speaking-escape-scope.md
@@ -1,5 +1,6 @@
 ---
+"@assistant-ui/core": patch
 "@assistant-ui/react": patch
 ---
 
-fix: keep Escape-to-stop-speaking active when message action bars are hidden
+fix: keep Escape-to-stop-speaking active when message action bars are hidden by moving the shortcut to `ThreadPrimitive.Root`; custom compositions must mount the root to enable it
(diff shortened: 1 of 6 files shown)


okisdev (approved):
the hide-the-bar leak is real, and ThreadPrimitive.Root is the right owner. i read the bubble-phase listener against ComposerInput's capture handler: official mention/slash Escape already preventDefault, so the defaultPrevented guard holds. the tests cover hidden control, idle Escape, and composer-first cancel.

CI Test Changed Packages is green on this head. JSDoc on Root still does not mention the shortcut; that is a docs follow-up, not a reason to keep the leak.

PR #5504, opened 2026-08-01
fix(react-langchain,react-langgraph): preserve file attachment filenames

### LINKED ISSUE #5501: react-langchain: file attachment filenames reach openai as LC_AUTOGENERATED
found while verifying assistant-ui/assistant-ui#5489 against a live provider (langchain-core 1.5.3, langchain-openai 1.4.1).

## what happens

both converters emit a base64 file block with the filename nested under `metadata`:

```json
{ "type": "file", "data": "...", "mime_type": "application/pdf",
  "metadata": { "filename": "memo.pdf" }, "source_type": "base64" }
```

langchain-core's openai translator does carry a backward compat branch for exactly that shape, and calling it directly with our block does return `filename: "memo.pdf"`. it never sees that shape on the real path, though. `invoke` first normalizes the v0 block into a v1 block, which nests the whole v0 `metadata` object under `extras`:

```json
{ "type": "file", "base64": "...", "mime_type": "application/pdf",
  "extras": { "metadata": { "filename": "memo.pdf" } } }
```

the translator then looks for `filename`, `extras.filename`, and `metadata.filename` in that order. the normalized block has none of the three, so it falls through to a placeholder and warns.

measured on the wire, same message, same process:

| block shape | filename openai receives |
| -- | -- |
| `metadata.filename` (what we ship) | `LC_AUTOGENERATED` |
| `extras.filename` | `LC_AUTOGENERATED` |
| top level `filename` | `memo.pdf` |

## impact

fidelity only, as far as i could measure. a real pdf sent with the placeholder filename is still read correctly by the model (same answer both ways). but the provider never sees the user's filename, and anything downstream keying on the name or its extension gets the placeholder instead.

## suggested direction

emit `filename` at the top level of the file block, keeping the existing `metadata` for consumers that read it. the top level key is the shape that survives normalization, and it is also what langchain-core's own warning tells callers to use.

affects `@assistant-ui/react-langchain` and `@assistant-ui/react-langgraph`; both build the same `metadata = { filename: part.filename ?? "file" }` block.

[Track in Rupic](<https://rupic.dev/s/s2.5vmpKRD0XB7K9ez6xSHeBY-U3Dv1In3NyIIAfpFw0HbYFTj8VCnMnCuNl5k?ref=github>)

<a href="https://rupic.dev/s/s2.5vmpKRD0XB7K9ez6xSHeBY-U3Dv1In3NyIIAfpFw0HbYFTj8VCnMnCuNl5k?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

Gujiassh:
closes #5501

## Summary

LangChain's v0-to-v1 normalization moves a legacy file block's `metadata` under `extras.metadata`, so provider translators no longer see the attachment filename and fall back to `LC_AUTOGENERATED`. Emit the existing filename at the top level of outbound id, URL, and base64 file blocks in both LangChain adapters while retaining `metadata` for existing consumers.

The tests cover all three source shapes in `@assistant-ui/react-langchain` and `@assistant-ui/react-langgraph`, including the existing default filename behavior and source-key semantics.

## Test plan

- [x] `pnpm --dir packages/react-langchain test`
- [x] `pnpm --dir packages/react-langgraph test`
- [x] `pnpm --dir packages/react-langchain run build`
- [x] `pnpm --dir packages/react-langgraph run build`
- [x] Changed-file Oxlint/Oxfmt and `git diff --check`
- [ ] Verify the filename on the wire with a live LangChain/OpenAI invocation

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5504?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.Xi41oNfcmwJg_l3TEnSQloIntUzBC5F1qKlGsqMIKn8vzgM3UWqENijuoN0?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>


okisdev:
verified the one box left open in the test plan, on a live OpenAI call through langchain-core 1.5.3 and langchain-openai 1.4.1, driving this branch's `getMessageContent` rather than a hand written block.

| attachment | filename on the wire | provider |
| --- | --- | --- |
| `filename: "vault-memo.pdf"` | `vault-memo.pdf` | pdf read correctly |
| no filename (default applies) | `file` | pdf read correctly |

before this change both cases arrived as `LC_AUTOGENERATED`. the unnamed case now sends the `"file"` default instead, and since neither placeholder carries an extension the provider behaves identically, so that is not a regression.

i also checked the round trip, since keeping `metadata` alongside the new top level key is what makes that safe: posting the block through a running `langgraph dev` server and reading the thread state back shows it persisted verbatim with both `filename` and `metadata` intact, and the inbound converter still resolves the name through `metadata.filename`. the v0 to v1 normalization behind #5501 happens inside `ChatOpenAI.invoke` at wire time and never touches what is stored, so retaining `metadata` costs nothing and keeps existing readers working. good call not dropping it.

merges cleanly onto current main (#5503 landed in these same two files after this branch forked); with main merged in, both suites stay green at 106 and 234.

one non blocking note, not for this PR: the inbound `contentToParts` reads only `metadata?.filename`, so a third party producer emitting just the v1 shape (top level `filename`, no `metadata`) would still fall back to `"file"`. that gap predates this change and is not made worse by it.

okisdev (approved):
lands the shape #5501 asked for, on all three source types in both packages, and keeps `metadata` so nothing downstream changes. verification evidence is in the comment above: the filename now reaches the provider on a live call, and the block still round-trips through a langgraph server with the name intact.

worth recording why this does not run into the objection that closed #5440: `filename` is not an arbitrary extension riding on the block, it is a first-class field of the data content block spec, read first by langchain-core's translator and named in its own warning text. `metadata` staying put is what keeps `contentToParts` working, which was the property that discussion cared about.

PR #5506, opened 2026-08-01
test(tap): add react-compiler compiled-dist test run

ShobhitPatra:
## What

Adds a `test:react-compiler` run to `packages/tap`, mirroring the harness core and store already have (#5411): a vitest config that aliases the test's relative imports to the built dist, plus a StrictMode replay contract test, wired into the root `test:react-compiler` script.

## Why

tap hosts the entire replay machinery that #5422 hit (memoCache reset and promote in `ResourceFiber`, the `c()` routing in `react-shim/compiler-runtime`), yet it was the only one of the three packages with no test against built dist. While building this I confirmed both failure directions with dist mutations: breaking the per-render cache reset fails the first test, breaking promote-at-commit fails the second, so a regression here now fails in tap instead of surfacing as a downstream store mystery.

## Impact

- CI already runs the root `test:react-compiler` script, so tap coverage kicks in with no workflow change.
- Test-only: no runtime code changes, no new public surface.

## Scope

- tap itself is deliberately never compiled by React Compiler (`x-buildutils` excludes it by design), so unlike store the test does not import a compiled tap module: it emulates what compiled consumer output does, calling `c()` from the dist shim inside a resource render with the sentinel pattern the compiler emits.
- The contract pinned is the #5422 semantics at its home: an uncommitted StrictMode replay re-seeds the `c()` cache from empty while state and ref cells survive on the fiber, and the committed cache persists across post-mount re-renders.
- The new file also matches tap's default vitest include and runs against source there too; that double-run is intentional and matches store, whose react-compiler test does the same under its default config.
- Tests: both assertions are mutation-verified against dist; not covered: the shim's non-tap fallback path (`reactC` outside a resource render), same scope line the store template drew.
- No changeset: test-only change, matching repo precedent for test-only PRs (#5366, #5189); the config and test are not published surface.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5506?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.lzvdx3jDo4bQ91Tb83acK2dq4DglFUdA221Shr5jgv_Bf1w7lml7a_DejV8?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "format:fix": "pnpm exec oxfmt",
     "sync-templates": "bash scripts/sync-templates.sh",
     "test": "turbo test && pnpm test:react-compiler && pnpm test:types",
-    "test:react-compiler": "turbo test:react-compiler --filter=@assistant-ui/core --filter=@assistant-ui/store",
+    "test:react-compiler": "turbo test:react-compiler --filter=@assistant-ui/core --filter=@assistant-ui/store --filter=@assistant-ui/tap",
     "test:types": "pnpm -r --workspace-concurrency=4 --no-bail exec sh -c '! test -f tsconfig.json || tsc --noEmit'",
     "prepare": "husky && pnpm --filter @assistant-ui/react-devtools run build:css",
     "deps:check": "npx taze major -f -r",
--- a/packages/tap/package.json
+++ b/packages/tap/package.json
@@ -41,6 +41,7 @@
     "build": "aui-build",
     "prepublishOnly": "pnpm turbo build",
     "test": "vitest run",
+    "test:react-compiler": "vitest run --config vitest.react-compiler.config.ts",
     "test:watch": "vitest"
   },
   "peerDependencies": {
--- a/packages/tap/src/react-shim/compiler-runtime.react-compiler.test.tsx
+++ b/packages/tap/src/react-shim/compiler-runtime.react-compiler.test.tsx
@@ -1,0 +1,94 @@
+// @vitest-environment jsdom
+
+import { StrictMode, useRef, useState } from "react";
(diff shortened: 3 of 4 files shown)


okisdev:
closing: the premise this PR was built on flipped with #5873, and what remains is covered.

the first test pins the pre #5873 behavior as the contract (distinct c() cache and distinct memoized instance per StrictMode replay pass); that re-seeding was the root cause of #5422/#5502 and #5873 removed it, so on current main the assertions describe the fixed bug. with them inverted the test duplicates packages/tap/src/__tests__/strictmode/compiled-memo-identity.test.tsx, which now pins the same seam (useTapHost plus hand written c(1) under StrictMode) at source level.

the compiled-dist premise also does not hold for tap: aui-build enables the React Compiler only for packages that depend on tap and react (x-buildutils src/index.ts), so tap's own build never runs the compiler and its dist is plain bundling with no src to dist semantic gap. the cross package composition that does matter, compiled consumer code against tap's built dist, is already exercised on every CI run by store's test:react-compiler, which imports useTapHost from the built workspace dist. the second test (post commit persistence) duplicates the existing "c() persists the memo cache across renders" coverage.

PR #5509, opened 2026-08-01
test(tap): drop stray argument from zero-arg Stream resource call

ShobhitPatra:
## What

Fixes a call-arity type error in `concurrent-version-replay.test.tsx`: `Stream({})` passes an argument to a resource whose hook takes none, now `Stream()`.

## Why

I hit this while typechecking tap for #5506: `tsc --noEmit` fails with TS2554 on this line, which breaks the root `pnpm test` locally (it runs `test:types`). CI never runs the typecheck, so it slipped through when the test landed in #5331.

## Impact

No behavior change: the argument was ignored at runtime and the test passes identically. Local root `pnpm test` goes back to green for tap.

## Scope &

- One line, test-only; runtime and published surface untouched.
- Getting CI to run `test:types` is a separate question, deliberately not touched here.
- No changeset: test-only change to a test file.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5509?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.NiogBT1vQVsZDT30sB0rSZL1J_bMe0IqwtTfWY40EbxJoOom9Yd8kI8lNZw?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/packages/tap/src/__tests__/react/concurrent-version-replay.test.tsx
+++ b/packages/tap/src/__tests__/react/concurrent-version-replay.test.tsx
@@ -60,7 +60,7 @@ describe("React-hosted reducer replay below the committed version", () => {
     let api!: { chunks: readonly string[]; push: (c: string) => void };
 
     function App() {
-      const value = useResource(Stream({}));
+      const value = useResource(Stream());
       api = value;
       return <div>{value.chunks.join(",")}</div>;
     }


okisdev:
closing as superseded rather than rejected: your diagnosis was right when you filed this — `Stream({})` did fail `tsc --noEmit` with TS2554, CI does not run `test:types`, and the one-line fix was the correct one. in the two weeks it sat, #5867 rewrote this test file as part of the tap replay work and every `Stream` call is already zero-arg on main (`git log -S 'Stream({})'` shows d7afb3dbd introducing it and d4b884535 removing it), so there is nothing left for this diff to change. the larger point you deliberately scoped out, that nothing in CI runs `test:types` so this class slips through, remains real and is tracked separately. thanks for the precise report.

PR #5553, opened 2026-08-03
fix(ui): render file parts in user messages and note the shipped default in docs

### LINKED ISSUE #5549: [BUG](react-langgraph, react-ai-sdk): attachment payloads exist twice on converted messages, blocking user-side file rendering
## Description

react-langgraph flattens attachment content into the wire message so the model input is self-contained, and separately reattaches the original attachments to the converted message so attachment UI can render chips. That leaves the same file on the converted message twice in the same session: once as a content part (file/image) and once inside `attachments`.

Today this is invisible because the shipped Thread renders no file parts. It became load-bearing on #5548: the moment the Thread gains a user-side `File` renderer, a langgraph user who attaches a PDF sees the attachment chip above the bubble and a duplicate File card inside it. The user-side renderer was built and then reverted there (commit 2303b46) specifically because of this; see the review thread on #5548 for the mechanism.

react-ai-sdk has the same class of bug through a different mechanism, and unconditionally: `convertParts` maps user file UI parts into `content`, and the user branch rebuilds `attachments` from the same `message.parts.filter((p) => p.type === "file")`, so the payload is doubled both in-session and on restored history. react-ag-ui avoids the class (user media is kept out of content and moved to attachments on restore). react-langchain avoids it too: it has no reattachment mechanism.

Because the fixes are adapter-shaped, this issue anchors three legs:

1. react-langgraph: payload-precise in-session dedupe (this cannot be a blanket strip; the fresh-reload path has no attachments, so content parts must keep rendering there).
2. react-ai-sdk: drop user file parts from content, since attachments are always rebuilt from them and fully represent them (the react-ag-ui shape).
3. Re-land the shipped Thread's user-side `File` arm (reverted in #5548 commit 2303b46), which is what makes the duplication visible.

## Reproduction

With `useLangGraphRuntime`, attach a file and send a message, then inspect the converted user message in the same session:

```ts
// useLangGraphRuntime.ts stages the attachments by staged human message id,
// while getMessageContent flattens the same attachment content into the wire message.
// The converter then produces BOTH:
message.content // includes { type: "file", data: "<base64>", ... } from the flatten
message.attachments // includes the same file as a CompleteAttachment
```

Observed: the payload is present in both `content` and `attachments` in-session. After a page reload the staging map is empty, so only the content part remains.

## Expected behavior

In-session, the converted message should not carry the flattened attachment copies in `content` when the attachments are reattached for that message. The dedupe should be payload-precise: drop a converted media part (file/image) only when its payload matches an attachment content part, so direct-content media sent alongside attachments keeps rendering, and the fresh-reload path (empty staging map) is untouched and keeps restored history renderable.

Text parts are out of scope: flattened text renders in the bubble today by design and the chip does not duplicate it visually.

## Evidence

- Flatten: `packages/react-langgraph/src/convertLangChainMessages.ts:376-379` (`getMessageContent` spreads `attachments.flatMap((a) => a.content)` into the wire message).
- Reattach: `packages/react-langgraph/src/useLangGraphRuntime.ts:92-105` (staging map, with a comment documenting the intent) and `convertLangChainMessages.ts:266-276` (human branch reattaches by message id).
- Media conversion arms that produce the duplicate content parts: `convertLangChainMessages.ts:144-179`.
- react-ai-sdk double-carry: `packages/react-ai-sdk/src/ui/utils/convertMessage.ts:326-333` (file UI parts into content) and `:402-417` (attachments rebuilt from the same parts).
- Sibling that avoids the class: react-ag-ui `conversions.ts:536-543` (user media kept out of content, moved to attachments on restore).
- Context: #5548 review thread (user-side File arm reverted pending this fix).

<a href="https://rupic.dev/s/s2.NtbwP0vMuz8FgcPh5zTC_XDOYesRajgF31p54CDApvqzCI0vuwe9MbY_6Jg?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

ShobhitPatra:
## What

User messages in the shipped Thread now render `file` parts through the `File` component (`components={{ File }}` on `MessagePrimitive.Parts`), and the File docs note that the shipped Thread renders file parts by default on both sides.

## Why

This re-lands the arm built and reverted on #5548 (commit 2303b46): back then, react-langgraph would have double-rendered every in-session attachment as chip plus card. #5550 fixed that at the source by deduping the flattened attachment copies, so restored-history user file parts (adk `file_url`, langgraph content parts) can finally render instead of dropping the whole bubble through `empty:hidden`. Closes #5549.

## Impact

- Restored user file parts render as File cards; previously the bubble showed nothing.
- Live composer attachments still render once, as chips: langgraph dedupes in-session (#5550), react-ai-sdk and react-ag-ui already keep user media out of content.

## Scope 

- Stacked on #5548 and merges after #5550 lands; will rebase once the parent merges.
- react-ai-sdk needs no change: `convertParts` already strips user file parts from content (convertMessage.ts:189-193, since #3695); the earlier claim otherwise on #5549 was corrected there.
- The `" "` text filler for attachment-only langgraph sends still renders an empty-looking bubble; pre-existing wire behavior, tracked separately, needs a shape call.
- Docs: one sentence in `file.mdx`; the manual wiring example stays as the custom-thread path.
- Tests: none in packages/ui (no harness); the langgraph side is pinned by #5550's runtime round-trip test.
- No changeset: `@assistant-ui/ui` and docs are private.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5553?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.6Zl7-SmTMGwcxwqXHXAy1HoW1qOeIp_LB5CNd-ylTe6ti2-b1A2CgYg4dKU?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/apps/docs/content/docs/ui/file.mdx
+++ b/apps/docs/content/docs/ui/file.mdx
@@ -22,7 +22,7 @@ import { FileSample } from "@/components/docs/samples/file";
 
 ### Use in your application
 
-Pass `File` to `MessagePrimitive.Parts`:
+The shipped `Thread` component renders `file` parts through `File` by default, in both assistant and user messages. In a custom thread, pass `File` to `MessagePrimitive.Parts`:
 
 ```tsx title="/components/assistant-ui/thread.tsx" {1,8}
 import { File } from "@/components/assistant-ui/file";
--- a/apps/registry/src/registry.ts
+++ b/apps/registry/src/registry.ts
@@ -181,6 +181,7 @@ export const registry: RegistryItem[] = [
     registryDependencies: [
       "button",
       "https://r.assistant-ui.com/attachment.json",
+      "https://r.assistant-ui.com/file.json",
       "https://r.assistant-ui.com/follow-up-suggestions.json",
       "https://r.assistant-ui.com/markdown-text.json",
       "https://r.assistant-ui.com/reasoning.json",
--- a/packages/ui/src/components/assistant-ui/file.tsx
+++ b/packages/ui/src/components/assistant-ui/file.tsx
@@ -212,7 +212,8 @@ const FileImpl: FileMessagePartComponent = ({
   sourceType,
 }) => {
   const kind = getFileDataKind(data, sourceType);
-  const showSize = kind === "base64" || kind === "data-uri";
+  const showSize =
+    typeof data === "string" && (kind === "base64" || kind === "data-uri");
 
   return (
     <FileRoot>
--- a/packages/ui/src/components/assistant-ui/file.tsx
+++ b/packages/ui/src/components/assistant-ui/file.tsx
@@ -183,6 +183,7 @@ function FileDownload({
   children,
   ...props
 }: FileDownloadProps) {
+  if (typeof data !== "string") return null;
   const kind = getFileDataKind(data, sourceType);
   if (kind === "id") return null;
   if (kind === "url" && !/^(https?:\/\/|blob:)/i.test(data)) return null;
--- a/packages/ui/src/components/assistant-ui/thread.tsx
+++ b/packages/ui/src/components/assistant-ui/thread.tsx
@@ -36,6 +36,7 @@ import {
   MessagePrimitive,
   SuggestionPrimitive,
   ThreadPrimitive,
+  type FileMessagePartComponent,
   type ToolCallMessagePartComponent,
   useAuiState,
 } from "@assistant-ui/react";
@@ -483,6 +484,12 @@ const AssistantActionBar: FC = () => {
   );
 };
 
+const UserFilePart: FileMessagePartComponent = (part) => (
+  <div data-slot="aui_user-message-file" className="py-1">
+    <File {...part} />
+  </div>
+);
+
 const UserMessage: FC = () => {
   return (
     <MessagePrimitive.Root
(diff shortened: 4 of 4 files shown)


okisdev (approved):
verified against post-#5548 main in a worktree: the effective delta collapses to exactly this PR's own commit (the parent squash landed identical content, so no rebase was needed), the ui build graph is green, and oxlint/oxfmt are clean. UserFilePart mirrors the assistant arm's slot naming and block-level spacing, and the components override is the right seam for the user side. the docs sentence matches shipped behavior now that both arms exist. the image-part gap flagged in review is real and out of scope; filing it as a follow-up.

PR #5559, opened 2026-08-03
feat(core): support main thread reload in tap runtimes

### LINKED ISSUE #5532: tap client: threads.reloadMainThread() is unreachable on the ExternalThread path
#5522 adds `threads.reloadMainThread()` and the `unstable_refetchThread` capability behind it. On the tap client the operation exists but can never do anything:

- `ExternalThread.ts` hardcodes `refetchThread: false` and `ExternalThreadProps` has no member to turn it on, so the capability can never be true.
- `SingleThreadList` and `InMemoryThreadList` answer `reloadMainThread` with `RESOLVED_PROMISE`, and neither has a runtime hook to remount, so there is no fallback path either.

The result is a call that resolves having done nothing, on the architecture the repo is migrating onto. That is defensible today, since neither of those thread lists holds remote state to refetch, but it stops being defensible once a tap-side remote thread list exists.

Wiring it means threading `onRefetchThread` through `ExternalThreadProps` to the same getter the legacy `ExternalStoreThreadRuntimeCore` uses, and deciding what `reloadMainThread` should mean on a thread list that has no remote state at all: resolve silently as now, or report that the runtime does not support it.

Sibling divergences: #5528 for `react-google-adk`, #5531 for the `react-langgraph` adoption.

<a href="https://rupic.dev/s/s2.GaLvUvlMzRUGzHZCIIkFfnzb1YEI_2fCYZ7nismRpEUlo6Mkz7BviEnUNYQ?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

Kinfe123:
## Summary

- add `onRefetchThread` to tap `ExternalThread` resources and expose the matching capability
- route `threads.reloadMainThread()` through both the automatic single-thread list and `InMemoryThreadList`
- preserve the existing resolved no-op when no refetch callback is configured, while propagating callback failures

## Why

Tap clients already expose `threads.reloadMainThread()`, but external tap threads always resolved without refreshing anything. Apps that need to reconcile remote state after polling, reconnects, or out-of-band updates therefore had to remount the thread and could lose in-progress UI state. This completes the tap path for the in-place refetch contract introduced in #5522.

Closes #5532

## Test plan

- `pnpm lint`
- `pnpm --filter @assistant-ui/docs generate:api-reference --strict`

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5559?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.M4uulNR5EexPI2zwRn3y4IiFfP5zKmxtdvBXaaqZv0UiNPcPbi-JPzFA074?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/tap-thread-refetch.md
+++ b/.changeset/tap-thread-refetch.md
@@ -1,0 +1,6 @@
+---
+"@assistant-ui/core": patch
+"@assistant-ui/react": patch
+---
+
+feat: support in-place main thread reloads in tap runtimes
--- a/api-surface/assistant-ui__core.ts
+++ b/api-surface/assistant-ui__core.ts
@@ -4649,6 +4649,7 @@ type ThreadMethods = {
   export(): ExportedMessageRepository;
   import(repository: ExportedMessageRepository): void;
   reset(initialMessages?: readonly ThreadMessageLike[]): void;
+  unstable_refetchThread?: (() => Promise<void>) | undefined;
   importExternalState(state: unknown): void;
   message(selector: {
     id: string;
--- a/api-surface/assistant-ui__core.ts
+++ b/api-surface/assistant-ui__core.ts
@@ -4649,7 +4649,7 @@ type ThreadMethods = {
   export(): ExportedMessageRepository;
   import(repository: ExportedMessageRepository): void;
   reset(initialMessages?: readonly ThreadMessageLike[]): void;
-  unstable_refetchThread?: (() => Promise<void>) | undefined;
+  unstable_refetchThread?(): Promise<void>;
   importExternalState(state: unknown): void;
   message(selector: {
     id: string;
@@ -5955,7 +5955,7 @@ declare const useActionBarFeedbackPositive: () => {
 
 declare const useActionBarReload: () => {
   reload: () => void;
-  disabled: any;
+  disabled: boolean;
 };
 
 declare const useActionBarSpeak: () => {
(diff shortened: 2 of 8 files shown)


okisdev:
superseded by #5828, which just merged; closing. the design here was right and converged with what shipped: `onRefetchThread` on the tap external thread props, the capability derived from its presence, and both lists delegating with the resolved-promise fallback, credit for getting there first.

the branch predates #5745, which moved `ExternalThread` and `SingleThreadList` into `@assistant-ui/core`, so the diff targets files that no longer exist and the PR sits CONFLICTING against main. the merged change lands the same wiring on the new layout, adds the `unstable_refetchThread` seam on the tap `ThreadMethods` (method-shorthand optionality; the store client schema rejects an explicit `| undefined`), and carries the test intent forward as `packages/core/src/tests/external-thread-refetch.test.tsx` plus `InMemoryThreadList` cases in `packages/react`.

PR #5572, opened 2026-08-03
fix(core): scope interactable persistence to its adapter

Kinfe123:
## summary

- queued interactable edits now flush through the adapter that observed them when the persistence adapter is replaced or removed (account or workspace switch, declarative unmount)
- an in-flight full-state save is not overlapped. the outgoing payload is snapshotted at switch time and drained after the current save settles, so an older write cannot land last and erase the queued edit
- retry timers are cancelled with the outgoing adapter so they cannot fire into the new context
- client and legacy implementations share that invariant

## why

interactable persistence is debounced. if the adapter changed during that window, `runPersistence` read the new `adapterRef` and wrote adapter A's state through adapter B.

the first version of this PR started a second `save` on switch while the first was still in flight. two full-state payloads then raced on A; a slower first write could overwrite the queued edit after the account had already switched. `flush()` and the debounce path already refuse to start a batch while one is in flight. the switch path now matches.

this does not replace already-loaded interactable state when the persistence context changes.

### already verified

- [x] `packages/core` vitest `Interactables.test.ts` + `interactables-legacy/Interactables.test.ts` -> 29/29, including mid-debounce switch, serialized in-flight drain (second save starts only after the first resolves; last A payload is the queued edit), retry-timer cancel, unmount flush, and post-switch edits going to B
- [x] the mid-debounce switch case fails on `origin/main` (`firstSave` empty, `secondSave` receives `{ n1: { v: 1 } }`)

### reviewer should verify

- [ ] a fetch-backed adapter: edit, switch accounts before 500ms, confirm the first account receives the edit and the second does not

## notes

- no linked issue. unstable API, patch changeset


okisdev (approved):
queued writes stay on the outgoing adapter, and a second full-state save no longer starts while one is in flight. the mid-debounce switch leaked to adapter B on main; the in-flight case now waits for the first save, then drains the snapshotted A payload, so the last write to A is the queued edit. client 23 and legacy 6 tests passed locally, including that serialization.

PR #5586, opened 2026-08-04
fix(react-data-stream): cancel responses after callback failures

Kinfe123:
## Summary

- cancel the fetched response body when `onResponse` rejects
- preserve the original callback error if response cancellation also fails
- cover the cleanup path with a cancellable streaming response regression test

## Why

`onResponse` is commonly used to validate response headers or perform application-side setup before decoding begins. If an asynchronous callback rejected after `fetch()` returned, the runtime rethrew the callback error without consuming or cancelling the response body. The server or model could therefore continue generating and transmitting a response that no consumer would read.

The response stream is now cancelled before the callback error is rethrown. Callback failures remain separate from request and decoding errors, so the existing `onError` behavior does not change.

## Validation

- reproduced the failure on unchanged `main`: the response body's `cancel()` hook was called zero times
- `pnpm --filter @assistant-ui/react-data-stream test` (22 tests)
- `pnpm --filter @assistant-ui/react-data-stream build`
- `pnpm --filter @assistant-ui/react-data-stream exec tsc --noEmit`

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5586?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.bjXk3hnofEthoE2VX1ZCrR7UevtkwUyki6qPfRw50iGsUev2T-t7ugcOYoo?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/clean-stream-responses.md
+++ b/.changeset/clean-stream-responses.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react-data-stream": patch
+---
+
+fix: cancel response streams when response callbacks fail
--- a/packages/react-data-stream/src/useDataStreamRuntime.test.ts
+++ b/packages/react-data-stream/src/useDataStreamRuntime.test.ts
@@ -86,15 +86,26 @@ describe("useDataStreamRuntime request errors", () => {
 
   it("keeps response callback failures separate from request errors", async () => {
     const error = new Error("response callback failed");
+    const cancel = vi.fn().mockRejectedValue(new Error("cancel failed"));
     const onResponse = vi.fn().mockRejectedValue(error);
     const onError = vi.fn();
-    vi.stubGlobal("fetch", vi.fn().mockResolvedValue(new Response()));
+    vi.stubGlobal(
+      "fetch",
+      vi.fn().mockResolvedValue(
+        new Response(
+          new ReadableStream({
+            cancel,
+          }),
+        ),
+      ),
+    );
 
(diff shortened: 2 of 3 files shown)


okisdev (approved):
the fix itself was right and both directions verified in a worktree; on the original head the extended test passed 12/12, and with the source reverted to main it failed with zero cancel calls, so the leak and the fix are both real.

i pushed one commit on top (873ea016c) reshaping the catch per the open review thread, which i confirmed independently before adopting: the abort listener now detaches before any async work, and the cancellation fires without being awaited, `void result.body?.cancel().catch(() => undefined)`. awaiting made the callback error's propagation depend on the body's cancel algorithm settling, which a hand-built stream from msw or a proxy can simply never do, and it widened the window where an abort during the round-trip invoked `onCancel` on an already-failing run. this also makes the guard byte-identical to the four sibling sites that converged on this shape in #5845 and #5938, so the reject-without-cancel family now has one form everywhere. your test needed no changes and still pins both the cancellation and the preserved original error.

PR #5592, opened 2026-08-04
feat(assistant-stream): opt-in heartbeat for DataStreamResponse

Yonom:
## What

Makes heartbeats opt-in for `DataStreamResponse` and moves keepalive dispatch onto the encoder.

- `DataStreamResponse(stream, heartbeat: HeartbeatOption = False)` — heartbeats are off by default for the data stream protocol. Opt in with `heartbeat=True` (15s default interval) or a number of seconds. The keepalive token is a bare `"\n"`, which the data stream line protocol ignores.
- SSE response classes are unchanged: `: heartbeat` comment keepalives remain on by default.
- Dispatch is now encoder-declared: `StreamEncoder.get_keepalive_token()` (default `None`) replaces the hardcoded `text/event-stream` media-type check in `AssistantStreamResponse`. SSE encoders return the comment line, `DataStreamEncoder` returns `"\n"`, encoders without a benign token return `None` and never emit keepalives.

## Why

The previous media-type check silently excluded non-SSE formats from keepalives even when the format has a safe idle token, and coupled the response class to a specific wire format. Each encoder knows whether its format tolerates a keepalive and what that token is.

## Tests

Contract tests in `python/assistant-stream/tests/test_heartbeat.py`: default-off for `DataStreamResponse`, opt-in emission of blank-line keepalives between real chunks (ordering asserted), and the existing SSE comment-heartbeat path. `uv run --extra dev --extra langgraph pytest`: 143 passed, 7 skipped (redis-gated).

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5592?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.mOSq0f7jTHIJHN1YAMaezK_9cvIpNpQ5R_J80iFTYHd06LI3xEQHz_kobXA?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/python/assistant-stream/src/assistant_stream/serialization/assistant_stream_response.py
+++ b/python/assistant-stream/src/assistant_stream/serialization/assistant_stream_response.py
@@ -1,7 +1,7 @@
 from assistant_stream.assistant_stream_chunk import AssistantStreamChunk
 from assistant_stream.serialization.heartbeat import (
     HeartbeatOption,
-    add_sse_heartbeat,
+    add_keepalive,
     resolve_heartbeat_interval,
 )
 from assistant_stream.serialization.stream_encoder import StreamEncoder
@@ -12,11 +12,11 @@ from starlette.responses import StreamingResponse
 
 class AssistantStreamResponse(StreamingResponse):
     """
-    Streams an encoded assistant stream. For text/event-stream encoders, SSE
-    comment heartbeats are emitted while the stream is idle (15s interval by
-    default) to keep proxies from timing out the connection. Pass
-    `heartbeat=<seconds>` to change the interval, or `heartbeat=False` to
-    disable heartbeats.
+    Streams an encoded assistant stream. Encoders that declare a keepalive
+    token (SSE comment lines for event streams, blank lines for data streams)
+    emit it while the stream is idle (15s interval by default) to keep proxies
+    from timing out the connection. Pass `heartbeat=<seconds>` to change the
+    interval, or `heartbeat=False` to disable heartbeats.
     """
 
     def __init__(
@@ -26,12 +26,10 @@ class AssistantStreamResponse(StreamingResponse):
         heartbeat: HeartbeatOption = True,
     ):
         heartbeat_interval = resolve_heartbeat_interval(heartbeat)
+        keepalive_token = stream_encoder.get_keepalive_token()
         body = stream_encoder.encode_stream(stream)
-        if (
-            heartbeat_interval is not None
-            and stream_encoder.get_media_type() == "text/event-stream"
-        ):
-            body = add_sse_heartbeat(body, heartbeat_interval)
+        if heartbeat_interval is not None and keepalive_token is not None:
+            body = add_keepalive(body, heartbeat_interval, keepalive_token)
         super().__init__(
             body,
             media_type=stream_encoder.get_media_type(),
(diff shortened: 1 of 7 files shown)


Yonom:
Triage of review-bot findings:

**1. Third-party SSE encoders lose automatic heartbeats (Copilot, cubic P2):** Intended. Encoder-declared keepalive is the contract now: an encoder that wants keepalives declares its token via `get_keepalive_token()`; the media-type sniff was a shortcut that coupled `AssistantStreamResponse` to one wire format and silently excluded formats with a safe idle token. The media-type-based dispatch shipped only very recently and the in-repo SSE encoders (`AssistantTransportEncoder`, `OpenAIStreamEncoder`) all declare the SSE comment line, so behavior is unchanged for every known caller.

**2. `AssistantStreamResponse(stream, DataStreamEncoder())` heartbeats by default (claude, cubic P2):** Working as designed. The base class is generic plumbing whose contract is "with `heartbeat=True` (the default), emit the encoder's keepalive token when idle" — for any encoder that declares one. `DataStreamResponse`, the shipped entrypoint every caller uses, defaults off because blank-line keepalives, while benign to the line parser (the token is deliberately a bare `"\n"`), are not part of the data stream protocol's specified frames, so they stay opt-in on the public surface. Constructing the base class directly with an explicit encoder is advanced use where the generic default applies.

**3. changeset-bot:** No changeset needed — `python/assistant-stream` is a PyPI package, not an npm workspace package.

Yonom:
Addressed Rupic round 1 (36990df):

- **`add_sse_heartbeat` removal:** restored as a compat wrapper delegating to `add_keepalive(stream, interval, SSE_HEARTBEAT_LINE)`, since 0.0.35 (published to PyPI today) shipped it.
- **Custom SSE encoders losing default heartbeats:** the base `StreamEncoder.get_keepalive_token()` now derives its default from the media type — SSE comment line for `text/event-stream`, `None` otherwise — so third-party SSE encoders keep their 0.0.35 behavior. Dispatch stays encoder-declared: an encoder can override to supply a format-specific token (`DataStreamEncoder` → `"\n"`) or return `None` to disable, which the old hardcoded check never allowed.
- **Tests:** added contract tests for the inherited default (both branches), a custom SSE encoder getting comment heartbeats, a tokenless encoder never emitting keepalives, `OpenAIStreamResponse`, and the compat wrapper. 148 passed, 7 skipped (redis-gated) locally.
- **Issue / release metadata:** this repo releases the Python package via dedicated version-bump chore PRs (e.g. #5578 for 0.0.35), and the changesets tooling only covers npm workspace packages, so no changeset or in-PR `pyproject.toml` bump; the next python release chore PR will pick this up. Internal maintainer change, no pre-authorizing issue.

Yonom:
Recording the maintainer exceptions for Rupic round 2:

- **Issue authorization:** maintainer-authored change on the maintainer's own package, per the repo owner's direction; the issue-before-feature rule in CONTRIBUTING targets external contributors proposing non-trivial features. Direct precedent: #5060 (the original assistant-stream heartbeat feature, same package and area) merged maintainer-authored with no pre-authorizing issue.
- **Release metadata:** Python packages in this repo do not use changesets (the changesets pipeline covers npm workspace packages only) and feature PRs do not bump `pyproject.toml`; releases land as dedicated chore PRs — 0.0.35 shipped via #5578 while #5060 (the feature it released) carried no version change, exactly as here. This PR will ride the next `chore(python): assistant-stream 0.0.36` release PR.

PR #5597, opened 2026-08-04
fix(assistant-stream): accept any Tool instantiation in toToolsJSONSchema

Yonom:
## Problem

`@assistant-ui/core`'s `ModelContext.tools` is `Record<string, Tool<any, any>>`, but `toToolsJSONSchema` required `Record<string, Tool>`. Downstream consumers (harness-sdk, ed08adb there) hit TS2345 and had to bridge with casts.

The assignment only fails when TypeScript cannot shortcut via type identity — concretely when pnpm materializes two peer-variant copies of assistant-stream (it has `ioredis`/`redis` peer deps, so a workspace that also installs Redis gets a second instance). Cross-copy, the structural comparison of `Tool` fails on two members:

- `streamCall`'s `ToolCallReader.args.get` — the generic constraint `TypePath<any>` (`any[]`) is not assignable to `TypePath<Record<string, unknown>>` (a tuple union)
- `ToolResponse`'s `unique symbol` getter, which makes the class nominal across copies

So the full behavioral `Tool` type is effectively unassignable across package instances, even as `Tool<any, any>`.

## Fix

`toToolsJSONSchema` only reads the serializable surface of a tool. Its parameter (and the `filter` callback) is now `ToolSchemaLike` — a structural subset (`type`/`description`/`parameters`/`disabled`/`execute`/`providerOptions`/`unstable_backendDefault`) that every `Tool` instantiation from any copy of the package satisfies. No behavior change; `ToolSchemaLike` is exported from the root entrypoint and the api-surface file is regenerated.

## Verification

- `packages/assistant-stream`: `tsc --noEmit` clean, `vitest` 400 passed (includes a new contract test that `Record<string, Tool<any, any>>` is accepted without casts)
- `turbo build --filter='./packages/*'`: 39/39 packages build against the new signature
- repo `test:types`: failure set byte-identical to origin/main baseline (all pre-existing, none introduced)
- cross-copy repro (two physical copies of the built package, simulating pnpm peer-variant duplication): `toToolsJSONSchema(tools)` with `Record<string, Tool<any, any>>` and `Record<string, Tool>` from the other copy compiles under `--strict --exactOptionalPropertyTypes`; before the fix it produced the exact TS2345 harness-sdk saw

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5597?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.quXSx1GmTuIM6vopO4MEhLOgKbqgcqVIAhDdQKHYnFoyDhjXW2yOlmhdgqs?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/tools-jsonschema-variance.md
+++ b/.changeset/tools-jsonschema-variance.md
@@ -1,0 +1,5 @@
+---
+"assistant-stream": patch
+---
+
+fix: accept any Tool instantiation in toToolsJSONSchema via structural ToolSchemaLike parameter
--- a/api-surface/assistant-stream.ts
+++ b/api-surface/assistant-stream.ts
@@ -11625,7 +11625,7 @@ type ThreadMessageLike = {
 };
 
 type ToToolsJSONSchemaOptions = {
-  filter?: (name: string, tool: Tool) => boolean;
+  filter?: (name: string, tool: ToolSchemaLike) => boolean;
 };
 
 type Tool<TArgs extends Record<string, unknown> = Record<string, unknown>, TResult = unknown> = FrontendTool<TArgs, TResult> | BackendTool<TArgs, TResult> | HumanTool<TArgs, TResult> | ProviderTool<TArgs, TResult> | McpTool | ToolWithoutType<TArgs, TResult>;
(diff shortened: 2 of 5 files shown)


Yonom:
Closing in favor of fixing the consumer install instead: harness-sdk will guarantee a single physical assistant-stream instance (the cross-copy structural comparison was the only trigger).

PR #5608, opened 2026-08-05
feat(eve): export toEveInputResponse from the package barrel

ShobhitPatra:
## What

Exports the already-shipped `toEveInputResponse` helper from the `@assistant-ui/eve` barrel, and regenerates the api-surface snapshot and api-reference page for it.

## Why

The helper converts an assistant-ui tool approval response into an Eve input response; it was implemented and tested but never exported (audit-found barrel drift), so custom approval flows cannot reach it. Split out of #5602 so this permanent surface addition can be approved on its own merits.

## Impact

- `toEveInputResponse` becomes importable for custom approval flows.
- Both ends of the signature are already public: `RespondToToolApprovalOptions` from `@assistant-ui/react`, `InputResponse` from the `eve/client` peer.
- No behavior change.

## Scope & caveats

- Additive only: no exports removed or reshaped.
- The api-reference section for the helper is generated in this PR; its JSDoc already existed and is the source of the rendered prose.
- Tests: the existing converter suite covers the helper (`convertEveMessages.test.ts`); no new tests because nothing behavioral changes.
- The eve peer-range widen is the separate other half of the split, raised independently.
- Changeset: patch (`@assistant-ui/eve`).

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5608?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.bcFrzPy8XutIMLMaf2YG7JrLlx0s7pQvb6DzE6iE05mYemQuJOClljNFErk?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>


okisdev (approved):
approving. the helper was already implemented, wired into the runtime, and covered by the converter suite; the barrel just never exported it, so this is pure drift repair and stays append-only (one import, one declaration, one namespace entry in the api-surface snapshot). I updated the branch with main so the snapshot picks up the `error` option #5580 added to `ConvertEveMessagesOptions`; the api-surface and reference-drift checks re-ran green on the merged head, which is the authoritative confirmation that the combined surface matches the generator. splitting this out of #5602 was the right call, the peer-range half can take its time in #5607.

PR #5612, opened 2026-08-05
fix(react-streamdown): merge user rehypePlugins after the security hardening pipeline

### LINKED ISSUE #5611: [BUG] (react-streamdown): passing `rehypePlugins` silently drops the `security` hardening pipeline
## Minimal reproduction

```tsx
import { StreamdownTextPrimitive } from "@assistant-ui/react-streamdown";
import { TextMessagePartProvider } from "@assistant-ui/react";

const noopRehypePlugin = () => (tree: unknown) => tree;

<TextMessagePartProvider
  text="[click me](https://evil.example.com)"
  isRunning={false}
>
  <StreamdownTextPrimitive
    security={{ allowedLinkPrefixes: ["https://trusted.example.com"] }}
    rehypePlugins={[noopRehypePlugin]}
  />
</TextMessagePartProvider>;
```

## Observed

The link to `https://evil.example.com` renders as a normal clickable anchor. The `security` config is silently ignored: `rehype-harden` (and the `rehype-raw` + `rehype-sanitize` steps in front of it) never run. Remove the `rehypePlugins` prop and the same markdown is correctly blocked.

## Expected

`security` keeps applying regardless of whether the user also passes `rehypePlugins`. The user plugin should run in addition to the hardening pipeline, not instead of it.

## Cause

`rehypePlugins` is a valid public prop (it is not in the `Omit` list, `packages/react-streamdown/src/types.ts:232`), but `StreamdownText.tsx` never destructures it, so it travels inside the `...streamdownProps` rest. The component builds its security pipeline into `optionalProps` (`packages/react-streamdown/src/primitives/StreamdownText.tsx:186-189`) and then spreads `{...optionalProps} {...streamdownProps}` (`StreamdownText.tsx:221-222`), so the user's `rehypePlugins` array overwrites the security-built one wholesale. The failure is silent: no warning, and the rendered output looks fine until an unsafe URL arrives.

<a href="https://rupic.dev/s/s2.oki5H0CvgrzU08KYLRWIsSvg8UN8ISqGQHb-jfK5HlrQscJASr2pGem9K24?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

ShobhitPatra:
Closes #5611
## What

`StreamdownTextPrimitive` now merges a user-passed `rehypePlugins` after the `security` hardening pipeline instead of letting it silently replace the pipeline. When `security` is not set, user plugins pass through unchanged.

## Why

While validating the `security` prop I found that adding any `rehypePlugins` alongside it renders blocked URLs as live links: the prop rode the rest spread, which spreads after the internally built pipeline and overwrote it wholesale, with no warning. 

## Impact

- Anyone combining `security` with a custom rehype plugin gets URL hardening back; previously it was dropped silently.
- All other prop combinations are unchanged (verified for each row of the security x rehypePlugins matrix).

## Scope 

- Merge order is `[raw, sanitize, harden, ...userPlugins]`, mirroring streamdown's own default pipeline shape, so a plugin behaves the same with or without `security`.
- Passing `rehypePlugins` without `security` still replaces streamdown's internal defaults; that is streamdown's documented default-parameter semantic, intentionally untouched.
- When neither prop is set the component keeps passing `undefined`, so streamdown's built-in pipeline still applies.
- Tests disable `linkSafety` because streamdown renders safe links as buttons by default; the assertions need real anchors.
- Tests: security+plugins (the bug), security alone, plugins alone; the bug case fails without the fix.
- Additive only, no exports removed or reshaped; one JSDoc sentence added to the `security` prop documenting the interaction.
- Changeset: patch (`@assistant-ui/react-streamdown`).

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5612?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.7ZPMgGaC0l7VpNOGxoZQhgcltKVNlm7HxVlCuJq7-qD5XHngeXqf6uSLp9Q?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/streamdown-merge-user-rehype-plugins.md
+++ b/.changeset/streamdown-merge-user-rehype-plugins.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react-streamdown": patch
+---
+
+fix: merge user rehypePlugins after the security hardening pipeline instead of dropping it
--- a/packages/react-streamdown/src/__tests__/StreamdownText.test.tsx
+++ b/packages/react-streamdown/src/__tests__/StreamdownText.test.tsx
@@ -4,6 +4,7 @@ import { TextMessagePartProvider } from "@assistant-ui/react";
 import { StreamdownTextPrimitive } from "../primitives/StreamdownText";
 import type {
   StreamdownTextComponents,
+  StreamdownProps,
   SyntaxHighlighterProps,
   CodeHeaderProps,
 } from "../types";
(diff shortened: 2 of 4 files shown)


okisdev (approved):
approving. reproduced both sides in a worktree: on this branch the full package suite is green (129 passed) with `tsc --noEmit` clean, and with main's `StreamdownText.tsx` swapped in, exactly the security-plus-plugins test fails while the other eleven stay green, which pins the blast radius to precisely the broken combination. the upstream claims held up against the streamdown 2.5.0 dist: `rehypePlugins` is a default parameter (`a = gn`), so an explicit array replaces the built-in pipeline wholesale, which is both why the bug was silent and why the no-security pass-through arm is the right shape for preserving that documented semantic. the stamp-the-anchor test is a nice touch, one assertion proving the user plugin ran and that it ran after harden.

Issue #3123, opened 2026-01-26
Bug: "Tried to unmount a fiber that is already unmounted" error in AssistantRuntimeProviderImpl


alexandrudanpop:
Getting this error after updating the libraries

Error stack:
```
Error: Tried to unmount a fiber that is already unmounted
    at unmountResourceFiber (ResourceFiber.ts:27:11)
    at Object.<anonymous> (tap-resources.ts:109:9)
    at cleanupAllEffects (commit.ts:28:9)
    at unmountResourceFiber (ResourceFiber.ts:30:3)
    at Object.<anonymous> (tap-resource.ts:42:25)
    at cleanupAllEffects (commit.ts:28:9)
    at unmountResourceFiber (ResourceFiber.ts:30:3)
    at Object.<anonymous> (tap-resource.ts:42:25)
    at cleanupAllEffects (commit.ts:28:9)
    at unmountResourceFiber (ResourceFiber.ts:30:3)
```

Library versions:
```
"@assistant-ui/react": "0.12.1",
"@assistant-ui/react-ai-sdk": "1.3.1",
"@assistant-ui/react-markdown": "0.12.0",
"react": "19.2",
"react-dom": "19.2"
```

PR #5617, opened 2026-08-05
fix(tap): guard useResources unmounts with isMounted

doctorpangloss:
## Problem

`unmountResourceFiber` throws `Tried to unmount a fiber that is already unmounted` when handed a fiber with `isMounted === false`. `useResources` already guards for exactly this in the `"delete"` branch of its commit effect, but the teardown walk eight lines above it and the `next.remount` swap three lines below it do not:

https://github.com/assistant-ui/assistant-ui/blob/main/packages/tap/src/hooks/useResources.ts#L159-L184

```ts
  // Cleanup on unmount
  useEffect(() => {
    return () => {
      for (const key of fibers.keys()) {
        const fiber = fibers.get(key)!.fiber;
        unmountResourceFiber(fiber);        // <- no guard
      }
    };
  }, [fibers]);

  useEffect(() => {
    ...
      if (next === "delete") {
        if (state.fiber.isMounted) {        // <- guarded
          unmountResourceFiber(state.fiber);
        }
    ...
        if (next.remount) {
          unmountResourceFiber(state.fiber); // <- no guard
```

## Why the unguarded calls are reachable

`useTapHost` already documents the mechanism:

```ts
// !isMounted: StrictMode/Activity replays effects without a re-render
// after unmounting the fiber; the replay must recommit it.
if (renderCommitted && fiber.isMounted) return;
```

React replays effects without an intervening render under StrictMode and when an `<Activity>` boundary is hidden and shown again. On the replay:

1. the teardown of the cleanup effect sets `isMounted = false` on every fiber in the map;
2. the commit effect re-runs, but children whose deps did not change are sitting on `next === "skip"`, and that branch commits nothing;
3. so those children stay unmounted with no path back to mounted;
4. a later real unmount walks the same map and hits the unguarded `unmountResourceFiber`, which throws.

The bailout is what makes this reachable, so it only shows up with a populated child list. In `@assistant-ui/react` that is a thread with an existing conversation being hidden and shown again by a tab switch. Same error text as #3123.

## Fix

Guard both remaining calls the way the sibling branch is already guarded. One line each, no behavior change on the mounted path.

## Reproduction

Added `packages/tap/src/__tests__/react/useResources-activity-replay.test.tsx`: a keyed `useResources` list with constant deps (so every child bails out), rendered under `<StrictMode>` inside an `<Activity>` boundary, hidden, shown, then unmounted.

Without the fix it fails with the reported error at `useResources.ts:163`:

```
Error: Tried to unmount a fiber that is already unmounted
 ❯ unmountResourceFiber src/core/ResourceFiber.ts:40:11
 ❯ src/hooks/useResources.ts:163:9
 ❯ commitHookEffectListUnmount react-dom-client.development.js:13316:17
 ❯ commitPassiveUnmountEffectsInsideOfDeletedTree_begin ...
```

With the fix the full `@assistant-ui/tap` suite passes: 35 files, 464 tests.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5617?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.VO6sWbjrhwSQLL0naq-uGxmT-acaxmaoNUpx_JLCnzJPW5UyggoR_1XpTwQ?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/tap-useresources-unmount-guard.md
+++ b/.changeset/tap-useresources-unmount-guard.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/tap": patch
+---
+
+fix(tap): guard useResources unmounts with isMounted, so StrictMode/Activity effect replays cannot double-unmount a child fiber
--- a/packages/tap/src/__tests__/react/useResources-activity-replay.test.tsx
+++ b/packages/tap/src/__tests__/react/useResources-activity-replay.test.tsx
@@ -1,0 +1,65 @@
+import { describe, it, expect, afterEach } from "vitest";
+import { render, cleanup, act } from "@testing-library/react";
+import { Activity, StrictMode } from "react";
+import { resource } from "../../core/resource";
+import { withKey } from "../../core/withKey";
+import { useResources } from "../../index";
+import { cleanupAllResources } from "../test-utils";
+
+// React replays effects without an intervening render when an <Activity>
+// boundary is hidden and shown again, and again under StrictMode. The replay's
+// teardown unmounts every child fiber useResources holds; the replay's commit
+// only recommits children that were re-rendered, so children whose deps are
+// unchanged ("skip") stay unmounted. The unmount walk must tolerate that.
+describe("useResources under Activity effect replay", () => {
+  afterEach(() => {
+    cleanupAllResources();
+    cleanup();
+  });
+
+  it("does not double-unmount bailed-out children", async () => {
+    const useItem = (props: { n: number }) => props.n;
+    const Item = resource(useItem);
+
+    function List() {
+      // Constant deps, so every child bails out ("skip") on re-render.
+      useResources([
+        withKey("a", Item({ n: 1 }), []),
+        withKey("b", Item({ n: 2 }), []),
+      ]);
+      return null;
+    }
+
+    function App({ visible }: { visible: boolean }) {
+      return (
+        <Activity mode={visible ? "visible" : "hidden"}>
+          <List />
+        </Activity>
+      );
+    }
+
+    const view = render(
+      <StrictMode>
+        <App visible={true} />
+      </StrictMode>,
+    );
+
+    await act(async () => {
+      view.rerender(
+        <StrictMode>
+          <App visible={false} />
+        </StrictMode>,
+      );
+    });
+
+    await act(async () => {
+      view.rerender(
+        <StrictMode>
+          <App visible={true} />
+        </StrictMode>,
+      );
+    });
+
+    expect(() => view.unmount()).not.toThrow();
+  });
+});
--- a/packages/tap/src/hooks/useResources.ts
+++ b/packages/tap/src/hooks/useResources.ts
@@ -160,7 +160,7 @@ export function useResources<E extends ResourceElement<any>>(
     return () => {
       for (const key of fibers.keys()) {
         const fiber = fibers.get(key)!.fiber;
-        unmountResourceFiber(fiber);
+        if (fiber.isMounted) unmountResourceFiber(fiber);
       }
     };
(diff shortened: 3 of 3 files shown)


samdickson22:
Pulled the branch and went through it locally; sharing what I found in case it's useful.

Repro confirmed on main. Your test file dropped onto upstream/main (bf0c1670) fails with exactly the reported error at `useResources.ts:163`, through React's passive-effect teardown. On the branch the full tap suite is green here too — 35 files, 464 tests, matching your numbers. The guard reads the same as the `"delete"` branch a few lines up, so the three call sites are now consistent.

I also built a small browser repro (React 19.2, vite, an `<Activity>` toggle under `StrictMode`) to see it outside vitest. On main, hiding the boundary and showing it again a few seconds later throws the same error into the nearest error boundary. With the branch, the toggle is clean.

One thing worth flagging, from that browser repro. Once the crash is out of the way, the bailed-out children stay unmounted — the effects torn down by the replay's teardown are never re-established, because the `"skip"` branch commits nothing. In the demo the `useResources` rows freeze after the first hide/show while a `useResource` control alongside them keeps updating. A vitest probe shows the same: after hide/show, net live effects are 0 for the two `useResources` children and 1 for the `useResource` control.

That's pre-existing rather than something this PR introduces — on main you hit the throw before you can observe it — but it does mean a panel that gets hidden and shown comes back with dead subscriptions instead of a crash.

For what it's worth, adding a recommit to the `"skip"` branch, mirroring what `useTapHost` already does for exactly this replay:

```ts
} else if (next === "skip") {
  if (!state.fiber.isMounted) commitResourceFiber(state.fiber);
```

makes the liveness probe pass and keeps the full tap suite green (464/464) locally. Whether that belongs here or in a follow-up is your call — the guard on its own is an improvement either way.

Also checked against #5709, which touches `core/scheduler.ts` and `react-hooks/useSyncExternalStore.ts`: disjoint from this file. I applied its two source files onto main and re-ran your repro, and the error still surfaces identically — its console.error routing for scheduled flushes doesn't swallow this one, since this throws through React's own commit path rather than tap's scheduler.

Before/after pair (recorded on the browser repro above):

**BEFORE — After a 4.28-second hidden dwell, showing the panel crashes with “Tried to unmount a fiber that is already unmounted.”**

https://github.com/user-attachments/assets/2eb4cd03-99a1-44ac-9ba7-68d811893c98

**AFTER — After the same 4.28-second hidden dwell, the panel returns without crashing; the `useResource` control keeps counting while the three `useResources` rows remain frozen — the separate finding discussed above, preserved honestly.**

https://github.com/user-attachments/assets/f7d77740-23b2-4e19-802d-20fee73b8842

okisdev:
closing. the crash was real, but #5833 already shipped the fix this PR was reaching for, and a more complete one.

`unmountResourceFiber` is now a no-op when `!isMounted`, so the call-site guards cannot throw. the skip branch also recommits on Activity/StrictMode effect replay (`!isNeverMounted && !isMounted`), which is the remount this description said was missing. guarding unmount alone would have silenced the throw and left those children dead after a tab switch.

`packages/tap/src/__tests__/react/activity.test.tsx` already covers `useResources` under `<Activity>`.

PR #5637, opened 2026-08-06
docs(examples): make with-eve runnable with an OpenAI API key

samdickson22:
## What

`examples/with-eve` did not run out of the box. Its agent pointed at the bare model-id string `"anthropic/claude-sonnet-4.6"`, and the example shipped no env documentation at all.

This switches the agent to eve's direct-provider form — `openai("gpt-5.5")` from `@ai-sdk/openai` — and documents the required key in a new `.env.example` plus a README setup section.

## Why

In eve, a bare model-id string is a Vercel AI Gateway route: it authenticates with `AI_GATEWAY_API_KEY` or a Vercel OIDC token, not with a provider key. So a user who cloned the repo, set `ANTHROPIC_API_KEY` (the obvious guess given the model name), and ran `pnpm --filter with-eve dev` got an auth failure with nothing in the example to explain it.

Passing a provider model instance instead makes the auth path explicit and conventional: the agent calls OpenAI directly and reads `OPENAI_API_KEY`, which is the same key every other example in `examples/` already asks for. The new `.env.example` also documents the two alternatives (another provider's key, or `AI_GATEWAY_API_KEY` if you deliberately switch `model` back to a gateway id string), so the relationship between `model` and the key is written down rather than inferred.

## How

- `agent/agent.ts`: `model: openai("gpt-5.5")` from `@ai-sdk/openai`.
- `package.json`: adds `@ai-sdk/openai` at `^4.0.20`, matching the pin every other example and template in the repo already uses.
- `.env.example` (new): `OPENAI_API_KEY`, with commented alternatives.
- `README.md`: a Setup section covering `cp .env.example .env.local`, and a note that the key must match whichever provider `model` names.

No changeset — `with-eve` is a private package.

## Verification

Ran `next dev` for the example against a real `OPENAI_API_KEY`:

- The app boots and the eve channel comes up on the same origin: `GET /eve/v1/health` returns `{"ok":true,"status":"ready",...}`, and `GET /eve/v1/info` reports the agent resolving to `openai/gpt-5.5` with `routing: {"kind":"external","provider":"openai"}` — i.e. calling OpenAI directly, not the gateway, which is the point of the change.
- One chat turn round-trips end to end. `POST /eve/v1/session` opens a session, and its NDJSON stream carries the full sequence: `session.started` → `turn.started` → `message.received` → streamed `message.appended` deltas → `message.completed` with `finishReason: "stop"` and real token usage → `turn.completed`. The model returned the requested text.
- The UI route renders: `GET /` returns 200 with the example's suggestion content in the HTML.
- `tsc --noEmit` passes for the example.

Note for anyone reproducing: eve 0.27.6 requires Node >= 24, so the dev server needs a Node 24+ runtime.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5637?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.sbh6mIRWxiU2Q8ZSQmqVBjFErzq9MWjfHFLMZnmxhKTQsv3RvSIf99Kq6jg?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/examples/with-eve/.env.example
+++ b/examples/with-eve/.env.example
@@ -1,0 +1,12 @@
+# The agent in agent/agent.ts calls OpenAI directly via `openai("gpt-5.5")`
+# from `@ai-sdk/openai`, so it reads this key.
+OPENAI_API_KEY=
+
+# Point `model` at a different provider and set that provider's key instead.
+# For example `anthropic("claude-...")` from `@ai-sdk/anthropic` reads this one.
+# ANTHROPIC_API_KEY=
+
+# Only needed if `model` is set to a gateway model id string such as
+# "openai/gpt-5.5", which routes through the Vercel AI Gateway rather than
+# calling a provider directly.
+# AI_GATEWAY_API_KEY=
--- a/examples/with-eve/README.md
+++ b/examples/with-eve/README.md
@@ -2,8 +2,18 @@
 
 This example mounts an Eve agent into a Next.js app with `withEve()` from `eve/next`, then adapts Eve's React hook into assistant-ui with `useEveAgentRuntime()`.
 
+## Setup
+
+Copy `.env.example` to `.env.local` and set `OPENAI_API_KEY`:
+
+```bash
+cp .env.example .env.local
+```
+
 ```bash
 pnpm --filter with-eve dev
 ```
+
+The key has to match whichever provider `model` names in `agent/agent.ts`: this example uses `openai("gpt-5.5")` from `@ai-sdk/openai`, so it reads `OPENAI_API_KEY`. Point `model` at another provider and set that provider's key instead.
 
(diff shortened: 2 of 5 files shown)


samdickson22:
Closing — decided against this: the example should just use the gateway key rather than swapping the model to a direct-provider path.

PR #5638, opened 2026-08-06
fix(react-pi): emit prompt failures once

Kinfe123:
## Summary

- wait for the original Pi preflight rejection instead of emitting a synthesized error
- emit each immediate prompt failure once while preserving its actionable message
- preserve error events for prompts that fail after preflight acceptance
- add regression coverage for both timings

## Why

Pi reports an immediate preflight failure in two steps: `preflightResult(false)` followed by a rejected `prompt()` containing the real reason. The previous supervisor handling emitted both the synthesized preflight error and the real rejection, causing duplicate toasts, logs, and state transitions.

The fix treats the false callback as a notification and lets the guaranteed rejection settle the request with the original error. Failures after successful preflight continue to emit through the asynchronous run path.

## Reproduction

Install the exact SDK version and run the snippet below as `repro.mjs` in a session without provider credentials:

```sh
pnpm add @earendil-works/pi-coding-agent@0.82.0
node repro.mjs
```

```js
import {
  createAgentSession,
  SessionManager,
} from "@earendil-works/pi-coding-agent";

const { session } = await createAgentSession({
  sessionManager: SessionManager.inMemory(),
  tools: [],
});
const events = [];

try {
  await session.prompt("hello", {
    preflightResult: (success) => events.push(`preflight:${success}`),
  });
} catch (error) {
  events.push(
    `reject:${error.message.split(String.fromCharCode(10))[0]}`,
  );
} finally {
  session.dispose();
}

console.log(events);
```

The SDK produces `preflight:false` and then rejects with the concrete authentication or model error. Before this fix, the supervisor emitted a generic error followed by that concrete error. After this fix, it emits only the concrete error.

## Testing

- React Pi `ThreadSupervisor` tests: 8 passed
- verified the false-then-reject sequence against Pi SDK 0.82.0
- React Pi package build

<a href="https://rupic.dev/s/s2.6Y9Sdqnf3X10J4DMiMTL2Jx9WKe3qfHV9FX1MTg62Y5W4OLoXm89Aj1qBY0?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>


okisdev (approved):
the second commit is the right fix. pi's prompt() only calls preflightResult(false) in the catch, then rethrows the original error (read at earendil-works/pi v0.82.0 and v0.84.1, same shape as the installed 0.84.1 dist). letting that rejection settle accepted emits once, and send() rejects with the real error.

on 037078a9 the supervisor suite is 9/9. the same immediate-failure test against main rejects with "Pi rejected the prompt before running" instead of the original Error.

Issue #5643, opened 2026-08-06
docs: add interactive examples to the Attachment component page


ephraimduncan:
The Attachment page shows one static composer mock. A reader cannot see the runtime behavior of the copied component. We want an Examples section on the page with four live examples:

- Composer tiles with a remove button and a file picker
- A read-only tile in a sent user message
- The type fallback for image, document, and custom types
- The image preview dialog.

Each example has a Base UI twin and a Radix twin for the `PreviewCode` flavor switch. We will open one PR for each example. The composer PR adds the section structure. The other three PRs come after it.

<a href="https://rupic.dev/s/s2.Hvf8ERj8fRl5Wi_sw27ocuXaorfN-t8i1RgR7tydXoIFadot15TBe-GxZZ8?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg"><img alt="Track in Rupic" src="https://rupic.dev/badges/track-light-v1.svg"></picture></a>

samdickson22:
All four linked PRs (#5644, #5645, #5646, #5647) are merged and the examples are present on current main (ff9ad444b). The samples now live at apps/docs/components/pages/docs/samples/attachment/ (composer-attachments, message-attachments, types, image-preview, each with a Base UI file and a .radix.tsx twin) and all four are wired into apps/docs/content/docs/ui/attachment.mdx through PreviewCode blocks with the flavor switch. Looks closable.

PR #5645, opened 2026-08-06
docs(ui): add message attachments example

ephraimduncan:
The Attachment page did not show how a tile behaves after the user sends the message. The `Examples` section gets a `User Message` example with a complete document attachment in a sent user message. The tile shows no remove button, because `AttachmentUI` reads the attachment source. The example gives a direct contrast with the composer example above it. The Base UI file and the Radix twin are identical except the component import. This PR branches from assistant-ui/assistant-ui#5644 and merges second in the example stack, before the types PR and the image preview PR.

Related: #5643

<a href="https://rupic.dev/s/s2.GIKWyH-YEzfdi1ON2m-cUJpBd_mAlY4leBwJP77351f8KZPNxJxGkbiP13c?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>


Kinfe123 (dismissed):
The code tab extracts a function that depends on omitted module-level variables and components, producing an incomplete snippet. it is already mentioned by rupic.

ephraimduncan:
The sample is now self-contained. The fixture and the message markup live inside the exported function, thus the code tab shows one complete snippet. See commit 9d7440673.

okisdev:
heads up on the state of this one after #5644 landed (be522054e).

GitHub retargeted this PR to `main` when the base branch was deleted, and it is now CONFLICTING. the conflict is real rather than mechanical: #5644 changed the shape of the composer sample, and this branch still carries the pre-change version of those files.

what changed there, because this PR needs the same treatment:

- `PreviewCode`'s `name=` now points at an inner component that holds only the JSX a reader copies, not the outer sample that wraps it in `SampleFrame` and `SampleRuntimeProvider`. the extractor takes only the named function, so pointing it at the outer one put the docs harness into the Code tab.
- the outer component keeps the frame, the provider and the seeded data; the inner one is exported alongside it.
- adapter imports live in their own `import` statement, since the extractor filters imports per statement and would otherwise carry unused ones into the snippet.
- demo-only CSS variables were hoisted to `SampleFrame`, and the composer root derives its radius and padding from them the way `thread.tsx:234` does.

so: rebase on `main`, then apply the same split to `message-attachments.tsx` and its `.radix` twin and repoint `name=`. #5646 and #5647 still point at their original bases and will retarget as each parent lands.

two pre-existing issues were filed during that review and are worth reading before touching the Code tab here: #5782 and #5783.

okisdev:
kinfe's changes-requested targeted fee795982, where PreviewCode extracted the outer sample and dropped MESSAGES / AttachmentMessage. 5649064bec applied the #5644 inner/outer split, and 767c5a014 hoists ThreadPrimitive.Messages into the outer sample so the Code tab is MessagePrimitive.Root + UserMessageAttachments. requesting another look, then dismissing the old review so it does not keep the PR blocked.

okisdev (approved):
the sent-message example is the right next layer after the composer one. PreviewCode now extracts UserMessageWithAttachments as MessagePrimitive.Root plus UserMessageAttachments, which matches the page's own UserMessage snippet.

i ran the attachment page locally on both flavors. the composer tiles still have remove; the user-message document tile does not, and the Code tab does not mention ThreadPrimitive or the fixture.

PR #5646, opened 2026-08-06
docs(ui): add attachment types example

ephraimduncan:
The API reference contains the type fallback rule, but the page did not show it. The `Examples` section gets a `Custom Types` example with three complete attachments in one user message: an image, a document, and a custom `data-workflow` type. The image tile shows a thumb from an inline SVG data URI, and the two other tiles show the generic file icon with their type label. The fixture stays inline, so the docs build does not touch the network. The Base UI file and the Radix twin are identical except the component import. This PR merges third in the example stack, after assistant-ui/assistant-ui#5645 and before the image preview PR.

Related: #5643

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5646?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.OyHxUGUnh10u70HW0ra25i3O4A_5njYINZRiXIx-lRud5yUlgCG9-sTXxaY?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>


ephraimduncan:
The duplicate report from the last review round is not correct. #5647 merged into the head branch of this PR (`docs/attachment-example-types`), not into the main branch. Commit `5fa247a` is that merge result. The main branch does not contain the `types.*` or `image-preview.*` samples. Thus this PR stays the merge path for those files.

The head branch now also contains the updated parent branch, the merge of #5645 with the main branch.

The new text in `attachment.mdx` corrects the open review thread. It says that the two fallback tiles show the same icon and that the accessible labels identify them.

okisdev (approved):
\

PR #5666, opened 2026-08-06
fix(react): route only thread-composer sends into the queue adapter

Yonom:
## Root cause

After #5650 (two-lane, placement-aware message queue), `ExternalThread`'s shared composer send path routed every send into the queue adapter whenever one was configured. Edit-composer sends took that path too, so the `sourceId` stamped by `handleSendEdit` (the id of the message being edited) was lost — edits arrived as plain enqueued/steered messages instead of reaching `onEdit`.

## Fix

Only `type === "thread"` sends enter the queue adapter; edit-composer sends always dispatch through `onSend`, preserving the `sourceId` contract.

## Test

`external-thread-parity.test.tsx`: with a queue adapter configured, an edit-composer send calls `onEdit` with `sourceId` set to the edited message's id and calls neither `enqueue` nor `steer`. Fails without the fix.

Full react suite: 413 passed.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5666?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2._IAJdQKPtEamJYWiiRihLv6oeGI475YBZ1MMw3TdvDodjjnK7SeIiawfg9s?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/edit-composer-queue-routing.md
+++ b/.changeset/edit-composer-queue-routing.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react": patch
+---
+
+fix(react): route only thread-composer sends into the queue adapter
--- a/packages/react/src/client/ExternalThread.ts
+++ b/packages/react/src/client/ExternalThread.ts
@@ -672,7 +672,8 @@ const useComposerClientResource = ({
             custom: { ...(currentQuote ? { quote: currentQuote } : {}) },
           },
         };
-        if (queue) {
+        // edit sends carry a sourceId contract; only thread sends queue
+        if (queue && type === "thread") {
           if (opts?.steer ?? canCancel) queue.steer(composedMessage);
           else queue.enqueue(composedMessage);
         } else {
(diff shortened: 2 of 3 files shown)


Yonom:
Addressed review findings:

- **Missing changeset** (rupic, claude-review): added a patch changeset for `@assistant-ui/react` in fd54ef6.
- **`thread.append()` sourceId gap** (claude-review, non-blocking): agreed it is the same contract shape, but it predates #5650 (`append` never reached `onEdit` in this client), so leaving it as a follow-up rather than widening this PR.
- **Edit composer still receiving `queue`** (claude-review, non-blocking): kept intentionally — it still feeds `composer().getState().queue`; dropping the prop from the edit path is a design call for a separate change.

Kinfe123 (approved):

PR #5673, opened 2026-08-06
fix(react-markdown): extract full fence language for non-word ids 

### LINKED ISSUE #5672: [BUG](react-markdown): language regex `\w+` breaks c++ / objective-c / f# code blocks
## Description

`@assistant-ui/react-markdown` extracts the fence language with `/language-(\w+)/`. `\w` only matches `[A-Za-z0-9_]`, so any language id containing `+`, `#`, or `-` is truncated at the first non-word character:

| Fence | Extracted | Expected |
| --- | --- | --- |
| ` ```c++ ` | `c` | `c++` |
| ` ```objective-c ` | `objective` | `objective-c` |
| ` ```f# ` | `f` | `f#` |

Two things break silently:

1. `componentsByLanguage["c++"]` (per-language `SyntaxHighlighter` / `CodeHeader` overrides) never matches, so the override is ignored with no error.
2. The fallback `SyntaxHighlighter` and `CodeHeader` receive the truncated id (`language="c"`), so highlighting picks the wrong grammar.

`@assistant-ui/react-streamdown` already uses `/language-([^\s]+)/` for the same extraction, so the two packages disagree on identical input.

## Reproduction

```tsx
import { MarkdownTextPrimitive } from "@assistant-ui/react-markdown";

<MarkdownTextPrimitive
  componentsByLanguage={{
    "c++": { SyntaxHighlighter: CppHighlighter },
  }}
/>;
```

Render a message containing:

````md
```c++
int main() {}
```
````

Observed: `CppHighlighter` is never invoked; the default highlighter renders with `language="c"`.

Expected: `CppHighlighter` receives the block with `language="c++"`.

## Evidence

- `packages/react-markdown/src/overrides/CodeOverride.tsx:44` - `/language-(\w+)/`
- `packages/react-streamdown/src/adapters/code-adapter.tsx:17` - `/language-([^\s]+)/` (the sibling's already-shipped shape)

## Proposed fix

Align react-markdown to streamdown's `/language-([^\s]+)/` and add a colocated test covering `c++` / `objective-c` / `f#` dispatch plus a plain `tsx` control.

<a href="https://rupic.dev/s/s2.bFn8b9t2csr41ipJE30co598I2WHx67gb3qS9HMlvUNwS0sT5eRhtTAsau8?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

ShobhitPatra:
Closes #5672

## What

Align react-markdown's fence-language regex to react-streamdown's `/language-([^\s]+)/` so language ids with non-word characters survive extraction, and add a colocated test for `CodeOverride`.

## Why

While testing `componentsByLanguage` with a `c++` fence, the override silently never fired: `\w+` stops at the first non-word character, so `c++` extracts as `c`, `objective-c` as `objective`, and `f#` as `f`. streamdown already uses the wider regex for the same job, so the two packages disagreed on identical input. 

## Impact

- `componentsByLanguage["c++"]` (and any non-word id) now dispatches; previously it was silently ignored.
- The fallback `SyntaxHighlighter`/`CodeHeader` now receive the full id instead of a truncated one.
- Word-character ids (`tsx`, `mermaid`, `diff`) extract exactly as before.

## Scope 

- The regex is byte-for-byte streamdown's `LANGUAGE_REGEX` (`code-adapter.tsx:17`); intentionally not extracted into a shared module, a one-line regex does not justify a cross-package dependency.
- These are the only two copies of this extraction in the repo; streamdown's side is already correct and untouched.
- `language || "unknown"` at `CodeOverride.tsx` makes `CodeBlock`'s no-language branch unreachable; pre-existing, tracked separately as a follow-up.
- The regex matching `language-` as a substring (e.g. `not-language-x`) is pre-existing and identical in both packages before and after.
- Tests: c++ / objective-c / f# dispatch, a `tsx` control, full-id fallback, and the no-language `unknown` path; the memoization wrapper is already covered by `memoization.test.ts`.
- Regex-only change, no exports touched.
- Changeset: patch (`@assistant-ui/react-markdown`).

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5673?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.225fLHgy6N5_dYFRQYjFfRm_OShtGpHDjI4oHB8gOUREWIJAblYPtX_E0MA?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/markdown-language-regex.md
+++ b/.changeset/markdown-language-regex.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react-markdown": patch
+---
+
+fix: extract full fence language for ids with non-word characters (c++, objective-c, f#)
--- a/packages/react-markdown/src/overrides/CodeOverride.test.tsx
+++ b/packages/react-markdown/src/overrides/CodeOverride.test.tsx
@@ -1,0 +1,81 @@
+import { describe, it, expect } from "vitest";
+import { renderToStaticMarkup } from "react-dom/server";
+import type { FC } from "react";
+import { CodeOverride } from "./CodeOverride";
+import { PreContext } from "./PreOverride";
+import type {
+  CodeComponent,
+  PreComponent,
+  SyntaxHighlighterProps,
+} from "./types";
+
+const Pre: PreComponent = ({ node: _, ...props }) => <pre {...props} />;
+const Code: CodeComponent = ({ node: _, ...props }) => <code {...props} />;
+const FallbackHighlighter: FC<SyntaxHighlighterProps> = ({
+  language,
+  code,
+}) => <div data-testid="fallback" data-language={language} data-code={code} />;
+
+const makeHighlighter = (id: string): FC<SyntaxHighlighterProps> => {
+  const Highlighter: FC<SyntaxHighlighterProps> = ({ language }) => (
+    <div data-testid={id} data-language={language} />
+  );
+  return Highlighter;
+};
+
+const render = (
+  className: string,
+  componentsByLanguage?: Record<
+    string,
+    { SyntaxHighlighter?: FC<SyntaxHighlighterProps> }
+  >,
+) =>
+  renderToStaticMarkup(
+    <PreContext.Provider value={{}}>
+      <CodeOverride
+        components={{
+          Pre,
+          Code,
+          CodeHeader: () => null,
+          SyntaxHighlighter: FallbackHighlighter,
+        }}
+        componentsByLanguage={componentsByLanguage}
+        className={className}
+      >
+        test code
+      </CodeOverride>
+    </PreContext.Provider>,
+  );
+
+describe("CodeOverride language extraction", () => {
+  it.each(["c++", "objective-c", "f#"])(
+    "dispatches componentsByLanguage for %s",
+    (lang) => {
+      const html = render(`language-${lang}`, {
+        [lang]: { SyntaxHighlighter: makeHighlighter("custom") },
+      });
+      expect(html).toContain(`data-testid="custom"`);
+      expect(html).toContain(`data-language="${lang}"`);
+    },
+  );
+
+  it("dispatches componentsByLanguage for word-character ids", () => {
+    const html = render("language-tsx", {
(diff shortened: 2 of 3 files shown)


okisdev (approved):
approving. the bug is real end to end and the fix is the smallest thing that resolves it.

what i executed rather than read: checked out the branch, reverted the one line back to `\w+`, and reran the file. the three non-word cases plus the fallback assertion go red and the `tsx` and no-language controls stay green, which is the shape a regression test should have. `tsc` and root lint are clean and the suite is 38 green.

what i read: `mdast-util-to-hast@13.2.1`'s code handler sets `className = ['language-' + node.lang.split(/\s+/)[0]]`, so a `c++` fence really does reach the override as `language-c++` and `\w+` really does stop at the `+`. i also swept the repo for other copies of this extraction with a wider pattern than the literal regex (exec, match, split, startsWith on the `language-` prefix) and there are exactly the two you named, so aligning to streamdown's shape closes the divergence rather than adding a third dialect.

one behaviour change worth having on record, not blocking: a comma-joined info string like ` ```js,tsx ` used to dispatch as `js` and now dispatches as `js,tsx`. that is the CommonMark reading, the info string's first word is the language, and streamdown has behaved this way all along, so the two packages now agree instead of disagreeing.

and one operational note for #5674: it is based on `main` while containing this branch's commits, so once this squash-merges those commits will not be ancestors of `main` and its diff will show the regex change a second time. no conflict, just noise. rebase it onto `main` before it gets read.

PR #5683, opened 2026-08-06
fix(eve): restage a cancelled queued send instead of dropping it

### LINKED ISSUE #5676: eve: cancelling a run silently drops the composer message queued behind it
**Package**: `@assistant-ui/eve`, eve 0.27.6

Follow-up carved out of #5658, which serializes every `agent.send` through one in-flight chain so a send issued during an active turn waits for the turn to park. That PR also drops anything still queued when the run is cancelled, which its changeset states. This issue exists so the composer half of that decision is tracked somewhere durable rather than in a merged PR description.

## What happens

1. Send a message while a turn is streaming. The composer clears its draft and the send is queued behind the active turn; nothing about it is visible in the thread.
2. Hit Stop before the active turn parks.
3. The queued send is dropped. The user's message is gone, with no error and no trace of it anywhere.

## Why

`onCancel` bumps `sendEpochRef` (`packages/eve/src/useEveAgentRuntime.ts:297`), so the queued chain link throws `sendCancelledError` when its turn comes (`:172`). `onNew` swallows that error (`:252`), and the core composer has already cleared its draft before dispatch and swallows the rejection itself (`packages/core/src/runtime/base/base-composer-runtime-core.ts`, `this._text = ""` before `void sendTask.catch(() => {})`), so there is nothing left to restore the message from.

## Why the two entry points should differ

Dropping on cancel is right for `onRespondToToolApproval` (`:306`): an approval for a turn the user just cancelled is meaningless, and re-sending it after the fact would act on a decision the user withdrew.

It is not right for `onNew`. That message was explicitly authored by the user and has already been taken out of the composer. Stopping the current run is a statement about the run, not about the message queued behind it, and "send a correction, then stop the current run so it takes effect sooner" is a natural sequence that loses the correction.

## Proposed fix

Re-stage the dropped message instead of discarding it. `stageUserMessage` (`:213`) already exists in this hook from #5652 and puts a user message into the thread as a staged draft, which is exactly the visible-and-recoverable state a cancelled send should land in. The user then sees their message in the thread and can reload to promote it, rather than watching it disappear.

## Not a regression

On `main` before #5658 the message was lost unconditionally on any send during a run. #5658 narrows the loss to the cancel path. This issue is about closing the remainder, not about restoring previous behaviour.

<a href="https://rupic.dev/s/s2.bTb3PNBwKND-WQpe5ATH22_mHMAEc3M5Oexewe5nh-Fzr2xUlqVtTyPVMjo?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

samdickson22:
Fixes #5676

## What

When a user message is queued behind an active turn and the run is then cancelled, `useEveAgentRuntime` now restages that message as a staged draft instead of discarding it. Cancelled tool-approval responses keep the discard behavior #5658 established.

## Why

The send-serialization machinery landed in #5658 (now merged into `main`). It serializes every `agent.send` through one chain, so a message sent during a streaming turn waits for the turn to park. Cancelling bumps `sendEpochRef`, and the still-queued link throws `sendCancelledError` instead of dispatching. `onNew` swallows that sentinel, and the core composer has already cleared its draft before it starts observing the send promise (`base-composer-runtime-core.ts` sets `this._text = ""` and then `void sendTask.catch(() => {})`), so the authored text exists nowhere afterwards: no thread message, no draft, no error.

That is the right outcome for `onRespondToToolApproval` — an approval for a run the user just cancelled is a withdrawn decision, and replaying it would act against the user's intent. It is the wrong outcome for `onNew`. The message was explicitly authored and has already left the composer, and "send a correction, then hit Stop so it lands sooner" is a natural sequence that currently loses the correction.

## How

`onNew`'s existing catch already narrows to the `sendCancelledError` sentinel; that sentinel is a module-level singleton thrown only by the epoch check *before* `agent.send` is called, so identity comparison is by construction a "this payload never reached the agent" test — a send that already dispatched can never take this branch and so can never double-appear. The branch now calls the existing `stageUserMessage(message)` (added in #5652) with the original `AppendMessage`, which puts the message in the thread with the staged status and records its `runConfig` in `stagedInputsRef`, making it visible and promotable through the normal `onReload` path. `onRespondToToolApproval` is untouched.

No new state, no new option, no export change.

## Tests

Colocated in `packages/eve/src/useEveAgentRuntime.test.tsx`, all under the existing "concurrent sends" describe:

- `restages a queued user send when the run is cancelled before it dispatches` (renamed from "drops a queued send…", whose old title now describes the opposite of the intended behavior) — asserts the queued message is not dispatched *and* is visible in the thread. Fails without the fix.
- `promotes a restaged send on reload without duplicating the dispatched one` — reloads the restaged message and asserts exactly two `send` calls total, the second being the restaged payload; the already-dispatched first send is never restaged. Fails without the fix.
- `discards a queued tool approval when the run is cancelled before it dispatches` — pins the deliberate asymmetry; passes with and without the fix, and guards the boundary from a future over-broad catch.

`pnpm --filter @assistant-ui/eve test`: 78 passed. `tsc --noEmit`, oxlint, and oxfmt clean.

## Changeset

Patch for `@assistant-ui/eve`, describing the restage behavior and the approval asymmetry.

This PR also makes a one-line edit to #5658's own unreleased changeset (`.changeset/eve-concurrent-send-guard.md`). That changeset claimed "Cancelling a run or unmounting the runtime drops sends still queued behind the active turn" — the cancel half of that sentence is no longer true once this lands, and both changesets ship in the same release. The sentence is narrowed to the unmount-only claim so the published changelog does not contradict itself. No behavior of #5658 changes.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5683?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.yK_CGHUOf_4Lc_vfy9OTAIxCWlHL97BPatsMYxLOzTTZgSQb2qazRnIYO0k?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

## Demo

**Before** (#5658's head, `demo/5676-before` @ 23f4e277f) — send during a streaming turn, hit Stop before it dispatches: the queued “Actually, stop at 10.” vanishes from both the thread and the composer, with no error:

https://github.com/user-attachments/assets/1b3193f1-06df-4f2e-b9ef-7927d6bd1f1e

**After** (this PR, 286a22172) — same steps: after Stop, “Actually, stop at 10.” appears in the thread as a staged, recoverable user message:

https://github.com/user-attachments/assets/b73a1433-82e1-4a8c-8903-765ed5983bd6


Kinfe123 (changes requested):
Looks good but could we unmount invalidation be distinguished from an explicit user cancellation here? Both currently produce the same sentinel, so queued sends rejected during unmount are restaged and trigger state updates after the component has unmounted, even though unmount is intended to discard them also merge conflict

okisdev:
closing this in favour of #5789, which fixes #5676 at the composer instead of in the thread.

the review thread about the trailing user leaf is right, and it reproduces. driving the runtime through cancel, drain, cancel and reading `thread.getState().messages` plus `composer.getState().text` at each step:

```
after first cancel    messages: [earlier, earlier answer, queued]   composer: ""
after second cancel   messages: [earlier, earlier answer, queued]   composer: "queued"
after resync          messages: [earlier, earlier answer, queued]   composer: "queued"
```

`cancelRun` deletes a trailing user leaf and puts its text back in the composer (`external-store-thread-runtime-core.ts`), the adapter never learns about the deletion, and the next snapshot re-adds the staged copy. the message ends up in both places, and sending the composer text dispatches a second copy while the ghost stays at the tail. a restaged draft is always that trailing leaf, so any later Stop hits it.

the other two points hold as well: `useActionBarReload` is disabled for `role !== "assistant"` and `MessageRuntime.reload()` throws `"Can only reload assistant messages"`, so nothing in the stock UI reaches `startRun({ parentId })`, and a restaged draft has no dismissal path.

#5789 takes the narrower route the thread suggests. core gains `MessageNotSentError`; a runtime that rejects `onNew` with it gets the composer draft restored under the same guard `cancelRun` already uses, and eve throws it when a cancel drops a queued send. no thread-visible draft, no promotion affordance to invent, and the capability lands for every runtime.

your unmount-versus-cancel fix carried over into it, and the changeset correction went with it too. thanks for the demo recordings, they made the before state easy to check.

PR #5710, opened 2026-08-07
fix(store): deliver swap-time events to default-scope listeners

### LINKED ISSUE #5699: default-scope useAuiEvent misses threadListItem.switchedTo emitted during a thread switch
## summary

`useAuiEvent("threadListItem.switchedTo", cb)` in the default listener scope (the form documented in `packages/store/src/useAuiEvent.ts`) never fires on a thread switch, while a star scope subscription (`{ scope: "*", event: "threadListItem.switchedTo" }`) receives the same emission.

## mechanism

`useAui.on` resolves the listener's scope binding at delivery time and compares it against the emitting client (`packages/store/src/useAui.ts`). the accessor closes over the methods of the render that produced it, and a switch builds a new accessor on a new building client. delivery runs one `queueMicrotask` after emit (`NotificationManager.ts`), while the rebind needs a commit plus effect flush, so at delivery the root's derived `threadListItem` binding still points at the pre-switch item and the identity comparison fails. `switchedTo` is emitted by the newly main item during the swap itself (`thread-list-item-runtime.ts` emits once it observes it became main), so the event always lands inside that window.

the store contract test "useAuiEvent tracks the derived selection across structural swaps" (`packages/store/src/__tests__/events.test.tsx`) flushes the swap with `flushTapSync` first and pings after, so it never exercises an event emitted by the swap itself; that is the coverage gap that let this ship.

## repro

on current main, this jsdom test (placed in `packages/core/src/tests/`) observes `{ default: 0, star: 1 }`:

```tsx
// @vitest-environment jsdom
import { act, render } from "@testing-library/react";
import { expect, it, vi } from "vitest";
import { AuiProvider, useAui, useAuiEvent } from "@assistant-ui/store";
import { RuntimeAdapter } from "../react/RuntimeAdapter";
import { AssistantRuntimeImpl } from "../runtime/api/assistant-runtime";
import { ExternalStoreRuntimeCore } from "../runtimes/external-store/external-store-runtime-core";
import type { ExternalStoreAdapter } from "../runtimes/external-store/external-store-adapter";

type DemoMessage = { id: string; role: "user" | "assistant"; text: string };

const createRuntime = () => {
  const threads = [
    { id: "t1", title: "one", messages: [{ id: "m1", role: "user" as const, text: "a" }] },
    { id: "t2", title: "two", messages: [{ id: "m2", role: "user" as const, text: "b" }] },
  ];
  let currentId = "t1";
  const makeAdapter = (): ExternalStoreAdapter<DemoMessage> => ({
    messages: threads.find((t) => t.id === currentId)!.messages,
    convertMessage: (m) => ({ id: m.id, role: m.role, content: [{ type: "text", text: m.text }] }),
    onNew: async () => {},
    adapters: {
      threadList: {
        threadId: currentId,
        threads: threads.map((t) => ({ status: "regular" as const, id: t.id, title: t.title })),
        onSwitchToThread: (threadId: string) => {
          currentId = threadId;
          sync();
        },
        onSwitchToNewThread: () => {},
      },
    },
  });
  const core = new ExternalStoreRuntimeCore(makeAdapter());
  const runtime = new AssistantRuntimeImpl(core);
  const sync = () => core.setAdapter(makeAdapter());
  return runtime;
};

it("delivers switchedTo to the default scope", async () => {
  const runtime = createRuntime();
  const defaultScope = vi.fn();
  const starScope = vi.fn();
  let aui!: ReturnType<typeof useAui>;
  const Harness = () => {
    aui = useAui({ threads: RuntimeAdapter(runtime) } as never);
    return (
      <AuiProvider value={aui}>
        <Consumer />
      </AuiProvider>
    );
  };
  const Consumer = () => {
    useAuiEvent("threadListItem.switchedTo" as never, defaultScope as never);
    useAuiEvent({ scope: "*", event: "threadListItem.switchedTo" } as never, starScope as never);
    return null;
  };
  render(<Harness />);
  await act(async () => {});

  await act(async () => {
    aui.threads.item({ index: 1 }).switchTo();
  });
  await act(async () => {});

  expect(starScope).toHaveBeenCalledTimes(1); // passes
  expect(defaultScope).toHaveBeenCalledTimes(1); // fails: 0 calls
});
```

the same asymmetry reproduces through the Vue bindings in `packages/vue`, so it is a store-level behavior, not a framework binding quirk.

## affected

these shipped listeners subscribe to `threadListItem.switchedTo` in the default scope and miss the event the same way per the mechanism above:

- `packages/react/src/primitives/composer/ComposerInput.tsx:349` (refocus on switch)
- `packages/react/src/primitives/thread/ThreadViewport.tsx:153` (clear top anchor turn)
- `packages/react/src/primitives/thread/useThreadViewportAutoScroll.ts:227` (scroll to bottom on switch)
- `packages/react/src/unstable/useComposerInputHistory.ts:75` (reset input history)
- `packages/react-native/src/primitives/thread/ThreadMessages.tsx:351`

`packages/vue` subscribes with `{ scope: "*" }` as a local workaround (#5698); once delivery resolves the scope correctly that can return to the documented default form, and the events contract test should gain a case where the ping is emitted from inside the swap rather than after it.

<a href="https://rupic.dev/s/s2.KQ4d_RbUvarrk-avIReFQTVZn4BIG5lzF6-N3COJGzt8pAUF9cMzsMkdZv8?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

samdickson22:
## What

A scoped event listener is filtered by comparing the emitting client stack against the listener's scope binding. Two windows around a *structural swap* (a derived scope resolving to a different instance) make that comparison miss an event that the swap itself emits, so a default-scope `useAuiEvent` hears nothing while a `scope: "*"` listener does.

Both are fixed here, with a regression test each.

**1. The listener set captured at emit time can be emptied before the flush.** `emit` captured `listeners.get(event)` and iterated that `Set` one microtask later. A consumer that unsubscribes and resubscribes in between — exactly what `useAuiEvent` does when the swap gives it a new client — empties the captured set, which is then deleted from the map, and the fresh subscription lands in a *new* set that the in-flight emission never sees. The event is dropped for that listener entirely, even though the documented live-set semantics ("listeners added between emit and flush are invoked") say it should be delivered. `emit` now re-reads the listener set at flush time; the emit-time "no listeners, drop it" check is unchanged.

**2. A listener that has not re-subscribed yet compares against the pre-swap binding.** Delivery is one microtask after the emit, but a React consumer's rebind needs a render, a commit and an effect flush. For an event emitted from inside the swap, the listener is still bound to the pre-swap instance when delivery runs, so the identity check fails. `useAuiEvent` now subscribes with a *live* binding: the scope is resolved through the host's current client at delivery time rather than through the client captured at subscription time. `aui.on` keeps its documented subscription-time binding (`listeners bind to the instance resolved at subscription time` still passes) — the binding mode is an internal parameter (`packages/store/src/utils/event-binding.ts`), not new public API.

The live mode leans on `clientRef.current`, which is why the ref is now assigned in the `useAuiRoot` render instead of in a commit effect: the tap root produces the post-swap client during the flush that performs the swap, which is before React can re-render anything. Assigning during render is safe here because the ref is only ever *read* at delivery time — inside a microtask, never during a render — and because `useCommittedClient` keeps the client identity stable across every non-structural render, which makes the assignment idempotent. A render that is discarded (StrictMode's double render, an interrupted transition) therefore either writes the same client the committed render would write, or is followed by the render that does. With the ref now written unconditionally there, the standalone root's null seed in `createAssistantClient` became unreachable and is deleted.

## Why this design

- *Re-resolve through the current root state* (chosen): the correction lands where the staleness is, keeps one code path for every binding, and needs no new scheduling.
- *Defer default-scope delivery until after the commit that rebinds* (rejected): delivery timing is a documented contract (microtask, live-set), the store cannot observe React commits, and non-React bindings (`createAssistantClient`, Vue) have no commit to wait for.
- *Match more loosely, e.g. on the source scope or on `"*"` plus a manual filter* (rejected): that is the workaround, not a fix — it delivers foreign instances' events to a listener that asked for one instance.

## Tests

`packages/store/src/__tests__/events.test.tsx` gains a thread-list-shaped harness where the client emits when it observes that it became the selected one, from inside the flush that performs the swap. Four new tests, each failing on `main`:

- the swap's own event reaches a default-scope listener (window 1), and the deselected instance's simultaneous emission still does **not** reach it — only the wildcard listener hears that one;
- the same, asserted before React re-renders (window 2);
- at the store level, with no React commit involved at all: a live-bound listener follows the new selection while a snapshot-bound `aui.on` listener stays on the instance it subscribed with;
- and one contract test for an interaction the flush-time re-read introduces: whether an emission survives is decided at emit time and asks only whether *anything* was listening, so when a wildcard listener is present, a scoped listener that subscribes between the emit and the flush now receives the emission (on `main` it was dropped). With no wildcard listener the existing "an emission with no listeners at emit time is dropped" contract is unchanged.

Verified: `@assistant-ui/store` (146), `@assistant-ui/tap` (473), `@assistant-ui/core` (947), `@assistant-ui/react` (423) and `@assistant-ui/vue` (65) suites green; store `tsc --noEmit`, oxlint and oxfmt clean; api-reference regenerated for the `useAuiEvent` JSDoc addition.

## Scope: this does not close #5699 on its own

Part of #5699. It fixes the store-level defect the issue's mechanism analysis describes, but I measured the end-to-end `threadListItem.switchedTo` case and it still does not fire, for a reason that lives outside the store: **at delivery time the store's runtime-client state has not caught up with the runtime**. With an external-store multi-thread runtime, right when the emission is delivered:

```
runtime mainThreadId = t2      (already switched)
client  mainThreadId = t1      (aui.threads.getState(), and the derived
                                threadListItem, still resolve to the old thread)
```

and a `flushTapSync` does not settle it — the runtime-driven state arrives through `useSubscribable` (`useSyncExternalStore`) and only lands on the next render, while the event is emitted synchronously from the runtime's own subscription. So the item announces `switchedTo` while every scope derived from the thread list still points at the previous thread, and no delivery-time comparison can match. That belongs in a follow-up on the core side (either settle the runtime-client state before the emission is delivered, or emit the switch from the store's own state transition instead of the runtime subscription).

Consequently the Vue wildcard workaround from #5698 stays for now: I reverted it locally against this branch and `scrolls the viewport to the bottom on thread switch` still fails. It becomes removable once the core-side ordering is addressed.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5710?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.B-SEYfr-zCte_RWq3zM_93t9HqYLX9DqobHucQ2pi7T3YeSZt8N6HANXSoo?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/mean-pugs-shout.md
+++ b/.changeset/mean-pugs-shout.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/store": patch
+---
+
+fix: deliver scoped events emitted during a structural swap to default-scope listeners
--- a/apps/docs/content/docs/(reference)/api-reference/hooks/state.mdx
+++ b/apps/docs/content/docs/(reference)/api-reference/hooks/state.mdx
@@ -54,6 +54,11 @@ scope or event name changes. The `callback` is wrapped in an effect-event
 shim, so the latest closure is invoked on each emission — you do not
 need to memoize it.
 
+A scoped subscription is filtered against the scope's binding at delivery
+time, so it follows a derived scope through a structural swap — including
+for an event the swap itself emits, such as `threadListItem.switchedTo` on
+a thread switch.
+
 {/* api-example:useAuiEvent:start */}
 ```tsx
 import { useAuiEvent } from "@assistant-ui/react";
--- a/packages/store/src/__tests__/events.test.tsx
+++ b/packages/store/src/__tests__/events.test.tsx
@@ -1,7 +1,7 @@
 // @vitest-environment jsdom
 
 import type { ReactNode } from "react";
-import { useState } from "react";
+import { useEffect, useRef, useState } from "react";
 import { act, cleanup, render, renderHook } from "@testing-library/react";
 import { afterEach, describe, expect, it, vi } from "vitest";
 import { flushTapSync, resource } from "@assistant-ui/tap";
@@ -12,6 +12,8 @@ import { useAuiState } from "../useAuiState";
 import { Derived } from "../Derived";
 import { useAssistantEmit } from "../utils/tap-assistant-context";
 import { useClientResource } from "../useClientResource";
+import { createAssistantClient } from "../createAssistantClient";
+import { onWithBinding } from "../utils/event-binding";
 
 type AnyClient = Record<string, any>;
 
@@ -43,6 +45,49 @@ const useThreadClient = () => {
   };
 };
 const ThreadClient = resource(useThreadClient);
+
+// Mirrors a thread list item: the client emits when it observes that it became
+// the selected one, from inside the flush that performs the selection swap
+const useSwitchingMessageClient = ({
+  id,
+  isSelected,
+}: {
+  id: string;
+  isSelected: boolean;
+}) => {
+  const emit = useAssistantEmit();
+  const wasSelected = useRef(isSelected);
+  useEffect(() => {
+    if (wasSelected.current === isSelected) return;
+    wasSelected.current = isSelected;
+    if (isSelected) emit("message.switchedTo" as never, { id } as never);
+    else
+      emit("message.pinged" as never, { id, value: "switched-away" } as never);
+  }, [id, isSelected, emit]);
+  return {
+    getState: () => ({ id, text: "" }),
+    ping: (value: string) =>
+      emit("message.pinged" as never, { id, value } as never),
+  };
+};
+const SwitchingMessageClient = resource(useSwitchingMessageClient);
+
+const useSwitchingThreadClient = () => {
+  const [selected, setSelected] = useState(0);
+  const m0 = useClientResource(
+    SwitchingMessageClient({ id: "m0", isSelected: selected === 0 }),
+  );
+  const m1 = useClientResource(
+    SwitchingMessageClient({ id: "m1", isSelected: selected === 1 }),
+  );
+  const messages = [m0, m1];
+  return {
+    getState: () => ({ selected }),
+    setSelected,
+    message: ({ index }: { index: number }) => messages[index]!.methods,
+  };
+};
+const SwitchingThreadClient = resource(useSwitchingThreadClient);
 
 const messageDerived = () =>
   Derived({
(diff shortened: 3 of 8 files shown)


Kinfe123 (changes requested):
Hmmm..could we avoid resolving live bindings through the parent clientRef for derived-only providers? These child clients reuse parent.on, so clientRef.current points to the parent rather than the child that subscribed. A MessageByIndexProvider listener can therefore resolve a missing or wrong scope and either throw or drop a valid event. Could derived-only clients keep their own live delivery reference, with a regression test using a derived-only provider?

okisdev:
superseded by #5769, which just merged and closes #5699 end to end. closing this one.

what carried over: your notification-manager flush-time re-read ships in #5769 essentially as written here, with a regression test for the unsubscribe-resubscribe window — that isolation was correct and load-bearing, thank you. your analysis of the two windows was also what mapped the store half of the problem.

why the surrounding approach was replaced, per maintainer direction on code paradigm: the binding-mode parameter gave the same selector different semantics depending on whether it was subscribed through aui.on or useAuiEvent, and the render-phase clientRef assignment needed a safety argument for breaking a rule rather than a seam that makes the rule irrelevant. #5769 instead emits the switch from the item client's own observed selection transition (after the flush that rebinds the derived scopes, which holds for react, vue, and standalone hosts alike), makes delivery-time scope resolution the single semantic for both entry points, and refreshes the client ref inside the existing flushTapSync notification closure, the same channel createAssistantClient already used. the core-half residue you correctly identified as out of store's reach is exactly what the emission-timing move fixes.

the derived-only gap from the review rounds is tracked in #5768 with the evidence matrix.

PR #5735, opened 2026-08-08
fix(react-opencode): isolate runtime error callbacks

Kinfe123:
## Summary

- isolate synchronous and asynchronous `onError` callback failures
- preserve the original OpenCode operation error for callers
- prevent initial-load callback rejections from becoming unhandled rejections
- document promise-returning error callbacks in the public runtime options

## Problem

`useOpenCodeRuntime` invokes the app-provided `onError` callback inside its own failure paths. If that callback throws during send, cancel, approval, or reload, its error replaces the real OpenCode error. During the initial `controller.load()` effect, a rejected callback promise is discarded and can become a global unhandled rejection. This is common when error reporting or telemetry has its own transient failure.

## Root cause

The callback ran directly inside each operation catch block, and the load effect discarded the promise returned by its catch handler. Callback failures therefore entered runtime control flow instead of remaining reporting failures.

## Fix

Route error reporting through a guarded helper that handles both synchronous throws and promise rejections. Callback failures are logged, each operation continues rejecting with its original error, and the public callback type now explicitly permits `Promise<void>`.

## Verification

- reproduced both the initial-load rejection and send-error replacement before the fix
- `pnpm --filter @assistant-ui/react-opencode test`
- `pnpm --filter @assistant-ui/react-opencode build`
- `pnpm exec oxlint packages/react-opencode/src/types.ts packages/react-opencode/src/useOpenCodeRuntime.ts packages/react-opencode/src/useOpenCodeRuntimeApproval.test.tsx`


okisdev (approved):
onError no longer replaces the OpenCode operation error, and a rejected callback on initial load is no longer an unhandled rejection.

the helper matches the already-shipped pi path. against this head the suite is 3/3; against main the two new tests fail with the reported symptoms (telemetry error instead of send failed, and no console.error on a rejected load callback).

PR #5740, opened 2026-08-09
feat: add 63 elements to the gallery

okisdev:
## summary

- adds 63 elements, taking the hand-authored gallery from 37 to 100 (122 pages counting the generative templates). the set came from a gap analysis of what the runtime already ships versus what the gallery had a picture of, so the bulk of it closes holes a user hits the day they adopt a feature: the message queue, realtime voice, read aloud, MCP elicitation and server state, trace waterfalls, the canvas, reviewable diffs, charts, and the thread shell.
- two sections join the taxonomy. Voice holds the realtime call and read aloud. Observability takes the run-economics elements (trace waterfall, cost meter, quota banner) out of Agents, which had grown to fifteen entries and was mixing supervising work with what the work cost.
- makes several otherwise-invisible packages visible for the first time: `@assistant-ui/react-o11y` (trace waterfall), react-mcp's elicitation primitives, `heat-graph`, `safe-content-frame`, and the realtime voice adapter.
- every element stays prop-driven with no runtime coupling, matching the contract of the original 37, so they compose into any runtime and copy cleanly through `shadcn add`.

### already verified

- [x] `pnpm lint` -> oxlint and oxfmt clean across 3,589 files (only pre-existing warnings in vue, react-ag-ui, and elements/surfaces.tsx remain)
- [x] `pnpm -C packages/ui exec tsc --noEmit` -> clean
- [x] `pnpm -C apps/docs exec tsc --noEmit` -> clean under `exactOptionalPropertyTypes`
- [x] `pnpm -C apps/registry build && pnpm -C apps/registry test` -> 95 element registry items emitted, 32/32 green
- [x] `pnpm -C packages/ui test` -> 13/13 green
- [x] `pnpm -C apps/docs test` -> 132/132 green
- [x] `pnpm -C apps/docs build` -> 122 `/elements/[slug]` paths prerendered. this is the real gate: the detail page reads each element's source off disk and throws when a file named in the registry is missing, so a full prerender proves every entry resolves.
- [x] rendered all 122 cards in a browser at 1280px in both themes, and swept every card for content escaping its own surface: 0 horizontal overflows, 0 vertical clips among the new elements.

### reviewer should verify

- [ ] open `/elements` and confirm the two new sections (Voice, Observability) read as coherent groupings rather than leftovers.
- [ ] open `/elements/activity-graph` and confirm the calendar window matches its caption. this element requires an explicit `start`/`end`; left unset, heat-graph draws a full year ending today, which silently disagrees with the data.
- [ ] spot-check a few detail pages for props tables that drifted from the component signature.

## notes

- three elements landed under a different name than the plan used, for a reason worth flagging in review. `diagram` takes `children` rather than a chart string, so nothing injects model-written markup and no renderer dependency lands in a copy-paste element. `activity-graph` builds on the first-party `heat-graph` package rather than hand-rolling date math, and is the only new element carrying a runtime dependency. `web-preview` is the safe-HTML-preview frame.
- thread search landed as its own element rather than a variant of `thread-list`, because adding search and grouping would have changed the props of an element already distributed through the registry.
- `createElementRegistryItem` gains a `usesElements` field so an element that imports another can declare it as a registry dependency. `reviewable-diff` is the first, pulling in `code-diff` for the `DiffLine` type; without it `shadcn add` would land a file with an unresolvable import.
- no changeset: `@assistant-ui/ui`, `@assistant-ui/docs`, and `@assistant-ui/shadcn-registry` are all `private: true`.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5740?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.kVXWSVOJgguueNm5m8wBEu_mXnlubkF0VWuCSfWREqT6z0Nif6d-QbxpXbc?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>


okisdev:
triaged all 45 threads. seven were real and are fixed in 1181057a8:

- `chart` read a falling delta only from U+2212, so an ASCII `-18%` rendered green (claude, cubic).
- `context-breakdown`, `reasoning-effort`, and the activity-graph demo formatted numbers with a bare `toLocaleString()`, which resolves against the runtime locale and so disagreed between the server render and the browser (claude, cubic).
- `comparison-card` compared traits against `false`, so an option with fewer traits than labels drew a green check with a blank label (cubic).
- `background-inbox` counted failed runs as in flight (cubic).
- `web-preview` described itself as though it created the sandbox. it renders `children` as given and isolates nothing, so the description and props table now say the caller has to pass an already-sandboxed frame (rupic, cubic). good catch; a copy-paste element that claims a boundary it does not enforce is worse than one that claims nothing.
- the props tables for `MessageAttachmentItem`, `MapPin`, and `FlowNode` omitted their required identity fields (rupic, cubic).

the rest are deferred, not rejected. they fall into three groups that all apply equally to the 37 elements already on main, so fixing them only inside this diff would leave the family inconsistent:

- keyboard and screen-reader affordances (arrow-key navigation in the command palette, live regions on feedback and document reference, hit targets on map pins).
- overflow hardening for long unbroken strings (timeline, code runner, command palette).
- guards for out-of-range props (zero context limit, negative `visibleLines`, out-of-range temperature).

the demo-interactivity notes (an inert Skip button, a paused timer, a no-op Authorize) are working as intended: these are gallery demos of presentational elements, and wiring every control to real behavior is not what the cards are for.

i will open a follow-up issue for the accessibility and overflow passes across all 100 elements.

okisdev:
round 4. three fixed in 959ae92bd, five deliberately not.

fixed:

- `voice-conversation`: the round-3 interrupt fix re-created the defect one level down. the orb was enabled whenever mode was speaking, regardless of whether `onInterrupt` was supplied, so a consumer copying the element still got an inert control promising an interrupt. good catch, and squarely my regression.
- `element-docs`: the MapAnswer table marked the optional `route` flag required and omitted `onSelect`.
- `read-aloud` demo: divided milliseconds by 100 and labelled the result seconds.

not fixed, and these are decisions rather than oversights:

- inert controls in gallery demos (Diagram's expand, the MCP Authorize row, ToolError's Skip, the plain div passed to WebPreview). every demo in this gallery, including the 37 already on main, stands the element up in a representative state rather than wiring real behavior. changing that for six of them would make the family inconsistent, and wiring all 100 is a different piece of work.
- mouse-only disclosure in `prompt-library` and `confidence-marker`. real, and it applies equally to the elements already shipped.

both belong in the accessibility and interaction pass across all 100 elements that i said i would open as a follow-up, and i would rather do that once against the whole family than partially here.

worth saying plainly: the bots have found real defects in four consecutive rounds, thirteen in total, one of which was a regression i introduced fixing round three. that is a fair signal that 63 elements in one PR is past the size i can self-review reliably. the reason i am merging rather than splitting is that the remaining findings have converged on the two categories above, both of which are pre-existing and repo-wide rather than introduced here.

okisdev:
deferred accessibility, overflow, and prop-range items are tracked in #5743, scoped to all 100 elements rather than the 63 added here, since the 37 already on main have the same gaps.

Issue #5743, opened 2026-08-09
elements: accessibility, overflow, and prop-range pass across all 100 elements


okisdev:
The elements gallery is 100 hand-authored pieces after #5740. Review of that PR surfaced three categories that apply to the whole family, including the 37 that predate it, so they were deliberately left out of that diff rather than fixed for the new elements only.

## keyboard and screen-reader access

- `confidence-marker` exposes the claim basis from `onMouseEnter`/`onMouseLeave` on non-focusable spans, so keyboard and touch users cannot reach it.
- `prompt-library` binds `onInsert` to `onDoubleClick` only, so a prompt can be selected but not inserted from the keyboard.
- `command-palette` styles `activeId` but implements no arrow-key or Enter handling, so the documented palette contract is the caller's to build.
- `mcp-server-panel` collapsed rows carry status only as a colour dot with no accessible name.
- `map-answer` pins are 10 to 14px targets and give screen-reader users no way to tell which place is selected.
- `feedback-dialog` and `document-reference` give no live-region confirmation when the selection changes.

## overflow with hostile content

Long unbroken strings escape their pane in `timeline`, `code-runner`, and `command-palette`. These need `overflow-wrap` or a scroll container rather than `whitespace-pre` in a fixed box.

## out-of-range props

`context-breakdown` and `settings-panel` were hardened in #5740 (zero limit, clamped temperature). The same treatment has not been applied to the rest: negative `visibleLines` in `canvas-split`, out-of-range `activeIndex` outside `conversation-search`, and similar.

## why one pass rather than per-element

Fixing these inside #5740 would have left the family inconsistent, with 63 hardened elements beside 37 that are not. Doing it as one sweep also makes it worth adding a lint rule or a test that catches the mouse-only-disclosure and unbounded-string shapes, so the next element cannot reintroduce them.

<a href="https://rupic.dev/s/s2.vPnCeDpHoYXnXaCu0o1z2ASzVYStr0VLQMA_WXTb3clQYEEJDz_N9_mfqqw?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg"><img alt="Track in Rupic" src="https://rupic.dev/badges/track-light-v1.svg"></picture></a>

okisdev:
@rupic-app cancel

okisdev:
[...]

PR #5744, opened 2026-08-09
refactor(ui): give the elements slots, parts, and the shadcn conventions

okisdev:
follows #5740. that PR added the elements; this one fixes their structure, after an audit of how all 95 are actually shaped.

## summary

- the audit's headline: elements have a **median of 5 props and 92 lines, 3 of 95 hold state, and 2 of 95 use an effect**. that is not the config-heavy monolith compound APIs exist to fix, so this deliberately does **not** rewrite the family into `Root`/`Item` parts. `ui/base/*` is compound because it wraps Base UI primitives that already own focus, portals, and keyboard; there is no primitive underneath a trace waterfall, and inventing `<TraceWaterfall.Root>` over a `spans` prop would buy nothing.
- **four elements did need it**, and two of those were simply wrong. `chat-panel` took a singular `userMessage` and `reply`, so it could not represent a conversation; `empty-state` hardcoded the place its consumer's composer has to go. both were demo compositions shipped as components. `canvas-split` fixed both panes, and `composer` was 613 lines behind 24 props holding six facets.
- the **real** divergence was not structure but convention. before this PR, `data-slot`, prop spreading, and `ComponentProps` typing appeared in 0 of 95 elements, against 19 of 20 in `ui/base`; this PR adds them. a copied element could not be restyled from outside or take an `id`, `ref`, aria attribute, or extra handler without editing the file. now 94 of 95 and 90 of 95.
- keyboard work landed on **4 elements, not the 12** the review flagged: 9 of them render their rows as `<button>` and were already Tab-reachable with a working Enter. only the three with a text input above a list need the element to own the arrow keys, plus `confidence-marker`, which had no focusable element at all.

### already verified

- [x] `pnpm lint` -> oxlint and oxfmt clean (only the pre-existing `surfaces.tsx` exhaustive-deps warning)
- [x] `pnpm -C packages/ui exec tsc --noEmit` -> clean
- [x] `pnpm -C apps/docs exec tsc --noEmit` -> clean under `exactOptionalPropertyTypes`
- [x] `pnpm -C packages/ui test` -> 13/13, `pnpm -C apps/docs test` -> 132/132, `pnpm -C apps/registry test` -> 33/33
- [x] `pnpm -C apps/docs build` -> 122 `/elements/[slug]` paths prerendered
- [x] drove the new arrow-key navigation in a real browser: it walks across group boundaries and wraps (`New thread` → … → `Stop the run` → back to `New thread`)
- [x] re-ran the overflow and clipping sweep over all 122 gallery cards: 0 horizontal overflow, 0 empty stages, and the only two clips are the pre-existing intentional ones (`number-ticker`'s roll strip, `composer-context`'s hover popover)

### reviewer should verify

- [ ] `/elements/composer` and its six sibling entries: each card should now show only the facet it demonstrates.
- [ ] `/elements/chat-panel`: the reply is rendered by composing `streaming-text` into the messages slot rather than by chat-panel re-implementing word streaming.
- [ ] tab into `/elements/confidence-marker` and confirm the basis pill appears on focus, not only on hover.

## notes

- `surfaces.tsx` gets no `data-slot`; it is a token module, not a component. the four `Collapsible`-rooted elements (`reasoning-panel`, `sources`, `tool-call`, `tool-timeline`) get no prop spread, because spreading `div` props onto a component would be wrong; they would need `ComponentProps<typeof Collapsible>`. that is 90 of 95 by choice, not by omission.
- collisions in the spread pass are avoided by omitting each component's own prop names from `ComponentProps`, so a prop called `title` or `value` does not clash with the HTML attribute.
- the accessibility work still outstanding in #5743 is unchanged by this PR. while auditing for it I found the static detector I wrote has real false negatives (it misses `map-answer`'s hit target, which lives on a child element), which is worth knowing before anyone treats a grep-based sweep as coverage.
- no changeset: `@assistant-ui/ui` and `@assistant-ui/docs` are both `private: true`.

### DIFF
--- a/apps/docs/components/elements/canvas-split-demo.tsx
+++ b/apps/docs/components/elements/canvas-split-demo.tsx
@@ -14,12 +14,12 @@ import { useStoryPhases } from "./use-demo";
 const MESSAGES = [
   {
     id: "u1",
-    role: "user" as const,
+    speaker: "user" as const,
     text: "Draft the migration note for 0.14",
   },
   {
     id: "a1",
-    role: "assistant" as const,
+    speaker: "assistant" as const,
     text: "Opened it in the canvas — editing now.",
   },
 ];
@@ -47,7 +47,7 @@ export function CanvasSplitDemo() {
     <CanvasSplit>
       <CanvasSplitThread>
         {MESSAGES.map((message) => (
-          <CanvasSplitMessage key={message.id} role={message.role}>
+          <CanvasSplitMessage key={message.id} speaker={message.speaker}>
             {message.text}
           </CanvasSplitMessage>
         ))}
--- a/apps/docs/components/elements/canvas-split-demo.tsx
+++ b/apps/docs/components/elements/canvas-split-demo.tsx
@@ -2,42 +2,75 @@
 
 import {
   CanvasSplit,
-  type CanvasMessage,
+  CanvasSplitBody,
+  CanvasSplitDocument,
+  CanvasSplitHeader,
+  CanvasSplitLine,
+  CanvasSplitMessage,
+  CanvasSplitThread,
 } from "@/components/elements/canvas-split";
 import { useStoryPhases } from "./use-demo";
 
-const MESSAGES: readonly CanvasMessage[] = [
-  { id: "u1", role: "user", text: "Draft the migration note for 0.14" },
+const MESSAGES = [
+  {
+    id: "u1",
+    role: "user" as const,
+    text: "Draft the migration note for 0.14",
+  },
   {
     id: "a1",
-    role: "assistant",
+    role: "assistant" as const,
     text: "Opened it in the canvas — editing now.",
   },
 ];
 
-const LINES: readonly string[] = [
-  "# Migrating to 0.14",
-  "The composer now owns its draft, so `useState` in the parent is no longer read.",
-  "Replace the local draft with `useDraft(threadId)` and drop the hydrate effect.",
-  "Threads that switch mid-edit keep their own draft instead of inheriting the last one.",
+const LINES: readonly { text: string; heading?: boolean }[] = [
+  { text: "Migrating to 0.14", heading: true },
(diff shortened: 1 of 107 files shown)


okisdev:
triaged the review. fixed in 2c1c1b9e7 and bba46d16b:

- `ComposerInput` set `onKeyDown` before the spread, and `onKeyDown` is a legal prop of the type it accepts, so a consumer passing one silently replaced Enter-to-submit. chained now, with `preventDefault()` as the opt-out. good catch: the type gave no signal at all.
- `confidence-marker` hardcoded `id="confidence-basis"`. two on a page emitted duplicate ids and every claim in both described the first. `useId` now.
- the guard-clause branches in `edit-message`, `error-state`, and `feedback-dialog` lost `data-slot` and the spread the moment their condition flipped, because the sweep patched only each component's first return. those were the only three in the tree.
- `typing-indicator` had the same shape through a variant rather than a guard, so `variant="bare"` dropped both.
- `ComponentProps` carries `children`, so 76 components advertised a `children` prop their own explicit root discarded. those signatures omit it now, which turns a lying type into a compile error.
- the composer demos cleared the input on Enter while a slash or mention menu was open, rather than taking the item the menu marks with a return-key hint.

two I am not acting on, with reasons:

- **`data-slot` placed before `{...props}`, so a caller can override it.** `ui/base/avatar.tsx` puts `data-slot` before the spread on all four of its parts. matching that was the point of this PR; deviating here would reintroduce the inconsistency it set out to remove.
- **"this PR adds the patterns its description says it removes."** the description reads "0 of 95 elements carried `data-slot` ... now 94 of 95", which is a before-and-after, not a goal. the reading is understandable though, so i have reworded it rather than argue.

on process: the resolve pass ran while this review was still landing and closed threads i had not read. that is the second time, so i went back and read every one before writing this. the findings above are from that read, not from the resolve.

PR #5776, opened 2026-08-11
feat(react-lexical): reparse drafts after formatter changes

Kinfe123:
## Summary

- detect composed directive parser changes and compare their parsed structure with the current Lexical tree
- reparse untouched restored drafts when a newly registered formatter changes their structure without demoting chips on removal or reparsing active edits
- skip no-op parser changes so active caret and composition state remain untouched
- preserve description and metadata only for the same directive occurrences during syntax changes and duplicate matches
- add regression coverage for empty-mount selection, multiline no-op registration, directive conversion, and metadata retention

## Why

`SyncPlugin` previously resynchronized only when the runtime text changed. If an unedited restored draft such as `[[alice]]` loaded before its formatter registered, Lexical created a plain text node. Registering the formatter later changed the parser but left the unchanged, unedited draft stale, so dynamic tool and mention registrations were not reflected until the text was edited.

The plugin now tracks which parser produced the editor state. When that parser changes, it compares normalized text/directive segments and rebuilds only if the resulting structure differs and the editor has not been edited or had its selection moved since the last runtime sync. Empty initial drafts retain the prior mount behavior, no-op formatter registrations preserve the existing nodes and caret, and parser-only rebuilds do not move DOM focus into an untouched composer. Hydrated node labels are preserved per occurrence when the old and new parsers agree, while genuine parser label changes still apply. Equivalent formatter wrappers reuse the composed parser when their parse and serialize functions are unchanged. When a parser-only rebuild is necessary, matching existing directive items retain richer metadata that was hydrated without changing the draft text. The existing sync tag prevents a genuine internal reparse from calling `composer.setText()`. Required parser-only rebuilds normalize inline line breaks to paragraph boundaries. For a custom formatter whose serialize and parse operations are not round-trip identical, the editor can show the new directive syntax while submission retains the prior runtime text until the next edit.

## Verification

- `pnpm --filter @assistant-ui/react-lexical test` \(45 tests\)
- `pnpm --filter @assistant-ui/react-lexical build`
- `pnpm exec oxlint packages/react-lexical/src/plugins/SyncPlugin.tsx packages/react-lexical/src/plugins/SyncPlugin.test.tsx`
- `pnpm exec oxfmt --check packages/react-lexical/src/plugins/SyncPlugin.tsx packages/react-lexical/src/plugins/SyncPlugin.test.tsx`

<a href="https://rupic.dev/s/s2.tdb4JpF9aEEisK6kudORaaX7nZY6jw_0Ou-UGS2bTR8QfeLElVFkGnh1_94?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/reparse-lexical-directives.md
+++ b/.changeset/reparse-lexical-directives.md
@@ -2,4 +2,4 @@
 "@assistant-ui/react-lexical": patch
 ---
 
-fix: reparse composer drafts when directive formatters change
+fix: reparse unedited composer drafts when directive formatters change
--- a/.changeset/reparse-lexical-directives.md
+++ b/.changeset/reparse-lexical-directives.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react-lexical": patch
+---
+
+fix: reparse composer drafts when directive formatters change
--- a/packages/react-lexical/src/plugins/SyncPlugin.test.tsx
+++ b/packages/react-lexical/src/plugins/SyncPlugin.test.tsx
@@ -46,15 +46,18 @@ vi.mock("@assistant-ui/store", async (importOriginal) => {
 (globalThis as Record<string, unknown>).IS_REACT_ACT_ENVIRONMENT = true;
 
 const createAui = (text: string) => {
+  let currentText = text;
   const runtime = {
-    getState: () => ({ text }),
+    getState: () => ({ text: currentText }),
     subscribe: () => () => {},
   };
 
   return {
     composer: {
       __internal_getRuntime: () => runtime,
-      setText: vi.fn(),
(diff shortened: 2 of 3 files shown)


okisdev:
superseded by #6088.

this head is 232 commits behind main. the selection-dirty guard also skips the common path: click into a restored `[[mention]]` draft, then register the formatter, and the text stays plain.

#6088 rebuilds on parser change unless the user edited text or the editor is composing. caret movement is not treated as an edit.

okisdev:
closed as superseded by #6088.

PR #5795, opened 2026-08-11
fix(store): publish hosted scope rebinds before layout effects

### LINKED ISSUE #5781: store: hosted scoped listeners can miss layout-effect events after a rebind
## Bug

A hosted provider can drop a scoped event when it structurally rebinds a derived scope and a descendant layout effect emits from the newly committed binding before passive effects run.

## Reproduction

Reproduced on current `main` at `417efee9` with a focused jsdom Store test:

1. Mount a root `thread` client with messages `m0` and `m1`.
2. Mount a hosted child provider with one root scope and `message: Derived(...)`, initially bound to index 0.
3. Keep a `useAuiEvent("message.pinged", callback)` subscription under that provider.
4. Rerender the hosted provider with index 1.
5. Emit `m1` from a descendant layout effect, then drain the queued event notification from an ancestor layout effect in the same commit (before passive effects).

Expected:

```text
callback({ id: "m1", value: "layout" })
```

Actual:

```text
callback call count: 0
```

## Cause

`useHostedAssistantClient` refreshes `clientRef.current` in a passive effect for an ordinary React-driven structural rebind. The new hosted client has committed by the time descendant layout effects run, but scoped delivery can still resolve against the previous binding until that passive effect executes, so the new event is filtered out.

The notification-driven `flushTapSync` path already refreshes the ref synchronously; this is the separate React commit path.

## Suggested direction

Publish the rendered hosted generation during the React commit phase from outside the tap host (for example, an insertion effect or an equivalent commit hook), while keeping render-phase publication out so an interrupted render cannot leak a speculative binding. Regression coverage should drain an event between layout and passive effects and assert the newly committed scope receives it.

Related to the hosted delivery-time rebinding work in #5769. This is separate from the derived-only provider bug in #5768.

<a href="https://rupic.dev/s/s2.KQgXSsskES8ZbCIYfLwLEoTpjOUjyE67bYIcFjD-7CL2P-ErzgJVWpGZjXM?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

Gujiassh:
## Summary

Hosted scoped listeners could miss an event emitted from a descendant layout effect immediately after a derived scope rebind. The listener still resolved the previous hosted client because the ref was published from a passive effect.

This change:

- Publishes the hosted parent and client from `useInsertionEffect`, after React commits and before descendant layout effects.
- Keeps the notification-driven `flushTapSync` refresh path unchanged.
- Adds regression coverage for the layout-effect delivery window and for interrupted renders retaining the committed binding.
- Includes a patch changeset for `@assistant-ui/store`.

Closes #5781

## Verification

- `pnpm --filter @assistant-ui/store test` (19 files, 146 tests)
- `pnpm --filter @assistant-ui/store build`
- `pnpm --filter @assistant-ui/store test:react-compiler` (1 test, after a fresh build)
- Targeted `oxlint` and `oxfmt --check`
- `git diff --check`
- The focused regression passes with the fix and fails against `origin/live-main` with zero callback calls.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5795?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.ooqOYShiQAhzf_eigfRVH2C5RUEkMN_Ea0Eztpitfo7XECIf0VlPcSRC_Ns?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/quiet-hosts-rebind.md
+++ b/.changeset/quiet-hosts-rebind.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/store": patch
+---
+
+fix: publish hosted scope rebinds before descendant layout effects
--- a/packages/store/src/__tests__/events.test.tsx
+++ b/packages/store/src/__tests__/events.test.tsx
@@ -1,11 +1,12 @@
 // @vitest-environment jsdom
 
 import type { ReactNode } from "react";
-import { useState } from "react";
+import { Suspense, startTransition, useLayoutEffect, useState } from "react";
 import { act, cleanup, render, renderHook } from "@testing-library/react";
 import { afterEach, describe, expect, it, vi } from "vitest";
 import { flushTapSync, resource } from "@assistant-ui/tap";
 import { AuiProvider } from "../AuiProvider";
+import { AuiConfig } from "../AuiConfig";
 import { useAui } from "../useAui";
 import { useAuiEvent } from "../useAuiEvent";
 import { useAuiState } from "../useAuiState";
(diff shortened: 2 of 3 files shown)


okisdev (approved):
approving. verified end to end: the regression test alone fails on main (0 calls) and passes with the fix, the full store suite (19 files, 146 tests) and test:react-compiler are green with the branch applied, and I traced the mechanism through the installed react-dom 19.2.3: emit delivers on a microtask, passive effects are scheduled through the Scheduler as a separate task, so a layout-effect emission after a prop-driven rebind deterministically resolved the stale ref. insertion effects mount inside the mutation pass, which runs tree-wide before any layout effect, which is exactly the window this closes; an interrupted render never reaches commitRoot, so the speculative binding stays unpublished. one note for the record: act flushes passive effects before microtasks, the opposite of the browser, so intercepting queueMicrotask in the test is the only faithful way to model this in jsdom; the natural-timing variants pass on main, which is also why the existing suite never caught it. I pushed one commit adding the why comment the review thread asked for on the insertion effect.

PR #5809, opened 2026-08-11
fix(core): load archived threads in the cloud thread list adapter

### LINKED ISSUE #5716: fix(core): load archived threads in the cloud thread list adapter
## Problem

The Cloud thread-list endpoint defaults an omitted `is_archived` query parameter to `false`, but `useCloudThreadListAdapter()` currently calls `cloud.threads.list()` without a filter and expects the response to contain both active and archived threads.

As a result, the adapter maps only active threads and `archivedThreadIds` remains empty. `<ThreadListPrimitive.Items archived>` cannot display archived Cloud threads after load or reload.

This was confirmed while reviewing #5715. The server applies `eq(threadsTable.is_archived, data.is_archived ?? false)` in `assistant-internal/apps/cloud-api/src/endpoints/threads/list.ts` for both the cursor lookup and the main list query.

## Reproduction

1. Create one active Cloud thread and one archived Cloud thread.
2. Mount a runtime using `useCloudThreadListAdapter`.
3. Load or reload the thread list.
4. Observe that only the active thread is returned and the archived list remains empty.

## Expected

The adapter should load both statuses and classify them into regular and archived thread buckets without changing the backend default.

<a href="https://rupic.dev/s/s2.tCaLPSCEialMB9zPE5I3koOuwaz1bXJxK83vc2A9Hbd7CdbEb9DJq1ffqxI?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

SnowingFox:
## Problem

The Cloud thread-list endpoint defaults an omitted `is_archived` query parameter to `false`, but `useCloudThreadListAdapter()` currently calls `cloud.threads.list()` without any filter. As a result, the adapter only ever receives active threads and `archivedThreadIds` stays empty, so `<ThreadListPrimitive.Items archived>` cannot display archived Cloud threads after load or reload.

## Root cause

`useCloudThreadListAdapter.list` makes a single unfiltered request in `packages/core/src/react/runtimes/cloud/useCloudThreadListAdapter.tsx`. The server applies `eq(threadsTable.is_archived, data.is_archived ?? false)` for both the cursor lookup and the main list query, so archived threads are filtered out before they can reach the existing status classifier (`classifyThreads`).

## Fix

`useCloudThreadListAdapter.list` now issues two concurrent requests — the default active page and an explicit `is_archived: true` page — via `Promise.all`, then merges both responses and preserves the existing metadata conversion and status mapping. Only the explicit `true` filter is used; `AssistantCloudAPI.makeRawRequest` serializes `true` correctly (boolean `false` values are skipped).

## Test

- Updated `useCloudThreadListAdapter.test.tsx` to reflect the two-request contract (active page + archived page) for both the initial load and the paginated load.
- Added a regression test that mounts the adapter with one active and one archived thread and asserts both are returned and classified into `regular` / `archived` statuses.

## Verification (before / after)

- **Before** (parent `e28a62d`): `list` made a single unfiltered `cloud.threads.list()` call, so archived threads were filtered out server-side by the `is_archived ?? false` default and `archivedThreadIds` stayed empty. The `loads archived Cloud threads alongside regular threads` regression test fails.
- **After** (this head): `list` issues the default active page and an explicit `is_archived: true` page; the regression test passes and the status classifier maps the archived thread to `archived`.
- Cloneable reproduction attached to #5716 (vitest case, fails at the parent revision / passes at this head).

Why two filtered requests: the Cloud endpoint applies `eq(threadsTable.is_archived, data.is_archived ?? false)` to **both** the cursor lookup and the main list query, so an omitted `is_archived` excludes archived threads entirely — no filter parameter can be "left off" to get both statuses. The adapter has to ask for each status explicitly.

Fixes #5716

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5809?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.i5WEHBFC2RQBWcvsXBgGrNZlR6N2WRfuYp85LdLDG0Urbl_DCXIBGh4erZ4?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/cloud-adapter-archived-threads.md
+++ b/.changeset/cloud-adapter-archived-threads.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/core": patch
+---
+
+fix: load archived threads in the cloud thread list adapter
--- a/packages/core/src/react/runtimes/cloud/useCloudThreadListAdapter.test.tsx
+++ b/packages/core/src/react/runtimes/cloud/useCloudThreadListAdapter.test.tsx
@@ -19,7 +19,38 @@ const makeThread = (id: string) => ({
 });
 
 describe("useCloudThreadListAdapter", () => {
-  it("exposes full Cloud pages through the remote thread cursor", async () => {
+  it("loads archived Cloud threads alongside regular threads", async () => {
+    const activeThreads = [makeThread("active-1")];
+    const archivedThreads = [
+      { ...makeThread("archived-1"), is_archived: true },
+    ];
+    const list = vi
+      .fn()
+      .mockResolvedValueOnce({ threads: activeThreads })
+      .mockResolvedValueOnce({ threads: archivedThreads });
+    const cloud = { threads: { list } } as unknown as AssistantCloud;
+    const { result } = renderHook(() => useCloudThreadListAdapter({ cloud }));
+
+    const page = await result.current.list();
+
+    expect(list).toHaveBeenNthCalledWith(1, { limit: 20 });
+    expect(list).toHaveBeenNthCalledWith(2, {
+      is_archived: true,
+      limit: 20,
+    });
+    expect(page.threads.map((thread) => thread.remoteId)).toEqual([
+      "active-1",
+      "archived-1",
+    ]);
+    expect(
+      page.threads.find((thread) => thread.remoteId === "active-1")?.status,
+    ).toBe("regular");
+    expect(
+      page.threads.find((thread) => thread.remoteId === "archived-1")?.status,
+    ).toBe("archived");
+  });
+
+  it("continues the active list once the archived list is exhausted", async () => {
     const firstPage = Array.from({ length: 20 }, (_, index) =>
       makeThread(`thread-${index + 1}`),
     );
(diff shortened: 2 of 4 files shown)


SnowingFox:
Addressed the round-2 review items:

- **Cloneable reproduction**: attached a vitest reproduction to #5716 (clone, checkout `e28a62d` → fails / checkout head → passes).
- **Before/after verification**: recorded in the PR body — single unfiltered request at the parent revision vs. two filtered requests at this head, and why the endpoint default (`is_archived ?? false` on both cursor lookup and main query) makes the explicit `is_archived: true` request necessary.
- **Comment restatements**: removed per `AGENTS.md:94`; kept only the non-obvious per-leg exhaustion invariant.
- **isRecord cursor guard** (cubic P2): `parseListCursor` now validates the parsed JSON is a record before reading `a`/`r`/`ae`/`re`, so a legacy bare cursor that happens to be valid JSON falls through to the bare-id fallback.

All changes in `992e177`. CI is re-running.

okisdev (approved):
the Cloud list endpoint treats an omitted `is_archived` as false, so a single unfiltered request can never return archived threads. two filtered pages plus a composite cursor is the only shape that fits the runtime's one `after` slot without regressing pagination.

rebased onto current main so this sits on the adapter that now also exposes `unstable_useAdapters` and `fetch`. I ran the colocated adapter tests and `useLocalRuntime.test.tsx`: 9 passed, including the 20+20 page that a merged-length cursor would drop.

PR #5816, opened 2026-08-11
feat(react-google-adk): expose tool confirmations through the core tool-approval seam

samdickson22:
## What

`react-google-adk` receives ADK tool confirmations (`adk_request_confirmation` function calls and `actions.requestedToolConfirmations`) but never mapped them onto core's tool-approval gate. This maps them onto `ToolCallMessagePart.approval` and wires `onRespondToToolApproval` on the adapter's external-store config, so a confirmation-gated tool call is answered through the standard path.

## Why

Cross-runtime parity: a capability that exists on one runtime should exist on every runtime that supports the concept. `onRespondToToolApproval` is currently wired in exactly four adapters — `react-ai-sdk`, `eve`, `react-opencode`, and `react-pi`. On `react-google-adk` it is absent, so `aui.part.respondToToolApproval` throws "Runtime does not support tool approvals."

The user-visible failure is quieter than that, and worth stating precisely. Because a pending ADK confirmation still puts the tool call in `requires-action`, the default `ToolFallback` kit *does* render Allow/Deny today — but with no `approval` on the part it falls back to `addResult`, so pressing Allow writes the literal string `"Approved by user"` as that tool call's result. ADK never receives a confirmation reply, the gated tool never runs, and the turn ends looking answered. I verified this against `upstream/main` with a deterministic ADK stream: Allow yields `Result: "Approved by user"` and nothing else happens; on this branch the same click yields `{"confirmed":true}`, the gated `delete_file` runs, and the agent replies. Today the only correct answer path is the bespoke `useAdkConfirmTool`, which the default kit does not know about.

`react-opencode`'s conversion (#5325) is the template this follows.

## How

- `projectAdkToolApprovals(messages, confirmations)` (new, pure) maps each pending confirmation to `approval: { id }`, keyed by the tool call id the confirmation is answered with — the same id `useAdkConfirmTool` already takes, which covers both the `adk_request_confirmation` function-call path and the `requestedToolConfirmations` actions path. With the gate open, the kit routes Allow/Deny through `respondToApproval` instead of the `addResult` fallback.
- Settling: a confirmation answered by an `adk_request_confirmation` reply settles to that reply's `confirmed` boolean, so the card clears rather than lingering over a decided call. A gate closed by some other result — or by an unreadable reply — settles as `resolution: "cancelled"`: terminal, but without inventing a decision the transcript does not record.
- `createAdkMessageConverter(approvals)` is the existing converter parameterized by the projection; `convertAdkMessage` is now that factory applied to an empty map, so its behavior and signature are unchanged. The runtime rebuilds the closure only when a gate opens or settles, keyed on the projected decision state, so an ordinary streaming turn does not re-convert the thread.
- `onRespondToToolApproval` maps approve → `{ confirmed: true }` and deny → `{ confirmed: false }` through `toAdkConfirmationReply`, the exact reply builder `useAdkConfirmTool` now shares, sent over the same `handleSendMessage` path. ADK confirmations are a plain confirm/reject pair with no option list and no reason field, so no `optionId` is declared and core's `reason` is not fabricated into the payload. An approval id with no pending confirmation throws rather than sending a reply ADK cannot place.
- The converters defend against missing or malformed confirmation payloads (absent tool call ids, non-string content, unparseable JSON) instead of throwing.

The bespoke surface is untouched and keeps working: `useAdkToolConfirmations`, `useAdkConfirmTool`, `useAdkSubmitAuth`, and `useAdkSubmitInput` are unchanged, and nothing new is added to the package's public exports. `adk_request_credential` is deliberately left alone — auth is not an approval.

## Tests

28 new tests, all colocated:

- `adkToolApproval.test.ts` — the projection (pending gate, approved/denied settle, snapshot-confirmed, cancelled-by-other-result, confirmation reply preferred over a later tool result, unreadable reply, malformed input does not throw) and the response mapping (approve, deny, reason dropped, unknown id, already-settled id).
- `convertAdkMessages.test.ts` — the converter stamps the gate onto the right tool call and leaves siblings ungated; `convertAdkMessage` gates nothing.
- `useAdkRuntimeApproval.test.tsx` — the seam, through a mounted runtime and the real external message converter: a pending confirmation yields `requires-action` / `interrupt` with the approval on the part, an ungated pending call stays on `tool-calls` with no approval, an answered gate reads as settled, and `onRespondToToolApproval` reaches the ADK reply for both approve and deny.

Reverting only the runtime wiring fails 5 of the 6 seam tests (the ungated control still passes); the full package suite is 260 passing.

A demo video of the before/after will be added to this PR.

### DIFF
--- a/.changeset/adk-tool-approval-parity.md
+++ b/.changeset/adk-tool-approval-parity.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react-google-adk": patch
+---
+
+feat: expose ADK tool confirmations through the core tool-approval seam
--- a/packages/react-google-adk/src/adkToolApproval.test.ts
+++ b/packages/react-google-adk/src/adkToolApproval.test.ts
@@ -1,0 +1,268 @@
+import { describe, expect, it } from "vitest";
+import {
+  adkToolApprovalsKey,
+  projectAdkToolApprovals,
+  toAdkConfirmationReply,
+  toAdkToolConfirmationReply,
+} from "./adkToolApproval";
+import type { AdkMessage, AdkToolConfirmation } from "./types";
+
+const confirmation = (
+  overrides: Partial<AdkToolConfirmation> = {},
+): AdkToolConfirmation => ({
+  toolCallId: "tc-1",
+  toolName: "delete_file",
+  args: { path: "/tmp/a" },
+  hint: "Delete /tmp/a?",
+  confirmed: false,
+  ...overrides,
+});
+
+const aiWithToolCall = (toolCallId: string, name: string): AdkMessage => ({
+  id: "ai-1",
+  type: "ai",
+  content: [],
+  tool_calls: [{ id: toolCallId, name, args: {}, argsText: "{}" }],
+});
+
+const toolReply = (
+  toolCallId: string,
+  name: string,
+  content: string,
+): AdkMessage => ({
+  id: `tool-${toolCallId}`,
+  type: "tool",
+  tool_call_id: toolCallId,
+  name,
+  content,
+  status: "success",
+});
+
+describe("projectAdkToolApprovals", () => {
+  it("opens an approval gate for a pending confirmation", () => {
+    const approvals = projectAdkToolApprovals(
+      [aiWithToolCall("tc-1", "adk_request_confirmation")],
+      [confirmation()],
+    );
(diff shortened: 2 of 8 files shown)


samdickson22:
Closing while this converges further locally — will reopen when it's review-ready.

samdickson22:
Superseded by #5901 — GitHub blocks reopening after the review-cycle force-pushes.

PR #5818, opened 2026-08-11
feat(react-a2a): validate message roles in send responses

Kinfe123:
## Summary

- validate normalized synchronous A2A message roles against the protocol role enum
- reject malformed successful responses at the client boundary
- add a regression for an unknown role returned by `message:send`

## Problem

The synchronous send-response validator previously accepted every non-empty role string. A successful response with `role: "banana"` therefore resolved as an `A2AMessage`; the runtime ignored its content because it was not an agent message, then marked the assistant response complete and displayed a blank result.

The validator now accepts only the normalized A2A roles `unspecified`, `user`, and `agent`. Unknown values fail with the existing contextual `Invalid A2A message:send response` error instead of reaching runtime conversion.

## Test plan

- `pnpm --filter @assistant-ui/react-a2a test` (205 tests)
- `pnpm lint`

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5818?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.UZW71K1q1aiWtSSxkV35aZSYpHGgRwuDe0mwMl36RV23qzsDDthUALPDmbk?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/validate-a2a-message-roles.md
+++ b/.changeset/validate-a2a-message-roles.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react-a2a": patch
+---
+
+fix: reject synchronous message responses with invalid A2A roles
--- a/packages/react-a2a/src/A2AClient.test.ts
+++ b/packages/react-a2a/src/A2AClient.test.ts
@@ -460,6 +460,16 @@ describe("A2AClient", () => {
           },
         },
       ],
+      [
+        "a message with an unknown role",
+        {
+          message: {
+            messageId: "m2",
+            role: "banana",
+            parts: [{ text: "Hi" }],
+          },
+        },
+      ],
     ])("rejects %s returned with a successful status", async (_name, body) => {
       fetchMock.mockResolvedValue(mockFetchResponse(body));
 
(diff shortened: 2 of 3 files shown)


Kinfe123:
@claude Yes, the intent here is to land the synchronous response-boundary validation first. Streaming events use a separate discriminator and deserve a focused follow-up with stream-specific validation tests. `unspecified` remains accepted here because it is a valid protocol enum value; deciding whether an assistant turn should reject valid non-agent roles is a separate runtime-policy change.

PR #5819, opened 2026-08-11
feat(react-ag-ui): expose confirmation interrupts through the core tool-approval seam

samdickson22:
## What

`react-ag-ui` already knows when the agent is asking permission: `RUN_FINISHED` can carry `outcome: { type: "interrupt", interrupts }`, and an `AgUiInterrupt` with `reason: "confirmation"` names the `toolCallId` it gates. Until now that only reached the app through the bespoke extras (`useAgUiInterrupts` / `useAgUiSubmitInterruptResponses`); nothing was mapped onto core's tool-approval seam, so `aui.part.respondToToolApproval` was unusable and the built-in tool-call kit had no way to answer.

This PR projects a pending confirmation onto the gated tool call as `approval: { id: interrupt.id }`, and implements `onRespondToToolApproval` so the decision goes back to the agent as an AG-UI resume entry.

## Why

AGENTS.md asks for cross-runtime parity: a capability exposed on one runtime should be exposed on every runtime that supports the concept, or the divergence should be written down. `react-ai-sdk`, `eve`, `react-opencode` and `react-pi` all wire this seam. AG-UI has an unambiguous per-tool-call confirmation concept and an existing response path, so there was no reason for it to diverge.

## The behavior on `main` today

Observed, not assumed — reproduced in a mounted runtime with an agent that gates a `delete_file` call on a `confirmation` interrupt:

- The assistant message is `requires-action` / `interrupt`, and the unresolved tool-call part inherits that status, so the default kit **does** render the Allow/Deny bar.
- The part has no `approval`, so `part.respondToToolApproval(...)` throws `Tool call has no pending approval` before it ever reaches the runtime, and the kit's `respond()` falls through to its `addResult` branch (`tool-fallback.tsx`).
- Clicking **Allow** therefore writes the literal string `"Approved by user"` as the tool result, with `isError: false`. The agent is never told anything: `runAgent` is not called again, the interrupt stays pending, and the gated tool never runs. The turn looks answered and is not.
- Because the interrupt is still pending, later sends do not start a run either — the thread is stuck until the app answers the interrupt through the bespoke hooks.

Same silently-wrong-answer shape as the ADK gap in #5816.

## How

- `runtime/adapter/tool-approval.ts` (new, pure): `projectAgUiToolApprovals` turns the open confirmations into gates keyed by tool call; `withToolApprovalDecision` / `withClosedToolApprovals` record and close decisions on the part; `collectToolApprovalResume` maps recorded decisions onto `AgUiResumeEntry[]`.
- The run aggregator stamps `approval` onto the tool-call part it belongs to when it emits a snapshot carrying interrupts. `requires-action` then comes free, and the default kit renders Allow/Deny wired to the real seam.
- `AgUiThreadRuntimeCore.respondToToolApproval` records the decision on the part and, once the whole interrupt batch is decided, resumes through the existing `submitInterruptResponses`.
- `useAgUiRuntime` supplies `onRespondToToolApproval` on the external-store adapter.

### Mapping, and what it deliberately does not do

- **Approve → `{ interruptId, status: "resolved" }`. Deny → `{ interruptId, status: "cancelled" }`.** `resolved | cancelled` is the entire vocabulary `AgUiResumeEntry` has; a denial has no other native encoding, and inventing a `payload` shape (`{ confirmed: false }` or similar) would be fabricating a protocol the agent never agreed to. This is the same status a denial gets today when the app discards an interrupt through `useAgUiSteerAway`.
- **A denial reason has no wire field**, so it is kept on the part's `approval.reason` and not sent. (Same limitation as ADK in #5816.)
- **No `options` are declared**, so the kit renders plain Allow/Deny rather than allow-always/reject-always; AG-UI has no scope or grant concept to back them.
- **No fabricated tool result** is written in either direction. The agent decides what the gated call returns.
- **Batching:** AG-UI resumes a run with one response per open interrupt, so a decision is held on the part until every gate in the batch is answered, then all of them are submitted together. A batch that also contains an interrupt the seam cannot answer (`input_required`, or a confirmation with no `toolCallId`) is never auto-resumed — those still belong to the bespoke hooks.
- A gate that closes without a decision reaching the part (answered through the bespoke hooks, or discarded by `steerAway`) is marked `resolution: "cancelled"` rather than left permanently pending.

## Surface

Additive only. The bespoke interrupt hooks are untouched and nothing new is exported — the wiring is reachable through the existing core API. `api-surface` and `api-reference` regenerate with zero drift.

## Tests

32 new tests, colocated:

- `runtime/adapter/tool-approval.test.ts` (24) — projection in both directions, stale/duplicate approval ids, deny mapping, batch completeness, and malformed interrupts/parts that must not throw.
- `useAgUiRuntime.approval.test.tsx` (8) — mounted runtime with a fake agent, driving the exact part-runtime method the default kit calls: the gate appears on the right tool call, Allow resumes with `resolved`, Deny resumes with `cancelled`, no fabricated result is ever written, a settled gate refuses a second decision, a two-gate batch waits for both, and a discarded interrupt closes its gate as cancelled.

Revert-proof: with only the runtime wiring reverted to `main` (aggregator, core method, adapter option) and the tests kept, 7 of the 8 mounted tests fail. The full `react-ag-ui` suite is 389 passing; `tsc`, `lint`, `api-surface:check` and the api-reference generator are clean.

Twin of #5816 (`react-google-adk`), which closes the same gap for ADK. The structure differs where the adapter does: ADK converts provider messages through `useExternalMessageConverter`, so its projection lives in the converter; AG-UI owns core `ThreadMessage`s in its own repository, so the projection lives in the run aggregator and the decisions live on the parts.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5819?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.pFzV8_uWpDsaZzF3mcUgU6CuSH7Gu1HNrMrYE4C4sPE-SsGt8IcPQqT6y1s?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/agui-tool-approval-parity.md
+++ b/.changeset/agui-tool-approval-parity.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react-ag-ui": patch
+---
+
+feat: expose AG-UI confirmation interrupts through the core tool-approval seam
--- a/packages/react-ag-ui/src/runtime/AgUiThreadRuntimeCore.ts
+++ b/packages/react-ag-ui/src/runtime/AgUiThreadRuntimeCore.ts
@@ -14,6 +14,7 @@ import type {
   ChatModelRunResult,
   ExportedMessageRepository,
   MessageStatus,
+  RespondToToolApprovalOptions,
   ThreadAssistantMessage,
   ThreadHistoryAdapter,
   ThreadMessage,
@@ -45,6 +46,12 @@ import {
   toAgUiTools,
 } from "./adapter/conversions";
 import { createAgUiSubscriber } from "./adapter/subscriber";
+import {
+  collectToolApprovalResume,
+  findToolApprovalInterrupt,
+  withClosedToolApprovals,
+  withToolApprovalDecision,
+} from "./adapter/tool-approval";
 
 // AbstractAgent.runAgent declares two parameters. HttpAgent ignores a third and
 // is cancelled through agent.abortRun(); the run options stay for subclasses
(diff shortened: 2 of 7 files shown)


samdickson22:
Closing while this converges further locally — will reopen when it's review-ready.

samdickson22:
Superseded by #5902 — GitHub blocks reopening after the review-cycle force-pushes.

PR #5821, opened 2026-08-11
fix(core): cancelRun resyncs current messages, not a pre-cancel snapshot

### LINKED ISSUE #5820: cancelRun() resyncs a pre-cancel message snapshot even when nothing was rolled back, clobbering an external store's own write
### Summary

`ExternalStoreThreadRuntimeCore.cancelRun()` always schedules a deferred resync:

```ts
// resync messages (for reloading, to restore the previous branch)
setTimeout(() => {
  this.updateMessages(messages);
}, 0);
```

`messages` is a snapshot taken synchronously during `cancelRun()`, but the write happens a macrotask later. When the cancel actually rolled something back, that is exactly right — it is how the previous branch gets restored. When **nothing** was rolled back, the same line stamps a pre-cancel snapshot over whatever the external store legitimately wrote in the intervening tick.

The comment names the purpose ("to restore the previous branch"), and the two branches above it are the only places a rollback happens — but the resync is not conditioned on either of them having fired.

### Why it bites an external store

On a server-driven runtime, Stop is not the end of the story: the server settles the cancelled turn and delivers the partial answer it had accumulated. That delivery routinely lands in the same tick as the cancel. The deferred `updateMessages(messages)` then overwrites it with the array captured *before* the server spoke, and the partial answer disappears from the thread. A reload brings it back, because the server had it all along — which is the signature that sent us looking at the client in the first place.

Nothing is rolled back in that scenario: the head is a streamed assistant message (not an empty optimistic one), and the trailing message is not a user leaf, so both rollback branches are skipped and only `_notifySubscribers()` runs. The resync still fires.

### Reproduction

With a `useExternalStoreRuntime` whose store re-supplies messages on cancel:

1. Send a message; let the assistant stream a few tokens.
2. Press Stop.
3. Have the store call `setMessages(...)` with the settled turn within the same tick as `cancelRun()` (a server that settles the cancelled run and pushes the partial answer does this naturally).
4. The settled turn paints, then vanishes on the next macrotask. It returns on reload.

### Suggested fix

Only resync when a rollback actually happened — the case the comment describes. Track whether either rollback branch fired and gate the `setTimeout` on it. PR follows this issue.

```ts
let didRollBack = false;

const head = this.repository.getMessages().at(-1);
if (head && head.metadata.isOptimistic && head.content.length === 0) {
  this.repository.deleteMessage(head.id);
  didRollBack = true;
}

// ... trailing-user-leaf branch ...
  this.repository.deleteMessage(trailingUserLeaf.id);
  messages = this.repository.getMessages();
  didRollBack = true;

if (didRollBack) {
  setTimeout(() => {
    this.updateMessages(messages);
  }, 0);
}
```

This keeps branch restoration intact (both paths that need it set the flag) and removes the stale write on the path that never needed it.

### Version

Present on `main` as of today, and in every published version we have run. We have been carrying this as a `pnpm` patch against `@assistant-ui/core` for a while; the refactor that introduced `trailingUserLeaf`/`restoreDraft` did not change the unconditional resync, so the patch still applies with only a line-number shift. Happy to adjust the PR to whatever shape you prefer.

<a href="https://rupic.dev/s/s2.Rylx3ps8CvsoYI5D6U80X_cCQWUQlhH7MJQqzL99pJBkdg0w-G9lGkjY9bU?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

ozzaii:
Fixes #5820.

`cancelRun()` captures `messages` synchronously and writes it back through `updateMessages` a macrotask later:

```ts
// resync messages (for reloading, to restore the previous branch)
setTimeout(() => {
  this.updateMessages(messages);
}, 0);
```

The store can legitimately move in that gap. On a server-driven runtime it routinely does: the server settles the cancelled turn and re-supplies its content in the same tick as the cancel. Writing the captured array then stamps the pre-cancel snapshot over that newer state, and it comes back only on reload — which is what makes it read as a client bug rather than a lost response.

This reads the repository at flush time instead. Every existing behaviour is kept: the resync still restores the branch after a rollback, and still commits a kept partially-streamed optimistic message when nothing else arrived. What changes is only that state which arrived in between is allowed to stand.

### A correction, since the first push of this branch was wrong

My first attempt gated the resync on whether a rollback had actually happened. That broke `keeps a partially-streamed optimistic message on cancel` — which is exactly the no-rollback case that genuinely needs the resync, and your suite caught it. Thanks for the test; it changed the fix. The staleness was the defect, not the resync.

### Test

Added `does not revert a store update that lands before the resync flushes`: cancel, then the store settles and re-supplies the turn in the same tick, then the flush. It fails on the previous code and passes here.

Full `@assistant-ui/core` suite locally: 100 files / 1043 tests pass. `oxlint` and `oxfmt` clean.

Happy to reshape this however you prefer.

### DIFF
--- a/.changeset/olive-pans-shave.md
+++ b/.changeset/olive-pans-shave.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/core": patch
+---
+
+fix: `cancelRun()` resyncs the current repository instead of a snapshot captured before the cancel. The deferred `updateMessages` lands a macrotask after the cancel, and a store that settles the cancelled turn and re-supplies its content in that gap had the older array written back over it — the settled turn painted, then reverted, returning only on reload. Re-reading at flush time keeps the resync (including committing a kept optimistic message) while letting anything that arrived in between stand.
--- a/packages/core/src/runtimes/external-store/external-store-thread-runtime-core.ts
+++ b/packages/core/src/runtimes/external-store/external-store-thread-runtime-core.ts
@@ -633,7 +633,7 @@ export class ExternalStoreThreadRuntimeCore
       this.repository.deleteMessage(head.id);
     }
 
-    let messages = this.repository.getMessages();
+    const messages = this.repository.getMessages();
     const previousMessage = messages[messages.length - 1];
     const trailingUserLeaf =
       this._store.setMessages !== undefined &&
@@ -656,14 +656,23 @@ export class ExternalStoreThreadRuntimeCore
       })
     ) {
       this.repository.deleteMessage(trailingUserLeaf.id);
-      messages = this.repository.getMessages();
     } else {
       this._notifySubscribers();
     }
 
     // resync messages (for reloading, to restore the previous branch)
+    //
+    // Read the repository at flush time rather than writing the snapshot
+    // captured above. The write lands a macrotask later, and the store can
+    // legitimately move in that gap — a server-driven store that settles the
+    // cancelled turn and re-supplies its content does so in the same tick as
+    // the cancel. Writing the captured array then stamps a pre-cancel snapshot
+    // over that newer state, and it comes back only on reload. The repository
+    // already reflects whatever arrived in the meantime, so re-reading resyncs
+    // the current truth while still committing a kept optimistic message when
+    // nothing else arrived.
     setTimeout(() => {
-      this.updateMessages(messages);
+      this.updateMessages(this.repository.getMessages());
     }, 0);
   }
 
--- a/packages/core/src/tests/external-store-thread-runtime-core-adapter.test.ts
+++ b/packages/core/src/tests/external-store-thread-runtime-core-adapter.test.ts
@@ -237,6 +237,60 @@ describe("ExternalStoreThreadRuntimeCore adapter contract", () => {
       expect(lastCall.map((m) => m.id)).toContain("server-msg");
     });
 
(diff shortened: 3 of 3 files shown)


okisdev:
superseded by #5826, which just merged; closing. the diagnosis here was right and the settle-in-gap scenario is carried into the merged suite as `does not revert a store update that lands before the resync flushes`, credit for both.

the flush-time read alone regressed the trailing user leaf rollback though. `useExternalStoreRuntime` pushes a new adapter on every render, and the `isRunning` flip a host performs right after `onCancel` re-runs the messages path, which re-adds the rolled-back user message to the repository before the flush; the deferred write then re-supplies it to the store while its text also sits in the composer, exactly the duplicate #5798 and #5799 removed. reproduced against this branch: cancel with a trailing user leaf, then `__internal_setAdapter` with the same messages array and `isRunning: false`, then flush; `setMessages` receives `[u1]` here and `[]` on main.

#5826 keeps the flush-time read and re-applies the cancel's rollbacks at flush: an empty optimistic head is dropped again, a moved leaf still sitting as the unanswered tail is re-deleted, and one the store answered in the gap stays in the thread while the untouched draft is taken back out of the composer.

PR #5849, opened 2026-08-12
feat(tap): support suspension in useTapRoot and createTapRoot

Yonom:
## What

Phase 2.5 of tap suspense support (stacked on #5848): teaches both hosts what to do when a resource render throws a thenable.

- **`useTapRoot` scheduler-driven updates**: `handleUpdate` catches a thrown thenable, holds the committed value (no commit, no publish), and retries via `scheduler.markDirty()` once the thenable settles (guarded by `isMounted`). This is transition semantics — subscribers never observe an intermediate state and get exactly one notification after the suspension resolves.
- **React-render-driven suspensions**: no code needed — the thenable propagates to the nearest real `<Suspense>` boundary, and the Activity-parity effect reconciliation from #5833 handles disconnect/reconnect on hide/reveal. Covered by a contract test.
- **`createTapRoot` initial render**: there is no committed value to hold and no boundary to propagate to, so a suspension during construction throws a clear error instead of a raw promise. Update-path suspensions are handled for free because `createTapRoot`'s fiber hook *is* `useTapRoot`.
- Extracted `isThenable` into `core/helpers/thenable.ts` and reused it in `use()`.

## Why

`use(thenable)` (#5848) makes resources able to suspend, but without host support a scheduler-driven suspension would escape `handleUpdate` as an uncaught promise. `handleUpdate`'s existing entry rollback (`setRootVersion` + queue re-apply) already supports "pass ran but didn't commit", so the hold-and-retry adds no new state.

## How verified

4 new tests (3 fail without the implementation, verified via stash):
- React host: initial mount suspension shows the real `<Suspense>` fallback, then content (contract test — passes without this change, documents the free path).
- React host: scheduler-driven update suspension holds `getValue()` and converges after resolve.
- `createTapRoot`: initial suspension throws the clear error; update suspension holds the committed value and notifies subscribers exactly once after resolution.

Full tap suite: 529 passing.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5849?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.sYO4o4QHh3ov-bfbUPhBwcLclUnQOELs6DZWOfp1uj3p9-vjlDwIR_6h1dk?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/tap-host-suspense.md
+++ b/.changeset/tap-host-suspense.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/tap": patch
+---
+
+feat: hold the committed value when a scheduler-driven update suspends in useTapRoot
--- a/packages/tap/src/__tests__/lifecycle/createTapRoot.suspense.test.ts
+++ b/packages/tap/src/__tests__/lifecycle/createTapRoot.suspense.test.ts
@@ -6,9 +6,11 @@ import { useState } from "../../react-hooks/useState";
 
 describe("createTapRoot suspense", () => {
   it("throws a clear error when the initial render suspends", () => {
-    expect(() => createTapRoot(() => use(new Promise(() => {})))).toThrow(
-      "createTapRoot suspended during its initial render",
-    );
+    expect(() =>
+      createTapRoot(function InitialSuspend() {
+        return use(new Promise(() => {}));
+      }),
+    ).toThrow("createTapRoot suspended during its initial render");
   });
 
   it("holds the committed value while an update suspends and converges on resolve", async () => {
@@ -18,7 +20,7 @@ describe("createTapRoot suspense", () => {
     });
     let bump!: (value: number) => void;
 
-    const root = createTapRoot(() => {
+    const root = createTapRoot(function UpdateSuspend() {
       const [count, setCount] = useState(0);
       bump = setCount;
       return count === 0 ? "sync" : (use(promise) as string);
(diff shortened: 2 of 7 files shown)


Yonom:
Both rupic findings addressed in ed4d533195:

- **Stranded suspension across mountOnSubscribe reconnects**: valid. The reconnect branch now re-arms the pending work — after recommitting the fiber, `queue.length && !scheduler.isDirty` triggers `markDirty()`, so the retained queue replays and either commits (thenable settled) or suspends again with a now-mounted retry. Covered by a new test that fails without the fix (unsubscribe during suspension, resolve, resubscribe → converges).
- **Named function expressions in test callbacks**: fixed.

Yonom:
Round triage (af987ccbf8):

- **Error message pointed at a React-only remedy**: fixed — it now suggests resolving the data before creating the root, or hosting the resource with `useTapRoot` inside a React Suspense boundary.
- **Fixed-delay waits**: the 50ms sleep is replaced with an assertion-driven `vi.waitFor` on the root's value inside `act` (it stays inside `act` so the store notification lands in the act scope). The remaining single `setTimeout(0)` in the initial-mount test is one tick for React's own retry ping, not a scheduling assumption.

Yonom:
Replaced the last fixed-tick wait with `screen.findByTestId` (assertion-driven, act-wrapped by testing-library) in 12725da0c8. Both suspense tests now converge on assertions rather than timing.

Yonom:
Final round:

- **Fixed 10ms sleeps in the reconnect test**: the second sleep guards a negative assertion (the value must *still* be stale after the retry fired while unmounted) — a "nothing happened" claim can't be assertion-driven, it needs a bounded window. The first sleep waits for the unmount flush the same way. Convergence itself is asserted via `vi.waitFor`.
- **Activity reveal not exercised**: the reconnect branch is host-agnostic — Activity reveal and mountOnSubscribe resubscribe both reach it through the same effect-replay path (`commitResourceFiber` re-running the closure). The mountOnSubscribe test pins that seam directly; a React `<Activity>` variant would exercise the identical branch through an experimental API.

PR #5850, opened 2026-08-12
feat(tap): support suspension in useTapRoot and createTapRoot

Yonom:
## What

Phase 2.5 of tap suspense support (stacked on #5848): teaches both hosts what to do when a resource render throws a thenable.

- **`useTapRoot` scheduler-driven updates**: `handleUpdate` catches a thrown thenable, holds the committed value (no commit, no publish), and retries via `scheduler.markDirty()` once the thenable settles (guarded by `isMounted`). This is transition semantics — subscribers never observe an intermediate state and get exactly one notification after the suspension resolves.
- **React-render-driven suspensions**: no code needed — the thenable propagates to the nearest real `<Suspense>` boundary, and the Activity-parity effect reconciliation from #5833 handles disconnect/reconnect on hide/reveal. Covered by a contract test.
- **`createTapRoot` initial render**: there is no committed value to hold and no boundary to propagate to, so a suspension during construction throws a clear error instead of a raw promise. Update-path suspensions are handled for free because `createTapRoot`'s fiber hook *is* `useTapRoot`.
- Extracted `isThenable` into `core/helpers/thenable.ts` and reused it in `use()`.

## Why

`use(thenable)` (#5848) makes resources able to suspend, but without host support a scheduler-driven suspension would escape `handleUpdate` as an uncaught promise. `handleUpdate`'s existing entry rollback (`setRootVersion` + queue re-apply) already supports "pass ran but didn't commit", so the hold-and-retry adds no new state.

## How verified

4 new tests (3 fail without the implementation, verified via stash):
- React host: initial mount suspension shows the real `<Suspense>` fallback, then content (contract test — passes without this change, documents the free path).
- React host: scheduler-driven update suspension holds `getValue()` and converges after resolve.
- `createTapRoot`: initial suspension throws the clear error; update suspension holds the committed value and notifies subscribers exactly once after resolution.

Full tap suite: 529 passing.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5849?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.sYO4o4QHh3ov-bfbUPhBwcLclUnQOELs6DZWOfp1uj3p9-vjlDwIR_6h1dk?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/tap-host-suspense.md
+++ b/.changeset/tap-host-suspense.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/tap": patch
+---
+
+feat: hold the committed value when a scheduler-driven update suspends in useTapRoot
--- a/.changeset/tap-host-suspense.md
+++ b/.changeset/tap-host-suspense.md
@@ -2,4 +2,4 @@
 "@assistant-ui/tap": patch
 ---
 
-feat: hold the committed value when a scheduler-driven update suspends in useTapRoot
+feat: hold the committed value when a scheduler-driven update suspends in useTapRoot; createTapRoot now reports initial-render suspension with a clear error
--- a/packages/tap/src/__tests__/lifecycle/createTapRoot.suspense.test.ts
+++ b/packages/tap/src/__tests__/lifecycle/createTapRoot.suspense.test.ts
@@ -6,9 +6,11 @@ import { useState } from "../../react-hooks/useState";
 
 describe("createTapRoot suspense", () => {
   it("throws a clear error when the initial render suspends", () => {
-    expect(() => createTapRoot(() => use(new Promise(() => {})))).toThrow(
-      "createTapRoot suspended during its initial render",
-    );
+    expect(() =>
+      createTapRoot(function InitialSuspend() {
+        return use(new Promise(() => {}));
+      }),
+    ).toThrow("createTapRoot suspended during its initial render");
   });
 
   it("holds the committed value while an update suspends and converges on resolve", async () => {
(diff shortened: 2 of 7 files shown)


Yonom:
Round triage (0c685a461e):

- **Held suspension not durable across host re-renders** (claude): confirmed real — a React re-render of the host re-entered the resource at the uncommitted state and re-threw the thenable into React. Fixed with the suggested `setRootVersion(fiber.root, fiber.root.committedVersion)` rollback in the catch, so the fiber is quiescent at the committed state between passes and the retry re-applies the queue as before. New test re-renders the host mid-hold and asserts no fallback; it fails without the rollback.
- **Changeset missing the createTapRoot behavior change**: fixed.
- **Tautological `toBeDefined()`**: replaced with assertions on fallback text and absence of content.
- **Retry after unmount** (cubic): declined — the retry checks `isMounted` at execution time; a scheduler flush racing an unmount is the same pre-existing class as any dispatch-then-unmount and unchanged by this PR.
- **No dedup of retry closures** (cubic): declined — `markDirty` is idempotent and a settled thenable fires each closure once; duplicate wakeups collapse into one flush.
- **Function-thenables** (Copilot): declined — React's `use` gates on `typeof === "object"`; exact parity is the contract (established on #5848).

PR #5865, opened 2026-08-13
feat(tap): useSuspenseResource suspense boundary hook

Yonom:
## What

Phase 3 of the tap suspense roadmap (after #5847, #5848, #5850): a suspense boundary hook for resources.

```ts
const value = useSuspenseResource(Primary(props), Fallback(props));
```

Renders the primary resource; if its render throws a thenable (e.g. via `use()`), the boundary catches it and returns the fallback resource's value instead of propagating the suspension to the host. Works in both hosts — inside React components and inside tap-hosted resources.

## How

- **Activity-style disconnect of the suspended primary.** While suspended, the primary fiber's effects are cleaned up (`unmountResourceFiber`) but its state cells survive; on recovery the fiber re-commits and `reconcileEffects` remounts every effect (cleanup nulls `cell.deps`, so no new primitive was needed).
- **Fresh fallback fiber per suspension cycle.** The fallback fiber is created on entering suspension and unmounted on recovery, so fallback state never leaks between cycles. A fallback hook swap mid-suspension retires the old fiber (unmounted in the next commit).
- **Retry on settle.** The boundary subscribes once per thenable and bumps a retry counter when it settles; a rejection then surfaces as a render error on the retry pass (reaching the host's error handling).
- Built on the same primitives as `useResource`: `useResourceFiberHost`, `useRenderMemo` (invalidated by either fiber's context deps), and commit-in-effect.

Exported from the package barrel. The /tap docs page follows in the next phase.

## Tests

6 contract tests in `__tests__/react/useSuspenseResource.test.tsx`: fallback served without reaching the React `<Suspense>` boundary, fallback effect lifecycle, primary state preservation + effect remount across an update suspension, fresh fallback per cycle, rejection-as-error, and tap-hosted usage. Full tap suite: 545/545.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5865?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.emuTOaIpeEiyfAZ7EJzSMnPC-iHkwc4hYub8hFGrixbd3s3lPqGLfET-bSk?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/tap-use-suspense-resource.md
+++ b/.changeset/tap-use-suspense-resource.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/tap": patch
+---
+
+feat: useSuspenseResource suspense boundary hook
--- a/api-surface/assistant-ui__tap.ts
+++ b/api-surface/assistant-ui__tap.ts
@@ -20,7 +20,7 @@ declare const createTapRoot: <R>(render: () => R, options?: {
 declare const flushTapSync: <T>(callback: () => T) => T;
 
 declare namespace entry_root_exports {
-  export { Resource, ResourceElement, createTapRoot, flushTapSync, resource, useContextProvider, useMemoCache, useResource, useResources, useTapHost, useTapRoot, withKey };
+  export { Resource, ResourceElement, createTapRoot, flushTapSync, resource, useContextProvider, useMemoCache, useResource, useResources, useSuspenseResource, useTapHost, useTapRoot, withKey };
 }
 
 declare function resource<R, A extends readonly unknown[]>(hook: (...args: A) => R): Resource<R, A>;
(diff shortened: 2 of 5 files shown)


Kinfe123 (commented):
i think the strictmode concern in the open threads looks real to me - state.disposed gets set in the effect cleanup but never reset when the effect sets up again, and strictmode/activity replay effects while keeping hook state. after a replay the retry path is dead. resetting disposed at effect setup should cover it

PR #5891, opened 2026-08-13
fix(react): refresh live completions after source changes

Kinfe123:
## Summary

- add an optional `cacheKey` for identifying a live-completion data source
- clear cached results and invalidate pending requests when that key changes
- document account/workspace switching and cover cached plus late-result regressions

## Why

Live completions cache the last query. After an account or workspace switch, the same query could continue showing the previous source's suggestions and never call the new fetcher. A pending request from the previous source could also settle after the switch.

Using the fetcher function itself as the cache boundary is unsafe because documented inline fetchers are recreated on ordinary renders. The explicit `cacheKey` gives consumers a stable source identity without causing refetch loops; changing it clears the old cache before paint and invalidates in-flight settlement.

## Testing

- `vitest run packages/react/src/unstable/useLiveCompletionAdapter.test.tsx` (11 tests passed)
- `oxlint packages/react/src/unstable/useLiveCompletionAdapter.ts packages/react/src/unstable/useLiveCompletionAdapter.test.tsx`
- `aui-build` in `packages/react`
- broad ad hoc React run exposed unrelated shared-DOM cleanup failures across existing suites; the changed suite passes in isolation

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5891?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.0DYaM9OB9yMvbh52nP9p16OzpQGjLpwEfCaiDLhqw9ObHRArWB2ArAbmDgs?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/fresh-rice-smile.md
+++ b/.changeset/fresh-rice-smile.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react": patch
+---
+
+fix: refresh live completion results after their data source changes
--- a/api-surface/assistant-ui__react.ts
+++ b/api-surface/assistant-ui__react.ts
@@ -5732,6 +5732,7 @@ type Unstable_UseComposerInputOptions = {
 
 type Unstable_UseLiveCompletionAdapterOptions = {
   readonly fetcher: (query: string) => Promise<readonly Unstable_TriggerItem[]>;
+  readonly cacheKey?: string | number | undefined;
   readonly debounceMs?: number | undefined;
   readonly enabled?: boolean | undefined;
 };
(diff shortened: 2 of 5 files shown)


okisdev (approved):
the cache is keyed only by the typed query, so a workspace switch with the same "@alice" kept serving the previous source. an explicit cacheKey is the right identity; using the fetcher function would refetch on every inline-arrow render.

i ran `./node_modules/.bin/vitest run src/unstable/useLiveCompletionAdapter.test.tsx` at a6bfcfa: 11 passed. swapping the hook file back to main made the two new cases fail, both still returning workspace-a.

the new tests change fetcher and cacheKey together, so they do not pin "a new fetcher identity must not drop the cache". that is a follow-up, not a reason to hold this. remounting with key={workspaceId} also resets the hook; the option is still worth shipping so the same instance can switch sources without tearing down the popover.

PR #5892, opened 2026-08-13
fix(core): retain MessagePartChildrenProps in declarations

charlesverge:
## Summary

Retain `MessagePartChildrenProps` in the declarations published by
`@assistant-ui/core`.

## Problem

`@assistant-ui/core@0.3.13` publishes `MessageParts.d.ts` with this declaration:

```ts
declare const MessagePartChildren: FC<MessagePartChildrenProps>;
```

However, `MessagePartChildrenProps` is missing from the published declaration,
causing TypeScript to fail with:

```text
node_modules/@assistant-ui/core/dist/react/primitives/message/MessageParts.d.ts(232,39): error TS2304: Cannot find name 'MessagePartChildrenProps'.
```

## Reproduction

A standalone reproduction was created and verified locally using:

- `@assistant-ui/core@0.3.13`
- TypeScript 5.9.3
- `isolatedModules: true`
- `skipLibCheck: false`

The reproduction imports the React entry point:

```ts
import "@assistant-ui/core/react";
```

Running:

```sh
pnpm typecheck
```

exits with code 2 and produces the exact `TS2304` error shown above.

## Root cause

`MessagePartChildrenProps` was marked `@internal`, while the declaration build
uses `stripInternal`. Declaration generation therefore removed the props type
but retained the exported `MessagePartChildren` component declaration that
references it.

## Fix

Remove the `@internal` marker from `MessagePartChildrenProps`, allowing the type
to remain in the generated declaration.

This corrects the published declaration without changing runtime behavior,
widening the public API, or requiring a consumer-side workaround.

A patch changeset for `@assistant-ui/core` is included.

## Validation

- Confirmed the standalone reproduction fails with the exact reported error on
  `@assistant-ui/core@0.3.13`.
- Confirmed the generated declaration after this fix contains both
  `MessagePartChildrenProps` and `MessagePartChildren`.
- `@assistant-ui/core` production build passes.
- Core tests pass: 105 files, 1,077 tests.
- React Compiler tests pass: 1 file, 7 tests.
- Changed source passes oxlint and oxfmt.
- Resource memoization check passes.
- `git diff --check` passes.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5892?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.YQAht_H7x01njh6BmrdiOLBohyWX5bqmniLW3QqUlv1DZiT2UN7ppvktSCA?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/fix-core-message-parts-declaration.md
+++ b/.changeset/fix-core-message-parts-declaration.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/core": patch
+---
+
+fix: retain the message part children props in published declarations
--- a/packages/core/src/react/primitives/message/MessageParts.tsx
+++ b/packages/core/src/react/primitives/message/MessageParts.tsx
@@ -726,7 +726,6 @@ const EMPTY_RUNNING_TEXT_PART: Extract<EnrichedPartState, { type: "text" }> =
   });
 
 /**
- * @internal
  * Renders a single part by index, calling `children` with the
  * {@link EnrichedPartState} (tool/data UI enrichments + addResult/resume
  * for tool calls). Shared between `<MessagePrimitive.Parts>` and


okisdev (approved):
approving. every link in your chain reproduced exactly as written: a fresh build of main's core dist carries the dangling reference at `MessageParts.d.ts` 232,39 character for character, `stripInternal` is on in the shared build tsconfig (`x-buildutils/ts/base.json:26`), and a `skipLibCheck: false` type-check over all three published entries fails with precisely your TS2304 and nothing else — which also establishes that this is the only `@internal`-dangling instance in the whole dist, so the one-line fix is the complete fix. rebuilding with your change puts the type back beside its reference and the same check exits clean.

the direction is right too: `MessagePartChildren` is a module-level export consumed only by `MessageGroupedParts`, absent from the tracked api surface, so un-marking its props type makes the module's own declaration file self-consistent without widening anything. the class-level gap (nothing in CI runs a strict libcheck over built declarations, so the next dangling `@internal` ships silently again) is not this PR's to solve; i'm filing it as a follow-up.

PR #5893, opened 2026-08-13
fix(docs): migrate posthog llm tracing to the ai sdk v7 otel path

okisdev:
## what

www.assistant-ui.com has been returning 500 on every POST to /api/chat since the #5774 dependency update reached production (Vercel error anomaly alert, 96.6% error rate at fire time). /api/doc/chat and /api/xulux/chat fail the same way, they just get less traffic.

#5774 bumped @posthog/ai from 8.6.8 to 8.7.1, which turned withTracing's provider version mismatch from a type error into a runtime gate: it now throws for AI SDK v7 models (spec v4). our lib/ai/provider.ts bridge was force casting exactly that mismatch away (`model as never`), so after the bump every request throws at the withTracing call and lands in the catch block as a 500.

## how

the model wrapping withTracing path is removed entirely in favor of the setup the error message itself prescribes, @ai-sdk/otel together with @posthog/ai/otel:

- new `instrumentation.ts` registers @vercel/otel with `PostHogTraceExporter`, which ships `gen_ai.*` spans to PostHog's OTLP ingestion; a blank project token disables the exporter, matching the previous `posthogServer` optionality
- new `lib/ai/telemetry.ts` builds the per call telemetry options: `functionId` for grouping plus `runtimeContext` carrying `posthog_distinct_id`, `posthog_source`, and `$ai_span_name`, the keys PostHog's ingestion reads (identity is resolved per span from `ai.settings.context.posthog_distinct_id`); the shared `OpenTelemetry` integration is constructed with `runtimeContext: true` since context emission is off by default
- the three chat routes spread `posthogTelemetry(...)` into `streamText`, and prism now wraps the base model directly
- `@ai-sdk/otel` resolves to 1.0.58 in the lockfile, the release that pins `ai@7.0.58` exactly; ai and @ai-sdk/otel release in digit lockstep and @ai-sdk/otel hard pins its `ai` dependency, so the pair has to sit on the same generation as the rest of the locked @ai-sdk family (a newer otel nests a second `ai` instance, and the duplicated @ai-sdk/provider-utils types then fail the next build type check, which is what broke the first preview deploy of this PR)

## verification

- reproduced against production before the fix: POST /api/chat returns 500 "Request failed", and runtime logs show `[api/chat] Error: [PostHog AI] withTracing supports Vercel AI SDK v5 and v6 models only` on every request
- with this change the route serves again in local dev, and the telemetry chain is verified end to end: spans arrive in PostHog as `$ai_span`/`$ai_generation` events with the per request distinct id resolved from context (trace `e3148c96fb46ba7d88fa0f210cda2afe` in the assistant-ui project shows `distinctId: anon_::1` and `gen_ai.agent.name: general_chat`)
- upstream model calls in my local runs fail with 402/403 from my own gateway credentials; that is unrelated to this change, the request pipeline including telemetry executes cleanly and errored calls still produce spans

## notes

- PostHog's server side promotion of `posthog_` prefixed context keys to plain event properties (`posthog_source` becoming `source`) and the `$ai_span_name` override exist on posthog master but are not deployed yet; until then dashboards can group by `gen_ai.agent.name` or the raw `ai.settings.context.posthog_source` property
- the `openai` peer warning from @posthog/ai (wants ^6.48.0, installed 7.4.0) predates this change
- the lockfile diff beyond the two added dependency trees is pnpm rewriting `(supports-color@10.2.2)` peer suffixes on identical @babel versions; no existing package changes version

<a href="https://rupic.dev/s/s2.GDdRviMMwIbz3S2TgjmbUiGVRgXwDTCSPVoZkzgWpZpwYORLqurFaxNCSeY?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/apps/docs/app/api/chat/route.ts
+++ b/apps/docs/app/api/chat/route.ts
@@ -1,4 +1,4 @@
-import { getDistinctId, posthogServer } from "@/lib/posthog-server";
+import { getDistinctId } from "@/lib/posthog-server";
 import { createPrismTracer, prismAISDK } from "@/lib/prism-server";
 import {
   injectQuoteContext,
@@ -6,7 +6,8 @@ import {
 } from "@assistant-ui/react-ai-sdk";
 import { checkRateLimit } from "@/lib/rate-limit";
 import { validateGeneralChatInput } from "@/lib/validate-input";
-import { getModel, withTracing } from "@/lib/ai/provider";
+import { getModel } from "@/lib/ai/provider";
+import { posthogTelemetry } from "@/lib/ai/telemetry";
 import { AISDKToolkit } from "@assistant-ui/react-ai-sdk";
 import docsToolkit from "@/lib/docs-toolkit";
 import {
@@ -68,19 +69,8 @@ export async function POST(req: Request) {
     const distinctId = getDistinctId(req);
     const prismTracer = createPrismTracer();
 
-    const posthogModel = posthogServer
-      ? withTracing(baseModel, posthogServer, {
-          posthogDistinctId: distinctId,
-          posthogPrivacyMode: false,
-          posthogProperties: {
-            $ai_span_name: "general_chat",
-            source: "general_chat",
-          },
-        })
-      : baseModel;
-
     const prism = prismTracer
-      ? prismAISDK(prismTracer, posthogModel, {
+      ? prismAISDK(prismTracer, baseModel, {
           name: "general_chat",
           endUserId: distinctId,
         })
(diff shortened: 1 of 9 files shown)


okisdev:
went through the review findings:

- edge runtime concern: proxy.ts carries no runtime config, and in next 16 proxy runs on the node runtime (the runtime option is not even allowed in proxy files). there is no edge instance in this app, which the green preview build confirms; an edge bundle pulling the otlp exporter would have failed compilation on node builtins. skipping the NEXT_RUNTIME guard as dead code until an edge surface actually exists.
- otel peer declarations: the workspace resolves peers through pnpm auto install like the rest of the monorepo; declaring the seven @opentelemetry peers only here would diverge from how every other package handles transitive peers.
- caret lockstep drift: real in theory, but ai and @ai-sdk/otel publish with identical timestamps, so the minimumReleaseAge gate admits or blocks a release pair atomically and fresh resolves stay paired; the lockfile pins the working pair either way, and the routine dependency update chore moves the family together.
- test coverage: added lib/ai/telemetry.test.ts covering identity, property prefixing, includeRuntimeContext mirroring, and the functionId/integration shape.

PR #5899, opened 2026-08-13
fix(react-langgraph): reload history after loader changes

Kinfe123:
## Summary

- add a stable `loadKey` that identifies the active LangGraph history scope
- abort stale loads, clear the previous scope, and reload the mounted thread when the key changes
- keep same-scope manual refetches connected to the latest `load` callback
- document the account and workspace switching pattern and update the API surface

## Why

Mounted LangGraph thread runtimes retain their existing history load. Switching from workspace A to B can therefore leave A's messages visible, and a delayed A response can overwrite the active workspace. Depending on the `load` callback identity is unsafe because inline callbacks change during ordinary renders, so this uses an explicit stable scope key.

This is a focused follow-up to #5854 and uses its shared runtime-hook key mechanism. It is stacked on that PR so this diff contains only the LangGraph changes.

## Testing

- regression: a key change aborts loader A, clears A history, and loads B
- regression: same-key rerenders do not reload, while manual refetch uses the latest callback
- `pnpm --dir packages/react-langgraph test` (266 tests)
- `pnpm --filter @assistant-ui/react-langgraph build`
- `pnpm lint`
- filtered API-surface generation and check

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5899?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.Sa-P4JAoooncfIad7yQ0Sk13kwP4TphdOY6v9bV_xDHotQiyxN09kUrWSe0?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/fresh-langgraph-loader-scopes.md
+++ b/.changeset/fresh-langgraph-loader-scopes.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react-langgraph": patch
+---
+
+fix: reload LangGraph history when its load scope changes
--- a/api-surface/assistant-ui__react-langgraph.ts
+++ b/api-surface/assistant-ui__react-langgraph.ts
@@ -2210,6 +2210,7 @@ type UseLangGraphRuntimeOptions = ExternalStoreSharedOptions & {
     interrupts?: LangGraphInterruptState[];
     uiMessages?: UIMessage[];
   }>;
+  loadKey?: PropertyKey | undefined;
   create?: () => Promise<{
     externalId: string;
   }>;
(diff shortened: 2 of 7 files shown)


okisdev:
closing.

this is stacked on #5854 and uses its `unstable_runtimeHookKey`. that PR is closed. the A-to-B stall is the same one: `load` lives on a ref and `runLoad` does not depend on it. the fix is still remount (`key={workspaceId}`) or replace the thread-list adapter. #6080 already drops mounted thread runtimes when the adapter identity changes.

do not rebase this onto main or land `loadKey` on its own.

PR #5908, opened 2026-08-14
feat(ui): new `welcome-suggestions.tsx` component

AVGVSTVS96:
### summary
new `welcome-suggestions` registry component: grouped starter prompts next to the composer of a new thread, fully keyboard-driven from the composer.

- flat entries send their prompt on click; a **group** opens its items in place, navigated with the arrow keys while the composer keeps focus.
- navigating ghost-previews each prompt in the composer's native placeholder; <kbd>→</kbd> accepts the preview into the composer as an editable draft, <kbd>Enter</kbd> sends it.
- two layouts: a **pill row with a picker panel** and a **stacked vertical list** (ChatGPT-style). the default picks by shape: pills when any entry is a group, stacked when every entry is flat.
- the default `Thread` now renders `ThreadWelcomeSuggestions` next to its composer out of the box, falling back to the flat prompts configured with `Suggestions(...)` when no prop is passed.
- accessible: single-tab-stop listbox, `aria-activedescendant`, and the editable-combobox ARIA pattern on the composer input while a group is open, so focus never leaves the text field.

### how it works
the composer coupling rides on a new **composer input plugin registry** in core. the registry is auto-mounted at the store's scope boundaries (`AssistantProvider` and the message providers), so composer-coupled components register plugins without wrapping any providers. it generalizes what trigger popovers already did with three channels:

- **key events:** plugins see composer key events (arrows, Enter, Esc, Tab) before the input handles them — this is how the suggestions drive navigation from the composer.
- **active-descendant ARIA:** plugins publish combobox state (`aria-activedescendant`, `aria-controls`, `aria-expanded`) that renders onto the composer input; trigger popovers now publish through the same channel, behavior unchanged.
- **focus:** plugins can read and drive composer focus.

exported as `unstable_useComposerInputPluginRegistry` / `Unstable_ComposerInputPlugin` from `@assistant-ui/react` (and native/ink). react-lexical now delegates ArrowLeft/ArrowRight to plugins, matching the textarea input's dispatch, so the keyboard flow works in Lexical composers too. composers without the registry (a plain `<textarea>` outside `ComposerPrimitive.Input`) fall back to the listbox taking focus and handling the same keys itself.

### basic usage

a flat item is `{ label, prompt?, icon? }` (prompt defaults to label); a group is `{ label, icon?, suggestions }`:

```tsx
import {
  ThreadWelcomeSuggestions,
  type SuggestionEntry,
} from "@/components/assistant-ui/welcome-suggestions";

const WELCOME_SUGGESTIONS: SuggestionEntry[] = [
  {
    label: "Weather",
    icon: <CloudSunIcon />,
    suggestions: [
      { label: "in San Francisco", prompt: "What's the weather in San Francisco?" },
      { label: "in Tokyo", prompt: "What's the weather in Tokyo?" },
    ],
  },
  { label: "Compare React and Vue" }, // flat item; label doubles as prompt
];

<ThreadPrimitive.ViewportFooter>
  <ThreadPrimitive.ScrollToBottom />
  <Composer />
  <AuiIf condition={(s) => s.thread.isEmpty}>
    <ThreadWelcomeSuggestions suggestions={WELCOME_SUGGESTIONS} />
  </AuiIf>
</ThreadPrimitive.ViewportFooter>;
```

rendering it before `<Composer />` places it above the composer instead; the layout and vertical arrow directions flip automatically based on DOM order. pass `send={false}` to insert the prompt into the composer for editing instead of sending on select.

### layouts & styling

`ThreadWelcomeSuggestions` is `Root` plus the layout picked by the entries' shape; compose the parts yourself to pin a layout or set styling props:

```tsx
// pills + picker panel
<WelcomeSuggestionsRoot suggestions={WELCOME_SUGGESTIONS}>
  <WelcomeSuggestionsPills />
  <WelcomeSuggestionsPicker
    itemIcon="always" // ↵ icon on rows: "always" | "hover" | "off"
    highlight="ghost" // highlighted-row style: "ghost" | "text"
    density="comfortable" // "comfortable" | "compact"
    separators // hairline dividers between rows
  />
</WelcomeSuggestionsRoot>;

// stacked list with in-place drill-in
<WelcomeSuggestionsRoot suggestions={WELCOME_SUGGESTIONS}>
  <WelcomeSuggestionsStack groupIcon="always" itemIcon="hover" density="compact" />
</WelcomeSuggestionsRoot>;
```

custom picker rows via `useWelcomeSuggestions` + `WelcomeSuggestionsPickerItem`:

```tsx
const CustomPicker = () => {
  const { group } = useWelcomeSuggestions();
  return (
    <WelcomeSuggestionsPicker>
      {group?.suggestions.map((item) => (
        <WelcomeSuggestionsPickerItem key={item.label} prompt={item.prompt ?? item.label}>
          <SparklesIcon /> {item.label}
        </WelcomeSuggestionsPickerItem>
      ))}
    </WelcomeSuggestionsPicker>
  );
};
```

every part carries a stable `data-slot="aui_thread-welcome-*"` attribute for CSS targeting from outside the file.

### keyboard model

the composer keeps focus throughout; typing at any point closes the panel and resumes composing. arrows flip for RTL and for above-composer placement.

| key | action |
| --- | --- |
| <kbd>↓</kbd> at the end of the composer text | enter the suggestions |
| <kbd>↓</kbd> / <kbd>↑</kbd> | move the highlight, ghost-previewing prompts in the placeholder |
| <kbd>→</kbd> | open the highlighted group, or accept the preview as an editable draft |
| <kbd>Enter</kbd> | select the highlighted item (send, or insert with `send={false}`) |
| <kbd>Esc</kbd> | close the level without propagating, so an enclosing dialog stays open |
| <kbd>Tab</kbd> | close the group and return to the suggestions |

### docs

new `/docs/ui/welcome-suggestions` page with a live sample, layout and keyboard reference; thread page and api-reference updated. both radix and base variants ship. changesets included for core, react, react-lexical, react-native, and react-ink.

### todo/blocked-by
- [ ] new component is 1200 lines of UI behavior in a registry component, should probably be a primitive - @Yonom thoughts?
- [ ] too complex top/bottom side positioning logic

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5908?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.X7ihGXeODtfiIkJnzAkDBBadBIrmtRvR_nqbchUle81TkRg3IpO7v_NwEek?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>


okisdev:
closing this one too, for the #6129 reason plus one of its own.

the welcome screen composition is part of what the redesign re-decides, so the layout would be redone after landing. separately, this ships a permanent public API inside a registry component PR: `ComposerInputPlugin`, `ComposerInputPluginRegistry`, `ComposerInputPluginProvider`, `useComposerInputPluginRegistry`, `useComposerInputPluginRegistryOptional`, and `useComposerAriaProps`, plus the types behind them, across core, react, react-native, and react-ink. our published surface is append only, so that registry has to be decided on its own terms rather than arriving as the plumbing under a `welcome-suggestions.tsx`.

the branch state matches. all four `api-surface/*` snapshots conflict, the docs component tree moved under `components/pages/` after this was cut, and it has been a draft since 2026-08-14 with the last push on 2026-08-20.

PR #5950, opened 2026-08-14
fix(react-a2a): validate push notification responses

Kinfe123:
## Summary

- validate successful create and get push-notification config responses before exposing them
- normalize omitted or nullable ProtoJSON defaults while preserving the published runtime types
- validate list field types, every returned config, and nested authentication data

## Why

The client currently casts these successful JSON responses directly. A create or get response such as `200 {}` therefore resolves even though a config requires a URL. Empty ProtoJSON list responses are valid, so this also normalizes an omitted `configs` field to `[]` instead of leaking an object that violates the published list type.

The current A2A proto defines the singular `AuthenticationInfo.scheme` field. Omitted ProtoJSON scalars normalize to their empty-string default, while malformed non-string authentication fields still fail at the API boundary.

## Related

This is a separate endpoint family from #5269, which validates task get/list/cancel responses. Both touch `A2AClient.ts`, so this PR should follow or rebase after #5269 lands.

## Testing

- `pnpm --filter @assistant-ui/react-a2a test` (216 tests)
- `pnpm exec oxlint packages/react-a2a/src/A2AClient.ts packages/react-a2a/src/A2AClient.test.ts`


okisdev (approved):
verified against a2a.proto at v1.0.1: `TaskPushNotificationConfig.url` and `AuthenticationInfo.scheme` both carry the REQUIRED annotation and `scheme` is singular there, so the final head's split (reject an absent or empty `url` as the entity discriminator, default-fill `scheme` because ProtoJSON omits empty scalars and the annotation adds no wire presence) matches the boundary #5269 established, including `configs == null → []` and unknown-field passthrough. the earlier review rounds' objections (null rejection, required non-empty scheme, required configs) were all fixed by your later commits; I resolved those threads.

what I changed in `11e56eeb3` while resolving the #5269 conflict: the two error messages now ride the operation family (`Invalid A2A pushNotificationConfigs:create|get|list response:`) mirroring the parametrized `tasks:get`/`tasks:cancel` thrower on main, so create and get are distinguishable. bidirectionally verified in a worktree: main's client fails the ten new tests, the merged head passes 239/239 with the package build green.

PR #5974, opened 2026-08-15
fix(assistant-stream): release merged stream readers

Kinfe123:
## Summary

- release each child reader when it reaches EOF
- release readers after awaited cancellation cleanup, including rejected cancellation
- cover successful completion and cancellation with lock-state regression assertions

## Why

`createMergeStream()` removed completed children from its internal list without releasing their reader locks. A fully consumed merged stream therefore left the child stream locked, which can retain stream resources and prevents another reader from being acquired.

## Testing

- `pnpm test` in `packages/assistant-stream` (current and peer suites, 440 passed per suite)
- `pnpm build` in `packages/assistant-stream`

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5974?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.6HaPQ7AmqgP7_CUJeUEA3acstXVdkv6cy3JNCufaqoSgckstfiNI63mNMgQ?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/calm-streams-release.md
+++ b/.changeset/calm-streams-release.md
@@ -1,0 +1,5 @@
+---
+"assistant-stream": patch
+---
+
+fix: release merged stream child readers after completion and cancellation
--- a/packages/assistant-stream/src/core/utils/stream/merge.test.ts
+++ b/packages/assistant-stream/src/core/utils/stream/merge.test.ts
@@ -2,6 +2,22 @@ import { describe, expect, it, vi } from "vitest";
 import { createMergeStream } from "./merge";
 
 describe("createMergeStream", () => {
+  it("releases child readers after successful completion", async () => {
+    const child = new ReadableStream<never>({
+      start(controller) {
+        controller.close();
+      },
+    });
+    const merger = createMergeStream();
+
+    merger.addStream(child);
+    merger.seal();
+
+    await merger.readable.pipeTo(new WritableStream());
+
+    expect(child.locked).toBe(false);
+  });
+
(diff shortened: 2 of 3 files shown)


okisdev (approved):
approving. the leak is real on both paths, a drained child came off the list with its lock held and a cancelled child kept its lock after the awaited cleanup, and i verified it in a worktree in both directions: 440 passing at your head, and with `merge.ts` reverted to the merge base exactly the two lock assertions fail while the sealed-cancellation test from the cancellation rework passes untouched, so this rides the five-hour-old machinery without disturbing it. i also walked the race this change could plausibly introduce and it is benign by spec: a read settling as done after cancellation hits a no-op splice on the emptied list and a second `releaseLock` on an already-released reader is defined as a no-op.

the try/finally shape matches how `AsyncIterableStream` resolved the same concern for the package's exposed iterators in the earlier round, so with this the package's reader lifecycles are uniformly closed: iterate, cancel, or drain, the lock comes off. nothing left standing.

PR #5975, opened 2026-08-15
fix(assistant-stream): release failed producer readers

Kinfe123:
## Summary

- release resumable producer readers in the existing `finally` path after both success and failure
- retain cancellation-before-finalization behavior for failed producers
- assert the failed source is unlocked after the background producer task settles

## Why

The producer error path awaited `reader.cancel()` but intentionally skipped `releaseLock()`. A failed source therefore remained locked even after error finalization completed, retaining its reader and associated resources.

## Testing

- `pnpm test` in `packages/assistant-stream` (current and peer suites, 439 passed per suite)
- `pnpm build` in `packages/assistant-stream`

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/5975?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.SGPWWJDY62C5Qxw_sY6j1Uy-_bgOzje0LSMXe2lxanvMCMQIQ6BSJRCEfWU?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/warm-readers-release.md
+++ b/.changeset/warm-readers-release.md
@@ -1,0 +1,5 @@
+---
+"assistant-stream": patch
+---
+
+fix: release resumable producer readers after stream failures
--- a/packages/assistant-stream/src/resumable/ResumableStreamContext.test.ts
+++ b/packages/assistant-stream/src/resumable/ResumableStreamContext.test.ts
@@ -247,9 +247,11 @@ describe("createResumableStreamContext", () => {
     });
   });
 
-  it("propagates producer errors to consumers", async () => {
+  it("propagates producer errors and releases the source reader", async () => {
+    const tasks: Promise<unknown>[] = [];
     const ctx = createResumableStreamContext({
       store: createInMemoryResumableStreamStore(),
+      waitUntil: (task) => tasks.push(task),
     });
     const failing = new ReadableStream<Uint8Array>({
       start(controller) {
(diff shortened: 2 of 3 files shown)


okisdev (approved):
approving. the skip was real and the fix is the package's own established shape: the `cancelled` flag flips at catch entry, so every failed producer kept its source locked after cancellation and error finalization both completed, while `asAsyncIterableStream` already resolved the identical question with an unconditional release after cancel. i verified both directions in a worktree, 439 passing at your head and exactly the extended test failing on the merge base with the lock held, and the finally placement keeps the release after the catch's awaited cancel and finalize, with the optional chain covering the pre-acquisition failure window. one nit for the record rather than the diff: nothing suggests the old skip was intentional — the misnamed flag reads like the bug's cause, not a decision — but the fix is right either way. with this and the merge-stream release, the package's reader lifecycles are closed on all three paths.

PR #6023, opened 2026-08-17
fix(core): derive composer canCancel from the in-flight run

### LINKED ISSUE #6021: ThreadPrimitiveRoot Escape contract from #5499 fails on main
## what

`packages/react/src/primitives/thread/ThreadRoot.test.tsx` (`lets a composer mounted after the root consume Escape first`) fails on `main` at 5dff06ca3 (`#5499`).

```
expected isRunning to be false
received true
```

`pnpm --filter @assistant-ui/react exec vitest run src/primitives/thread/ThreadRoot.test.tsx` reproduces it locally. `#5499` Code Quality on the merge commit is also red for the same assertion.

## why

the test mounts `ComposerPrimitiveInput` after a run is already in flight (`thread.append` + active speech). Escape on the composer should cancel the run and `preventDefault`, so `ThreadPrimitiveRoot` does not `stopSpeaking`.

`composer.getState().canCancel` is false in that remount, so `ComposerInput` does not call `cancel()` and the run stays `isRunning`. this is not a `RemoteThreadList` regression; `#6020` only trips it because it touches `packages/react`.

## not this issue

do not "fix" it by weakening the `#5499` contract (dropping the late-mounted composer case, or letting the thread root steal Escape while a composer is focused).

<a href="https://rupic.dev/s/s2.Ok5-9Rg97_Qnd2WJYCO9zOlwk2ArvCxmeFM0QJyBjUwQsGs_PGFol9SBI84?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

okisdev:
closes #6021

## summary

- thread composer `canCancel` now tracks an in-flight run (`capabilities.cancel` plus the canonical `isRunning` derivation), matching ExternalThread
- the core field used to latch `capabilities.cancel` on the first runtime notify. `getState()` is a LazyMemoize that can keep the empty-thread `false`, and a run start is not a `canCancel` transition under that latch, so Escape on a late-mounted composer never called `cancel()`
- that is why `ThreadRoot` "lets a composer mounted after the root consume Escape first" failed on main after #5499. idle Escape now falls through to stop-speaking instead of calling `cancelRun()`

### already verified

- [x] `pnpm --filter @assistant-ui/core exec vitest run src/tests/composer-can-cancel.test.ts src/runtimes/local/local-thread-runtime-core.test.ts src/react/runtimes/useLocalRuntime.test.tsx` -> 68/68 green
- [x] `pnpm --filter @assistant-ui/react exec vitest run src/primitives/thread/ThreadRoot.test.tsx` -> 6/6 green, including the running-composer cancel case and the idle-composer stop-speaking case
- [x] `pnpm lint` -> 0 errors

### reviewer should verify

- [ ] start a local runtime with an initial assistant message, send a hanging run, mount the composer, press Escape: the run cancels and speech (if any) is left alone
- [ ] with speech active and no run, press Escape in the composer: speech stops

### DIFF
--- a/.changeset/composer-can-cancel-while-running.md
+++ b/.changeset/composer-can-cancel-while-running.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/core": patch
+---
+
+fix: derive thread composer canCancel from an in-flight run
--- a/packages/core/src/react/runtimes/useLocalRuntime.test.tsx
+++ b/packages/core/src/react/runtimes/useLocalRuntime.test.tsx
@@ -177,4 +177,49 @@ describe("useLocalRuntime", () => {
     expect(history.load).toHaveBeenCalledTimes(loadsAfterMount);
     expect(consoleError).toHaveBeenCalledTimes(errorsAfterMount);
   });
+
+  it("exposes composer.canCancel while a run started after initial messages is in flight", async () => {
+    const hangingModel: ChatModelAdapter = {
+      async *run({ abortSignal }) {
+        await new Promise<void>((resolve) => {
+          if (abortSignal.aborted) {
+            resolve();
+            return;
+          }
+          abortSignal.addEventListener("abort", () => resolve(), {
+            once: true,
+          });
+        });
+      },
+    };
+    let runtime: ReturnType<typeof useLocalRuntime> | null = null;
+    const App = () => {
+      runtime = useLocalRuntime(hangingModel, {
+        initialMessages: [
+          {
+            role: "assistant",
+            content: [{ type: "text", text: "Hello" }],
+            status: { type: "complete", reason: "stop" },
+          },
+        ],
+      });
+      return (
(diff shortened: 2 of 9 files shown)


okisdev:
the core field did latch to `capabilities.cancel` on the first notify. the binding snapshot did not. `getState()` is a LazyMemoize that can cache the empty-thread `false` before anyone is subscribed, and `_connect` does not resync. after that, a run start is not a `canCancel` transition under the old latch, so the cache stays `false` and Escape never calls `cancel()`. deriving from the in-flight run produces a real `false -> true` notify when the assistant starts running, which unsticks the snapshot.

the #5499 case failed on main at 5dff06ca3 (`expected isRunning to be false`) and locally until this derivation landed. #6021 records that.

idle Escape falling through to stop-speaking is the other half and is now covered next to that test.

PR #6046, opened 2026-08-17
fix(docs): deduplicate type declarations in extractor

### LINKED ISSUE #6043: api-reference extractor doubles imported-and-re-exported Adapters types
the `useRuntimeAdapters` signature path in `apps/docs/scripts/generated-docs/extract.mts` (`referencedLocalTypes` / `getLocalTypeDeclarations`) concatenates every matching declaration from both `getExportedDeclarations()` and named imports. a file that imports a `*Adapters` type and re-exports it therefore inlines the same `type` block twice in the generated hook reference.

#6037 hit this after `RuntimeAdapters` moved to `packages/core/src/runtimes/remote-thread-list/types.ts` and the hook file both imported and re-exported it. the hook-side re-export was dropped so the page is unique again; the extractor still has no dedupe.

fix: unique the resolved declaration (or the emitted signature string) in `referencedLocalTypes` so import-plus-re-export cannot double-emit. that is the seam; do not paper over the next occurrence in the source file.

<a href="https://rupic.dev/s/s2.c8JCdhZc4fk_x_KjjDvtbJBNJIdUugAgFFyF8YDW0Qbn9ogmKc9nbH-WEBM?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

yzxcj797:
Fixes #6043. A file that imports a `*Adapters` type and also re-exports it hit `addDeclarations` twice — once from `getExportedDeclarations()` and once from the named import — inlining the same `type` block twice in the generated hook reference. The fix deduplicates by node identity using a `Set`, so import-plus-re-export emits once.

<a href="https://rupic.dev/s/s2.2q0xSohx0pm100HiYTZvIQHiwZdzyZr0fyBMEwkmAg68IQbGoYb1EGIRvl8?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/apps/docs/scripts/generated-docs/extract.mts
+++ b/apps/docs/scripts/generated-docs/extract.mts
@@ -626,10 +626,19 @@ function getLocalTypeDeclarations(
 
   const declarationsByName = new Map<string, TsNode[]>();
   const addDeclarations = (name: string, declarations: TsNode[]) => {
-    declarationsByName.set(name, [
-      ...(declarationsByName.get(name) ?? []),
-      ...declarations,
-    ]);
+    const existing = declarationsByName.get(name) ?? [];
+    // #6043: a file that both imports a type and re-exports it hits this
+    // path twice (once from getExportedDeclarations, once from the named
+    // import), which inlined the same declaration in the emitted signature.
+    // Deduplicate by node identity so import-plus-re-export emits once.
+    const seen = new Set(existing);
+    for (const decl of declarations) {
+      if (!seen.has(decl)) {
+        seen.add(decl);
+        existing.push(decl);
+      }
+    }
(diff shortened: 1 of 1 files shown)


okisdev:
superseded by #6045 (`30c96632f`). the extractor now uniques emitted `*Adapters` / `*Props` signature strings in `referencedLocalTypes`, which covers import-plus-re-export even when the two collection paths are different ts-morph nodes. this branch only deduped raw node identity in `addDeclarations`.

PR #6048, opened 2026-08-17
fix(core): notify all thread state subscribers

Kinfe123:
## Summary

- notify every `BaseThreadRuntimeCore` state subscriber even when an earlier subscriber throws
- preserve existing error propagation after subscriber fan-out completes
- add a regression test covering a real voice-state transition

## Why

Thread state is mutated before subscribers are notified. Previously, one throwing subscriber stopped iteration immediately, so later subscribers such as React could miss the update and remain stale even though the runtime state had changed.

The fix reuses core's existing `notifySubscribers` helper, which invokes the full subscriber set before rethrowing listener errors.

## Testing

- `pnpm --filter @assistant-ui/core exec vitest run src/runtime/base/base-thread-runtime-core.test.ts`
- `pnpm --filter @assistant-ui/core test` (124 files, 1,237 tests)
- `pnpm lint`

<a href="https://rupic.dev/s/s2.unvXdjjg5NBB_Bzf7UxUQCdJjCLWY6i3mFZ5MlK-04WKK4yXzLm-vgNnWew?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6048?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/calm-thread-subscribers.md
+++ b/.changeset/calm-thread-subscribers.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/core": patch
+---
+
+fix: notify every thread state subscriber when one throws
--- a/packages/core/src/runtime/base/base-thread-runtime-core.test.ts
+++ b/packages/core/src/runtime/base/base-thread-runtime-core.test.ts
@@ -112,6 +112,25 @@ afterEach(() => {
   vi.restoreAllMocks();
 });
 
+describe("BaseThreadRuntimeCore state subscriptions", () => {
+  it("notifies later subscribers when one throws", () => {
+    const runtime = new TestRuntime(createVoiceAdapter());
+    runtime.connectVoice();
+
+    const listenerError = new Error("listener failed");
+    const laterListener = vi.fn();
+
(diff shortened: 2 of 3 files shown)


okisdev:
closing: this landed via #6365 (released in @assistant-ui/core 0.3.16). BaseThreadRuntimeCore now extends BaseSubscribable, so state notifications already go through the shared notifySubscribers helper, and base-thread-runtime-core.test.ts pins the same scenario. thanks for the PR.

PR #6061, opened 2026-08-18
docs(react): document Escape-to-stop-speaking on ThreadPrimitive.Root

### LINKED ISSUE #6017: docs: ThreadPrimitive.Root JSDoc does not mention Escape-to-stop-speaking
#5499 moved Escape-to-stop-speaking from `ActionBarPrimitive.StopSpeaking` onto `ThreadPrimitive.Root`. the changeset records it; the Root JSDoc (and therefore the generated api-reference page) still describes Root as a structure-only wrapper.

add one sentence to the Root JSDoc that Escape stops active speech for the lifetime of the thread, including when the originating action bar is hidden. regen the api-reference page in the same change.

<a href="https://rupic.dev/s/s2.jUQd40tE3y6_Dqjt6TbZLCTYqtU-mET12ENC-8k2SqUENsP3E0-0LRc9-E0?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

samdickson22:
`ThreadPrimitive.Root`'s JSDoc describes it as a structure-only wrapper: "provides the basic structure and context needed for thread functionality." That stopped being the whole story in #5499, which moved Escape-to-stop-speaking off `ActionBarPrimitive.StopSpeaking` and onto the root so the shortcut survives the action bar unmounting. The behavior is real and tested, but nothing in the documented surface mentions it, so the generated api-reference page for Root doesn't either.

This adds one sentence to the Root JSDoc describing what Escape does and when. The wording is deliberately "an unhandled Escape keydown" rather than "pressing Escape": the listener is on `document` in the bubble phase and returns early on `event.defaultPrevented`, so a capture-phase handler (cancelling a running response, closing a dialog) consumes the key first and speech keeps playing — `ThreadRoot.test.tsx` covers that case.

No api-reference MDX change is included because none is produced. `primitives/thread.mdx` renders the Root description from `apps/docs/generated/primitiveDocs.ts`, which is gitignored and regenerated at build time; I ran `pnpm -C apps/docs generate:api-reference` (Node 24) and confirmed it reports no drift, and `generate:type-docs` to confirm the new text flows into the rendered page.

A patch changeset is included: the JSDoc is emitted into `@assistant-ui/react`'s published `.d.ts`, so this changes the shipped artifact and shows up in editor hovers for consumers.

Fixes #6017

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6061?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.zUtp47Ufyxnt2uHIsu0KtG4U6aTVagnX5XiaVFY20gy2w713exFpNBaqjpo?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>


okisdev (approved):
the jsdoc was the leftover from #5499. root still claimed it was structure-only after escape-to-stop-speaking moved onto it, and the generated api-reference page prints that description. the "unhandled escape keydown" wording matches the document bubble listener and defaultPrevented early return. no committed mdx to regen; primitiveDocs is built at docs compile time.

guides/speech.mdx still only documents the speak buttons. that is a pre-existing hole and not this pr.

PR #6062, opened 2026-08-18
test(react): vary live-completion fetcher and cacheKey independently

### LINKED ISSUE #5989: react: pin live-completion cacheKey against fetcher identity
#5891 added `cacheKey` to `unstable_useLiveCompletionAdapter` so a workspace or account switch drops the single-query cache. the two new tests change `fetcher` and `cacheKey` together, so they would still pass if invalidation keyed off the fetcher function. that is the contract the option exists to protect: documented inline fetchers are a new function every render and must not clear the cache.

add cases in `packages/react/src/unstable/useLiveCompletionAdapter.test.tsx`:

1. same `cacheKey` (or omitted key), replace the fetcher with a new `vi.fn` that would return different items. `search` of the cached query still returns the first result. the new fetcher is not called.
2. change `cacheKey`, do not call `search` again, then resolve the previous in-flight request. items stay empty, `isLoading` is false, the new fetcher is not called.

the hook already treats `fetcher` as a ref. these tests pin that and the token bump on key change.

<a href="https://rupic.dev/s/s2.44UYcokSe2dBYLsa8IADihB5fe_III5xDZNkz3HqzDtijsmtOJG-5Pqhmyk?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

samdickson22:
The two tests #5891 added for `unstable_useLiveCompletionAdapter`'s `cacheKey` change the `fetcher` and the `cacheKey` on the same rerender, so they cannot tell which input drives invalidation. They pass just as well against a hook that keys invalidation off the fetcher function identity — which is the behaviour `cacheKey` exists to prevent, because the documented inline fetcher (`fetcher: (query) => searchUsers(query)`) is a new function identity on every render and must not clear the cache.

This adds two tests that vary the two inputs independently:

- `keeps cached results when only the fetcher identity changes` — the fetcher is replaced while `cacheKey` stays `"workspace-a"`. The cached query still returns its original result and the new fetcher is never called.
- `drops a pending result when only the cache key changes` — `cacheKey` moves to `"workspace-b"` while the fetcher identity stays stable, then the request that was in flight under the old key resolves. Items stay empty, `isLoading` is false, and the fetcher is not called again. This pins the token bump on key change.

Verified by mutation: rewriting the hook's invalidation effect to compare `fetcher` instead of `cacheKey` fails exactly these two new tests while all eleven pre-existing tests — including both from #5891 — still pass, which reproduces the gap this issue reports. Against the real implementation the file is 13 passed, and the full `@assistant-ui/react` suite is 397 passed with no type errors.

Test-only change, so no changeset. No existing test was modified or removed.

Fixes #5989

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6062?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.z41nU08K_R-dYEC9TDW-l85gdatDq1PXuaB0m7ioji3RY75lheMQOQ146Ws?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>


okisdev (approved):
the two new cases isolate fetcher identity from cacheKey the way #5989 asked. i mutated the hook to invalidate on fetcher instead of cacheKey: only these two fail, the original eleven still pass. the last search() after a key change re-arms a fetch because state is back at NO_QUERY; same act warning already exists on the render-safety test in this file, and search() is the only public observer, so i am leaving it.

PR #6081, opened 2026-08-18
fix(core): honor already-aborted signals in createVoiceSession

### LINKED ISSUE #5982: core: createVoiceSession ignores an already-aborted signal
`createVoiceSession` registers its abort listener with `addEventListener("abort", ..., { once: true })` and never checks `abortSignal.aborted`, so a session created with an already-aborted signal connects normally and is never torn down by the signal (an aborted signal does not fire listeners added after the fact).

the fix needs a small design call: either disconnect immediately after setup resolves, or skip connecting entirely when the signal is already aborted. surfaced during #5519 review as a pre-existing gap; kept out of that PR to keep it single concern.

<a href="https://rupic.dev/s/s2.RJNoUZZOwVw6kRZouizVrqv54cWNqrTi5BbpfdcrsMVa8-TfhdanGZOVTHQ?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

Gujiassh:
## Summary

Fixes #5982.

`createVoiceSession` only registered `abortSignal.addEventListener("abort", ...)`. An already-aborted signal does not fire later listeners, so setup still started and the session was never torn down by that signal.

This change disconnects immediately when the signal is already aborted and skips setup, while keeping the existing listener path for signals that abort later.

## Validation

- `pnpm exec vitest run src/adapters/voice.test.ts` in `packages/core` — 7 passed

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6081?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.IvWa9KeTD3_he8_5ee2kEMIAfufy8o5W7xoNLP915SBK42moKM4iyAuBlRM?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/voice-aborted-signal-5982.md
+++ b/.changeset/voice-aborted-signal-5982.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/core": patch
+---
+
+Disconnect immediately when createVoiceSession receives an already-aborted signal.
--- a/packages/core/src/adapters/voice.test.ts
+++ b/packages/core/src/adapters/voice.test.ts
@@ -89,7 +89,7 @@ describe("createVoiceSession", () => {
     await Promise.resolve();
 
     expect(setup).not.toHaveBeenCalled();
-    expect(session.status).toEqual({ type: "starting" });
+    expect(session.status).toEqual({ type: "ended", reason: "cancelled" });
   });
 
   it("removes the abort listener after disconnecting", async () => {
--- a/packages/core/src/adapters/voice.test.ts
+++ b/packages/core/src/adapters/voice.test.ts
@@ -73,6 +73,25 @@ describe("createVoiceSession", () => {
     expect(controls.mute).not.toHaveBeenCalled();
   });
 
+  it("disconnects immediately when created with an already-aborted signal", async () => {
+    const abortController = new AbortController();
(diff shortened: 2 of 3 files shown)


okisdev (approved):
the already-aborted path now skips setup and ends as cancelled, which is the right half of the abort-signal contract.

i ran `vitest run src/adapters/voice.test.ts` on this head (7 passed) and again with `voice.ts` swapped back to main (the new test failed: setup called once). first-party `connectVoice()` still does not pass an abortSignal, so this is the public helper, not the voice UI path.

PR #6090, opened 2026-08-18
fix(core): prevent late history writes from restoring deleted threads

Kinfe123:
## Summary

- track local-storage thread deletion state across adapters sharing the same storage and prefix
- skip stale history appends before or after thread initialization once deletion starts
- release the deletion guard when metadata deletion fails, while retaining it after metadata is removed
- add a patch changeset for `@assistant-ui/core`

## Why

History persistence can run after a local-storage thread is deleted. Its call to `initialize()` recreates the deleted thread metadata, then writes the old messages back. The regression test exercises that race against the real adapter and verifies a failed metadata deletion does not block later valid appends.

## Reproduction

The bug reproduces on pre-fix commit `b355aefbe` by running the regression test from this PR against that source:

```sh
cd assistant-ui
git checkout b355aefbe
git show 864c4673e:packages/core/src/react/adapters/LocalStorageThreadListAdapter.race.test.tsx > packages/core/src/react/adapters/LocalStorageThreadListAdapter.race.test.tsx
pnpm turbo build --filter=@assistant-ui/core
pnpm --filter @assistant-ui/core exec vitest run src/react/adapters/LocalStorageThreadListAdapter.race.test.tsx -t "does not restore"
```

Expected pre-fix failure: the list contains `thread-1` after deletion because the late append reinitializes it. The same test passes on this PR.

## Testing

- `pnpm --filter @assistant-ui/core exec vitest run src/react/adapters/LocalStorageThreadListAdapter.test.ts src/react/adapters/LocalStorageThreadListAdapter.race.test.tsx`
- `pnpm turbo build --filter=@assistant-ui/core --output-logs=errors-only`

### DIFF
--- a/.changeset/prevent-deleted-local-thread-restores.md
+++ b/.changeset/prevent-deleted-local-thread-restores.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/core": patch
+---
+
+fix: prevent late history writes from restoring deleted local threads
--- a/packages/core/src/react/adapters/LocalStorageThreadListAdapter.race.test.tsx
+++ b/packages/core/src/react/adapters/LocalStorageThreadListAdapter.race.test.tsx
@@ -1,0 +1,128 @@
+// @vitest-environment jsdom
+
+import { act, renderHook } from "@testing-library/react";
+import { describe, expect, it, vi } from "vitest";
+import type { AsyncStorageLike } from "./LocalStorageThreadListAdapter";
+
+const mocks = vi.hoisted(() => ({
+  aui: undefined as unknown,
+}));
+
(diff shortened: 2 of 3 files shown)


Kinfe123:
Closing this after tracing the real RemoteThreadList lifecycle. The regression test mocked threadListItem.initialize() as a direct adapter.initialize() call, but the runtime cannot take that path after deletion: deleted thread data is removed and initialize() rejects instead of recreating metadata. The remaining possible race is an orphaned messages blob from an append already past initialization, which is a different and narrower issue. This PR should not ship the deletion tracker for the originally stated resurrection bug.

PR #6091, opened 2026-08-18
fix(cloud-ai-sdk): clear deleted thread selections after refresh

Kinfe123:
## Summary

- verify a previously listed selected thread omitted from a refresh page with `cloud.threads.get()`
- clear the selection only when that request confirms deletion with a 404
- preserve valid off-page, newly-created, and newer selections made while refresh is pending
- finish loading and clear prior errors after the refreshed list commits while deletion verification continues
- add a patch changeset for `@assistant-ui/cloud-ai-sdk`

## Why

A successful refresh can leave `threadId` pointing at a thread that has been deleted remotely. A list page is not authoritative because the endpoint is paginated and can lag after creation, so absence alone must not clear an active conversation. This change uses a positive 404 signal before clearing while retaining the existing selection for off-page, archived, or temporarily unlisted threads.

The verification probe is limited to IDs observed in an earlier successful list. This keeps the remote-deletion fix while preventing a just-created thread from being cleared if an immediate `GET /threads/{id}` is not yet read-after-write consistent with creation.

The deliberate tradeoff is that a host that deep-links `selectThread(id)` to an already-deleted thread before the ID ever appears in a list will preserve that selection. There is no reliable signal that distinguishes it from a new thread whose list entry is still propagating, so preserving the active conversation is the safer behavior.

## Reproduction

The confirmed-deletion bug reproduces on pre-fix commit `b355aefbe` by running the regression test from this PR against that source:

```sh
cd assistant-ui
git checkout b355aefbe
git show b19aaf758:packages/cloud-ai-sdk/src/threads/useThreads.test.ts > packages/cloud-ai-sdk/src/threads/useThreads.test.ts
pnpm turbo build --filter=@assistant-ui/cloud-ai-sdk
pnpm --filter @assistant-ui/cloud-ai-sdk exec vitest run src/threads/useThreads.test.ts -t "confirms it was deleted"
```

Expected pre-fix failure: `threadId` remains `"thread-1"` even though fetching that selected thread returns 404. The same test passes on this PR.

## Testing

- `pnpm --filter @assistant-ui/cloud-ai-sdk exec vitest run src/threads/useThreads.test.ts` (26 tests)
- `pnpm --filter @assistant-ui/cloud-ai-sdk test` (90 tests on each peer configuration plus peer-v6 typecheck)
- `pnpm turbo build --filter=@assistant-ui/cloud-ai-sdk --output-logs=errors-only`


okisdev (approved):
this is ready to merge. a list page is not a delete; only a 404 from get() on an id that already appeared in a successful list should clear the selection. the new tests fail on main's hook (threadId stays "thread-1") and pass on the pre-merge head. the merge-from-main commit does not touch the hook.

PR #6103, opened 2026-08-18
fix: make ThreadIf empty agree with ThreadEmpty on react-native and react-ink

samdickson22:
**The bug.** `@assistant-ui/react-native` and `@assistant-ui/react-ink` derived emptiness in `ThreadIf` from `thread.messages.length === 0`. Everything else in the runtime uses `thread.isEmpty`, which is `messages.length === 0 && !isLoading` — including those same two packages' `ThreadEmpty` primitive, which reads `useThreadIsEmpty()`. A thread that is still loading its first page has no messages but is not empty, so `<Thread.If empty>` rendered while `<Thread.Empty>` did not, in the same pass. Web was already correct.

**The fix.** Rather than alias the web primitive (which would narrow RN/Ink's permissive props, since web requires at least one filter), the shared predicate moves down: the web primitive's local `useThreadIf` is lifted verbatim into `packages/core/src/react/primitive-hooks/useThreadIf.ts`, exported from `@assistant-ui/core/react` alongside the existing `useThreadIsEmpty` and `useThreadIsRunning`, and consumed by all three distributions. RN and Ink additionally gain the optional `disabled` filter web already had, closing the other divergence between the implementations.

**Surface.** Strictly additive. Core gains `useThreadIf` and `UseThreadIfFilters`; react-native and react-ink gain an optional `disabled` prop. `@assistant-ui/react`'s api-surface is unchanged, and no prop type narrows — web keeps its `RequireAtLeastOne` props and its `@deprecated` marker.

**Verification.** New tests in all three packages. Reverting either component to the old `messages.length` implementation fails 6 react-native cases and 4 react-ink cases; both pass with the fix. Full suites green, `tsc --noEmit` matches the upstream baseline exactly, `pnpm lint` clean, api-surface regenerated under Node 24 with only the additive entries above.

Out of scope, filed as #6105: the `ExternalThread` tap client computes `isEmpty` without the `isLoading` term, so it disagrees with `thread-runtime-client.ts` and with `ThreadState`'s documented contract. (`useExternalStoreRuntime` is unaffected — it runs through the producer that already carries the term.)

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6103?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.FAlvZbbdt4WHVnSY1yYMe0dcbG-EDiHRsF_WguCRmjbrDpXnRci_U1INraw?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>


samdickson22:
> P1: When this release is installed with `@assistant-ui/core@0.3.14`, this named import is unavailable and loading `@assistant-ui/react` fails. Raise the React package's minimum core dependency to the release containing `useThreadIf`.

Not a real issue — that resolution can't happen, because the changeset in this PR bumps `@assistant-ui/core` alongside `@assistant-ui/react`, and Changesets rewrites the internal range as part of versioning.

Verified by running the release step on this branch (`8cf571f`):

```
BEFORE: core 0.3.14, react deps["@assistant-ui/core"] = ^0.3.14
AFTER:  core 0.3.15, react deps["@assistant-ui/core"] = ^0.3.15
        react-native  ^0.3.15
        react-ink     ^0.3.15
```

So the published `@assistant-ui/react@0.15.16` depends on `^0.3.15` — the release that contains `useThreadIf` — and `0.3.14` no longer satisfies it. `updateInternalDependencies` is unset in `.changeset/config.json`, so it takes its default of `patch`, which is what performs the rewrite.

The finding would apply to a changeset that bumped only `@assistant-ui/react` while consuming a new core export; this one lists all four affected packages.

Kinfe123 (commented):
the fix itself looks right to me, using thread.isEmpty instead of messages.length === 0 makes ThreadIf finally agree with ThreadEmpty during initial load, and sharing one useThreadIf across the three distributions is the right shape.

two things before i can approve though:

1. in packages/react-ink/src/primitives/thread/ThreadIf.test.tsx the mock is a plain `const mockUseAuiState = vi.fn()` at module scope, while the core and react-native tests in this same PR use vi.hoisted. vi.mock gets hoisted above the import so this can hit the TDZ — can you switch it to the vi.hoisted pattern like the others?

2. packages/react now imports useThreadIf from @assistant-ui/core/react, but the core dep range still allows older core versions that don't export it. someone with an old core in their lockfile would get a broken named import. please bump the core minimum to the release that ships useThreadIf

okisdev:
closing: ThreadIf is a leftover. empty-state rendering already goes through `AuiIf condition={(s) => s.thread.isEmpty}`, which reads the loading-aware field. RN and Ink already export AuiIf; the docs and the v0.15 codemod already migrate `ThreadPrimitive.If` onto it.

this PR would add a new public `useThreadIf` hook and disabled-filter parity on that deprecated surface. that is the opposite of the AuiIf consolidation.

PR #6104, opened 2026-08-18
chore(react-native): drop two unreachable re-export shims

samdickson22:
`src/runtimes/RemoteThreadListHookInstanceManager.tsx` and `src/runtimes/RemoteThreadListThreadListRuntimeCore.tsx` were one-line re-exports of `@assistant-ui/core/react`. Nothing in the repo imported them, neither appears in `src/index.ts` or `src/internal.ts`, and the package's `"."`/`"./internal"` exports map left them unreachable to consumers — a deep import would fail with `ERR_PACKAGE_PATH_NOT_EXPORTED`.

Because the exports map is `"."`/`"./internal"` only, the four emitted `dist/runtimes/RemoteThreadListHookInstanceManager.*` and `RemoteThreadListThreadListRuntimeCore.*` files were unreachable to any exports-aware resolver; they disappear with this change, and the api-surface snapshot is unchanged. (An earlier version of this description claimed the dist was byte-identical — a clean baseline-vs-branch build diff shows those four files were emitted and are now gone; corrected in the comments.) Deliberately not indexing these instead of deleting them: they are implementation-level core React exports, and re-exporting them from react-native would expand an append-only public surface for no consumer.

Verified: react-native tests, `tsc --noEmit`, `pnpm lint`, and `pnpm api-surface` (Node 24) with zero drift.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6104?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.v1zHl7StkswMfqgkAgeQqSdRUVTKzUISHqqmtO6A8xeBKSXu90-N97w0uRg?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/drop-rn-remote-thread-list-shims.md
+++ b/.changeset/drop-rn-remote-thread-list-shims.md
@@ -1,0 +1,12 @@
+---
+"@assistant-ui/react-native": patch
+---
+
+chore: drop two unreachable re-export shims from react-native
+
+`src/runtimes/RemoteThreadListHookInstanceManager.tsx` and
+`src/runtimes/RemoteThreadListThreadListRuntimeCore.tsx` each re-exported one
+symbol from `@assistant-ui/core/react`, but nothing imported them, they were
+absent from `src/index.ts` and `src/internal.ts`, and the `"."`/`"./internal"`
+exports map made them unreachable to consumers. The public API surface is
+unchanged.
--- a/packages/react-native/src/runtimes/RemoteThreadListHookInstanceManager.tsx
+++ b/packages/react-native/src/runtimes/RemoteThreadListHookInstanceManager.tsx
@@ -1,1 +1,0 @@
-export { RemoteThreadListHookInstanceManager } from "@assistant-ui/core/react";
--- a/packages/react-native/src/runtimes/RemoteThreadListThreadListRuntimeCore.tsx
+++ b/packages/react-native/src/runtimes/RemoteThreadListThreadListRuntimeCore.tsx
@@ -1,1 +1,0 @@
-export { RemoteThreadListThreadListRuntimeCore } from "@assistant-ui/core/react";


samdickson22:
The correction in the review is right and my claim was wrong. Verified rather than conceded:

I built `packages/react-native` clean from a worktree whose react-native source is untouched base, then clean from this branch, and diffed the emitted trees:

```
=== BASELINE dist/runtimes ===
RemoteThreadListHookInstanceManager.d.ts
RemoteThreadListHookInstanceManager.js
RemoteThreadListThreadListRuntimeCore.d.ts
RemoteThreadListThreadListRuntimeCore.js
useLocalRuntime.d.ts / useLocalRuntime.js
useRemoteThreadListRuntime.d.ts / useRemoteThreadListRuntime.js

=== delta (baseline → this branch) ===
< ./runtimes/RemoteThreadListHookInstanceManager.d.ts
< ./runtimes/RemoteThreadListHookInstanceManager.js
< ./runtimes/RemoteThreadListThreadListRuntimeCore.d.ts
< ./runtimes/RemoteThreadListThreadListRuntimeCore.js
```

So `dist/` is **not** byte-identical: four emitted files disappear, exactly as predicted from `aui-build`'s glob entry with `unbundle: true`, where every non-test source file is its own entry.

**How I got it wrong**, since it's worth recording: I listed `dist/runtimes` only *after* deleting the sources and rebuilding, saw the two modules absent, and concluded the bundler had never emitted them. I never established a baseline. That's a post-state observation dressed up as a before/after — and it's the same methodological gap I'd flag in any finding I was reviewing. On the sibling deletion PR I did run a proper clean-baseline-vs-clean-branch diff; I simply failed to apply it here.

**The verdict is unaffected, for the reasons the review already gives.** The exports map is `"."`/`"./internal"` only, so those dist paths were never reachable by an exports-aware resolver; both symbols remain exported from `@assistant-ui/core/react`; and the api-surface snapshot is unchanged. If anything the patch changeset is now more clearly warranted, since the published tarball genuinely changes rather than only shedding two `src/` lines.

Requesting the "byte-identical dist" sentence be struck from the description so the merge record doesn't rest on a false fact. I'd rather leave the commit message as-is than force-push a green branch over wording, unless a maintainer prefers otherwise.

okisdev (approved):
the two shims are unreachable through `.` / `./internal` and neither barrel indexes them. repo grep has no callers, react-ink never copied the pair, and the react-native api-surface snapshot does not name the classes. deleting them rather than promoting them keeps the append-only surface intact.

i grepped the tree, read the barrels and the exports map, and listed `dist/runtimes` on main. i did not re-run the package tests. the earlier "byte-identical dist" claim was wrong and is already struck from the description; the commit message still has it, not worth a force-push.

update-branch only brought in #6121. net diff is still the two deletions plus the patch changeset.

PR #6106, opened 2026-08-18
fix(core): ExternalThread no longer reports an empty thread while loading

### LINKED ISSUE #6105: external-store: thread.isEmpty ignores isLoading
### Summary

`ExternalThread` computes `thread.isEmpty` without the `isLoading` term, so it disagrees with the other producer of the same state field. A thread that has loaded no messages yet but is still loading reports `isEmpty: true`, and empty-state UI renders underneath the loading indicator.

The two producers:

- `packages/core/src/store/clients/external-thread.ts:1088` — `isEmpty: messages.length === 0`
- `packages/core/src/store/runtime-clients/thread-runtime-client.ts:97` — `isEmpty: messages.state.length === 0 && !runtimeState.isLoading`

`isLoading` is already destructured and in scope at the first site (`external-thread.ts:902`) and is already a dependency of the surrounding `useMemo`, so only the expression differs.

### Scope

This affects the `ExternalThread` client, which is public API on both `@assistant-ui/core` and `@assistant-ui/react`, and its documented `isLoading` prop ("Whether the adapter is in a loading state. Displays a loading indicator instead of the composer.").

`useExternalStoreRuntime` is **not** affected — it runs through `ExternalStoreRuntimeCore` and `thread-runtime-client.ts`, which already carries the `!isLoading` term. That difference between the two paths is the inconsistency being reported.

### Reproduction

```tsx
const aui = useAui({ thread: ExternalThread({ messages: [], isLoading: true }) });

// expected false (still loading), actual true
aui.thread.getState().isEmpty;
```

Every consumer reading that field follows it, so with `isLoading: true` and no messages both of these render their empty state:

```tsx
<AuiIf condition={(s) => s.thread.isEmpty}>…</AuiIf>
useThreadIsEmpty(); // → true, and so <Thread.Empty> renders
```

Verified against `357151b85` with a test at the producer seam and through `useThreadIsEmpty` / `AuiIf`: the no-messages-and-not-loading and the has-messages cases pass, and only the loading case fails.

### Fix direction

Add the `isLoading` term to the `ExternalThread` producer so both producers define `isEmpty` identically.

Found during #6103, which fixed the mirror-image inconsistency on the consumer side (`ThreadIf` on react-native and react-ink read `messages.length` instead of `thread.isEmpty`).

<a href="https://rupic.dev/s/s2.qFy7b5xulKW18mcCzfP0KccmirT5arhpCJOKPhoX_1FoOiTaOg_7LAR8BqY?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

samdickson22:
`ExternalThread` computed `thread.isEmpty` as `messages.length === 0`, omitting the `isLoading` term. `ThreadState` documents the intended contract on the field itself — "A thread is considered empty when it has no messages and is not loading" — and the other producer, `store/runtime-clients/thread-runtime-client.ts:97`, already applies it.

A thread with `isLoading: true` and no messages yet therefore reported `isEmpty: true`, so `<ThreadPrimitive.Empty>`, `useThreadIsEmpty()`, and `AuiIf` on `s.thread.isEmpty` rendered their empty state underneath the loading indicator.

`isLoading` was already destructured in the same function and already a dependency of the surrounding `useMemo`, so only the expression changes and there is no reactivity change. Both producers now define `isEmpty` identically.

**Behavior change.** Consumers that pass `isLoading` to the `ExternalThread` client and render empty-state UI will now see that UI stay hidden until loading finishes. `useExternalStoreRuntime` is unaffected — it runs through `ExternalStoreRuntimeCore` and `thread-runtime-client.ts`, which already carried the term.

Tests pin the behavior at the producer seam and end-to-end through `useThreadIsEmpty` and `AuiIf`, including both loading transitions. Removing the new term fails 4 of the 6 cases.

Fixes #6105

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6106?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.wkhPC0fyQW2RgmGqqVYGStPKHfVtHoB5k-CQbCGXlGUgrCly7z3Yb9UYGbA?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>


samdickson22:
On the test-placement note — the rule citation is correct, and the repo is split on it in this exact spot, so I'd rather surface the evidence than quietly pick a side.

AGENTS.md does say colocate, and `packages/core/src/store/clients/` already hosts colocated tests (`model-context-client.test.ts`, `thread-message-client.test.ts`), so moving this file would not be unprecedented.

Against that: every existing test for this module lives in `packages/core/src/tests/` —

```
external-thread-attachments.test.tsx
external-thread-branches.test.tsx
external-thread-feedback.test.tsx
external-thread-parity.test.tsx
external-thread-refetch.test.tsx
external-thread-speech.test.tsx
external-thread-suggestions.test.tsx
```

and across `packages/core/src` the split is 53 files under `src/tests/` versus 76 colocated, so neither placement is anomalous package-wide.

I put `external-thread-isEmpty.test.tsx` beside those seven deliberately: it reuses the same `ExternalThread` + `AuiProvider` harness shape they use, and moving it alone would separate it from its own module's test family without moving them. Following the written rule strictly here would argue for relocating all eight, which is a separate cleanup and not this PR's concern.

Happy to move just this one if a maintainer would rather the new file follow the documented rule and the older ones get swept later — it's a rename with no behavior change. Flagging the trade-off rather than deciding it unilaterally.

okisdev (approved):
this is ready to merge. ExternalThread never had the `!isLoading` term; the kit's startup exception lives on `isNewChatView`, not on `isEmpty`. this lines the tap producer up with `ThreadState` and `thread-runtime-client`.

PR #6107, opened 2026-08-18
fix(core): preserve falsy assistant transport artifacts

Kinfe123:
## Summary

- preserve tool artifacts when their value is `false`, `0`, `""`, or `null`
- continue omitting the artifact field only when it is `undefined`
- cover the public message-part API through the serialized Assistant Transport request

## Why

`artifact` is a JSON value, so falsy primitives are valid payloads. The existing truthiness check silently dropped those values before sending the `add-tool-result` command.

## Validation

- `pnpm --filter @assistant-ui/core test` (128 files, 1,295 tests)
- `pnpm turbo build --filter=@assistant-ui/core...`
- targeted `oxlint` and `oxfmt`

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6107?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.PJ5xcwIjaKSbZDSHF4GT8uVGuLS_oK3iYJzo5LXOt096Z1F5j3PZMnjPSBI?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/calm-tools-respond.md
+++ b/.changeset/calm-tools-respond.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/core": patch
+---
+
+fix: preserve falsy assistant transport artifacts
--- a/packages/core/src/react/runtimes/assistant-transport/useAssistantTransportRuntime.test.tsx
+++ b/packages/core/src/react/runtimes/assistant-transport/useAssistantTransportRuntime.test.tsx
@@ -4,6 +4,7 @@ import { act, render, waitFor } from "@testing-library/react";
 import type { FC } from "react";
 import { afterEach, describe, expect, it, vi } from "vitest";
 import { useAui } from "@assistant-ui/store";
+import { ToolResponse } from "assistant-stream";
 import { AssistantRuntimeProvider } from "../../AssistantRuntimeProvider";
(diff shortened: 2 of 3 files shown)


okisdev (approved):
`artifact` is a JSON value, so `false`, `0`, `""`, and `null` are real payloads. the truthy spread was the last `artifact &&` in the repo; everything else already uses `!== undefined`.

i ran `useAssistantTransportRuntime.test.tsx` at `749a90dcf` (18 passed), then swapped the runtime back to `origin/main` with the PR tests kept: the four new cases all failed because the serialized `add-tool-result` command had no `artifact`. `undefined` is still omitted.

PR #6112, opened 2026-08-18
fix(core): contain a synchronously throwing clipboard writer in useActionBarCopy

### LINKED ISSUE #6111: core: a synchronously throwing clipboard writer escapes useActionBarCopy and kills the ink process
`useActionBarCopy` invokes the caller-supplied clipboard writer as an argument to `Promise.resolve(...)` (`packages/core/src/react/primitive-hooks/useActionBarCopy.ts:45`):

```js
Promise.resolve(copyToClipboard(valueToCopy)).then(onOk, () => {});
```

The writer runs before `.then` attaches a rejection handler, so a **synchronous** throw escapes it and propagates out of `copy()` into the caller's press handler. The comment above that line says the rejection handler swallows "permission denied, API unavailable" failures — for the synchronous case it does not.

**Reachability.** `@assistant-ui/react` is unaffected (its first-party writer guards `!navigator.clipboard` and returns a rejected promise). React Native and Ink take the platform clipboard writer as a public prop, so the writer is user code; the common trigger is an unlinked clipboard native module, where `Clipboard.setString` throws `TypeError` synchronously.

**Severity on Ink**: `Pressable` calls `onPress()` from inside ink's `useInput` stdin handler — outside React render and any error boundary — so the throw is an uncaught exception that terminates the Node process: the TUI exits on a keypress. On React Native it surfaces as a redbox.

**Repro** (fails on current main, `useActionBarCopy.test.ts` shape):

```ts
// packages/core/src/react/primitive-hooks/useActionBarCopy.test.ts
const copyToClipboard = vi.fn(() => {
  throw new TypeError("Cannot read properties of undefined (reading 'writeText')");
});
const { result } = renderHook(() => useActionBarCopy({ copyToClipboard }));

expect(() => result.current.copy()).not.toThrow(); // throws on main
```

The Svelte implementation of this same primitive already guards this case with a try/catch around the writer (`packages/svelte/src/primitives/actionBar.ts:139-144`), added in #5921 — core diverged from that shape. Aligning core with the existing Svelte guard looks like the minimal fix.

<a href="https://rupic.dev/s/s2.zSSHhDBhT6Izz28avhJJeuzkocr3wEN8IACLJLmMrF0wTv7LN9bsq_2wTu0?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

samdickson22:
`useActionBarCopy` invokes the caller-supplied clipboard writer as an argument to `Promise.resolve(...)`:

```js
Promise.resolve(copyToClipboard(valueToCopy)).then(onOk, () => {});
```

The writer therefore runs before `.then` attaches a rejection handler, so a **synchronous** throw escapes it entirely and propagates out of `copy()` into the caller's press handler. The comment above that line claims the rejection handler swallows "permission denied, API unavailable" failures; for the synchronous case it does not.

`@assistant-ui/react` is unaffected — its first-party writer already guards `!navigator.clipboard` and returns a rejected promise. The reachable paths are React Native and Ink, where `ActionBarCopy` takes the platform clipboard writer as a public prop, so the writer is user code. The common trigger is a clipboard native module that isn't linked, where `Clipboard.setString` throws `TypeError` synchronously rather than rejecting.

On Ink this is worse than a failed copy: `Pressable` calls `onPress()` from inside ink's `useInput` stdin handler, outside React's render and outside any error boundary, so the throw is an uncaught exception that terminates the process — the TUI exits on a keypress. On React Native it surfaces as a redbox.

AGENTS.md requires that code "must not throw on … unavailable platform APIs".

**Fix.** Invoke the writer inside a `try`, and pass its result to `Promise.resolve`. This mirrors the guard already present in the Svelte implementation of the same primitive (`packages/svelte/src/primitives/actionBar.ts`), including keeping the writer on the synchronous stack — deferring it into the promise chain would push it past the event-dispatch stack that WebKit scopes clipboard user-activation to, breaking copy in Safari. The comment is mirrored from that sibling verbatim so the relationship between the two implementations stays visible.

A synchronous throw now behaves exactly as an async rejection already does: no copied state, no timer. That does mean a developer whose clipboard module isn't linked gets a button that silently does nothing instead of a crash. That trade-off is deliberate — it matches both the existing async path and the Svelte implementation, and crashing the Ink process on a keypress is the worse failure. Surfacing a dev-mode warning would be a behavior change and belongs in its own PR.

**Testing.** Added a case to `useActionBarCopy.test.ts` covering a writer that throws synchronously; it fails on `main` with the TypeError escaping `copy()` and passes with the fix. Full `@assistant-ui/core` suite passes (128 files, 1292 tests).

Fixes #6111

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6112?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.TSMAst05gPTObaudH39VKwn4ssz2eNFpDcw27oq28ml9zgPtore2PanjLhE?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/copy-sync-throw.md
+++ b/.changeset/copy-sync-throw.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/core": patch
+---
+
+fix: contain a synchronously throwing clipboard writer in useActionBarCopy
--- a/packages/core/src/react/primitive-hooks/useActionBarCopy.test.ts
+++ b/packages/core/src/react/primitive-hooks/useActionBarCopy.test.ts
@@ -82,6 +82,21 @@ describe("useActionBarCopy", () => {
     expect(mocks.setIsCopied).not.toHaveBeenCalled();
   });
 
+  it("does not report copy success when the clipboard handler throws synchronously", async () => {
+    const copyToClipboard = vi.fn(() => {
+      throw new TypeError(
+        "Cannot read properties of undefined (reading 'writeText')",
+      );
+    });
+    const { result } = renderHook(() => useActionBarCopy({ copyToClipboard }));
+
+    expect(() => result.current.copy()).not.toThrow();
+    await Promise.resolve();
+
+    expect(copyToClipboard).toHaveBeenCalledWith("Hello");
+    expect(mocks.setIsCopied).not.toHaveBeenCalled();
+  });
+
   it("keeps copy success visible for the full duration after copying again", async () => {
     vi.useFakeTimers();
     const copyToClipboard = vi.fn();
--- a/packages/core/src/react/primitive-hooks/useActionBarCopy.ts
+++ b/packages/core/src/react/primitive-hooks/useActionBarCopy.ts
@@ -40,9 +40,19 @@ export const useActionBarCopy = ({
     if (!valueToCopy) return;
     const scopeGeneration = scopeGenerationRef.current;
 
-    // The rejection handler swallows clipboard write failures (permission denied,
-    // API unavailable) so they don't surface as unhandled promise rejections.
-    Promise.resolve(copyToClipboard(valueToCopy)).then(
+    // The writer runs synchronously inside the press so the user-gesture gating
+    // of navigator.clipboard survives (WebKit scopes it to the stack); the try
+    // contains a synchronously throwing caller-supplied writer, and the rejection
+    // handler swallows clipboard failures (permission denied, API unavailable) so
+    // they don't surface as unhandled rejections.
+    let write: void | Promise<void>;
+    try {
+      write = copyToClipboard(valueToCopy);
+    } catch {
+      return;
+    }
+
+    Promise.resolve(write).then(
       () => {
         if (scopeGeneration !== scopeGenerationRef.current) return;
 


samdickson22:
On the open question about hardening ink's `useInput` dispatch: deliberately not in this PR. A blanket try/catch in ink's `Pressable` would change error semantics for every consumer `onPress` (swallowing throws that today crash loudly during development), which is a design call about what the dispatch seam should promise — not a defect containment like this one, where the hook's own comment already claimed the safety it didn't deliver. If maintainers want the general guard, happy to open an issue for it so the trade-off (silent-by-default vs crash-by-default outside error boundaries) gets decided deliberately.

okisdev (approved):
the writer was evaluated as an argument to Promise.resolve, so a synchronous throw escaped copy() before the rejection handler attached. that is reachable: react-native and react-ink pass copyToClipboard through as a public prop, and ink's Pressable calls onPress from useInput, so the throw kills the process.

i ran the new test against main's hook (fails at not.toThrow with the TypeError) and against this head (8/8 pass). the try stays on the press stack, matching the svelte sibling, so webkit user-activation is preserved. swallowing the sync throw is the same silent no-op the async path already has.

update-branch only merged main. net diff is still the hook, the test, and the patch changeset.

PR #6116, opened 2026-08-18
fix(react-o11y): flatten deeply nested traces iteratively

Kinfe123:
## Summary

- replace recursive span-tree flattening with an explicit traversal stack
- preserve depth-first traversal, sibling start-time ordering, and collapsed-subtree behavior
- add regression coverage for a 20,000-span parent chain and traversal ordering

## Why

`SpanResource` normalized parents and calculated depth iteratively, but flattened the resulting tree recursively. A valid 20,000-span parent chain therefore threw `RangeError: Maximum call stack size exceeded` while preparing the visible span list. Using an explicit LIFO stack removes the JavaScript call-stack limit; children are pushed in reverse order so the rendered order remains unchanged.

## Validation

- reproduced the stack overflow with the new regression test before the implementation change
- `pnpm --filter @assistant-ui/react-o11y test` (2 files, 14 tests)
- `pnpm turbo build --filter=@assistant-ui/react-o11y...`
- `pnpm lint`

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6116?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.9amEwQWaRW_eWPEZ0rsljxwFaROfWjVfSIPQzsp6XOHG24cH4cI4UmLnqW4?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/deep-spans-flatten.md
+++ b/.changeset/deep-spans-flatten.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react-o11y": patch
+---
+
+fix: flatten deeply nested observability traces without overflowing the call stack
--- a/packages/react-o11y/src/resources/SpanResource.test.tsx
+++ b/packages/react-o11y/src/resources/SpanResource.test.tsx
@@ -4,14 +4,18 @@ import { AuiProvider, useAui } from "@assistant-ui/store";
 import * as SpanPrimitive from "../primitives/span";
 import { SpanResource, type SpanData } from "./SpanResource";
 
-const createSpan = (id: string, parentSpanId: string | null): SpanData => ({
+const createSpan = (
+  id: string,
+  parentSpanId: string | null,
+  startedAt = 0,
+): SpanData => ({
   id,
   parentSpanId,
   name: id,
   type: "test",
   status: "completed",
-  startedAt: 0,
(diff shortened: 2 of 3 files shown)


okisdev (approved):
the recursive flatten was the remaining call-stack path. a 20,000-span parent chain still blew up in `dfs` on main; the iterative LIFO with children pushed in reverse is the same DFS, just not on the JS stack.

i ran `SpanResource.test.tsx` at `175d12d86` (6 passed, deep chain 1144ms), then swapped `SpanResource.tsx` back to `origin/main` with the PR tests kept: the 20k case threw `RangeError: Maximum call stack size exceeded`, and the start-time order test still passed.

PR #6120, opened 2026-08-18
fix(react-native,react-ink): expose ThreadListNew active state

### LINKED ISSUE #6118: react-native/react-ink: ThreadListNew exposes no active state — assistive tech cannot announce the current control
Web's `ThreadListPrimitive.New` marks itself `data-active` + `aria-current` while the new thread is the current one (`packages/react/src/primitives/threadList/ThreadListNew.tsx:22-31`, reading `s.threads.newThreadId === s.threads.mainThreadId`). The react-native and react-ink versions wire only the switch action and read no state at all, so neither platform can announce nor style the currently-active control — assistive tech on RN has no way to know the "new thread" item is the current one.

Repro: on current main, render ThreadListNew on RN/ink with the new thread active — no state is exposed to children and (RN) no `accessibilityState.selected` is set; the same scenario on web emits `aria-current`.

Related, likely the same shape: `ThreadListItemRoot`/`ThreadListItemTrigger` have the same active-state gap on both runtimes ("which existing thread is selected"), and the shared `newThreadId === mainThreadId` selector can report active during the pre-init window when both are undefined (pre-existing, identical on web).

<a href="https://rupic.dev/s/s2.oprDoKaRFU2zheJ_4HH13wZTsNX1DktYG0U3bFnhqb52oyHr1vq5xZlwBXg?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

samdickson22:
the web `ThreadListNew` marks itself `data-active` + `aria-current` while the new thread is the current one; the react-native and react-ink versions only wired the action, so neither could announce nor style the current control.

both now read the same selector (`s.threads.newThreadId === s.threads.mainThreadId`). react-native sets `accessibilityState.selected` (the closest native equivalent to `aria-current`) ahead of the caller spread, merged so callers can override per-key, and keeps `accessibilityRole="button"` from #6119. function children receive `isActive` *alongside* the platform Pressable state (`pressed` on react-native, `isFocused`/`disabled` on ink) instead of replacing it. that keeps the #6114 children contract additive.

deliberately out of scope: `ThreadListItemRoot` and `ThreadListItemTrigger` have the same active-state gap on both runtimes and should land as their own PR. the shared `newThreadId === mainThreadId` selector can report active during the pre-init window when both are undefined, which is pre-existing and identical on web.

fixes #6118

<a href="https://rupic.dev/s/s2.fCXsepJPU1rtA4L1RIZTDZUeCRuUF6xAF6RK1oBoW2PMMOkJD2t16XENa3I?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>


okisdev (approved):
this is ready to merge. replacing Pressable children with `{ isActive }` only would have undone the render-function contract from #6114, so the merge keeps `pressed` / `isFocused` and adds `isActive` on top. I ran the package vitest files locally: react-native `ThreadListNew` + `accessibilityRole` 54/54, react-ink `ThreadListNew` 3/3. the trap is treating `ActionBarCopy` as the sibling; that primitive replaces press state on purpose. the same gap on `ThreadListItemRoot` / `ThreadListItemTrigger` is #6131.

PR #6135, opened 2026-08-19
fix(core): suppress stale cancellation resyncs

Kinfe123:
## Summary

- stop deferred external-store cancellation resyncs after the runtime is invalidated
- preserve the existing same-runtime resync behavior for cancellation/store races
- add a focused teardown regression and patch changeset

## Why

`cancelRun()` defers its final `setMessages()` resync to the next macrotask so a server-side cancellation update can land first. If the runtime is unmounted or replaced during that gap, the deferred callback still writes through the detached adapter.

A focused reproduction on current `main` called `cancelRun()`, invalidated the runtime, and observed one stale `setMessages()` call. The callback now captures the runtime generation and exits when teardown advances it.

## Verification

- 15/15 cancellation-path tests
- `@assistant-ui/core` build
- `git diff --check`

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6135?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.Rc4e8CQxAcNKhDiqRCAL0Kdl5TEDwgp52S5w3R1NW57JwxI92Sif5HHZkik?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/quiet-runs-rest.md
+++ b/.changeset/quiet-runs-rest.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/core": patch
+---
+
+fix: prevent cancelled external-store runs from resyncing after teardown
--- a/packages/core/src/runtimes/external-store/external-store-thread-runtime-core.ts
+++ b/packages/core/src/runtimes/external-store/external-store-thread-runtime-core.ts
@@ -674,6 +674,8 @@ export class ExternalStoreThreadRuntimeCore
     if (!this._store.onCancel)
       throw new Error("Runtime does not support cancelling runs.");
 
+    const generation = captureThreadRuntimeGeneration(this);
+
     // Abort any in-flight client-side tool executions. Fire-and-forget —
     // the abort resolves once executions settle, but we don't gate the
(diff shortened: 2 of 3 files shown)


okisdev (approved):
the deferred cancel resync is the same generation hole append already closed. unmount invalidates the runtime during that macrotask, and the callback would still write the detached adapter. the flush-read from #5826 stays intact.

I read cancelRun against the append generation guard and the new teardown test. I did not rerun vitest; CI already ran Test Changed Packages on this head.

PR #6141, opened 2026-08-19
fix(react-devtools): handle conversation load failures

Kinfe123:
## Summary

- consume rejected custom-client thread switches from the DevTools Load conversation button
- contain synchronous client errors as well as asynchronous rejections
- add focused coverage for both failure modes

## Why

`DevToolsClient.switchToThread()` may return a promise, but React does not observe promises returned from click handlers. A custom or remote DevTools client that rejects therefore produces a global `unhandledRejection`.

The default in-process client already treats thread-switch failures as non-fatal. The button now applies the same behavior for every client without changing successful switching.

## Verification

- regression before the fix: `Unhandled Rejection: switch failed`
- `@assistant-ui/react-devtools` tests: 111 passed
- `@assistant-ui/react-devtools` build passed
- targeted oxlint and oxfmt passed

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6141?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.QI9eRR3FOS8AvyquIHMU3_5MLhzPbmEC914MPCzVy5EPS5GIwqu6PBoGNLc?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/tidy-devtools-switch.md
+++ b/.changeset/tidy-devtools-switch.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react-devtools": patch
+---
+
+fix: handle failed DevTools conversation switches
--- a/packages/react-devtools/src/shell/tabs/ThreadTab.test.tsx
+++ b/packages/react-devtools/src/shell/tabs/ThreadTab.test.tsx
@@ -1,0 +1,59 @@
+import type { ComponentProps } from "react";
+import { renderToStaticMarkup } from "react-dom/server";
+import { describe, expect, it, vi } from "vitest";
+
+let capturedOnClick: (() => unknown) | undefined;
+
+vi.mock("../../views/ui", async (importOriginal) => ({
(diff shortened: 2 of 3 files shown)


okisdev:
superseded by #6163.

the rejection has to be consumed on useDevToolsClient so every tab and plugin caller is covered. swallowing it only on the load conversation button leaves the same unhandledRejection on any other switchToThread call.

PR #6142, opened 2026-08-19
fix(core): stop superseded local runs before tool continuation

Kinfe123:
## Summary

- stop a LocalRuntime run from entering another tool roundtrip after a newer run replaces it
- preserve the replacement run's active abort controller
- add a regression test covering a delayed tool-call result from the superseded run

## Why

A chat-model adapter may finish after its abort signal is raised. If that stale result requests another automatic tool step, the old run currently enters `performRoundtrip()` again. That method aborts the shared current controller, so it cancels the newer replacement run.

The runtime already tracks the active run. This change checks that ownership after each roundtrip and exits before tool continuation when the run has been superseded.

## Verification

- regression without the guard: 3 adapter calls instead of 2, and the replacement signal is aborted
- `local-thread-runtime-core.test.ts`: 60 tests passed
- `@assistant-ui/core` build passed
- targeted oxlint and oxfmt passed

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6142?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.BhnKA_XhU_UsDJLIFDs6qa-NQ9u6KqESRJ9FYJghhDuaFjLZP13Gg1xCcSk?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/tidy-runs-continue.md
+++ b/.changeset/tidy-runs-continue.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/core": patch
+---
+
+fix: stop superseded local runs before tool continuation
--- a/.changeset/tidy-runs-continue.md
+++ b/.changeset/tidy-runs-continue.md
@@ -2,4 +2,4 @@
 "@assistant-ui/core": patch
 ---
 
-fix: stop superseded local runs before tool continuation
+fix: settle aborted and superseded local runs without continuing them, while keeping approval pauses answerable after cancel
--- a/packages/core/src/runtimes/local/local-thread-runtime-core.test.ts
+++ b/packages/core/src/runtimes/local/local-thread-runtime-core.test.ts
@@ -783,6 +783,30 @@ describe("LocalThreadRuntimeCore cancellation", () => {
     });
   });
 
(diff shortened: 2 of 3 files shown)


okisdev (approved):
approved. the original ownership guard fixed the replacement-abort symptom but left cancel, pause, terminal-status, and same-message races. this head now cancels only stale results that would continue, preserves terminal and approval pauses, and rejects stale repository/history writes by message-object ownership. I reproduced the two original failures before the fixes, then ran the final LocalThreadRuntimeCore suite at 67/67 and built core successfully.

Issue #6056, opened 2026-08-17
core: the tool-approval seam models only a decision, and carries no failure signal


samdickson22:
Two gaps in core's tool-approval surface surfaced independently while working on the eve adapter (#5626, #5627). Both are core-shape questions, so this issue asks for direction rather than proposing a patch.

## 1. The approval surface models a decision, not a question

`ToolCallMessagePart.approval` (`packages/core/src/types/message.ts:213`) carries only decision-shaped fields — `id`, `approved`, `reason`, `options`, `optionId`, `resolution`. There is no field for the human-facing prompt or for how the request should be presented.

Providers do distinguish these. Eve, for one, separates `confirmation` / `select` / `text` input requests and carries the question text with the request; #5627 had to smuggle that through `providerMetadata.eve.inputRequest` because core has nowhere to put it. The consequence is visible in the shipped kit: `packages/ui/src/components/assistant-ui/tool-fallback.tsx` filters `approval.options` down to the four known kinds and otherwise renders a generic Allow/Deny pair, so a free-form question renders exactly like a tool approval and offers no way to type an answer. #5904 tracks the kit-side half of this (custom-kind options being filtered out); it is a rendering fix within today's types. This issue is about the missing model underneath it — with no prompt or display field, no renderer, default or custom, can tell a question from an approval without reading provider-specific metadata.

## 2. The approval seam has no first-class failure signal

`respondToToolApproval` returns `void` on both the runtime-core interface (`packages/core/src/runtime/interfaces/thread-runtime-core.ts:159`) and the public `MessagePartRuntime` API (`packages/core/src/runtime/api/message-part-runtime.ts:31`). The adapter callback may return a promise (`onRespondToToolApproval` in `packages/core/src/runtimes/external-store/external-store-adapter.ts:152`), but the external-store core calls it and discards the result (`external-store-thread-runtime-core.ts:641`), so a rejection is unobserved.

The practical effect is that "this response cannot be submitted" can only be reported by throwing synchronously before any `await` — which is what #5627 ended up doing, and it works only because the mapping happens before the send is enqueued. Anything that fails after that point, or any adapter that wants to report a rejection asynchronously, has no channel back to the UI that rendered the controls. The kit reflects this: `respond()` in `tool-fallback.tsx` calls `respondToApproval` and unconditionally sets `submitted`, disabling both controls, because there is no outcome to wait on.

## Direction

[...]

PR #6154, opened 2026-08-19
fix(assistant-stream): decode AI SDK v6 tool chunks in UIMessageStreamDecoder

### LINKED ISSUE #6153: assistant-stream: UIMessageStreamDecoder drops every AI SDK v6 tool chunk (tool-input-*/tool-output-*)
### Summary

`UIMessageStreamDecoder` (`packages/assistant-stream/src/core/serialization/ui-message-stream/UIMessageStream.ts`) silently drops every tool call streamed by a real AI SDK v6/v7 backend. Its switch handles the chunk names `tool-call-start` / `tool-call-delta` / `tool-call-end` / `tool-result` — names the AI SDK UI Message Stream protocol never emits. The real wire names, per the `UIMessageChunk` union in both `ai@6.0.257` and `ai@7.0.66` installed in this repo, are `tool-input-start`, `tool-input-delta`, `tool-input-available`, `tool-input-error`, `tool-output-available`, and `tool-output-error`. All of them fall through to `default: // ignore unknown types`, so the assistant message renders with no tool part and no warning.

This is the path real backends hit: `useDataStreamRuntime` (`@assistant-ui/react-data-stream`) selects the `ui-message-stream` protocol from the `x-vercel-ai-ui-message-stream` header that `streamText().toUIMessageStreamResponse()` sets — and `ui-message-stream` is also the detection fallback.

The decoder's SSE-mapping stage was already shimmed for the real v6 names on other chunk families (`delta` → `textDelta`, `source-url` → `source`, `file`-with-`url`), but the tool family was missed, and the colocated tests only exercise the invented names.

### Reproduction

Temporary vitest colocated next to the decoder, using the suite's own SSE harness:

```ts
const events = [
  JSON.stringify({ type: "start", messageId: "msg_1" }),
  JSON.stringify({ type: "tool-input-start", toolCallId: "call_1", toolName: "weather" }),
  JSON.stringify({ type: "tool-input-delta", toolCallId: "call_1", inputTextDelta: '{"city":' }),
  JSON.stringify({ type: "tool-input-delta", toolCallId: "call_1", inputTextDelta: '"Berlin"}' }),
  JSON.stringify({ type: "tool-input-available", toolCallId: "call_1", toolName: "weather", input: { city: "Berlin" } }),
  JSON.stringify({ type: "tool-output-available", toolCallId: "call_1", output: { temp: 20 } }),
  JSON.stringify({ type: "finish", finishReason: "tool-calls", usage: { inputTokens: 1, outputTokens: 1 } }),
  "[DONE]",
];

const chunks = await collectChunks(
  createUIMessageStream(events).pipeThrough(new UIMessageStreamDecoder()),
);
expect(chunks.filter((c) => c.type === "part-start").length).toBeGreaterThan(0); // fails: 0
```

The decoded stream contains only `step-start` and `message-finish` — no tool-call part chunks at all. The equivalent stream using the invented `tool-call-*` names decodes fine, which isolates the name mismatch as the cause.

### Impact

Any AI SDK v6/v7 server used through `useDataStreamRuntime` loses all tool calls silently: no argument streaming, no tool UI, no results — text and reasoning still render, so the failure looks like "the model chose not to call tools".

### Fix shape

Additive decoder branches (matching this repo's protocol-coupling convention): handle `tool-input-start/delta/available/error` and `tool-output-available/error` alongside the existing names — `tool-input-available` must also cover the no-prior-deltas case (servers may emit it without streaming deltas first) — with regression tests using the real wire fixtures. The existing invented-name cases stay for back-compat. PR incoming.

<a href="https://rupic.dev/s/s2.seauWgSAPHIZySF3QVxd-8NAgoHFOg-jsgYvOmdOONnR46rm4e2MzRLJUmU?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

Kinfe123:
## Summary

- handle the real AI SDK v6 UI Message Stream tool chunk names in `UIMessageStreamDecoder`: `tool-input-start`, `tool-input-delta`, `tool-input-available`, `tool-input-error`, `tool-output-available`, `tool-output-error`
- `tool-input-available` covers both the streamed case (deltas already delivered — just closes the args text, no double-append) and the non-streamed case (no prior `tool-input-start` — creates the part and writes the full input JSON)
- `tool-input-error` and `tool-output-error` settle the part with `isError: true`; `preliminary: true` outputs are skipped so the part settles on the final output
- the previously handled `tool-call-start/delta/end` / `tool-result` names stay untouched for back-compat; the legacy single-active-args-controller behavior is isolated to those names — `tool-input-start` does not close the previous tool's args stream, so parallel v6 tool input streams decode independently, and out-of-order or duplicated chunks degrade gracefully instead of aborting the strict args stream: late deltas and repeated `tool-input-available`/`tool-input-error` for a settled id are dropped, and a `tool-output-*` for an id this decoder instance never saw (e.g. the resumed response of an approval flow, where the input streamed in the previous response) is ignored rather than thrown
- `tool-output-denied` / `tool-approval-request` / `tool-approval-response` remain intentionally unhandled: they map onto the tool-approval seam, which is a separate concern (see #6056)

## Why

Fixes #6153. The decoder's switch only knew tool chunk names the AI SDK never emits (`tool-call-start/delta/end`, `tool-result`), so every tool chunk from a real `streamText().toUIMessageStreamResponse()` backend fell into the ignore-unknown default — the assistant message rendered with no tool part and no warning. The `UIMessageChunk` union in both installed majors (`ai@6.0.257`, `ai@7.0.66`) confirms the real family is `tool-input-*`/`tool-output-*`. The text-side names were already shimmed (`delta`→`textDelta`, `source-url`, `file`-with-`url`); the tool family was the remaining gap, and this lands as additive decoder branches per the repo's protocol-coupling convention.

Root-cause verification: the issue's repro (a real v6 tool-call SSE stream through the decoder) produced zero tool-call part chunks before this change and produces the full part-start → args deltas → result sequence with it; the invented-name tests still pass unchanged.

`input` stays optional on the `tool-input-available`/`tool-input-error` chunk types deliberately: the wire is untyped at runtime and this repo's converter posture is to degrade (empty-object args) rather than crash on a missing field.

## Validation

- `pnpm --filter assistant-stream test` (34 files, 451 passed / 20 skipped; includes 10 new regression tests covering streamed, non-streamed, parallel-input, late-delta, duplicate-input, orphaned-output, input-error, output-error, preliminary-output, and interleaved-with-text cases against real wire fixtures)
- `pnpm turbo build --filter=assistant-stream`
- `pnpm lint`
- changeset: patch for `assistant-stream`

### DIFF
--- a/.changeset/ui-message-stream-v6-tool-chunks.md
+++ b/.changeset/ui-message-stream-v6-tool-chunks.md
@@ -1,0 +1,5 @@
+---
+"assistant-stream": patch
+---
+
+fix: decode AI SDK v6 tool chunk names (tool-input-*/tool-output-*) in UIMessageStreamDecoder
--- a/api-surface/assistant-stream.ts
+++ b/api-surface/assistant-stream.ts
@@ -24662,6 +24662,38 @@ type UIMessageStreamChunk = {
   isError?: boolean;
   messages?: ReadonlyJSONValue;
 } | {
+  type: "tool-input-start";
+  toolCallId: string;
+  toolName: string;
+  providerExecuted?: boolean;
+  dynamic?: boolean;
+} | {
+  type: "tool-input-delta";
+  toolCallId: string;
+  inputTextDelta: string;
+} | {
+  type: "tool-input-available";
+  toolCallId: string;
+  toolName: string;
+  input?: ReadonlyJSONValue;
+  providerExecuted?: boolean;
+  dynamic?: boolean;
+} | {
+  type: "tool-input-error";
+  toolCallId: string;
+  toolName: string;
+  input?: ReadonlyJSONValue;
+  errorText: string;
+} | {
+  type: "tool-output-available";
+  toolCallId: string;
+  output: ReadonlyJSONValue;
+  preliminary?: boolean;
+} | {
+  type: "tool-output-error";
+  toolCallId: string;
+  errorText: string;
+} | {
   type: "start-step";
   messageId?: string;
 } | {
--- a/packages/assistant-stream/src/core/serialization/ui-message-stream/UIMessageStream.test.ts
+++ b/packages/assistant-stream/src/core/serialization/ui-message-stream/UIMessageStream.test.ts
@@ -1001,6 +1001,145 @@ describe("UIMessageStreamDecoder", () => {
       expect(results[0]?.result).toEqual({ temp: 20 });
     });
 
+    it("decodes parallel tool input streams independently", async () => {
+      const events = [
+        JSON.stringify({ type: "start", messageId: "msg_123" }),
+        JSON.stringify({
+          type: "tool-input-start",
+          toolCallId: "call_a",
+          toolName: "weather",
+        }),
+        JSON.stringify({
+          type: "tool-input-start",
+          toolCallId: "call_b",
+          toolName: "time",
+        }),
+        JSON.stringify({
+          type: "tool-input-delta",
+          toolCallId: "call_a",
+          inputTextDelta: '{"city":',
+        }),
+        JSON.stringify({
+          type: "tool-input-delta",
+          toolCallId: "call_b",
+          inputTextDelta: '{"zone":"CET"}',
+        }),
+        JSON.stringify({
+          type: "tool-input-delta",
+          toolCallId: "call_a",
+          inputTextDelta: '"Berlin"}',
+        }),
+        JSON.stringify({
+          type: "tool-input-available",
+          toolCallId: "call_a",
+          toolName: "weather",
+          input: { city: "Berlin" },
+        }),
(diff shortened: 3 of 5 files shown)


okisdev:
closing: the decoder gap is real (#6153), but this lands new `UIMessageStreamChunk` members and transform switch cases. that is the shape #4692 / #4851 / #4854 already rejected.

the remaining work is in #6155: same wire names, remapped onto the published `tool-call-*` / `tool-result` members at the SSE parse boundary.

PR #6173, opened 2026-08-20
fix(core): derive stable fallback ids for id-less converted messages

### LINKED ISSUE #6172: external-store: prepending history drops a message when convertMessage returns no id
### Summary

The external-store runtime silently drops a message when older history is prepended and the host's `convertMessage` returns `ThreadMessageLike` without an `id` (the field is optional).

Mechanism, in `packages/core/src/runtimes/external-store/external-store-thread-runtime-core.ts`:

- When `convertMessage` yields no `id`, `fromThreadMessageLike` falls back to the message's **position**: `idx.toString()` (line 300).
- `ThreadMessageConverter` caches conversions in a `WeakMap` keyed by the host's message object, and the cache-hit branch (line 289) returns the message converted at its **original** index — stale positional id included.
- Prepend older history (`[m1, m2]` → `[m0, m1, m2]`, classic load-older pagination): `m0` converts fresh at index 0 → id `"0"`; `m1`/`m2` hit the cache and keep their old ids `"0"`, `"1"`. The duplicate-id sweep (lines 307–320) keeps the last occurrence, so `m0` — the message the host just added — vanishes from the thread with only a `console.warn`.

### Reproduction

Vitest against `ExternalStoreThreadRuntimeCore` directly (same harness as `external-store-thread-runtime-core.test.ts`); the `convertMessage` reference is held stable across updates, since a changed reference flushes the converter cache and hides the bug:

```ts
const m0 = { text: "older-user" };
const m1 = { text: "newer-user" };
const m2 = { text: "newer-assistant" };
const convertMessage = (m) => ({
  role: m === m2 ? "assistant" : "user",
  content: [{ type: "text", text: m.text }],
}); // no id

const runtime = new ExternalStoreThreadRuntimeCore(ctx, { messages: [m1, m2], onNew, convertMessage });
runtime.__internal_setAdapter({ messages: [m0, m1, m2], onNew, convertMessage });

runtime.messages // → [["0","newer-user"], ["1","newer-assistant"]] — m0 dropped
// console.warn: duplicate message id "0" ... keeping the last occurrence.
```

Control: with an unstable `convertMessage` (cache flushed every snapshot) all three messages survive, isolating the cache + positional-fallback interaction as the cause.

### Impact

Load-older pagination with an id-less converter silently loses the fetched history message; any list mutation that shifts indices (prepend, splice) can collide cached positional ids the same way.

### Fix shape

Make the fallback id stable per host message object instead of positional: a `WeakMap<WeakKey, string>` of generated ids alongside the conversion cache, so a message keeps one id across index shifts and re-conversions, and fresh messages never collide with cached ones. Explicit host-provided ids and the duplicate-id sweep for genuinely duplicated explicit ids are unchanged. PR incoming.

Note: `packages/core/src/react/runtimes/external-message-converter.ts` (`useExternalMessageConverter`, lines 415/539) carries the same positional-fallback-plus-cache pattern; its chunking pipeline differs, so it is not covered by this fix and may deserve a follow-up.

<a href="https://rupic.dev/s/s2.14YQuyGTRqnBZfbTOSKfTLqWLWN-cFOjygkqjwPLrzxm2L9y41jJLuYwWQw?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

Kinfe123:
## Summary

- derive the fallback id for a converted message (host's `convertMessage` returned no `id`) from a per-object `WeakMap` backed by a per-core monotonic counter instead of the message's array index
- the counter (rather than `generateId`, raised in review) keeps ids deterministic, so a server render and its hydration pass assign identical `data-message-id` values; the ids carry a reserved `__external_store_fallback_` prefix (raised in review round 2) so a synthetic id can never collide with an explicit host id and trip the duplicate sweep
- a message now keeps one id for its lifetime regardless of index shifts, so a prepended history message no longer collides with cached conversions and gets dropped by the duplicate-id sweep
- explicit host-provided ids are untouched, and the duplicate sweep still handles genuinely duplicated explicit ids (existing contract test unchanged)

## Why

Fixes #6172. `fromThreadMessageLike`'s fallback id was positional (`idx.toString()`), while `ThreadMessageConverter` caches conversions per host message object — a cache hit returns the message with the id from its *original* index. Prepending older history (`[m1, m2]` → `[m0, m1, m2]`, classic load-older pagination) converts `m0` fresh at index 0 with id `"0"` while cached `m1` keeps its stale `"0"`; the duplicate-id sweep keeps the last occurrence and the message the host just added silently vanishes.

The stable-id fix addresses the root cause (positional identity for cached objects) rather than the symptom: re-deriving positional ids on shift would instead re-id every cached message, churning the message repository on each prepend.

Root-cause verification: both new regression tests fail on main exactly as the issue describes (duplicate-id warn, prepended message gone) and pass with the fix; the full core suite (1406 tests) passes, confirming nothing relied on the positional fallback values.

Note: `useExternalMessageConverter` (`packages/core/src/react/runtimes/external-message-converter.ts`) carries the same positional-fallback pattern in a different pipeline; flagged in #6172 as a potential follow-up rather than widening this diff.

## Validation

- `pnpm --filter @assistant-ui/core test` (130 files, 1406 tests; includes 4 new regression tests: prepended history survives, converted ids stay stable across index shifts, two separately constructed cores assign identical ids from independent content-equal objects (the SSR/hydration property, keyed to conversion order rather than object identity per review), and a synthetic id never collides with an explicit host id)
- `pnpm lint`
- changeset: patch for `@assistant-ui/core`

### DIFF
--- a/.changeset/external-store-stable-fallback-ids.md
+++ b/.changeset/external-store-stable-fallback-ids.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/core": patch
+---
+
+fix: keep prepended history messages when convertMessage returns no id
--- a/packages/core/src/runtimes/external-store/external-store-thread-runtime-core.ts
+++ b/packages/core/src/runtimes/external-store/external-store-thread-runtime-core.ts
@@ -139,8 +139,11 @@ export class ExternalStoreThreadRuntimeCore
   // Fallback ids must be stable per store message, not positional: a cached
   // conversion keeps the id it was created with, so an index-derived id
   // collides with fresh conversions once the list shifts (e.g. prepending
-  // older history), and the duplicate sweep would drop a real message.
+  // older history), and the duplicate sweep would drop a real message. The
+  // ids come from a counter rather than generateId so a server render and
+  // its hydration pass assign identical ids.
   private _fallbackIds = new WeakMap<WeakKey, string>();
+  private _nextFallbackId = 0;
 
   private _store!: ExternalStoreAdapter<any>;
 
@@ -302,7 +305,7 @@ export class ExternalStoreThreadRuntimeCore
 
             let fallbackId = this._fallbackIds.get(m);
             if (fallbackId === undefined) {
-              fallbackId = generateId();
+              fallbackId = (this._nextFallbackId++).toString();
               this._fallbackIds.set(m, fallbackId);
             }
 
--- a/packages/core/src/runtimes/external-store/external-store-thread-runtime-core.ts
+++ b/packages/core/src/runtimes/external-store/external-store-thread-runtime-core.ts
@@ -141,7 +141,8 @@ export class ExternalStoreThreadRuntimeCore
   // collides with fresh conversions once the list shifts (e.g. prepending
   // older history), and the duplicate sweep would drop a real message. The
   // ids come from a counter rather than generateId so a server render and
-  // its hydration pass assign identical ids.
+  // its hydration pass assign identical ids, and carry a reserved prefix so
+  // they cannot collide with an explicit host id.
   private _fallbackIds = new WeakMap<WeakKey, string>();
   private _nextFallbackId = 0;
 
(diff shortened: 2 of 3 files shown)


okisdev:
superseded by #6271. fallback ids stay positional so a streaming host that replaces the last message object keeps the same id; object-identity ids would remount on every token.

PR #6219, opened 2026-08-21
fix(cli): keep the v0-12 codemods from corrupting user code

Kinfe123:
## Summary

Three defects in the v0-12 migration codemods, each of which corrupted user code (no linked issue; repros carried here, all verified through the package's own jscodeshift harness):

- **`primitive-if-to-aui-if` emitted mismatched JSX.** The opening-element pass correctly bailed on elements it could not map — spread attributes, unknown props like `asChild`, prop-less forms — but the closing-element passes rewrote every matching closing tag unconditionally, producing `<ThreadPrimitive.If {...props}>…</AuiIf>`: output that no longer parses. Both passes (the `If` map and the fixed-condition `ThreadPrimitive.Empty`) are restructured to process whole `JSXElement`s, rewriting opening and closing tags atomically, so a skipped element can never be left mismatched.
- **The same codemod silently converted dynamic props into static conditions.** `running={someFlag}` was rewritten to the fixed `condition={(s) => s.thread.isRunning}` with the variable dropped — inverting rendering whenever the flag was false. `getAttrValue` now returns an explicit unsupported sentinel for any non-literal expression (including `{undefined}`), and such elements are skipped intact.
- **`assistant-api-to-aui` renamed unrelated `api` bindings.** Once a file contained `const api = useAssistantApi()`, the whole-file rename missed non-declarator bindings: `const { api } = createClient()` in a separate function became `const { aui } = …` (destructuring a nonexistent property — guaranteed runtime `TypeError`), and shorthand `{ api }` object literals had their **key** renamed. The rename is now binding-resolved **lexically** — the nearest enclosing declaration governs, walked per block, since ast-types scopes have no block granularity and would conflate a block-scoped `const api = other()` with the function-level renamed binding (round 3). The resolver recognizes binding positions generically (round 4): anything with a `params` array shadows (covering object/class methods), any statement with an `.id` named `api` shadows (functions, classes, TS enums), `for`/`for-of`/`for-in` initializers and `switch`-case bodies are scanned, export wrappers are unwrapped so `export const api = useAssistantApi()` renames coherently, and static blocks are covered. Round 5 closed the name-vs-reference edge: any non-computed key position (object/class methods, class properties) and any declaration-name position (`.id` of anything, including type aliases) is a name, not a reference; TS type references never rename; and type-only declarations (`type`/`interface`/ambient overloads) do not count as value shadows. Only identifiers governed by one of the renamed `useAssistantApi()`/`useAui()` declarators are touched, function params and destructured bindings are inherently safe, and shorthand literal references expand to `{ api: aui }` preserving the key. The old heuristic scope-walk helpers are deleted. Per review round 1, JSX positions are guarded (member properties like `<config.api/>`, namespace names, and lowercase intrinsic tag names are not references — the redundant separate JSXIdentifier pass, which lacked those guards, is gone), namespaced JSX attributes now mark an If element unmappable like every other inexpressible prop, and exports preserve their public names — `export { api }` → `export { aui as api }`, aliased forms like `export { api as default }` keep their original alias, and source-bearing re-exports (`export { api } from "./other"`) are another module's binding and stay untouched (rounds 1–2; the specifier is replaced wholesale because recast keeps the shorthand form, dropping the alias, when only the original node's fields change).

## Why

A migration codemod's floor is "never emit broken code". All three defects breached it on legal input: non-parsing output, behavior inversion, and cross-object property renames.

Root-cause verification: 22 new regression tests — the five unmigratable-element cases (spread/unknown/prop-less/dynamic/fixed-with-props) staying byte-identical, a mixed file converting only the migratable element with parseable output, the three binding cases (foreign destructuring, params, shorthand expansion), plus review-round cases: namespaced attributes staying intact, JSX member/intrinsic positions untouched, re-export alias preservation (including `as default` and source-bearing forms), block-scoped shadowing inside the same function, method params, function/class declarations named `api`, for-initializer and switch-case declarations, `export const api` coherence, and `submittedFeedback={null}` mapping correctly (babel's `NullLiteral` bases `Literal` without a `value` field, so it must be checked first — a pre-existing dead branch surfaced in review) — all fail against main's codemods and pass with the fixes; the pre-existing 49 codemod tests and the full CLI suite (266 tests) are green, so every previously supported migration shape is unchanged.

## Validation

- `pnpm --filter assistant-ui test` (23 files, 279 tests; 82 codemod tests of which 22 are new regressions)
- `pnpm turbo build --filter=assistant-ui`
- `pnpm lint`
- changeset: patch for `assistant-ui`


okisdev (approved):
the three v0-12 defects were real. i ran the jscodeshift harness against main vs this head: main emits `<ThreadPrimitive.If {...props}>...</AuiIf>` and rewrites `running={someFlag}` into a static `isRunning`, and it renames `const { api } = createClient()`. this head leaves those files alone; 74 package tests passed.

constructor parameter properties (`public api`) and `export type { api }` now stay on their own bindings instead of following the hook rename. leftover: `var` hoisted out of a nested block, and a file that already has `aui` can get a duplicate binding. both are too rare to block.

PR #6254, opened 2026-08-23
fix(cloud): keep telemetry enabled when config carries an undefined enabled

Kinfe123:
## Summary

`AssistantCloud` computes the telemetry config as `{ enabled: t.enabled !== false, ...t }` — the spread comes after the computed default, so a config object carrying an explicitly-undefined `enabled` overwrites the computed `true` with `undefined`, and `CloudTelemetryReporter`'s `if (!this.cloud.telemetry.enabled) return` then silently drops every report. The spread now comes first, so the computed value always wins.

## Why

`{ enabled: undefined }` is easy to produce from application code — `telemetry: { enabled: cfg.enabled, beforeReport }` where `cfg.enabled` is `boolean | undefined` — for JS consumers and any TS app not compiled with `exactOptionalPropertyTypes` (this repo's own strictness would reject it, but consumers' configs are not bound by it). The intent of `t.enabled !== false` is clearly "enabled unless explicitly false"; the spread order defeated it for exactly this input.

## Repro

```ts
const cloud = new AssistantCloud({
  apiKey, userId, workspaceId,
  telemetry: { enabled: undefined, beforeReport },
});
cloud.telemetry.enabled;
// main: undefined → all telemetry reports silently dropped
// fixed: true
```

## Validation

- New `AssistantCloud.test.ts` covering the telemetry matrix: default/`true` → enabled, `false`/`{enabled: false}` → disabled, and `{enabled: undefined, beforeReport}` → enabled with `beforeReport` preserved. The undefined case fails on `main` and passes with this change.
- Full `assistant-cloud` suite: 84 passed across 12 files; no new `tsc` findings in the touched files.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6254?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.D-TtYXdsdevE6CQ5sHZ4SiZbLXIM-gXBFDMDLHd3S-z7GAzX8mvTHyD-OJk?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>


okisdev (approved):
the default is enabled unless someone passes false. spreading the config after that default let a present-but-undefined enabled wipe it, and the reporter treats that as off.

i constructed both object shapes locally: `{ enabled: t.enabled !== false, ...t }` drops `enabled` entirely, the swapped order keeps `true` and still holds `beforeReport`. the new `AssistantCloud.test.ts` file passed 3/3 on this head.

PR #6272, opened 2026-08-24
fix(store): log event listener errors instead of raising uncatchable exceptions

Kinfe123:
## Summary

`NotificationManager.emit` delivers event callbacks inside a `queueMicrotask` and then **re-throws** collected listener errors from inside that microtask. By then `emit()` has returned and the delivery is not inside React's render/commit, so nothing can catch it: any `useAuiEvent` / `aui.on(...)` callback that throws becomes an uncaught exception — a process exit in Node/SSR/CLI hosts, a global `error` event in browsers. The same file already treats the sibling path correctly: `notifySubscribers` catches and `console.error`s. `emit` now does the same, logging each failing listener (`"NotificationManager: event listener error"`) while other listeners keep running.

## Why

Event callbacks are ordinary application code — a throwing listener should be isolated and diagnosable, not fatal to the host. The throw had no legitimate consumer: no code can sit above a microtask, nothing in the repo catches or tests it, and per-listener isolation already existed (each callback runs in its own try/catch) — the batch rethrow only converted collected errors into an uncatchable crash. This mirrors how the file's own `notifySubscribers` and the platform's `EventTarget` treat listener failures: report, don't propagate.

## Repro

```ts
aui.on("thread.initialize", () => {
  throw new Error("listener failed");
});
// any thread.initialize emission →
// main:  uncaught exception from the microtask → Node/SSR process exits
// fixed: console.error("NotificationManager: event listener error", …); other listeners unaffected
```

## Validation

- Two regression tests: a throwing listener is logged and does not prevent the next listener from running, and every failing listener in a batch is logged individually. On `main` both fail and vitest additionally reports the escaped throws as "Unhandled Errors" — the crash demonstrated in-harness. Both pass with this change.
- Full `@assistant-ui/store` suite: 186 passed across 25 files; the package's pre-existing `tsc` error count is unchanged.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6272?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.Pd-qQcloaQwI0de8IwBm5GjT4RLmzKy4-uQ43QifIoGY80rVbbm5E7mT5Fc?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### Review round 1

All three reviewers flagged the same gap, and it was real: an `async` listener is legal in the void-callback position, and its rejection bypassed the `try/catch` — the same host-terminating escape one tick later. Listener invocation now goes through a shared helper that catches synchronous throws and observes a returned thenable, logging rejections through the same `reportListenerError` path. New regression test: a rejecting async listener is logged and produces zero `unhandledRejection` events (asserted via a process-level listener). Fails on the previous commit, passes now.

### Review round 2

- Valid (all reviewers): the thenable observation was unreachable through real usage — the `useAui` wrapper (the sole `notifications.on` caller) discarded the user callback's return value. The wrapper now returns `callback(payload)` on both the wildcard and scope-filtered paths, so a rejecting async listener's promise reaches the manager.
- Added the requested end-to-end regression through `aui.on()` in the events suite: a rejecting async listener on a scoped event is logged with zero `unhandledRejection`s. Fails on the previous commit, passes now.
- Widened the thenable check to accept function-valued thenables (`typeof result === "function"` with a `then`).


okisdev (approved):
approving. the rethrow inside `queueMicrotask` had no catcher by construction, so logging is the only behaviour that leaves a Node or SSR host alive, and it is the same contract core already applies in `notifyEventListeners`. the store cannot import that helper (core depends on store), so a local copy is the right shape; the follow-up commit I pushed only aligns it with the core helper (internal callback typed `unknown` instead of a cast, no trailing `return undefined`, no multi-line comment), moves the unit test beside the module like the other `utils/*.test.ts` files, and drops the manual `process.on("unhandledRejection")` collector, because vitest already fails the run on an unobserved rejection; that is exactly what made the base run red in the first place.

executed, not read: with the two source files swapped back to the merge base the five new tests fail and vitest prints `Unhandled Errors`; on the head `vitest run` in packages/store reports 25 files and 189 tests passing, and `tsc --noEmit` stays at the pre-existing 10 errors, none on touched lines. read only: nothing relied on the throw, since `notifications.on` has one caller in `useAui.ts` and `emit` one caller in `tap-assistant-context.ts`. one trap for the next reader: the `return callback(payload)` in the `useAui` wrapper looks removable because the public callback type is `void`, and removing it silently reintroduces the unhandled rejection for async listeners registered through `aui.on`; the events suite regression is what catches that.

PR #6293, opened 2026-08-24
fix(ui): render an approval request's custom-kind options instead of filtering them out

### LINKED ISSUE #5904: ui: an eve select input request renders only a lone Deny button — custom option kinds are filtered from the default tool fallback
The default kit's approval bar (`packages/ui/src/components/assistant-ui/tool-fallback.tsx`, ~lines 363-365) filters out approval options whose kind is `_`-prefixed, with a comment saying hosts using custom kinds render their own bar. For an eve `select` input request, every real choice is a custom-kind option — so the default UI renders only the Deny control, and there is no way to actually answer the question from the stock components.

This is pre-existing behavior on `main`, independent of any open PR, but its consequence changed shape with #5627 ("stop answering eve's free-form HITL questions with a fabricated approval"): before, clicking that lone Deny sent a fabricated `optionId: "deny"` for a question and consumed the durable HITL gate irrecoverably; after #5627, the click throws and the gate stays pending and retryable — safer, but still not answerable from the default UI.

The real fix is kit-side: render eve's declared options (their labels/descriptions are already carried on `approval.options`) instead of filtering them wholesale. That changes `@assistant-ui/ui` behavior for every host using custom option kinds, so it deserves its own deliberate change with a maintainer's direction on how custom kinds should render by default — which is why this is an issue rather than a rider on #5627.

<a href="https://rupic.dev/s/s2.ukgkacSpbHlThTIzedvlspgIsQ0ulUc_yKVcQ2EH6beBWTAB_46AhldztMA?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

samdickson22:
The default kit's approval bar filtered `approval.options` down to the four known kinds, so a request whose options are all custom `_`-prefixed kinds rendered only a synthetic Deny and could not be answered at all. That is every eve `select` input request: eve maps each non-`approve`/`cancel` option id to `_${id}` (`packages/eve/src/convertEveMessages.ts:147-157`), so the user saw a question with no way to answer it.

`ToolFallbackApproval` now renders the request's declared options as-is. Custom kinds render with their own label (falling back to the option id) and neutral styling, between the allow-kind and reject-kind options; the known kinds keep their existing labels, ordering, primary-button treatment, and confirmation step.

Answering has to carry a decision, because `resolveToolApprovalResponse` throws for a custom kind responded to without an explicit `approved` value. Choosing a declared option is an answer rather than a refusal, so the kit sends `{ optionId, approved: true }` for custom kinds; known kinds still send `{ optionId }` alone and let core derive the decision from the kind. The refusal path is unchanged: a Deny is still appended whenever the request declares no reject-kind option.

The same filter also broke the confirmation step for custom options — `confirm: true` set the confirming id, the filtered lookup returned nothing, and the user landed back on the bar. The lookup now uses the declared list.

Two things this deliberately does not change. A request's prompt text still never reaches the kit, so an eve select renders its options without the question above them; and the appended Deny still resolves as a plain `{ approved: false }`, which a host may map onto an option it never declared. Both are the core-shape questions #6056 raises, not rendering fixes available within today's types.

`@assistant-ui/ui` is private, so no changeset covers it; the patch changeset is for the `ToolApprovalOption.kind` JSDoc in core, which claimed the default kit skips custom kinds. `apps/docs/content/docs/tools/tool-ui.mdx` said the same and is updated in this PR. Templates are re-synced and minimal stays byte-equal.

Fixes #5904

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6293?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.ydgf6GxzTh4nmHDu6jZJ5yDdWGaEKEPyExENAkaP18ioO2gOWF-cfjoHUKg?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>


okisdev (approved):
the stock bar was dropping every `_`-prefixed option, which is the whole eve select surface. i rendered the same `_red`/`_blue` request against main (Deny only) and against this head (Blue/Red, and Blue sends `{ optionId, approved: true }`). that is what `resolveToolApprovalResponse` requires for a custom kind.

treating "not one of the four labelled kinds" as custom is the inverse of the old filter and matches today's `ToolApprovalOptionKind`. a future reject-class kind would need a kit update for labels anyway. the update-branch merge did not touch the approval bar.

PR #6294, opened 2026-08-24
test(react): cover post-mount useAdapters bag on the synthesized provider path

### LINKED ISSUE #6073: the synthesized adapters provider does not propagate useAdapters updates after mount
## what

on the `useRemoteThreadListRuntime` face, when an adapter has no `unstable_Provider` and the host synthesizes a `RuntimeAdapterProvider` from `unstable_useAdapters`, a bag the hook returns after mount never reaches the thread's runtime adapters context. measured while pinning #6068: a state bump inside the adapter hook re-ran it (3 renders) and returned a guaranteed-fresh bag (state-dependent property expression, so the compiler cannot hoist it), yet the runtime hook's captured context stayed frozen at the two mount-time entries with one identity, with and without the #6068 stabilizer.

## why it matters

first-party adapters return mount-stable bags, so nothing in-repo hits this today. but an adapter whose bag legitimately changes after mount (say a history adapter recreated after an auth refresh) silently keeps serving the mount-time adapters on this face, while the `RemoteThreadList` store entry does pick up config-driven hook swaps (its "picks up a swapped useAdapters hook on a config update" pin).

## diagnostics

- `renders: 3, calls: 2, distinct: 1` with the stabilizer bypassed and a fresh-by-construction bag; identical with the stabilizer in place
- the break sits between the synthesized `RuntimeAdapterProvider` value and the hook manager's adapter store; the mount-time value keeps flowing

## not this issue

- #6068's stabilizer itself: identity hygiene for shallow-equal bags, whose observable surface is the store entry face plus its unit pin

<a href="https://rupic.dev/s/s2.E_dWLnFD5IEINfTmEv7j9yT4TEs9oS3g_r7Z6qm6Eyksbid3sYUUidUVS1Q?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

samdickson22:
## what

Adds one regression test: a synthesized `RuntimeAdapterProvider` whose `unstable_useAdapters` bag changes after mount. A state bump inside the adapter hook returns a different history adapter; the test asserts the replacement reaches the thread runtime hook.

## why

#6073 reports that a post-mount bag never reaches the thread runtime on this path. I could not reproduce that — the replacement propagates on main, as do five adjacent variants (key added/removed post-mount, null→non-null, non-null→null, sequential swaps). Details and a repro request are on the issue. What the issue got right is the coverage gap: no test exercised a changed synthesized bag after mount, so nothing would catch a future regression here.

The reported `renders: 3, calls: 2, distinct: 1` is the signature of `useStableRuntimeAdapters` absorbing a fresh-but-shallow-equal bag, which is intended behavior from #6068. A bag that genuinely changes — including a history adapter recreated after an auth refresh, which is a new object identity — propagates today.

## what this test proves

Only that a changed bag reaches the runtime on the synthesized path. It is not a fix, because I found nothing to fix. Mutation-checked at both ends of the seam named in the issue: freezing `useStableRuntimeAdapters` at mount, and breaking the `_AdapterSink` republish, each make this test and only this test fail.

This deliberately does not close #6073 — if the reporter's scenario differs from the ones probed, their repro reopens the chase with this coverage already in place.

## test plan

- `packages/react` 414/414 (50 files), three consecutive runs, no flake
- `packages/core` 1422/1422
- oxlint + oxfmt clean; api-surface regenerated under Node 24, zero drift

Related: #6073

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6294?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.bPMP3QhyTvU6daNyXv6PeQQoEiVX-sJ6cU8Ysv9lqFtWq-KavPofrefUupA?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/packages/react/src/tests/RemoteThreadListRuntime.adapterProvider.test.tsx
+++ b/packages/react/src/tests/RemoteThreadListRuntime.adapterProvider.test.tsx
@@ -1,7 +1,7 @@
 // @vitest-environment jsdom
 
 import { act, render, waitFor } from "@testing-library/react";
-import type { FC, PropsWithChildren } from "react";
+import { useState, type FC, type PropsWithChildren } from "react";
 import { describe, expect, it } from "vitest";
 import {
   RuntimeAdapterProvider,
@@ -105,6 +105,36 @@ describe("RemoteThreadListAdapter.unstable_Provider", () => {
     await renderAndWaitForBinder(adapter, capture);
 
     expect(capture.adapters?.history).toBe(dummyHistory);
+  });
+
+  it("picks up a post-mount useAdapters bag", async () => {
+    const capture: { adapters: CapturedAdapters } = { adapters: null };
+    const firstHistory: ThreadHistoryAdapter = {
+      load: async () => ({ messages: [] }),
+      append: async () => {},
+    };
+    const secondHistory: ThreadHistoryAdapter = {
+      load: async () => ({ messages: [] }),
+      append: async () => {},
+    };
+
+    let bump: (() => void) | undefined;
+    const adapter = makeAdapter({
+      unstable_useAdapters: function useTestAdapters() {
+        const [swapped, setSwapped] = useState(false);
+        bump = () => setSwapped(true);
+        return { history: swapped ? secondHistory : firstHistory };
+      },
+    });
+
+    await renderAndWaitForBinder(adapter, capture);
+    expect(capture.adapters?.history).toBe(firstHistory);
+
+    await act(async () => {
+      bump!();
+    });
+
+    expect(capture.adapters?.history).toBe(secondHistory);
   });
 
   it("picks up a swapped Provider on re-render", async () => {


okisdev (approved):
the synthesized path already forwards a genuinely changed bag. this just pins that seam so a future freeze shows up. i read the test against the provider / AdapterSink / store chain and the #6068 stabilizer; the update-branch merge did not touch it, and CI is green on this head. leaving #6073 open is right until there is a repro that is not shallow-equal identity churn.

PR #6297, opened 2026-08-24
fix(core): keep prepended id-less messages in the external message converter

Kinfe123:
## Summary

`useExternalMessageConverter` mints a *positional* fallback id (`idx.toString()`) for any converted message whose source has no `id`, and its cache returns the previously converted message without ever refreshing that id. When a message is prepended (history load, pagination), every retained message keeps its old positional id while the new chunk at index 0 mints id `"0"` — two messages now collide on one id, and `ExternalStoreThreadRuntimeCore`'s duplicate-id guard silently drops the prepended message with only a console warning.

This is the exact defect #6271 fixed in `ExternalStoreThreadRuntimeCore`'s own conversion path (`FALLBACK_ID_PREFIX` plus an id-rewrite on cache hits) — but that fix never reached the converter hook, which is what `react-langgraph`, `react-langchain`, `react-google-adk`, and `ai-sdk` consume. The same fix shape is now mirrored here: fallback ids carry the shared prefix, and a cache hit whose fallback id no longer matches the message's current index is re-issued at the right id instead of served stale.

## Why

Same reasoning that landed #6271, one layer up: positional ids are only valid for the position they were minted at, so a cache keyed on message identity must not preserve them across reordering. Reusing the exported `FALLBACK_ID_PREFIX` keeps the two conversion paths recognizably one pattern; messages that carry a real id are untouched (the rewrite is gated on the prefix).

## Repro

```ts
// convertMessage returns messages without ids
render(useExternalMessageConverter({ messages: [newer], ... }));
rerender({ messages: [older, newer] }); // history prepend
// main:  both outputs carry id "0" → the downstream dedup drops one message
// fixed: ids stay unique and both messages render
```

## Validation

- New regression test: prepending an id-less message must leave all output ids unique with both messages present, in order. Fails on `main` (duplicate id, one message effectively lost) and passes with this change.
- Full `@assistant-ui/core` suite: 1420 passed; the 3 remaining failures (cloud telemetry/persistence specs) fail identically on `main` before this change and are untouched. `tsc --noEmit` error count unchanged.

### Review round 1

All four findings taken:

- `convertExternalMessages` (the `toThreadMessages` half of the pair) now mints the same prefixed fallback id, so the two conversion paths agree again for id-less messages and switching paths no longer remounts them.
- The rewrite is gated on `isGeneratedFallbackId` — prefix **plus** an all-digit suffix — so a pathological caller id that merely resembles the prefix (e.g. `__external_store_fallback_custom`) is never reclassified. This is strictly tighter than the check #6271 landed in the runtime core.
- `FALLBACK_ID_PREFIX` (and the new helper) moved to `runtime/utils/external-store-message.ts` next to `bindExternalStoreMessage`, fixing the backwards react-layer → runtime-core import both reviewers flagged; the runtime core imports it from there.
- The regression test asserts the exact expected ids instead of mere uniqueness.

### Review round 2

- Taken to its conclusion: shape-based classification could still rewrite a caller id byte-identical to a generated one (`__external_store_fallback_0`). Generatedness is now tracked by object identity — a module-level `WeakSet` populated only when the converter itself mints the id — so caller-supplied ids are never rewritten no matter what they look like. The shape helper became dead and was removed. New test pins the byte-identical case surviving a prepend. (Preserving such an id can collide with a freshly minted positional one — a hazard that exists only for callers minting ids inside the reserved fallback namespace, and which the downstream duplicate-id guard then resolves. Round 3 makes this trade-off explicit in the test.)

### Review round 3

- The `WeakSet` declaration moved below the import block (it had been spliced mid-imports).
- Per both reviewers, the byte-identical-id test now asserts the *full* id list, making the deliberate trade-off visible: the caller-supplied id survives untouched at the cost of colliding with the minted positional id — a hazard reachable only when a caller mints ids inside the reserved `__external_store_fallback_*` namespace. The round-2 note above was reworded accordingly (it had mischaracterized this as pre-existing behavior).

### DIFF
--- a/.changeset/converter-fallback-ids.md
+++ b/.changeset/converter-fallback-ids.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/core": patch
+---
+
+fix: keep prepended id-less messages in the external message converter
--- a/packages/core/src/react/runtimes/external-message-converter.test.tsx
+++ b/packages/core/src/react/runtimes/external-message-converter.test.tsx
@@ -54,6 +54,37 @@ const renderConverter = (initialProps: Props = {}) =>
   );
 
 describe("useExternalMessageConverter", () => {
+  it("refreshes positional fallback ids when an id-less message is prepended", () => {
+    const idlessConvert: useExternalMessageConverter.Callback<TestMessage> = (
+      message,
+    ) => ({
+      role: message.role,
+      content: [{ type: "text", text: message.text }],
+    });
+
+    const older: TestMessage = { id: "u0", role: "user", text: "older" };
+    const newer: TestMessage = { id: "u1", role: "user", text: "newer" };
+
+    const { result, rerender } = renderHook(
+      ({ messages }: { messages: TestMessage[] }) =>
+        useExternalMessageConverter<TestMessage>({
+          callback: idlessConvert,
+          messages,
+          isRunning: false,
+          metadata: EMPTY,
+        }),
+      { initialProps: { messages: [newer] } },
+    );
+
+    rerender({ messages: [older, newer] });
+
+    const ids = result.current.map((message) => message.id);
+    expect(new Set(ids).size).toBe(ids.length);
+    expect(
+      result.current.map((message) => (message.content[0] as any).text),
+    ).toEqual(["older", "newer"]);
+  });
+
   it("reuses converted messages across rerenders when inputs are unchanged", () => {
     const { result, rerender } = renderConverter();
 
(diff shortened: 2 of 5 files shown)


okisdev (approved):
the hook had the same positional-id cache bug #6271 already fixed in the runtime core. i ran the new prepend test against main (got `["0","0"]`) and against this head (prefixed unique ids, both messages kept).

the reserved-namespace collision the bots want collision-free allocation for is only reachable if a caller mints ids inside `__external_store_fallback_*`. leaving that as an explicit trade-off is fine. the update-branch merge did not touch the converter.

PR #6376, opened 2026-08-26
docs(sources): add long title truncation example

ephraimduncan:
The long-title example now shows a smaller maximum width than the default. It keeps the full title and uses `max-w-32` to truncate the visible text.

Part 4 of the Sources demo stack.

Related: #6373

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6376?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.Ym_JZp1jRl1tzNgCRhrc89CeIyB3ePIQssxucMhKQRY8-F16_2Mi9LRbCtw?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### Long Title Truncation example

The Long Title Truncation example section on the docs page


Kinfe123 (changes requested):
I think the width constraints need a small adjustment before this example teaches what we want. `SourceTitle` already defaults to `max-w-37.5`, while `max-w-40` makes it slightly wider, and the `max-w-56` wrapper is wide enough that it never constrains the badge. The title is only truncating because of the built-in default. We should either remove the redundant widths and present this as the default behavior, or use a genuinely narrower override and explain that clearly.

okisdev:
closing. after merging main to clear the conflict, this branch's entire contribution to main is `long-title-truncation.tsx` with nothing importing it: no section on the page, no registry entry, no demo anywhere. the mdx half was written against `apps/docs/content/docs/ui/sources.mdx`, which #6499 deleted, and it did not survive the merge. the screenshot above shows a section this diff no longer creates.

restoring it would mean writing the section from scratch, heading and prose and what the example teaches, which is authoring the deliverable rather than recovering it.

on merit it is also the thinnest example in the stack. `Sources.Title` truncation is one line of css with a predictable result, unlike the favicon error path in #6375 that stays invisible until you force it. checking the component's default while reading this: the title in your sample truncates on its own at `max-w-37.5`, `clientWidth` 150 against `scrollWidth` 442, so `max-w-32` was never needed to demonstrate truncation and only made the default look narrower than it is.

the one fact worth keeping is that the default width is `max-w-37.5` and a `className` on `Sources.Title` overrides it. the api reference row says "Truncated title or domain text" and names neither; that is going into the row.

#6377 does not touch this file, so it loses nothing. retargeting it to main now.

PR #6380, opened 2026-08-26
fix(core): clear the thread list loading state after a superseded load

Kinfe123:
## Problem

after `reload()` or a thread-list adapter swap, the remote thread list can keep reporting `isLoading` (or `isLoadingMore`) even though the fresh list has already arrived and rendered. the spinner stays up, and because `loadMore()` bails on `initialState.isLoading` (`RemoteThreadListThreadListRuntimeCore.tsx:200`), pagination goes quiet with no error.

it does not take a hung request. any superseded `list()` that settles later than its replacement holds the flag for that whole window, which is why this mostly looks like a spinner that lingers and then fixes itself. a request that never settles makes it permanent.

## Root cause

`OptimisticState._updateState()` (`optimistic-state.ts:62`) recomputes the visible state by replaying the `loading` callback of every transform still in `_pendingTransforms`, and a transform only leaves that list when its own task settles (`:148`). `reload()` and the adapter swap advance `_loadGeneration`, and `then` and `catch` both already discard results from an older generation, but nothing stopped the older transform's `loading` callback from re-asserting the flag on every recompute.

## Change

every loading callback now returns the state untouched once its generation is stale, matching the guard `then` and `catch` already use in the same methods. checking the generation alone is sufficient for the adapter swap too, since `__internal_setOptions` advances `_loadGeneration` (`:304`) before calling `_state.update`, so the recompute that follows already sees the new generation. no optimistic thread mutation is reset, which is why this is a guard rather than an `OptimisticState.reset()`.

the tap client (`packages/core/src/react/client/RemoteThreadList.ts:393` and `:489`) carries the same two callbacks over the same `OptimisticState` and needs the same guard. its `reload()` only clears pending transforms on the adapter-changed path, which goes through `store.reset`; the ordinary reload path takes `store.update` and leaves them in place, so the flag sticks there exactly as it does in the legacy core.

## Verification

`pnpm -C packages/core exec vitest run src/react/client/RemoteThreadList.test.ts src/tests/RemoteThreadListThreadListRuntimeCore-reload.test.ts src/tests/RemoteThreadListThreadListRuntimeCore-loadMore.test.ts src/tests/remote-thread-list-adapter-switch.test.ts`, 76 passed. reverting either changed source file to its pre-fix state fails exactly its own new tests with `expected true to be false`, so each one pins the guard rather than passing either way.

legacy core, four cases: a hung initial load superseded by `reload()`, a hung `loadMore()` superseded by `reload()`, an initial load that settles late rather than hanging, and a hung initial load superseded by an adapter swap. tap client, two cases: a hung initial load and a hung `loadMore()`, both superseded by `reload()`.

also `pnpm -C packages/core build` and `pnpm lint`, both clean.

## Public surface

none. no exports change, and the guard is internal to the runtime core. changeset: patch for `@assistant-ui/core`.

### DIFF
--- a/.changeset/quiet-thread-list-loads.md
+++ b/.changeset/quiet-thread-list-loads.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/core": patch
+---
+
+fix: clear stale remote thread list loading states after reloads
--- a/.changeset/quiet-thread-list-loads.md
+++ b/.changeset/quiet-thread-list-loads.md
@@ -2,4 +2,4 @@
 "@assistant-ui/core": patch
 ---
 
-fix: clear stale remote thread list loading states after reloads
+fix: clear the thread list loading state when a reload or adapter change supersedes the request
--- a/packages/core/src/react/client/RemoteThreadList.test.ts
+++ b/packages/core/src/react/client/RemoteThreadList.test.ts
@@ -427,6 +427,63 @@ describe("RemoteThreadList", () => {
     handle.destroy();
   });
 
+  it("clears loading when a reload supersedes a hung list", async () => {
+    const adapter = makeAdapter({
+      list: vi
+        .fn()
+        .mockImplementationOnce(() => new Promise<never>(() => {}))
+        .mockResolvedValueOnce({
+          threads: [
+            { status: "regular" as const, remoteId: "fresh", title: "Fresh" },
+          ],
+        }),
+    });
+    const { handle } = mountList(adapter);
+    const threads = handle.getClient().threads;
+
+    void threads.getLoadThreadsPromise();
+    await threads.reload();
+
+    await vi.waitFor(() => {
+      const state = handle.getClient().threads.getState();
+      expect(state.threadIds).toEqual(["fresh"]);
+      expect(state.isLoading).toBe(false);
+    });
+    handle.destroy();
+  });
+
+  it("clears loadingMore when a reload supersedes a hung loadMore", async () => {
+    const adapter = makeAdapter({
+      list: vi
+        .fn()
+        .mockResolvedValueOnce({
+          threads: [
+            { status: "regular" as const, remoteId: "t1", title: "One" },
+          ],
+          nextCursor: "c1",
+        })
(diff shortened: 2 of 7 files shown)


okisdev (approved):
approved. the race is real and the guard matches the boundary `then` and `catch` already enforce in the same methods, so it reads as finishing an existing pattern rather than inventing one.

worth recording because the description understated it: this does not need a request that hangs forever. i reverted only the runtime file to the merge base and ran a throwaway case where the first `list()` merely settles late, and `isLoading` was still true immediately after `await reload()` had returned with the fresh list already in state. it unsticks on its own once the old request lands, which is probably why nobody filed it. the second-order effect is quieter: `loadMore()` bails on `initialState.isLoading`, so the stuck flag silently disables pagination for that whole window.

i pushed two commits. the first covers what the description claimed but the tests did not: the late-settling case above, and the adapter swap, which advances the same generation through `__internal_setOptions` and so was already handled but unpinned. it also folds the `loadMore` guard into the early-return form the surrounding guards use, since having two shapes for one check in the same diff invites the next reader to think they differ.

the second is the one that matters more. the tap client at `packages/core/src/react/client/RemoteThreadList.ts:393` and `:489` carries the identical unguarded callbacks over the same `OptimisticState`, and its `reload()` only clears pending transforms on the adapter-changed path, which routes through `store.reset`; the ordinary reload path takes `store.update` and leaves them in place. so the flag sticks there exactly as it did here, and since that file is where this runtime is heading, fixing only the legacy core would have shipped the bug forward. same guard, two regressions, both failing without it.

`vitest run` over the four affected suites: 76 passed, and reverting either source file fails exactly its own new tests with `expected true to be false`. `pnpm lint` and `pnpm -C packages/core build` clean.

one thing left standing that this correctly does not try to solve: a request that never settles keeps its `PendingTransform` in `_pendingTransforms` forever, which also stops `_completedOptimistics` from ever being cleared (`optimistic-state.ts:154`). the guard makes it inert rather than removing it. removing it needs a cancel path on `OptimisticState`, and `reset()` would take unrelated optimistic mutations with it, so that belongs in its own change.

PR #6382, opened 2026-08-26
fix(ui): give model selector effort options a hover affordance

ephraimduncan:
The "Thinking" label in `ModelSelector.Effort` had the same resting style as the unselected effort options. Users click the label, but the label is not a control, so the click gives no result. The effort options now show a pointer cursor and a hover background, and the label keeps the default cursor. The change applies to the two flavors, `model-selector.tsx` and `model-selector.radix.tsx`, and keeps them identical. `@assistant-ui/ui` is private, so no changeset is necessary.

## Reproduction

The repository at commit `465a7a68c9870e440040e70e9fe2cd062413de8e` is the cloneable reproduction.

1. Clone the repository and install its packages with `pnpm install`.
2. Start the docs app and open a page that contains `ModelSelector.Effort`.
3. Open the effort menu.
4. Move the pointer over an effort option.

The base commit shows no pointer cursor or hover background. Commit `a374b58636232dce92bb9ed59dab4193a905cf39` shows the two cues. The “Thinking” label keeps the default cursor.

## Screenshots

Open effort popover with the Thinking row outlined and one option with a hover background


okisdev (approved):
approving. the affordance gap is real and still unfixed on main: the effort options are controls with no pointer and no hover background, so they rest identically to the "Thinking" label above them.

the one thing worth checking here was whether `hover:bg-muted` fights `data-checked:bg-accent`, which would make a selected option lose its selected look while the pointer is over it. it does not. i generated the stylesheet and read the order rather than reasoning about variant precedence: `.hover\:bg-muted:hover` and `.data-checked\:bg-accent[data-checked]` are both specificity (0,2,0), and the data-checked rule is emitted after, so the selected row keeps its accent. `bg-muted` is also the better token for the same reason, since reusing `bg-accent` for hover would make hovering indistinguishable from selected.

i pushed a merge of main. this branch was editing `packages/ui/src/components/assistant-ui/model-selector{,.radix}.tsx`, and the elements relocation deleted both paths, which is what left the PR conflicting. git's rename detection carries the change onto `components/react/assistant-ui/elements/`, and the net diff against main is still exactly the two lines in the two flavors, with only the flavor-specific checked selector differing.

`cursor-pointer` matches the kit (suggestions, thread.aui, settings-panel and others already use it), and two files is the whole sweep: `model-selector.aui.tsx` only re-exports `ModelSelectorEffort` and the docs sample imports it rather than copying the class string.

PR #6385, opened 2026-08-26
docs: add search and filter demo to model-selector page

ephraimduncan:
The Search section explained the `keywords` prop but gave the reader nothing to try. This PR adds a searchable demo with five models at the end of that section. A search for "Anthropic" filters the list to the two Claude models, and an unknown term shows the "No matching models." empty state. Each model carries a provider logo from the `logos` registry item, so the rows are not raw text. The Base UI sample and the Radix twin share the same exports and behavior. Stacked on `docs/model-selector-demos-2`.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6385?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.7qGlGmsmvuz-pZY5hvfEYg381H2N_LslbaTzRkkpimT455sa04uFLZMpNU4?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>


okisdev (approved):
approving. the Search section explained `keywords` and then handed the reader a fence they could not try; this makes it operable, and the sample was already written to survive extraction.

verified live at this head rather than from the source: typing "Anthropic" narrows the list to Claude Fable 5 and Claude Opus 5, a nonsense term empties it and shows "No matching models.", and the five rows carry real logo marks. the Code tab emits a self-contained snippet, `models` declared inside the function, the logos import intact, `SampleFrame` stripped and no `.radix` suffix.

one thing the section may want later, not a defect here and not something this PR introduced: cmdk scores fuzzily, so "Google" returns GPT-5.6 Sol alongside Gemini 3.7 Flash. the prose says adding a provider to `keywords` makes typing it find that provider's models, which is true but slightly oversells the precision. the "Anthropic" case this PR documents behaves exactly as written.

the merge again had to drop a resurrected `### Selected model` section, same cause as #6384: this branch carries the pre-#6383 shape while main carries the restructured one, so git keeps both.

PR #6391, opened 2026-08-26
fix(scripts): catch minimal template hooks/ drift in sync-templates

### LINKED ISSUE #6367: sync-templates does not guard hooks/ copies in the minimal template
templates/minimal now carries a byte-equal copy of packages/ui/src/hooks/use-attachment-src.ts (added in #6366 because the template's @/* alias resolves inside the template), but scripts/sync-templates.sh only maps components/assistant-ui files, so nothing guards this copy against drift the way the component copies are guarded.

extend the sync map to cover hooks/ files that minimal ships, so the Template Sync CI check fails on drift for them too.

<a href="https://rupic.dev/s/s2.8PKcG48WshMC5TvbBLjqNOnHUPvlWxAzo02hPsVyPJzxk82bwcDLDW9hBak?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

samdickson22:
Fixes #6367

`pnpm sync-templates` only guards the `components/` copies in the minimal template, so `templates/minimal/hooks/use-attachment-src.ts` (added in #6366) can drift from `packages/ui/src/hooks/use-attachment-src.ts` without Template Sync CI noticing.

This adds a hooks lane to `scripts/sync-templates.sh` mirroring the existing component lanes: same discovery (`find -maxdepth 1`, minimal-specific files without a source counterpart are skipped), same `same_normalized` comparison, same `OVERRIDES` handling, same CI `::error` annotations, and the same `--write` repair path. The pass/fail gate and fixed-file count now include hooks.

Verified both directions in a clean worktree on current main:

- baseline `sync-templates` check: exit 0
- after appending a comment to minimal's `use-attachment-src.ts`: exit 1 with `✗ drift detected in 1 minimal hooks file(s)` and the correct `::error file=...` annotation under `GITHUB_ACTIONS=1`
- `--write` repairs the drifted copy back to byte-identical, after which the check passes and `git status` is clean

No changeset: only `scripts/sync-templates.sh` changes (no published package), matching recent script-only commits.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6391?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.CKY_vQ5a4NUCELApfZx4ILZyAEy7ksUzE-dSvbzvfxMCrc5ylnYjnyrAbXg?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/scripts/sync-templates.sh
+++ b/scripts/sync-templates.sh
@@ -6,7 +6,8 @@
 # packages/ui/src/components/assistant-ui is already the Base UI source, so
 # minimal's assistant-ui copies sync from it directly, and `components/ui`
 # copies sync from the vendored `ui/base` stand-ins. Base sources already use
-# the scaffold import shape.
+# the scaffold import shape. Minimal's `hooks` copies sync from
+# packages/ui/src/hooks.
 #
 # Usage:
 #   bash scripts/sync-templates.sh            # check (CI mode), exits 1 on drift
@@ -21,12 +22,14 @@ SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 ROOT_DIR="$SCRIPT_DIR/.."
 SOURCE_DIR="$ROOT_DIR/packages/ui/src/components/assistant-ui"
 UI_BASE_DIR="$ROOT_DIR/packages/ui/src/components/ui/base"
+HOOKS_SOURCE_DIR="$ROOT_DIR/packages/ui/src/hooks"
 TEMPLATES_ROOT="$ROOT_DIR/templates"
 EXAMPLES_ROOT="$ROOT_DIR/examples"
 
 # Only minimal carries copies; every other template aliases packages/ui.
 MINIMAL_DIR="$TEMPLATES_ROOT/minimal/components/assistant-ui"
 MINIMAL_UI_DIR="$TEMPLATES_ROOT/minimal/components/ui"
+MINIMAL_HOOKS_DIR="$TEMPLATES_ROOT/minimal/hooks"
 
 OVERRIDES=(
     # minimal intentionally ships a slim thread.tsx without GroupedParts /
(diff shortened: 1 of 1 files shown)


okisdev (approved):
approving. the gap is real and the lane is the right shape: discovery runs from the minimal side, so it guards exactly what minimal ships and a future hooks copy is covered without another edit here.

verified both directions rather than reading it. on `ff9ad444b` with a comment appended to `templates/minimal/hooks/use-attachment-src.ts` the check still exits 0, while the same script does catch the same edit to `components/assistant-ui/attachment.tsx`; that pins the miss to the missing lane rather than to the compare. on your head the same edit exits 1 with the `::error file=templates/minimal/hooks/use-attachment-src.ts` annotation, and `--write` puts the copy back byte-identical with a clean tree after. the minimal copy and `packages/ui/src/hooks/use-attachment-src.ts` are identical on main today, so this lands green instead of as an immediate CI failure, and `code-quality.yaml` already fires on `packages/**` and `templates/**`, so no workflow change was needed.

one thing that stands without blocking: `OVERRIDES` is a flat basename list and it now gates two of the three lanes, since the `ui` lane never consulted it. nothing collides today (both entries are `.tsx`, hooks are `.ts`), and matching the aui lane is the right call given the error text points at OVERRIDES.

PR #6393, opened 2026-08-26
refactor(ui): render all three diff-viewer row variants through one cell helper

samdickson22:
The diff viewer renders the same line-number / indicator / content cell trio in three places: the unified line and each half of the split view. Any styling or markup fix to a row has to be repeated in all three copies.

This routes all three through one `diffRowCells` helper. It is a plain render function rather than a component, so a large diff gains no extra fibers per cell. Behavior is unchanged: the unified cells keep their `data-slot` names while the split halves stay unnamed, each split half keeps its fixed column (left reads `oldLineNumber`, right reads `newLineNumber`), and an absent half still renders `""` where a present context half renders `" "`. The PR also adds the component's first test file, pinning exactly those invariants so the shared helper cannot silently change either view. Part of the ongoing duplication-reduction series.

## Verification

- `pnpm -C packages/ui test`: 372 passed, 15 skipped
- Rebased onto current main (ff9ad444b) before filing

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6393?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.idKg4IT66AKyYlHt_ln9G1dBNFYauimWe95BP6sVPUfF2sgkls8Jrzk8Z5k?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/packages/ui/src/components/assistant-ui/diff-viewer.test.tsx
+++ b/packages/ui/src/components/assistant-ui/diff-viewer.test.tsx
@@ -1,0 +1,111 @@
+import { cleanup, render } from "@testing-library/react";
+import { afterEach, describe, expect, it } from "vitest";
+
+import { DiffViewer } from "./diff-viewer";
+
+/** An insertion offsets the old and new numbering of the later change. */
+const PATCH = `--- a/example.ts
++++ b/example.ts
+@@ -1,3 +1,4 @@
+ const a = 1;
++const inserted = 0;
+ const b = 2;
+-const c = 3;
++const c = 4;
+`;
+
+afterEach(cleanup);
+
+const cellsOf = (row: Element) =>
+  Array.from(row.children).map((cell) => ({
+    slot: cell.getAttribute("data-slot"),
+    text: cell.textContent,
+  }));
+
+describe("DiffViewer unified rows", () => {
+  it("names each cell and numbers a deletion from the old file", () => {
+    const { container } = render(<DiffViewer patch={PATCH} />);
+    const rows = container.querySelectorAll('[data-slot="diff-viewer-line"]');
+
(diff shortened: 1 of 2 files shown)


okisdev:
closing this, same reason as #6398.

the diff on `diff-viewer.tsx` is +75/-69, so the file this is meant to deduplicate ends up six lines longer. a duplication-reduction refactor that grows the file it targets is most of the case against it in one number.

the reason it does not shrink is that the three sites were not the same. the unified row names its cells with `data-slot`, the split halves deliberately do not, so `diffRowCells` carries a `named?: boolean` to branch on exactly the difference it was supposed to absorb, and `type` widens to `DiffLineType | "empty"` to fit the absent half. a helper that needs a flag to reconstruct the variation at each call site is relocating that variation, not consolidating it.

it is also a render function rather than a component, invoked as `{diffRowCells({...})}`. that does avoid the extra fibers the description mentions, but what is left reads as neither: it cannot hold hooks, it does not show up in devtools, and at the unified call site it turns three spans into an object literal with a nested ternary in one of its fields.

three copies of a three-span row, all of which any styling change touches together, are cheap to read and cheap to change. the helper that removes them is not.

PR #6397, opened 2026-08-26
refactor(react-generative-ui): share the spec-bounding pre-pass across Slack and Teams

samdickson22:
The Slack converter carried a byte-identical copy of the Teams `boundSpec` walk (the pre-normalization pass that caps hostile specs by depth, children count, node budget, and cycles), and its inline clamp-reason strings duplicated Teams' `clampReasonDetail` word for word. Two copies of a security-relevant traversal means a future hardening fix can land in one protocol and silently miss the other.

This moves the walk, `clampReasonDetail`, and the four bounding constants they share to a single `src/convert/boundSpec.ts`, next to the `takeRun` and `isElement` helpers #6392 put there. Every import site points at that module directly; the protocol-specific caps (Slack block caps, Teams attachment caps) stay in their own constants modules. No output change: constants, warning strings, and traversal order are identical.

The walk also gains a colocated `convert/boundSpec.test.ts`. Its coverage previously existed only through the two converter suites, so the four clamp paths (depth, children, budget, cycle) plus the fabricated-`length` case are now pinned at the seam itself, the way `convert/takeRun.test.ts` pins its helper.

Part of a series of small deduplication refactors across the converters.

## Verification

- `pnpm -C packages/react-generative-ui test`: 601/601 pass, including the new seam suite and both converter suites' clamp coverage.
- `oxlint` and `oxfmt --check` clean; `aui-build` green; `tsc --noEmit` reports the same 19 pre-existing test-file errors as `main`, none in the touched sources.
- Changeset included (patch).


okisdev (approved):
approved. the dedup premise holds: i extracted both walks and diffed them, and Slack's inline `boundNode`/`boundSpec` block is byte-identical to `teams/boundSpec.ts` apart from where `BoundState` sits and the `export` keyword, so there really was one implementation wearing two hats.

i pushed onto the branch rather than round-tripping. the substantive commit drops the re-export shims. `slack/constants.ts` and `teams/constants.ts` were re-exporting the moved names to keep import sites untouched, but neither module is on the public surface (`src/slack.ts` and `src/teams.ts` export only converters and types, and the exports map has no path to either), so the shim bought no compatibility while leaving `toAdaptiveCard.ts` importing `boundSpec` from `../convert/boundSpec` and `clampReasonDetail` from `./constants` in the same file. three of the re-exported names had no consumer left at all once the move landed: `MAX_ELEMENT_DEPTH` in both modules, and the `ClampReason` type in teams. #6392 landed `takeRun` and `isElement` into that same directory while this was open and pointed every import site at `src/convert` directly, so this just follows it. i merged main in as well, since your import block collided with that PR.

the other two commits add `convert/boundSpec.test.ts`, and how it was checked matters more than that it exists. i mutation-tested it: broke the source five ways and confirmed each break turns the suite red. four were caught by the file as first written (dropping `ancestors.delete`, an off-by-one on the children cap, loosening the depth guard, removing the `exhausted` latch). the fifth was not, and it is the one worth knowing about: replacing `Array.prototype.slice.call(value, 0, CHILDREN_CAP)` with the obvious `value.slice(0, CHILDREN_CAP)` passed the entire suite. the fabricated-`length` proxy does not separate the two, because forwarding `slice` through the get trap lands on the same function anyway. that is the whole reason the last commit adds a proxy that hijacks its own `slice`; without that case a future cleanup deletes the hardening the JSDoc promises and CI stays green.

what i executed: `pnpm -C packages/react-generative-ui test` prints 602/602 across 28 files, `oxlint` and `oxfmt --check` are clean, `aui-build` completes, and `tsc --noEmit` reports the same 19 errors as main, all pre-existing casts in test files with none in the touched sources.

one thing left over, out of scope here: `fromSlackBlocks.ts:48`'s `clampArray` bounds with `value.slice(0, cap)`, so the inbound path lacks the hardening the outbound path now has pinned. i'll open a separate issue for it.

PR #6398, opened 2026-08-26
refactor(core): share the indexed-items iterator across list primitives

samdickson22:
Four list primitives (ChainOfThoughtParts, ComposerAttachments, ComposerQueue, MessageAttachments) each spelled out the same inner iterator by hand: read a length off the store, `Array.from` over it building an index-keyed provider, a `RenderChildrenWithAccessor`, and a lazy getter handed to the render function. Only the length selector, the provider component, the accessor path, and the getter's property name differed between them.

They now build that inner component through a shared `createIndexedItems` factory, which owns the memoized iteration and the accessor wiring while each call site keeps its own selector, provider, accessor path, and value shape. The published render-prop signatures are unchanged and the generated API surface does not move. `ThreadMessages` and `ThreadSuggestions` are deliberately left out: both return `null` rather than an empty array at length zero, an observable difference (MessageAttachments' own test pins the empty-array behavior), and encoding it as a factory flag would cost more than it saves.

Part of a series of small deduplication refactors.

## Verification

- `pnpm -C packages/core test`: 1483 tests pass, including a new `createIndexedItems.test.tsx` covering the factory directly.
- 7 files, +195/−124, rebased onto current main (ff9ad444b).
- Changeset included (patch).

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6398?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.Pp9UTUpVIKx8GWdt6SFhru5mjMQ4N6foQAUDp7yps79TOJ0K1sMGTwX229c?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/olive-moons-shave.md
+++ b/.changeset/olive-moons-shave.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/core": patch
+---
+
+refactor: share the indexed-items iterator across the list primitives
--- a/packages/core/src/react/primitives/chainOfThought/ChainOfThoughtParts.tsx
+++ b/packages/core/src/react/primitives/chainOfThought/ChainOfThoughtParts.tsx
@@ -5,9 +5,10 @@ import {
   type ReactNode,
   useMemo,
 } from "react";
-import { RenderChildrenWithAccessor, useAuiState } from "@assistant-ui/store";
+import { useAuiState } from "@assistant-ui/store";
 import type { PartState } from "../../../store/scopes/part";
 import { ChainOfThoughtPartByIndexProvider } from "../../providers/ChainOfThoughtPartByIndexProvider";
+import { createIndexedItems } from "../utils/createIndexedItems";
 import { MessagePartComponent } from "../message/MessageParts";
 import type {
   ReasoningMessagePartComponent,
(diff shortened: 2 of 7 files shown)


okisdev:
closing this. the extraction does not earn its place.

what these four call sites shared is not machinery, it is wiring. the machinery is `RenderChildrenWithAccessor`, and it is already shared: it owns the accessor, the memoization, and the subscription gating. what stayed inline at each site is a length read and an `Array.from` over indices, roughly twenty lines of flat JSX you take in at a glance. `createIndexedItems` replaces that with a factory returning a component, configured by five keys, one of which (`getValue`) receives an accessor rather than a value and needs a doc comment to explain why. reading `MessagePrimitiveAttachmentsInner` now means reading two files instead of one.

the accounting is about twenty five net lines of source, paid for with a new directory, a forty line factory, and a hundred line test for the factory.

two things read as the abstraction working against the grain rather than with it. the generated component loses its identity, so all four primitives collapse to `IndexedItems` in devtools and in react's own component stacks, and restoring the names costs a config key whose only job is to undo a regression the factory introduced. and it absorbs four of six candidates: `ThreadMessages` and `ThreadSuggestions` stay out because they return `null` rather than `[]` at length zero. an abstraction that cannot take a third of its own call sites without a flag is not paying for the indirection it adds.

the duplication here is cheaper to read than the indirection that removes it. i pushed a commit to your branch while reviewing and have reset it back to `7ec3290b2`, so the branch is as you left it.

PR #6399, opened 2026-08-26
fix(registry): drop the unused tooltip-icon-button dependency from thread-list

samdickson22:
`shadcn add thread-list` pulls `tooltip-icon-button` into the user's project even though `thread-list.tsx` never imports or renders it — the component only uses `button`, `input`, and `skeleton`. The stale registry entry costs every consumer an extra component file (plus its tooltip dependency chain) they never asked for.

This drops the `https://r.assistant-ui.com/tooltip-icon-button.json` entry from the thread-list registry item's `registryDependencies`. No other item reaches `tooltip-icon-button` through thread-list: `threadlist-sidebar`, its only dependent, does not use it either, and the items that do use it declare it themselves.

## Verification

- 44 registry tests pass, including the self-containment validation that checks every `@/components/*` import in an item's files resolves through its declared `registryDependencies`.
- Semantically cross-checked against #6366, which reworks the same registry area, to confirm no overlap.
- Rebased onto current main (ff9ad444b).

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6399?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.5CqZj6HSeexQ7HEx2HKNx-FXYVKLlT3J7BesNX8On6puvMvJYp8nzX_Ao9o?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/apps/registry/src/registry.ts
+++ b/apps/registry/src/registry.ts
@@ -1151,12 +1151,7 @@ export const registry: RegistryItem[] = [
           "../../packages/ui/src/components/assistant-ui/thread-list.tsx",
       },
     ],
-    registryDependencies: [
-      "button",
-      "input",
-      "skeleton",
-      "https://r.assistant-ui.com/tooltip-icon-button.json",
-    ],
+    registryDependencies: ["button", "input", "skeleton"],
     dependencies: ["@assistant-ui/react", "lucide-react"],
   },
   {


okisdev (approved):
approving. the removal is correct, though not for the reason the description gives.

the cited evidence does not hold. those 44 tests do not look at this item: `build-registry.test.mjs:1211` filters to items named `elements-*` and matches only relative sibling imports, so alias imports like `@/components/ui/*` are never examined. i confirmed by mutation, deleting `"skeleton"` from `thread-list` while `thread-list.tsx:5` still imports it, and all 44 stayed green.

what does check it is `validateRegistryInstallMetadata` in the registry build. it fails that same mutation with the exact file and specifier, and it is green on this head, which is the verification worth citing. the removal is safe: `thread-list` ships one file whose only alias imports are `button`, `input`, and `skeleton`, it has no CSS import, and `threadlist-sidebar`, its only dependent, pulls `icons/github`, `ui/sidebar`, and `thread-list` and never reaches `tooltip-icon-button`. the saving is real, since `tooltip-icon-button` declares `["tooltip", "button"]` plus `radix-ui`, so `shadcn add thread-list` currently lands two component files and an npm dependency for something it never renders.

i tried to widen this into the full sweep the repo asks for and was wrong, which is worth recording. `mcp-config` declares `input` and `mcp-config.tsx` has no literal `@/components/ui/input` in source, so it reads as the same staleness; dropping it fails the build, because the validator reads the emitted item and the emitted file does import it. `sources` and `directive-text` look stale for a different reason and are not either: both take `badge` relatively at `sources.tsx:8` and `directive-text.tsx:7`. so `thread-list` is the only genuine instance, and this PR is already the complete sweep.

the reason it went stale is that the validator runs on `push` to `main` only and never on a pull request. filed as #6443.

PR #6400, opened 2026-08-26
refactor(react-mcp,react-opencode): generate ids from the shared core generator

samdickson22:
react-mcp repeated the same `crypto.randomUUID()`-or-hand-rolled-fallback ternary in two places (the custom-server record and the elicitation request id), and react-opencode carried its own `createLocalId` helper for a single ephemeral `clientId`. Core already ships `generateId`; three private copies of the same fallback logic is drift waiting to happen.

react-mcp keeps `crypto.randomUUID()` first, so a persisted `MCPCustomServerRecord` keeps the id format it has always had; only the no-randomUUID fallback changes, becoming a prefixed `generateId()`. One `createMcpId` helper now serves both sites. The ids are opaque: `assertValidServerId` only rejects empty ids, `"__"`, and `\x1f`, and nothing parses them. react-opencode's `clientId` keys an in-memory `pendingUserMessages` map and is never persisted or parsed, so it takes `generateId()` directly.

Two nearby generators are deliberately left alone: the OAuth state nonce in `createOAuthProvider` is a CSRF value whose fallback would lose entropy, and cloud-ai-sdk's `createSessionId` would need a new `@assistant-ui/core` dependency to reach `generateId`.

Note on the import: core's `generateId` carries an `@deprecated` JSDoc tag, but in this codebase that tag is used as an experimental-API marker, not a superseded-API marker; #6342 established the same convention of importing it from the public `@assistant-ui/core` entry, which this PR follows.

## Verification

- `pnpm -C packages/react-mcp test`: 132 tests pass (including a new `createMcpId` test covering both the UUID path and the fallback).
- `pnpm -C packages/react-opencode test`: 132 tests pass.
- 6 files, +47/−12, rebased onto current main (ff9ad444b).
- Changeset included (patch).

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6400?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.ntUpZqEzE-kH20OVNlzJHfd2fvs5f33MFKXXax02wXpirXe89GEYGipeyDw?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/shared-id-generators.md
+++ b/.changeset/shared-id-generators.md
@@ -1,0 +1,6 @@
+---
+"@assistant-ui/react-opencode": patch
+"@assistant-ui/react-mcp": patch
+---
+
+refactor: generate ids from the shared core generator instead of per-call-site implementations
--- a/packages/react-mcp/src/resources/McpManagerResource.ts
+++ b/packages/react-mcp/src/resources/McpManagerResource.ts
@@ -8,6 +8,7 @@ import {
 } from "@assistant-ui/store";
 import { useAssistantScopeEffect } from "@assistant-ui/store/client";
 import { ModelContext } from "@assistant-ui/core/store";
+import { createMcpId } from "../utils/createMcpId";
 import type { Tool } from "assistant-stream";
 import { McpServerResource } from "./McpServerResource";
 import { McpLocalStorage } from "./storage/McpLocalStorage";
@@ -275,10 +276,7 @@ const useMcpManagerResource = (
       elicitation,
     }) => {
       const record: MCPCustomServerRecord = {
-        id:
-          typeof crypto !== "undefined" && "randomUUID" in crypto
-            ? crypto.randomUUID()
-            : `mcp-${Date.now()}-${Math.random().toString(36).slice(2)}`,
+        id: createMcpId(),
         name,
         url,
         auth: auth as MCPAuthConfig,
--- a/packages/react-mcp/src/resources/McpServerResource.ts
+++ b/packages/react-mcp/src/resources/McpServerResource.ts
@@ -24,6 +24,7 @@ import type {
   MCPServerState,
   MCPToolInfo,
 } from "../mcp-scope";
+import { createMcpId } from "../utils/createMcpId";
 
 export type McpServerResourceProps = {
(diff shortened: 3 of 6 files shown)


okisdev (approved):
approved. the invariant this rests on is that `mcp-${generateId()}` can never trip `assertValidServerId`, and it holds: 500k samples of core's generator came back exactly 7 chars, alphanumeric every time, so neither `__` nor `\x1f` can appear and the id is never empty. that matters because custom server records do reach the validator, at `McpServerResource.ts:47` on mount and `McpLocalStorage.ts:58` on read-back.

ran both suites on 26c12ae and got your numbers (132/132 each). also mutated the fallback to `mcp__${generateId()}` to confirm the new test is load bearing rather than tautological; it fails on the `toMatch`. on the opencode side i traced `clientId` to its two consumers and both re-namespace it as `local:${clientId}`, and nothing splits that back apart, so dropping the `local_` prefix is safe.

both exclusions check out. the oauth nonce especially: `generateId` is `nanoid/non-secure`, so swapping it in there would be a real entropy loss on a csrf value. and cloud-ai-sdk genuinely has no `@assistant-ui/core` dependency to reach through.

one thing left standing that isn't yours to fix here: `packages/core/src/store/clients/external-thread.ts` still builds composer attachment ids with `Math.random().toString(36).substring(7)`. correcting myself on the earlier version of this comment, which called it a collision risk: it isn't one at any realistic composer size (0% through n=100 in simulation). it is just the same ad hoc idiom this PR is removing, sitting in the package that owns `generateId`, and it emits variable length ids (1 to 10 chars). filed separately since it's core, not these two packages.

PR #6402, opened 2026-08-26
fix(ui): route the image and mermaid zoom views through the kit Dialog

samdickson22:
The image and mermaid zoom views each hand-roll their own modal: a portal, an Escape listener, a body scroll lock, and (for mermaid) a `querySelector`-based focus trap. The image view has no focus trap or focus restoration at all, so a keyboard user can tab behind the overlay. `flow-expand.tsx` already shows the alternative: the kit Dialog owns portalling, dismissal, focus trapping and restoration, and scroll locking, so each view keeps only its own pan/zoom math.

This PR routes both views through the kit Dialog. Behavior kept: full-screen styling, the transform reset on close, and click-anywhere-to-close on the zoomed image; the image dialog keeps an explicit close button (the kit's own, styled `text-white` so it stays visible on the dark overlay), while the mermaid view keeps its toolbar Close instead. The mermaid trigger becomes a kit `DialogTrigger`, which supplies the aria wiring and focus return on close, and both views gain a real accessible name via a screen-reader-only `DialogTitle`. Both registry items declare the new `dialog` dependency so `shadcn add` still lands a self-contained file. Part of a series replacing hand-rolled plumbing with the shared kit primitives.

This adds the first tests for either view: opening a titled dialog, Escape / image-click / surrounding-area dismissal and the labelled close control on the image side; the id-rewritten SVG copy, focus return to the trigger, and the transform reset on reopen on the mermaid side.

## Verification

- `pnpm --filter @assistant-ui/ui test` — 376 passed, 15 skipped, rebased onto current main and re-verified today
- Registry validator passes with the added `registryDependencies` (and fails without them)

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6402?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.T1-57WVkcz6uXhRFkWru4cQGmECL5l8tD8mvoycDaYuV4vPsD0UzGxGr0CA?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/apps/registry/src/registry.ts
+++ b/apps/registry/src/registry.ts
@@ -1402,6 +1402,7 @@ export const registry: RegistryItem[] = [
       "@assistant-ui/react",
       "@assistant-ui/react-markdown",
     ],
+    registryDependencies: ["dialog"],
   },
   {
     name: "diff-viewer",
@@ -1497,7 +1498,7 @@ export const registry: RegistryItem[] = [
       "lucide-react",
       "class-variance-authority",
     ],
-    registryDependencies: [],
+    registryDependencies: ["dialog"],
   },
   {
     name: "file",
--- a/packages/ui/src/components/assistant-ui/image.test.tsx
+++ b/packages/ui/src/components/assistant-ui/image.test.tsx
@@ -8,7 +8,7 @@ import {
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import type { ImageMessagePart } from "@assistant-ui/react";
 
-import { ImageActions } from "./image";
+import { ImageActions, ImageZoom } from "./image";
 
 class FakeClipboardItem {
   constructor(public readonly items: Record<string, Blob>) {}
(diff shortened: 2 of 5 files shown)


okisdev:
closing this. the diagnosis is half right and the remedy is the thing these two files are written the way they are to avoid.

the kit Dialog is not an option for `image.tsx` or `mermaid-diagram.tsx`. both ship as standalone registry items with `registryDependencies: []`, and this PR flips both to `["dialog"]`, so `shadcn add image` and `shadcn add mermaid-diagram` would start landing the dialog component in every consumer's project. that is the cost the hand-written portal, escape handler, scroll lock, and trap exist to avoid. `mermaid-diagram.tsx` spelling `role="dialog"` and `aria-modal="true"` by hand at `:185-187` rather than importing Dialog is that decision, written down in code.

so the mermaid half of this is a rewrite of working, deliberate code. it swaps a self-contained implementation for a dependency, and 87 added against 134 removed lines in that file is not the win it looks like once the dependency is counted.

the image half identifies something real. `ImageZoom` has Escape and a scroll lock but no dialog role, nothing moving focus into the overlay, no trap, and no focus restoration, so Tab does walk into the page behind it. that gap stands on its own and does not need Dialog to close: the sibling in the same package already does all four without it. tracked in #6442 with the specific lines to mirror.

PR #6403, opened 2026-08-26
refactor(assistant-stream): share one SSE encoder and decoder across the transport pipelines

samdickson22:
`assistant-transport` and `ui-message-stream` each rebuild the SSE plumbing that `SSE.ts` already owns: the same `data: ${JSON.stringify(chunk)}` frame on the way out, and the same `TextDecoderStream` → `SSEEventDecoderStream` → per-event switch on the way in, down to identical unknown-event and missing-`[DONE]` error text. Gorp already builds on `SSE.ts`; the other two pipelines carry ~100 lines of divergence-prone copies.

This PR moves both onto the shared classes. `SSEEncoder` gains an optional `doneSentinel`, written verbatim on flush so `[DONE]` never passes through JSON. On the decode side, a new `SSEMessageDecoder` owns the SSE half — byte decoding, event framing, the strict-versus-lenient unknown-event policy, and the terminating sentinel — and hands each message payload to the protocol's own `onMessage`. An optional `onDone` hook runs before the sentinel terminates the stream, which is how `ui-message-stream` keeps flushing its pending tool calls on `[DONE]` exactly as before (#6155). What a payload means stays with each protocol: assistant-transport keeps `parseChunk` and its warn-once-per-reason drops, ui-message-stream keeps its shape validation and field normalizations, and `SSEDecoder` becomes the plain-JSON specialization of the same base. Part of a series replacing duplicated plumbing with the shared primitives.

No wire change: every frame and error message is byte-identical, and the cross-language interop suite passes untouched.

## Verification

- `pnpm --filter assistant-stream test` — 490 passed, 20 skipped in both test configs, rebased onto current main and re-verified today
- `AssistantTransport.interop.test.ts` and its Python encoder fixture pass unmodified, as do the four decoder suites
- `api-surface` regenerated: the new `SSEEncoder` option and `SSEMessageDecoder` types, plus the two transport classes re-declared on the shared bases

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6403?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.RcSfvHcxuA66PTB06p8D1O0bjQu7cedzyG4yRD8DsdvMtZeOUU8co7YkEM8?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/sse-shared-pipeline.md
+++ b/.changeset/sse-shared-pipeline.md
@@ -1,0 +1,5 @@
+---
+"assistant-stream": patch
+---
+
+refactor: build the assistant-transport and UI-message-stream SSE pipelines on one shared encoder and decoder
--- a/api-surface/assistant-stream.ts
+++ b/api-surface/assistant-stream.ts
@@ -234,13 +234,13 @@ type AssistantTransformerStartCallback = (controller: AssistantStreamController)
 
 type AssistantTransformerTransformCallback<I> = (chunk: I, controller: AssistantStreamController) => void | PromiseLike<void>;
 
-declare class AssistantTransportDecoder extends PipeableTransformStream<Uint8Array<ArrayBuffer>, AssistantStreamChunk> {
+declare class AssistantTransportDecoder extends SSEMessageDecoder<AssistantStreamChunk> {
   constructor(options?: {
     strict?: boolean | undefined;
   });
 }
 
-declare class AssistantTransportEncoder extends PipeableTransformStream<AssistantStreamChunk, Uint8Array<ArrayBuffer>> implements AssistantStreamEncoder {
+declare class AssistantTransportEncoder extends SSEEncoder<AssistantStreamChunk> implements AssistantStreamEncoder {
   headers: Headers;
   constructor();
 }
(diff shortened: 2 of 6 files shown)


okisdev:
closing this as superseded, not as the wrong idea. the SSE pipeline is real shared machinery and deduplicating it was the right call.

#6414 landed the same consolidation on 2026-08-27. `SSEJson.ts` now owns `createSSEJsonEncoder(doneMarker?)` and `createSSEJsonDecoder({ parse, done, strict, onUnknownEvent })`, and both pipelines already route through them: `AssistantTransport.ts:5-8,86,102` and `UIMessageStream.ts:5,221`. neither file still carries a hand-rolled `TextDecoderStream` to `SSEEventDecoderStream` chain, so the roughly hundred lines of divergence-prone copies this describes no longer exist on `main`. that is also why this is `CONFLICTING`: it rewrites the three files that rewrite already touched.

the part worth writing down is the difference in approach, because it would have decided this even without the timing. #6414 added zero public exports, keeping the helpers package-private. this adds twenty one lines to `api-surface/assistant-stream.ts`: `SSEMessageDecoder`, `SSEMessageDecoderOptions`, a new `SSEEncoder` constructor signature taking `doneSentinel`, and both transport classes re-declared on the new bases. the published surface is append-only, so that shape would be frozen as a contract with users forever, in exchange for an internal refactor. a plumbing change should not be visible from outside the package.

PR #6447, opened 2026-08-27
fix(react-mcp): scope oauth callbacks to committed renders

Kinfe123:
## Summary

- publish OAuth lifecycle callback options only after a React render commits
- prevent an interrupted prop update from replacing callbacks used by an in-flight OAuth completion
- add a concurrent-render regression test and patch changeset

## Why

`useMcpOAuthCallback()` updated its options ref during render. If one mounted hook instance started `completeAuth()`, then received new callback props in a concurrent render that suspended and never committed, the discarded render's callback remained published. When authentication completed, the hook invoked that uncommitted callback instead of the callback belonging to the UI that remained committed. Account or workspace transitions are one way an application can trigger this sequence.

The ref now publishes from `useInsertionEffect`, which keeps reads tied to the latest committed render while making the callbacks available before other commit-phase effects.

## Verification

- reproduced on `main`: the discarded render's callback ran once and the committed callback ran zero times
- `pnpm --filter @assistant-ui/react-mcp test`
- `pnpm lint`
- `pnpm turbo build --filter=@assistant-ui/react-mcp...`

<a href="https://rupic.dev/s/s2.V3Y3Nd9X4SV9-A4oaqBBuLgPXQTCazuK-JJVrzj1iUdUMEuEYbOWfTQOTBs?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/clean-rice-connect.md
+++ b/.changeset/clean-rice-connect.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react-mcp": patch
+---
+
+fix: keep OAuth lifecycle callbacks scoped to committed renders
--- a/packages/react-mcp/src/hooks/useMcpOAuthCallback.test.ts
+++ b/packages/react-mcp/src/hooks/useMcpOAuthCallback.test.ts
@@ -1,6 +1,12 @@
 // @vitest-environment jsdom
 
-import { renderHook, waitFor } from "@testing-library/react";
+import {
+  createElement,
+  Suspense,
+  startTransition,
+  type PropsWithChildren,
+} from "react";
+import { act, renderHook, waitFor } from "@testing-library/react";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 
 const mocks = vi.hoisted(() => {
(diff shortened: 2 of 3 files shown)


Labeeb2339 (commented):
Reviewed current head 58d50123ddc1b1e52ed574f8a7061c6d7f8b0430. No blocking findings: publishing optsRef in useInsertionEffect keeps callback options tied to committed renders, and the Suspense/startTransition regression test covers the interrupted-render case. Relevant current-head checks are green: Test Changed Packages, Build Changed Packages, Oxlint + Oxfmt, and JavaScript/TypeScript CodeQL.

okisdev (approved):
approving. this is the same change class as #6473, and the part that decides it is that the two `optsRef.current` reads sit inside the async body of the `useEffect`, after `await completeAuth`. moving the write to `useInsertionEffect` makes render-phase reads see the previous committed value, so that placement is what makes this safe rather than a trap. i checked every `optsRef` site in the file for exactly that.

the missing dependency array is also right, not an oversight: nothing writes `optsRef.current` imperatively anywhere in the hook, which is the condition that separates `useA2ARuntime` (no deps) from the eve runtime (needs `[agent, messages]` because staging assigns the ref directly and a dep-less effect would clobber the optimistic value).

what i ran: the suite on this head, 7 passing, then reverted the hunk to the render-phase assignment and reran, which fails the new test with `expected "vi.fn()" to not be called at all, but actually been called 1 times`, matching the reported symptom that the discarded render's callback fires. the test itself is worth calling out as the shape these should take: real Suspense plus a thrown pending promise under `startTransition`, and it asserts its own premise with `expect(interruptedRender).toHaveBeenCalled()` so it cannot silently degrade into a plain rerender that proves nothing.

one thing i did not act on: `Rupic code review` is red, but it posted "Review failed" with zero findings, is absent from the required set, and never registered a review on this head. that is the reviewer erroring, not a code signal; cubic reported no issues across the three files on this same sha.

landing note for #6517, which touches this file too: the hunks are disjoint (it removes the inline `error` parameter handling further down the async body) and both branches share the same pre-image, so this should merge under it cleanly. smaller guard first.

PR #6493, opened 2026-08-28
fix(runtime): publish adapter refs after commit

### LINKED ISSUE #6486: runtime adapters publish render-derived values into refs during render
several runtime adapters write render-derived values into a ref at the top level of the hook body, then read that ref later from callbacks or effects:

```ts
const messagesRef = useRef(messages);
messagesRef.current = messages;
```

react may abandon a render (a suspended tree, an interrupted transition), and an abandoned render still runs that assignment. the ref then holds the abandoned render's value while a different tree stays committed, so any callback reading it acts on the wrong snapshot.

#6473 fixed exactly this in `packages/eve/src/useEveAgentRuntime.ts`, where an interrupted workspace switch left `agentRef.current` pointing at the uncommitted agent and sent a thread refetch to the wrong workspace, and left `messagesRef.current` holding the other workspace's history so a staged draft was appended to it. the fix moves publication into `useInsertionEffect`, which is already the idiom here: `packages/react-a2a/src/useA2ARuntime.ts:51`, `packages/cloud-ai-sdk/src/chat/useCloudChatCore.ts:53`, `packages/cloud-ai-sdk/src/chat/useChatRegistry.ts:56`, `packages/store/src/useAui.ts:355`, and the `useEffectEvent` shim at `packages/tap/src/react-shim/useReactEffectEvent.ts:11`.

the same shape is still in place elsewhere, each one a render-derived value read from a callback:

- `packages/ai-sdk/src/runtime/useAISDKRuntime.ts:144-145`, read at `:175`
- `packages/react-google-adk/src/useAdkRuntime.ts:140-141`, read at `:257` and `:271`
- `packages/react-google-adk/src/useAdkRuntime.ts:189-190`, read at `:342` and `:395`
- `packages/react-google-adk/src/useAdkMessages.ts:64-70`

two things make this per-site judgment rather than a blind codemod.

a render-phase ref write is only a hazard when the value differs between renders. one that is identity-stable for the lifetime of the hook instance is inert: `runtimeRef.current = runtime` at `packages/eve/src/useEveAgentRuntime.ts:482` looks like the same pattern but `useExternalStoreRuntime` returns `useMemo(() => new AssistantRuntimeImpl(runtime), [runtime])` over a `useState` core, so every render writes the identical object. that hunk was deliberately left alone in #6473.

and a site that also writes the ref imperatively needs a dependency array on the insertion effect. `useA2ARuntime` passes none because nothing else touches `headersRef`, but the eve runtime passes `[agent, messages]` because staging and reload assign `messagesRef.current` directly, and a dep-less effect would clobber that optimistic value on any unrelated commit.

<a href="https://rupic.dev/s/s2.UipzeAhworOmtAQVfLaINH-uNU3bJkGyxERYWAKqbwJjYKBkeZ6ig0nfaac?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

Gujiassh:
## Summary

- publish changing runtime adapter refs from committed renders instead of during render
- keep nested remote-thread runtimes on their render-local hook while hosted runtimes use a committed stable callback
- preserve Google ADK staged, edited, and immediate message ref updates
- add interrupted-render regressions for nested runtime changes and repeated ADK reloads

Closes #6486

## Testing

- `pnpm --filter @assistant-ui/core test` (1522 tests)
- `pnpm --filter @assistant-ui/ai-sdk test` (254 tests)
- `pnpm --filter @assistant-ui/react-google-adk test` (324 tests)
- core and AI SDK React Compiler tests
- core, AI SDK, and Google ADK package builds
- full lint/format, scoped checks, changeset validation, and `git diff --check`

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6493?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.jlt39ESb6Y5OXst4HoviexF6zJjzviWV6fLH-FlOw0l37D_u3DKAprlWVFI?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/committed-adapter-refs.md
+++ b/.changeset/committed-adapter-refs.md
@@ -1,0 +1,7 @@
+---
+"@assistant-ui/core": patch
+"@assistant-ui/ai-sdk": patch
+"@assistant-ui/react-google-adk": patch
+---
+
+fix: keep adapter callbacks scoped to committed renders
--- a/.changeset/stable-adk-approval-converter.md
+++ b/.changeset/stable-adk-approval-converter.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react-google-adk": patch
+---
+
+fix: preserve message conversion caches while approval state is unchanged
--- a/api-surface/assistant-ui__core.ts
+++ b/api-surface/assistant-ui__core.ts
@@ -3736,6 +3736,7 @@ declare class RemoteThreadListHookInstanceManager extends BaseSubscribable {
   __internal_subscribeRuntimeReplaced(callback: () => void): Unsubscribe$1;
   stopThreadRuntime(threadId: string): void;
   setRuntimeHook(newRuntimeHook: RemoteThreadListHook): void;
+  __internal_republishRuntimeHook(): void;
   __internal_setDefaultAdapters(adapters: RuntimeAdapters | null): void;
   __internal_setThreadAdapters(threadId: string, adapters: RuntimeAdapters | null): void;
   __internal_dispose(): void;
@@ -3788,6 +3789,7 @@ declare class RemoteThreadListThreadListRuntimeCore extends BaseSubscribable imp
   getLoadThreadsPromise(): Promise<void>;
   loadMore(): Promise<void>;
   constructor(options: RemoteThreadListOptions, contextProvider: ModelContextProvider);
+  __internal_republishRuntimeHook(): void;
   __internal_setOptions(options: RemoteThreadListOptions): void;
   __internal_load(): void;
   reloadMainThread(): Promise<void>;
--- a/packages/ai-sdk/src/runtime/useAISDKRuntime.ts
+++ b/packages/ai-sdk/src/runtime/useAISDKRuntime.ts
@@ -1,6 +1,12 @@
 "use client";
 
-import { useEffect, useMemo, useRef, useState } from "react";
+import {
+  useEffect,
+  useInsertionEffect,
+  useMemo,
+  useRef,
+  useState,
+} from "react";
 import type {
   UIMessage,
   useChat,
@@ -142,9 +148,11 @@ const useGeneratedSuggestions = (
   const controllerRef = useRef<AbortController | null>(null);
   const wasRunningRef = useRef(false);
   const messagesRef = useRef(messages);
-  messagesRef.current = messages;
   const adapterRef = useRef(suggestionAdapter);
-  adapterRef.current = suggestionAdapter;
+  useInsertionEffect(() => {
+    messagesRef.current = messages;
+    adapterRef.current = suggestionAdapter;
+  }, [messages, suggestionAdapter]);
   const hasAdapter = suggestionAdapter != null;
 
   useEffect(() => {
@@ -563,7 +571,9 @@ export const useAISDKRuntime = <UI_MESSAGE extends UIMessage = UIMessage>(
   });
 
   const setMessagesRef = useRef(chatHelpers.setMessages);
-  setMessagesRef.current = chatHelpers.setMessages;
+  useInsertionEffect(() => {
+    setMessagesRef.current = chatHelpers.setMessages;
+  }, [chatHelpers.setMessages]);
 
   useEffect(() => {
     if (hasSeededRepositoryRef.current) return;
--- a/packages/core/src/react/runtimes/RemoteThreadListHookInstanceManager.tsx
+++ b/packages/core/src/react/runtimes/RemoteThreadListHookInstanceManager.tsx
@@ -46,6 +46,7 @@ type AdapterSnapshot = {
 type HostSnapshot = {
   threads: readonly { id: string; generation: number }[];
   hookEpoch: number;
+  renderEpoch: number;
 };
 
 const ProviderRenderDetector: FC<{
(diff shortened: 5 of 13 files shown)


Kinfe123 (commented):
I think the hosted runtime path uses a renderEpoch counter to force another render, which conflicts with the repository’s architecture rules and lacks a real hosted-path test. which is mention by rupic bot as well.

okisdev (changes requested):
requesting changes. the ref sweep itself is right and most of this should survive; the blocker is the republish mechanism it needs on the hosted path.

**1. undisclosed render behavior change.** every consumer builds its options inline, so `options.runtimeHook` is a fresh closure on every render (`useLocalRuntime.ts:113`, `useChatRuntime.ts:42`, `useAdkRuntime.ts:472`, `useStreamRuntime.ts:631`, `useAssistantTransportRuntime.ts:476`). the guard at `useRemoteThreadListRuntime.ts:50` therefore never short circuits, and `__internal_republishRuntimeHook()` fires on every parent render. i measured it with a probe counting hosted `runtimeHook` executions across 5 parent rerenders, using a consumer-shaped inline hook:

- main: 0
- this branch: 5

the hosted thread runtime goes from never re-rendering on a parent render to re-rendering on every one, and during streaming the parent re-renders constantly. that may well be a trade worth making, and main's captured closure going stale is a real bug this would incidentally fix, but it is a core render-path change that the title and description do not mention. it needs to be argued on its own terms rather than arriving inside "publish adapter refs after commit".

**2. `renderEpoch` is the pattern AGENTS.md names.** the "Don't introduce" list has "a `notifyUpdate` plus version-counter re-render hack (bridging non-React state with `useSyncExternalStore`)". `renderEpoch` is incremented and never read (`RemoteThreadListHookInstanceManager.tsx:279` destructures only `threads` and `hookEpoch`); its entire job is to make the zustand snapshot differ so the host re-renders. a replacement should publish the committed hook rather than count versions.

the mechanism does work as written, so this is a design objection rather than a correctness one: `resource()` returns `{hook, args}` with no `deps`, `withKey` adds none without a third argument, and tap's `canReuse` requires `deps !== undefined`, so the child never bails out and the body re-executes.

**3. please confirm the oxlint suppression survives the build.** `messageConverter`'s cache semantics at `useAdkRuntime.ts:198` rest on `react/exhaustive-deps` being suppressed, and `aui-build` runs React Compiler over the package. the comment explains the invariant well, but the shipped `dist` is what users get, so it is worth checking against the built output rather than the source.

notes, not blocking:

i pushed `c3337f09d`, a test that swaps `runtimeHook` on the hosted path and asserts the new one renders. without the republish effect the swapped hook is never invoked at all and the recorded list stays empty. keep it against whatever mechanism replaces this one; it pins the contract rather than the implementation.

rupic is right that neither that test nor the nested one is an interrupted-render regression. both are plain rerenders. they are worth having, but the description should not claim the #6486 failure mode is covered.

what i would not change: the `useInsertionEffect` moves in `useAISDKRuntime`, `useAdkMessages` and the ADK refs are a faithful application of the #6473 precedent, and `toolApprovalsRef.current` to `toolApprovals` at `:199` is a genuine catch, since a `useMemo` reading a commit-written ref would see the previous value. i swept the other `Ref.current` reads across the four changed files; the remaining ones (`useAdkRuntime.ts:364/417/456`) are all inside async adapter callbacks, which is correct.

okisdev:
closing. the ref sweep in `useAISDKRuntime` and `useAdkMessages` was right, but this PR bundles three concerns and the one it is named after is the smallest of them.

the blocker is the core change. `renderEpoch` is incremented and never read (`RemoteThreadListHookInstanceManager.tsx:279` destructures only `threads` and `hookEpoch`); its only job is to dirty the zustand snapshot so the host re-renders, which is the notifyUpdate plus version-counter hack AGENTS.md names in "Don't introduce". it also adds two `__internal_*` methods to the published core surface, which is append-only, on a patch changeset. and it takes hosted `runtimeHook` executions from 0 to 5 across 5 parent rerenders, a render hot-path change during streaming that neither the title nor the description mentions.

`runtimeHookRef` is not in #6486's bug class to begin with. the issue's criterion is a render-derived value read from a callback; that ref is read during render, at `RemoteThreadListHookInstanceManager.tsx:271`. a render-phase write is the correct mechanism for a value consumed during render, and the republish machinery here exists only to compensate for moving it off one. main's captured closure going stale is a real bug, but it needs its own issue and a mechanism that publishes the committed hook rather than counting versions.

the branch is also conflicting, and the two core tests are plain rerenders, so they do not exercise the abandoned-render path they are filed under.

#6486 stays open. #6605 landed the react-langgraph and react-langchain sites the issue body did not name; the ai-sdk and google adk sites are still unfixed.

PR #6504, opened 2026-08-28
fix(langgraph): restrict credential proxy to same origin

Kinfe123:
## What changed

- reject cross-origin browser requests in both bundled LangGraph proxies
- remove wildcard CORS responses and strip upstream CORS and cookie headers
- preserve the request content headers and disable caching of proxied responses
- stop following upstream redirects, so a 3xx cannot carry `LANGCHAIN_API_KEY` to the redirect target, and answer 502 instead (304 still passes through, since it is not a redirect)
- fail closed when the upstream URL is missing
- add focused regression tests and deployment guidance

## Why

The proxy attaches the server-side LangChain API key to upstream requests. Previously it also returned wildcard CORS headers, so another website could use a deployed example as a credentialed LangGraph proxy.

This keeps the local example flow unchanged for same-origin requests. Origin validation limits browser abuse but is not user authentication, so the docs now call out the need for application auth and durable rate limiting in production.

## Reproduction

Before this change, a request with a foreign `Origin` reached LangGraph with `x-api-key` attached and the response carried `Access-Control-Allow-Origin: *`. The new tests reproduce that request and verify it is rejected before the upstream fetch.

## Coverage

`templates/langchain/app/api/[..._path]/route.ts` and `examples/with-langgraph/app/api/[..._path]/route.ts` are identical apart from the template's `export const runtime = "edge"`, so the example suite covers the logic in both. No template in the repo carries a test script, and adding one would ship a test runner into every project scaffolded from it, so the template stays untested by design.

## Verification

- `pnpm --filter with-langgraph test` (8 tests)
- `pnpm exec turbo build --filter=assistant-ui-starter-langchain... --filter=with-langgraph...` (19 tasks)
- `pnpm lint`

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6504?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.MSxcs0RYrWPogQ65lK8ZXrbLY64g1ZjdAezudfFKdD1Lqtcsu7Gao4w3ZMs?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>


okisdev (approved):
approving. the wildcard CORS on a credentialed proxy is a real hole, and i confirmed it rather than taking the description's word for it: running the new tests' request shape against `origin/main`'s route reaches LangGraph with `x-api-key` attached and answers with `Access-Control-Allow-Origin: *`, so a deployed example was a free credentialed LangGraph proxy for any site. `isAllowedRequestContext` here is the function #6505 shipped, and the README keeps the same "not user authentication" framing, so this lands where that one settled rather than reopening the trusted-proxy-origin detour the commit list passes through on the way.

i pushed 3652c8f7d. cubic's two findings were both real and i verified them before acting: an upstream 304 came back as a 502, and the Origin fallback's accept path was unpinned, which a mutation confirmed by forcing a matching same-origin `Origin` to reject and watching all six tests stay green. the 304 guard is restored in both copies, and both branches now have a test that fails on its own mutation and on nothing else.

i also dropped `pnpm --filter assistant-ui-starter-langchain test` from the verification list, because that package has no test script and the command ran nothing. the template stays untested deliberately: its route is identical to the example's apart from `export const runtime = "edge"`, no template in the repo carries a test script, and adding one would ship a test runner into every project scaffolded from it. the description records that now, along with the redirect change, which was shipping unmentioned.

executed: `pnpm --filter with-langgraph test`, `pnpm exec turbo build` over both packages, `pnpm lint`, and the mutation and `origin/main` probes above. the rest i read.

PR #6505, opened 2026-08-28
fix(livekit): isolate and shorten voice tokens

Kinfe123:
## What changed

- reject cross-origin browser requests to the LiveKit token route
- issue ten-minute tokens instead of relying on the longer SDK default
- create a unique room and participant identity for each voice session
- mark token responses as non-cacheable and surface token request failures in the client
- add real JWT regression tests and deployment guidance

## Why

The example previously minted publish and subscribe credentials for one shared room. Anonymous callers could consume voice capacity, and unrelated sessions could receive credentials for the same room.

The example still supports the same voice flow: the configured room name is now a prefix, while the LiveKit agent is dispatched to each isolated room. Origin validation limits browser abuse but production deployments still need application auth and durable rate limiting.

## Reproduction

Before this change, two calls to `/api/livekit-token` produced credentials for the same configured room. The regression test decodes real generated JWTs and verifies a ten-minute lifetime, publish and subscribe grants, and different room names per session.

## Verification

- `pnpm --filter with-livekit test`
- `pnpm exec turbo build --filter=with-livekit...`
- repository lint and changed-file formatting checks

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6505?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.xkN0WLbYXy7aUjBB8D2zUaStc8o1PPVHoOCyXUuoFUG23dMt9MGo6wyFzz8?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/apps/docs/content/docs/guides/voice.mdx
+++ b/apps/docs/content/docs/guides/voice.mdx
@@ -297,7 +297,11 @@ const runtime = useChatRuntime({
 
 Clone the example for the adapter source, token endpoint, and agent worker rather than re-implementing the full LiveKit event map from scratch.
 
-The example rejects browser requests marked same-site or cross-site. For clients without Fetch Metadata, reverse proxies must preserve the public scheme and host in the request URL for the `Origin` fallback. Request-context checks are not user authentication. Before deploying, require your application session in the token route and apply a durable rate limit so anonymous callers cannot consume voice capacity.
+The token route accepts a request only when `Sec-Fetch-Site` is `same-origin` or `none`, and falls back to comparing the `Origin` header against the request URL for clients that omit Fetch Metadata. Behind a reverse proxy, that fallback needs the public scheme and host preserved in the request URL.
+
+<Callout type="warn">
+  A request-context check is not authentication. Before deploying, require your application session in the token route and apply a durable rate limit, so anonymous callers cannot consume voice capacity.
+</Callout>
 
 You can also scaffold it with the CLI:
 
--- a/apps/docs/content/docs/guides/voice.mdx
+++ b/apps/docs/content/docs/guides/voice.mdx
@@ -297,7 +297,7 @@ const runtime = useChatRuntime({
 
 Clone the example for the adapter source, token endpoint, and agent worker rather than re-implementing the full LiveKit event map from scratch.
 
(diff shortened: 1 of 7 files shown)


okisdev (approved):
approving. the real fix here is per-session rooms: before this, `LIVEKIT_ROOM_NAME` was the room, so every visitor landed in `assistant-room` together, and `user-${Date.now()}` collided at millisecond granularity, which LiveKit resolves by evicting the earlier participant. the ten-minute ttl is a genuine second win; `livekit-server-sdk` ships `defaultTTL = '6h'` (`dist/AccessToken.js:3`).

what i wanted certainty on before approving was whether the agent still reaches the new rooms, since turning a fixed room into a prefix would silently break the example if the worker were on explicit dispatch. it isn't. `agent/agent.py` passes `WorkerOptions(entrypoint_fnc=entrypoint)` with no `agent_name`, and pinned upstream `livekit-agents` 1.7.0 (`agent/uv.lock:1015`) documents that field at `worker.py:219` as "Set agent_name to enable explicit dispatch. When explicit dispatch is enabled, jobs will not be dispatched to rooms automatically." automatic dispatch stands, `LIVEKIT_AGENT_NAME` is not in `.env.example`, and the worker never reads `LIVEKIT_ROOM_NAME` at all. that is a reading of the pinned source, not something i ran a live agent against.

i ran the suite twice, once with `LIVEKIT_*` cleared and once with them populated, since the ambient-env dependency is what churned across the earlier rounds; five passed both times. i pushed one commit on top. it adds a test for the `Origin` fallback accept branch, which previously had only a reject case, so a regression that accepted every origin would have stayed green (flip the comparison at `route.ts:16` and the new case turns 403). the same commit splits the guide paragraph so the proxy caveat and the deployment requirement read separately, the latter in a warn callout like the sibling guides use.

one thing worth stating plainly rather than leaving implicit: this does not close anonymous consumption. `isAllowedRequestContext` returns true when there is no `Origin` and no `Sec-Fetch-Site` (`route.ts:13`), and it accepts `sec-fetch-site: none`, both of which a non-browser caller controls, so `curl -X POST` still mints a token. it is browser-side CSRF hardening, which is the right scope for an example, and the docs now say so. flagging it because it is the sentence someone will otherwise read backwards.

PR #6511, opened 2026-08-28
feat(store): per-scope React contexts with selector-only linked child state

Yonom:
## What

New state-update mechanism for `@assistant-ui/store`: every scope's state is published through its own React context, and components read it with `useAuiState("scope")`. Updates ride React's scheduler instead of `useSyncExternalStore`, which makes them concurrent-mode friendly. All internal library primitives in core/react/react-native/react-ink are migrated to the scoped overload; templates and the `@assistant-ui/ui` kit are untouched, and the legacy `useAuiState(selector)` overload keeps working.

Child scope state is no longer embedded in the parent. `useAuiState("thread")` / `aui.thread.getState()` return the thread's own fields plus `messageIds`; `thread.messages`, `thread.composer`, `message.parts`, `message.composer`, `threads.main` and `threads.threadItems` are **linked state**: prototype getters that resolve from the child scopes and are only readable inside a `useAuiState(selector)` callback (they throw elsewhere, pointing at the ids field or the child scope). The `AssistantState` types still include them, so existing selectors type-check and keep working.

## How

- `AuiProvider` renders a fixed chain of per-scope context providers. The scope set is frozen at mount (dev error if it changes); reserve a scope with `undefined` in the config to fill it in later.
- Each scope publishes two contexts: a **state** context (what `useAuiState(scope)` reads; changes only when the state identity changes) and a **render-output** context (what derived scopes resolve their `get()` against, so a lookup added in the same render is visible). A client that re-renders with unchanged state republishes only its output, so state readers bail out.
- Derived scopes resolve against a view of the parent whose source accessor reads the source's render-phase output (`SYMBOL_GET_RENDER_OUTPUT` on the client facade), including sibling scopes of the same config and derived scopes in the hosted tap root. A post-commit check schedules one more render when the committed instance moved.
- `useLinkedState(scope, ownState, resolvers)` (store internal) builds the scope state: own fields live on the object, linked fields are non-enumerable prototype getters. `useAuiState(selector)` runs the selector inside `runInStateSelector`, which is the only place the getters resolve. Linked arrays are cached while shallow-equal so uSES snapshots stay stable; own state is shallow-stabilized so a scope whose fields did not change keeps its identity across store updates.
- `useClientLookup` / `useClientList` expose `keys`; `thread.messageIds` is derived from it. `getThreadMessages(thread)` (core internal) is the non-React way to read the full message list.
- `RenderChildrenWithScope` replaces `RenderChildrenWithAccessor` in every list primitive and reads the item state from the scope context. `MessageBy{Index,Id}Provider`'s `composer` scope now resolves through its declared `message` source.
- Primitives that counted or typed parts/messages now read `message.content` / `thread.messageIds`; `MessageRoot`'s neighbour-role checks resolve the neighbour through `aui.thread.message({ id })` inside a selector instead of materializing `thread.messages`.

## Perf comparison

Harness: `packages/ui/src/__bench__/thread-perf.test.tsx` (vitest + jsdom, React 19.2, `React.Profiler`), rendering the real `@assistant-ui/ui` `<Thread>` over `useExternalStoreRuntime` with a hoisted `convertMessage`. 100 messages mounted, then 300 streaming tokens appended to the last message (one `act` per token), then 20 message appends. `useSmooth` is mocked to a pass-through. Two modes: **inline** re-creates `<Thread />` on every store update (the whole tree re-renders each token, worst case), **stable** passes a stable `children` element (the usual provider-with-children shape, where React bails out of unchanged subtrees). 2 runs per cell, dists of store/core/react rebuilt per variant.

| variant | mode | mount (ms) | stream (ms / token) | append (ms / message) |
| --- | --- | --- | --- | --- |
| `main` (uSES + selectors) | inline | 164–168 | **9.5** | 13.5–15.4 |
| `main` | stable | 154–171 | **3.1–3.4** | 8.0–8.7 |
| this PR before the context split (`8528ae9252`) | inline | 164–167 | 21.5–22.0 | 26.0–26.6 |
| this PR before the context split | stable | 169–171 | 10.7–12.0 | 14.3–16.0 |
| this PR HEAD | inline | 162–163 | **14.8–15.8** | 23.6–24.4 |
| this PR HEAD | stable | 161–174 | **6.3–7.1** | 13.1–14.1 |

With stable children, a streamed token now re-renders only the streaming message's subtree (render counts per token: `MessageRoot` ×1, `MessageGroupedParts` ×1, `MessageParts` ×1, the text part ×1; no untouched message renders). The remaining gap to `main` is fixed per-update cost, from a CPU profile of the stable run (≈6.6 ms / token):

| what | ms / token |
| --- | --- |
| tap host: thread client re-render, incl. `useClientLookup` over 100 message clients (legacy runtime state churn) | ~1.3 |
| 100 `MessageByIndexProvider` derived scopes re-rendering on the `thread` output context | ~0.8 |
| tap commit effects | ~0.7 |
| GC | ~0.7 |
| legacy `useAuiState(selector)` subscribers (`getSnapshot` per notification) | ~0.4 |
| React scheduling / reconciliation of the above | rest |

Inline mode is dominated by the full-tree re-render that both variants pay; the extra ~5 ms over `main` there is the per-message provider chain (`ScopeStateProviders`, derived scope mounts) plus the same fixed costs.

### Next steps for perf

- Skip the 100 derived-provider renders per token: derive `message` from a keyed view that only changes when the message's own client changes, or memoize `RenderChildrenWithScope`'s children by key.
- Stop rebuilding the thread client's message lookup on every legacy-runtime notification (bail out when `messageIds` and each message's runtime state are unchanged).
- Give legacy selectors a cheaper path (e.g. a per-scope change channel) so the remaining `useAuiState(selector)` call sites stop paying `getSnapshot` per notification.

## Status

Ready for review of the mechanism. Stable-children streaming is ~2× `main` (6.3–7.1 vs 3.1–3.4 ms / token, 100 messages); inline is ~1.6×. The fixed costs above are the follow-up.

https://claude.ai/code/session_01Ad1CaRW3TA9SP3Rdr85cfm


Kinfe123 (commented):
Some of the reviews are valid and yeah could we preserve linked fields for imperative getState() calls? This getter now throws outside useAuiState(selector), but existing clients already use aui.thread.getState().messages, including the AISDKChat contract tests. Since TypeScript still exposes messages, consumers will compile successfully and then crash at runtime. Imperative getState() needs a way to resolve child state, or this requires an explicit breaking migration with all downstream usage updated.

PR #6515, opened 2026-08-28
docs(tw-shimmer): use compositor markup in examples

ephraimduncan:
## Problem

The tw-shimmer guide and the demo page showed one-node markup. Users who copied the examples did not get the compositor path from #6514. The guide also listed a reduced-motion claim, defaults, and a `shimmer-width` utility that the CSS does not contain.

## Root cause

The pages were written against an earlier CSS revision. The compositor rework changed the markup contract and some defaults, and the pages did not move with it.

## Change

Every text example in the guide and the demo uses the direct inert clone markup. The demo builds each example from one `ShimmerText` component and one `shimmerCode` source. As a result, the rendered markup and the shown code cannot drift apart.

The customization table now matches `src/index.css`. The corrected values are speed 200, the `--shimmer-track-width` variable in place of the absent `shimmer-width-{n}` utility, spread `calc(4ch + 80px)` and `480px`, angle 15, and repeat-delay 100 and 20. The reduced-motion paragraph now says that only the clone animation stops, and that one-node markup and `shimmer-bg` keep the paint animation. The browser-support section names the load-bearing features (`@property`, relative `oklch()`, `color-mix()`, the `@supports` gate). The floors are Chrome 119+, Firefox 128+, and Safari 16.4+.

## Verification

- We compared each table value and each claim with `packages/tw-shimmer/src/index.css` on this head.
- `tsc` on apps/docs reports no errors in the changed demo page.
- The Firefox 128 and Chrome 119 floors come from the ship versions of `@property` and relative color syntax.

## Public surface

Docs only: `apps/docs/content/docs/utilities/tw-shimmer.mdx` and the tw-shimmer demo page. No published package changes and no changeset.

Stacked on #6514.


okisdev:
closing: this layer only exists to document the `shimmer-clone` markup from #6514, which is closed in favor of the host mask direction in #6825, and the branch carries #6514's commits so it cannot be retargeted to main.

the default and naming corrections here are real and are tracked on their own in #6827 so they land against main without the clone markup. the new browser support sentence does not hold: below the floor the text is invisible, not readable, see #6826.

Issue #6396, opened 2026-08-26
Proposal: expose WebMCP tools (searchDocs, getDoc, getExample) on assistant-ui.com


samdickson22:
Since Aug 25, the ChatGPT desktop app's built-in browser (both ChatGPT and Codex) automatically uses WebMCP tools on any site that registers them, and Shopify flipped WebMCP on for every Liquid storefront on Aug 21. Chrome's origin trial (149 through 156) is live. Our docs site currently offers browsing agents nothing but HTML to scrape, even though we already ship the exact tools they'd want as a stdio MCP server in `packages/mcp-docs-server`.

I'd like to register three tools on `document.modelContext` on assistant-ui.com:

- `searchDocs`, backed by the same search logic as `assistantUISearch`
- `getDoc`, returning a docs page as markdown, like `assistantUIDocs`
- `getExample`, like `assistantUIExamples`

Implementation: a small client component mounted in the docs app root (apps/docs, Next.js 16 + fumadocs) that feature-detects `document.modelContext`, registers the tools, and calls back into the docs content, either through a thin API route or a prebuilt index reusing mcp-docs-server's search and mdx utilities. Plus the Chrome 149 origin-trial token in headers. Nothing changes for normal visitors, and there's no new package or public API. My guess at effort is one to three days.

Why bother: an agent that lands on our docs inside ChatGPT desktop gets structured search instead of scraping, today. And for the project whose pitch is chat UIs and frontend tools, having the docs site itself be agent readable seems like the obvious move.

Honest caveats. The spec is an early W3C CG draft and has already churned once (the attachment point moved from `navigator.modelContext` to `document.modelContext`), and the origin trial could end at Chrome 156 without graduating. So everything stays behind feature detection, touches only the docs app, and can be deleted in one commit if the spec dies. This deliberately does not touch the published packages; a library-level bridge is a separate, bigger conversation.

If this direction is wanted I'll put up a PR.

<a href="https://rupic.dev/s/s2.f8x4xPfnN73YY6rPc7b7vyRuLT-9zj_P_3_yDuWt_IiGyl-3msY7__DSOOI?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg"><img alt="Track in Rupic" src="https://rupic.dev/badges/track-light-v1.svg"></picture></a>

Yonom:
I approve a prototype PR here

[...]

PR #6538, opened 2026-08-29
feat(react-webmcp): bridge runtime tools to the page's WebMCP model context

samdickson22:
## What this adds

A bridge that publishes an assistant-ui runtime's tools to the page's WebMCP model context (`navigator.modelContext`), so a browser agent talking to the page can call the same tools the in-page assistant can. Two new workspace packages, both `private: true`:

- `packages/react-webmcp` — the bridge itself. Public surface is `useWebMcpBridge`, `useWebMcpApprovals`, and their types.
- `examples/with-webmcp` — a task-board demo that exercises registration, the approval dialog, and the status readout.

Nothing published changes. `@assistant-ui/react-webmcp` is `0.0.0` and private, so this is additive on disk and invisible to npm consumers.

## Why

The runtime already knows its tool set. Today that set is reachable only from inside the app. WebMCP is the emerging browser-side surface for handing page capabilities to an agent, and the runtime is the right place to source them from: it already has the tool descriptors, the schemas, and the execution path. Doing this in userland means every app re-implements descriptor conversion and the approval prompt.

## How it works

`useWebMcpBridge` reads `aui.modelContext.getModelContext().tools`, converts each tool to a WebMCP descriptor, and diffs that against what it currently has registered. Adds register, removals dispose, and a changed description or input schema is treated as a re-registration.

Four pieces, kept separate on purpose:

- `webmcp-adapter.ts` owns the `navigator.modelContext` handle and the ownership bookkeeping (which registered names are ours).
- `convertTools.ts` is a pure descriptor converter, both directions.
- `diffRegistrations.ts` is a pure diff over name to descriptor maps.
- `approval-gate.ts` owns the human-in-the-loop prompt, its timeout, and allow-always memory.

The bridge orchestrates. The pure modules convert. That follows the adapter shape in AGENTS.md, with the caveat below.

A few behaviors worth calling out because they are not obvious from the diff:

- **Name collisions are not stolen.** If the page already has a tool under a name we want, we skip it and warn rather than overwrite. A skipped registration is not ours, so its disposer never unregisters that name.
- **Allow-always grants are scoped to a descriptor.** If a tool's description or input schema changes, the prior grant is dropped and the user is asked again. A grant given for one prompt should not silently transfer to a different one.
- **The approval wait is bounded** at 120s, and a settled approval does not re-request user attention.
- **Registration survives StrictMode remounts** and the async window between dispose and the page reporting the name gone.

## Review history

This went through three adversarial review rounds, each run by two independent model families so a finding only counted as load-bearing when both reached it separately. Round 3 produced two confirmed regressions, both traced to the same root cause: ownership state in `webmcp-adapter.ts` was keyed by name alone, so it could not tell "this name is ours" from "this name is ours right now."

Both are fixed in 8a9719de9 by replacing the `ownNames: Set<string>` with `ownTokens: Map<string, object>` plus an is-current guard on release, and by marking a rejected registration rejected before `onError` runs so its disposer never calls name-based `unregisterTool`. Four regression tests cover it; they fail 4 of 4 against the parent commit and pass with the fix.

Two items were deliberately documented rather than fixed, because fixing either would have meant guessing at spec behavior that is not settled: the approval re-prompt on descriptor change (intended, but arguably too aggressive for a description-only edit) and the collision guard's use of a synchronous `hasTool` check against a spec whose `getTools` is async. Both are noted in the example README.

## Testing

`packages/react-webmcp` is fully covered: converter both directions, the diff, the approval gate, the adapter, and each hook, colocated per AGENTS.md.

Suite is green on Node 24, which is what CI pins and what the root `engines` requires (`>=24`). One test, `convertTools.test.ts`'s merge-error case, fails on Node 22 because `AbortSignal.any` there accepts a duck-typed signal instead of throwing. That is a Node version artifact, not a defect in this change, and it is worth knowing if you run the suite on an older local Node.

## Notes for review

- **No changeset.** Both packages are `private: true` and `.changeset/config.json` sets `privatePackages.version: false`, so a changeset file would be inert. AGENTS.md exempts private packages.
- **No prior issue for the bridge package.** #6396 covers the docs-site tools, which is a different PR. AGENTS.md asks for an issue before a non-trivial feature PR, and I did not open one for this. If the direction is not wanted, say so and I will close this rather than have you review 4k lines out of politeness.
- **One deliberate divergence from the adapter shape in AGENTS.md.** The bridge keeps two structures keyed by tool name rather than one: `allowAlwaysMemory: Set<string>`, which is handed to the approval gate, and `grantedSignatures: Map<string, string>`, which never leaves the bridge. Collapsing them would put descriptor-signature knowledge inside the gate, whose entire input is `toolName`. `ToolInvocationTracker` keeps five structures keyed by tool call id for the same kind of reason, so this is consistent with how core already does it, but it is a judgment call and I would rather flag it than have it found.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6538?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.vWIgmKiKFEluORkbI6Mo9sTTBxZACsKvLEDD1pwnv_Q2tVxmjRaZ_mAiofQ?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>


samdickson22:
Closing this at the author's request to re-open it as separate, smaller PRs.

Worth recording before it goes: CI caught a real defect here that local runs did not. Both failures are `exited (127)`, command not found (`vitest: not found` in `packages/react-webmcp`, `next: not found` in `examples/with-webmcp`), even though both are declared devDependencies. The branch's `pnpm-lock.yaml` was generated 96 commits behind main, so the new workspace importers never resolved on a clean CI install. That gets fixed before anything is re-opened.

PR #6559, opened 2026-08-29
fix(core): preserve tool renderer hook boundaries

Kinfe123:
## Summary

- render replacement tool UIs as React components instead of invoking them as plain functions
- preserve the stable wrapper returned by `useInlineRender`
- add a regression test for switching from a renderer with no hooks to one that uses hooks

## Why

`useInlineRender` directly invoked the current renderer inside its stable wrapper. That merged the renderer's hooks into the wrapper's hook list. Replacing a renderer with one that uses a different hook layout then caused React to throw `Rendered more hooks than during the previous render`.

Rendering the selected UI as a child component gives each renderer its own hook boundary. Replacing the renderer now follows normal React component semantics and safely initializes the replacement's state, while the wrapper identity remains stable for registration consumers.

## Testing

- `vitest run packages/core/src/react/model-context/useInlineRender.test.tsx`
- `aui-build` from `packages/core`
- `git diff --check`

The repository lint command could not be reproduced locally because the existing installed oxlint binary predates rules in the current configuration; CI will run the pinned toolchain.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6559?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.brp70uJcWOLI1_ofij47y5z8ht0S5-iijnau5u1G0d6TuYd_6tx_peYpjEM?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/quiet-ravens-juggle.md
+++ b/.changeset/quiet-ravens-juggle.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/core": patch
+---
+
+fix: preserve React hook boundaries when tool renderers change
--- a/packages/core/src/react/model-context/useInlineRender.test.tsx
+++ b/packages/core/src/react/model-context/useInlineRender.test.tsx
@@ -9,25 +9,19 @@ afterEach(() => cleanup());
 
 type Props = ToolCallMessagePartProps<any, any>;
 
-// The point of useInlineRender is a component whose identity survives a
-// changing toolUI, so inner state must not be lost when toolUI swaps.
-const makeToolUI = (label: string): FC<Props> =>
-  function Inner() {
-    const [n, setN] = useState(0);
-    return (
-      <button onClick={() => setN((v) => v + 1)}>
-        {label}:{n}
-      </button>
-    );
-  };
-
 describe("useInlineRender", () => {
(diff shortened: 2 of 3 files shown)


Kinfe123:
Closing after review identified a valid contract regression.  is documented to preserve state for fresh inline renderer closures; rendering the closure as a child component would remount that subtree on ordinary host rerenders. The hook-order crash is real only when the renderer implementation changes hook layout, and React cannot preserve state across that change without altering the helper's shipped behavior. This needs an API/design decision rather than this patch.

Kinfe123:
Correction: `useInlineRender` is the documented helper whose state-preservation contract this patch would regress.

PR #6561, opened 2026-08-29
fix: preserve composer attachment component identity

Kinfe123:
## Summary

- key composer attachment renderers by stable attachment IDs instead of array indices
- keep React and Vue attachment primitives at behavior parity
- add regressions that remove the first of two attachments while the rendered child holds local state

## Why

The attachment primitives used the array index as the component key. Removing an attachment before another caused React or Vue to reuse the removed attachment's component instance for the surviving file. Any local preview, error, or other component state could then be displayed for the wrong attachment.

Selecting the stable attachment IDs and using them as keys preserves component identity while the index-based provider continues resolving the attachment's current position.

## Verification

- `packages/core`: 150 test files, 1,549 tests passed
- `packages/vue`: 14 test files passed, 87 tests passed, 1 existing skip
- `packages/core`: `aui-build` passed
- `packages/vue`: `aui-build` passed

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6561?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.fHDEMZMB1e1xDvDL17XQQjXrB-5SiHVyDzD7PO3ITVaGGcENHmFX7A3bsM8?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/quiet-files-remain.md
+++ b/.changeset/quiet-files-remain.md
@@ -1,0 +1,6 @@
+---
+"@assistant-ui/core": patch
+"@assistant-ui/vue": patch
+---
+
+fix: preserve composer attachment component identity after removals
--- a/packages/core/src/react/primitives/composer/ComposerAttachments.test.tsx
+++ b/packages/core/src/react/primitives/composer/ComposerAttachments.test.tsx
@@ -1,0 +1,76 @@
+/** @vitest-environment jsdom */
+import { render, screen } from "@testing-library/react";
+import { useState, type ReactNode } from "react";
+import { describe, expect, it, vi } from "vitest";
+import type { Attachment } from "../../../types/attachment";
+
+const mocks = vi.hoisted(() => ({
+  attachments: [] as Attachment[],
+  aui: {
+    composer: {
+      attachment: ({ index }: { index: number }) => ({
+        getState: () => mocks.attachments[index],
+      }),
+    },
+  },
+}));
+
+vi.mock("@assistant-ui/store", async (importOriginal) => ({
+  ...(await importOriginal<typeof import("@assistant-ui/store")>()),
+  useAui: () => mocks.aui,
+  useAuiState: (selector: (state: unknown) => unknown) =>
+    selector({ composer: { attachments: mocks.attachments } }),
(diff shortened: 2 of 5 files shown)


okisdev (approved):
approving. same bug class as #6562 and the same fix shape, but the attachment-specific risk is different enough to be worth naming: queue ids are core-generated, attachment ids come out of userland `AttachmentAdapter.add`, so keying on them looks like it puts a fresh uniqueness burden on adapter authors. it doesn't. `base-composer-runtime-core.ts` already resolves both `removeAttachment` and the streaming upsert through `findIndex(a => a.id === ...)`, so an adapter handing back duplicate ids is broken today, independently of anything here.

ran both halves rather than reading them. the core test passes on `0a0140f40` and fails on `80a34c405` with `expected null not to be null`; the vue test passes and fails on base with `[ 'first.txt' ]` where `[ 'second.txt' ]` was expected, which is precisely the survivor inheriting the removed chip's instance. vue is 14 files and 87 tests green, core's react suite 48 files and 310 green, and the typecheck baselines match on both packages (core 91 pre-existing, vue 20 against 20, none of them in anything you touched).

one thing i would have done differently, not blocking. `ThreadPrimitiveMessages` next door keeps a shallow-stable id array and renders off that, which is the vue counterpart of the `useShallowSelector` you reached for on the react side; `composerAttachments.ts` renders straight off `s.composer.attachments` instead, so the two halves are not actually symmetric even though the description calls them parity. i went looking for a re-render delta to justify changing it and could not produce a publish where the id set holds still while the array identity moves, so it stays a consistency note rather than a cost i can demonstrate.

PR #6562, opened 2026-08-29
fix(core): preserve composer queue item identity

Kinfe123:
## Summary

- key composer queue item renderers by stable queue item IDs instead of array indices
- retain the index-based provider for resolving each item's current position
- add a regression that removes the first queued message while the rendered child holds local state

## Why

`ComposerPrimitive.Queue` keyed removable queue items by their array index. Removing an item before another caused React to reuse the removed item's component instance for the surviving message. Local component state could therefore be displayed or applied to the wrong queued message.

Queue items already have stable IDs and all move/remove operations use those IDs. Using the ID as the React key preserves component identity without changing the existing positional lookup contract.

## Verification

- regression fails on current `main` and passes with this change
- `packages/core`: 150 test files, 1,549 tests passed
- `packages/core`: `aui-build` passed

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6562?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.UXvuLCjnaBV07hehz-468Mahb1HpX9yzAm_shoDLrTvOnkTjW-V6BYpartQ?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/calm-queues-remain.md
+++ b/.changeset/calm-queues-remain.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/core": patch
+---
+
+fix: preserve composer queue item component identity after removals
--- a/packages/core/src/react/primitives/composer/ComposerQueue.test.tsx
+++ b/packages/core/src/react/primitives/composer/ComposerQueue.test.tsx
@@ -1,0 +1,72 @@
+/** @vitest-environment jsdom */
+import { render, screen } from "@testing-library/react";
+import { useState, type ReactNode } from "react";
+import { describe, expect, it, vi } from "vitest";
+import type { QueueItemState } from "../../../runtime/queue/queue-item";
+
+const mocks = vi.hoisted(() => ({
+  queue: [] as QueueItemState[],
+  aui: {
+    composer: {
+      queueItem: ({ index }: { index: number }) => ({
+        getState: () => mocks.queue[index],
+      }),
+    },
+  },
+}));
+
+vi.mock("@assistant-ui/store", async (importOriginal) => ({
+  ...(await importOriginal<typeof import("@assistant-ui/store")>()),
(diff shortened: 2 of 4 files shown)


okisdev (approved):
approving. the index key was the whole bug, and it is more reachable than the description lets on: `composer.queue` is `[...steerItems, ...laneItems]`, so `steer()` reorders that flattened array just as readily as `remove()` shortens it, and either one hands a surviving item the displaced item's component instance.

verified by running it rather than reading it. your regression test passes on `e2ce32c4` and fails on `80a34c405` with the survivor still rendering `first message`; `packages/core`'s react suite is 48 files and 310 tests green; `tsc --noEmit` on core returns the same 91 pre-existing errors as base with none in the two files you touched. keying by id while still resolving by index reads odd in isolation, but it is the same split `MessageGroupedParts` and `MessageParts` already use with `useShallowSelector`, and ids come from `generateId()` with each item in exactly one lane, so there is no duplicate key to worry about.

the react-ink mock change is load bearing rather than cosmetic, since `{ length: 3 }` has no `.map` and that test would throw without it. six more list primitives still key by index the same way, `ThreadListPrimitive.Items` the worst of them; that is #6570 rather than something this diff should grow to cover.

PR #6588, opened 2026-08-29
fix(ci): run changeset semver checker from the PR base

### LINKED ISSUE #6582: ci: Changeset Semver Check runs from the PR head, so it fails on any branch older than the script
`.github/workflows/changeset-semver-check.yaml` (added in #6574) checks out the PR head rather than the merge ref:

```yaml
- name: Checkout PR head
  uses: actions/checkout@...
  with:
    ref: ${{ github.event.pull_request.head.sha }}
```

and then runs `node --test scripts/check-changesets.test.mjs` and `node scripts/check-changesets.mjs` out of that tree. two consequences:

**1. it hard-fails on every open PR whose branch predates #6574.** the scripts do not exist at those heads, so the step dies with `Could not find 'scripts/check-changesets.test.mjs'` and the job goes red for a reason that has nothing to do with the PR. hit on #6581, whose branch forked at `c161e2247`; the only way out was `gh pr update-branch`. any other PR opened before `dc90e7739` is in the same state and will look broken to its author.

**2. the gate runs the PR's own copy of itself.** because the checker and its test come from the head, a PR that edits `scripts/check-changesets.mjs` is validated by the edited version rather than by main's. that is the wrong direction for a check whose job is to catch a bad changeset.

both go away by taking the checker from the base: either drop the `ref:` so `actions/checkout` uses the default merge ref, or keep the head checkout for the changeset files and run the scripts from a separate base checkout. the `git diff BASE_SHA HEAD_SHA` step already needs `fetch-depth: 0`, so the base commits are present either way.

<a href="https://rupic.dev/s/s2.ZmwcGt9SeJEdi9eGLcrlyVsS9eIt34OTBE2eTCr9DS1lDhrIE4yWrIFD8nI?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

BetterAndBetterII:
## Summary
Changeset Semver Check was checking out the PR head and running `scripts/check-changesets.*` from that tree. Branches forked before the checker was added fail with a missing-file error, and a PR can rewrite the validator that is supposed to gate it.

This keeps the full-depth PR-head checkout for changeset files and `git diff`, checks out the PR base into `.changeset-checker-base`, and runs the base checker against the PR-head workspace via `CHANGESET_CHECK_ROOT`.

Fixes #6582

## Test plan
- [x] `node --test scripts/check-changesets.test.mjs` (Node >= 24) — 15 pass, including two new workflow-contract tests

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6588?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.e1hsyD3kHHkAKS57hs84Un5zqUocqGPAyOZwXo-IB7IungTOAEKgW-Tz9o8?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.github/workflows/changeset-semver-check.yaml
+++ b/.github/workflows/changeset-semver-check.yaml
@@ -25,6 +25,12 @@ jobs:
           ref: ${{ github.event.pull_request.head.sha }}
           fetch-depth: 0
 
+      - name: Checkout PR base checker
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # ratchet:actions/checkout@v7.0.1
+        with:
+          ref: ${{ github.event.pull_request.base.sha }}
+          path: .changeset-checker-base
+
       - name: Setup pnpm and node.js
         uses: pnpm/setup@703c52620218391530e48b9e8870d5c0082e1b9b # ratchet:pnpm/setup@v2.1.0
         with:
@@ -32,10 +38,12 @@ jobs:
           install: false
 
       - name: Test the changeset checker
-        run: node --test scripts/check-changesets.test.mjs
+        run: node --test .changeset-checker-base/scripts/check-changesets.test.mjs
 
       - name: Verify changesets only name releasable packages
-        run: node scripts/check-changesets.mjs
+        env:
+          CHANGESET_CHECK_ROOT: ${{ github.workspace }}
+        run: node .changeset-checker-base/scripts/check-changesets.mjs
 
       - name: Check changeset bump types against package versions
         id: check
--- a/scripts/check-changesets.test.mjs
+++ b/scripts/check-changesets.test.mjs
@@ -1,6 +1,12 @@
 import assert from "node:assert/strict";
 import { spawnSync } from "node:child_process";
-import { mkdirSync, mkdtempSync, rmSync, writeFileSync } from "node:fs";
+import {
+  mkdirSync,
+  mkdtempSync,
+  readFileSync,
+  rmSync,
+  writeFileSync,
+} from "node:fs";
 import { tmpdir } from "node:os";
 import path from "node:path";
 import test from "node:test";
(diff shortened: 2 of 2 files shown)


okisdev:
closing in favor of #6593, which fixes #6582 by deleting the `ref:` line so `actions/checkout` takes the default `refs/pull/N/merge`. the merged tree already carries the base's `scripts/` and the PR's `.changeset/` at once, and every other `pull_request` job in this repo already checks out that way, so `changeset-semver-check.yaml` was the only one diverging.

two reasons the two-checkout approach was not taken.

the second checkout does not make the checker trusted. for `pull_request`, GitHub runs the workflow definition from the merge commit, so a PR that wants to defeat this gate edits `changeset-semver-check.yaml` itself and drops the base-checkout step. that half of #6582's rationale does not hold, and the issue was wrong to state it.

it also removes the only CI coverage the checker has. `repoRoot` in `scripts/check-changesets.test.mjs:15` is `path.resolve(import.meta.dirname, "..")`, so `node --test .changeset-checker-base/scripts/check-changesets.test.mjs` runs the base test against the base script. `changesets:test` runs in no other workflow, and `code-quality.yaml`'s `paths:` filter lists `scripts/check-built-declarations*` but not `scripts/check-changesets*`, so a PR breaking `scripts/check-changesets.mjs` would have merged green. the two contract tests added here never executed in this PR's own CI for the same reason.

the diagnosis of #6582 was right; the fix landed as a deletion instead.

PR #6614, opened 2026-08-31
refactor(react): dispose WebMCP registrations through the abort signal alone

### LINKED ISSUE #6585: webmcp: unregisterTool is not in the explainer; disposal should rely on the abort signal alone
the WebMCP provider carries a name-keyed `unregisterTool` disposal path that the explainer does not define. at `webmachinelearning/webmcp@41d12f05` (2026-08-26) the README mentions `unregisterTool` zero times; the documented interface is `registerTool` / `getTools` / `executeTool` / `toolchange`, and unregistration is spelled out as aborting the signal you passed in:

```js
await document.modelContext.registerTool({...}, { signal: controller.signal });
// To unregister the tool later, abort the signal.
// controller.abort();
```

`webmcp-host.ts` already aborts that controller unconditionally on dispose, so disposal is correct on a spec host today. the name-keyed call is extra, and the machinery guarding it is not free:

- `ownersByContext` / `ownersOf`, the per-registration `owner` token, `displaced`, `disownName`, and `releaseName` exist only so a name-keyed unregister cannot delete a registration that has since taken the name over
- three regression tests exist for that machinery (`does not unregister a name a later registration has taken over`, `still releases the name when the registration that displaced it is refused`, `still releases the name when the registration that displaced it throws`)

every one of those failure modes is unreachable on a host that follows the explainer, because nothing removes a tool by name.

removing the path would delete the ownership map and its three tests and leave disposal as the single abort, which is what the spec describes. the reason it is not done in #6540 is that it changes the disposal contract of the module at the end of a long review, and the host tests are written against `unregisterTool` spies. it wants its own change with its own review.

worth confirming first whether any shipping implementation exposes `unregisterTool` in practice; if one does, the tolerance stays and this issue becomes a comment instead.

surfaced during review of #6540.

<a href="https://rupic.dev/s/s2.ehVw_7CI59H-FExjPvYOLn9fG2cfLSvfBlgVjeEcAXgcVd94ypkV-xihrLY?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

okisdev:
closes #6585

## Problem

`webmcp-host.ts` carried two tool-removal paths that the WebMCP explainer does not define: a name-keyed `unregisterTool(name)` on the model context, and an `unregister()` handle on the `registerTool` return value.

the name-keyed one was not free. because it addresses a tool by name rather than by identity, a disposal could delete a registration that had since taken the name over. guarding that needed a per-host `WeakMap` ownership registry (`ownersByContext`, `ownersOf`, a per-registration `owner` token, `displaced`, `disownName`, `releaseName`) plus a `settled` / `disposed` deferred-release state machine, and roughly two thirds of the host test file existed to cover machinery that only ever ran on the non-spec path.

## Root cause

the tolerance was written speculatively. checking the pinned upstream settles it.

at `webmachinelearning/webmcp@41d12f05` (2026-08-26, current HEAD), `index.bs` declares the whole interface:

```webidl
interface ModelContext : EventTarget {
  Promise<undefined> registerTool(ModelContextTool tool, optional ModelContextRegisterToolOptions options = {});
  Promise<sequence<RegisteredTool>> getTools(optional ModelContextGetToolOptions options = {});
  Promise<DOMString> executeTool(RegisteredTool tool, optional object inputObject = {}, optional ModelContextExecuteToolOptions options = {});
  attribute EventHandler ontoolchange;
};
```

there is no `unregisterTool`, and `registerTool` resolves to `undefined` rather than to an unregister handle. the spec states the mechanism directly: `ModelContextRegisterToolOptions.signal` is "An `AbortSignal` that unregisters the tool when aborted."

nothing implements the tolerated paths either. `@mcp-b/webmcp-types@5.0.3`, which models the pre-standard MCP-B runtime, declares `ModelContext` as `registerTool` / `getTools` / `ontoolchange`, and puts its non-standard surface in two clearly labelled extension interfaces: `ChromeModelContextExtensions` (`executeTool?`) and `ModelContextExtensions` (`registerTool` overloads plus `listTools()`). neither carries a name-keyed removal. its `RegistrationHandle { unregister(): void }` is documented as the return of MCP-B's own registration helpers, not of `modelContext.registerTool`.

## Change

disposal is now the abort and nothing else, so `createHost` collapses to its actual job:

```ts
const createHost = (context: WebMcpModelContext): WebMcpHost => ({
  available: true,
  registerTool: (def, onError) => {
    const handle = context.registerTool(def, { signal: controller.signal });
    if (isThenable(handle)) {
      handle.then(
        () => {},
        (error) => onError?.(error),
      );
    }
    return () => controller.abort();
  },
});
```

the ownership map and the deferred-release state machine go with it. this is not only dead-code removal: abort is identity-keyed, so the displacement hazard the map guarded against is gone by construction rather than defended against. the spec has the same re-registration pattern the provider uses on a signature change (abort the old signal, register the replacement), under its unregistration/execution race note.

`registerTool` keeps its `Promise<void> | void` tolerance and the `document` / `navigator` lookup order; only the two undefined removal paths are dropped. the fulfilled handler stays a no-op rather than being dropped to `then(undefined, …)`, because a host returning a hand-rolled thenable calls it with no argument and a real `Promise` would not.

`createAsyncModelContext` in the test fake now removes on abort, asynchronously and identity-keyed, which is what a conforming async host does.

## Verification

- `pnpm exec vitest run src/unstable/webmcp` in `packages/react`: 3 files, 62 passed.
- the host suite is rewritten around the abort: it asserts the signal is handed to `registerTool` unaborted, aborted on dispose, aborted even when the registration rejected, and that dispose is idempotent, plus the bare-thenable, returns-nothing, and synchronous-throw arms.
- `keeps the tool exposed across a description change on an async host removal` covers the displacement case end to end against the abort-based fake, and passes without any ownership bookkeeping.
- `pnpm exec tsc --noEmit` in `packages/react`: three pre-existing errors under `unstable/webmcp` remain, byte-identical to the same run on `main`; none added.
- `pnpm exec oxlint` and `pnpm exec oxfmt --check` on the touched paths: clean. `pnpm changesets:check`: passes.
- repo-wide grep for `unregisterTool`: no references remain outside the changeset.

## Public surface

`WebMcpModelContext` and `WebMcpHost` are internal; `packages/react/src/index.ts` exports only `unstable_useWebMcpProvider`, `Unstable_WebMcpProviderOptions`, `Unstable_WebMcpProviderResult`, and `unstable_defaultWebMcpFilter`, none of which change. no export is added or removed.

docs: `apps/docs/content/docs/tools/webmcp.mdx` had a lifecycle bullet whose "the provider never unregisters a tool it does not own" clause documented the ownership map; it now states the structural reason (the provider only ever aborts its own registration signal).

changeset: `@assistant-ui/react` patch.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6614?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.UEMyqcdkXVIyM57Km5EzQASq87XnHCVoqpPdDsRFd-aB1MsVrlRQMaJvRaE?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>


okisdev:
addressed the fixture-fidelity note. the spec's abort steps run `unregister a tool`, whose map removal is synchronous (only the toolchange notification is `in parallel`), `registerTool` rejects with `InvalidStateError` when the name already exists, and the returned promise resolves on a queued task rather than at insertion. `createAsyncModelContext` now models all three instead of deferring removal past a microtask and letting a duplicate overwrite the entry.

`keeps the tool exposed across a description change on a conforming host` is load-bearing as a result: dropping the removal from the abort listener makes it fail, because the replacement registration is then refused for a name the host still holds.

changeset trimmed to its one-line summary. left it as its own file rather than folding into `tame-donkeys-repeat.md`, which is already merged and unreleased rather than sitting in an open PR; editing another change's changelog entry to describe this one reads worse than the extra line.

Issue #6576, opened 2026-08-29
docs: finish the css cleanup behind the fumadocs-ui exit


okisdev:
#6575 removed fumadocs-ui and kept the diff behavior-preserving, leaving css debt in `apps/docs/styles/`. the original framing of this issue was wrong on two counts, corrected below.

## what is actually true

#6575 did not remove fumadocs prose. it swapped `fumadocs-ui/css/preset.css` for `@plugin "@fumadocs/tailwind/typography"` plus a `--color-fd-*` alias block (`styles/globals.css:60` to `:79`). the plugin still emits `.prose`, 365 lines of it.

the `!important`s in `styles/fumadocs.css` are redundant for a different reason than the one this issue originally gave: cascade layer position, not the plugin's absence.

| rule source | position | outcome |
| --- | --- | --- |
| plugin `.prose` | `@layer utilities` | loses |
| `styles/fumadocs.css` | unlayered | wins |

unlayered normal declarations beat every layered declaration regardless of specificity, so the 138 `!important`s change nothing.

verified in a browser: stripping all 138 and recomputing styles over 40 elements (h1 through h4, p, ul, ol, li, blockquote, pre, code, hr, img, table, thead, th, td, tr, not-prose, tabpanel, prose-no-margin, prose-blog, preview-code-preview, `#nd-page`, `#nd-toc`) across 33 properties produces a zero diff. the only rules whose behavior changes are ones competing with an inline `style=`, and shiki's dual-theme output writes only `background-color`, `color` and `--shiki-dark`, none of which fumadocs.css sets on `pre` or `pre code`. mdx content carries no inline css.

## the fd alias block cannot be dropped

the generated `.prose` reads 27 `var(--color-fd-*)` references over 6 tokens: foreground (10), border (9), primary (3), muted (2), muted-foreground (2), card (1). deleting the `@theme inline` block in `globals.css` leaves `--tw-prose-body`, `--tw-prose-headings`, `--tw-prose-hr` and `--tw-prose-th-borders` unresolved.

sweeping the 39 `fd-*` utility sites across 6 tsx files to house tokens is still worth doing. the alias block stays until `@fumadocs/tailwind/typography` itself goes, which is a much larger job: the plugin is the only source for `dl` / `dt` / `dd`, `figure` / `figcaption`, `kbd`, `strong`, nested lists, `tfoot`, `code` and `strong` inside headings, and the `prose-<element>:` variants the grok and perplexity example pages use.

[...]

PR #6616, opened 2026-08-31
refactor(docs): finish the css cleanup behind the fumadocs-ui exit

okisdev:
## Problem

#6575 removed fumadocs-ui and kept the diff behavior-preserving, leaving css debt in `apps/docs/styles/`: 138 `!important`s, a `fd-*` token vocabulary in app code, and a stylesheet named after a package the app no longer depends on. #6576 tracks it.

## Root cause

The `!important`s were needed while `fumadocs-ui/css/preset.css` was imported. They are now redundant for a reason worth writing down, because the obvious reading is wrong: fumadocs prose is still here. #6575 swapped the preset for `@plugin "@fumadocs/tailwind/typography"`, which still emits `.prose`, 365 lines of it, into `@layer utilities`. `apps/docs/styles/fumadocs.css` is imported unlayered, and unlayered normal declarations beat every layered declaration regardless of specificity. Cascade layer position is what makes the `!important`s inert, not the plugin's absence.

Two blocks in that stylesheet had also gone dead: `.preview-code-block` (53 lines, 31 `!important`s) and `.steps-container` / `.step-number` have zero references anywhere in the repo.

The `#nd-*` ids are in the same category. `nd-` is next-docs naming, but fumadocs-ui no longer emits those elements: `docs-page.tsx` and `table-of-contents.tsx` hand-write `#nd-docs-layout`, `#nd-page` and `#nd-toc`, and nothing else in the repo or in fumadocs-core references them.

## Change

Delete the two dead blocks, then strip the remaining `!important`s.

Sweep the 39 `fd-*` utility sites across six components to the house tokens. Every `--color-fd-x` is defined as `var(--x)`, the same expression as its `--color-x` counterpart, so each swap is exact. The `--color-fd-*` alias block itself stays, with a comment saying why: the generated `.prose` reads 27 `var(--color-fd-*)` references over six tokens, so deleting the block would leave `--tw-prose-body`, `--tw-prose-headings`, `--tw-prose-hr` and `--tw-prose-th-borders` unresolved. It can only go when the plugin does, which is a much larger job (the plugin is the only source for `dl`/`dt`/`dd`, `figure`/`figcaption`, `kbd`, `strong`, nested lists, `tfoot`, and the `prose-<element>:` variants).

Move the `#nd-*` rules onto the components:

- `#nd-page`'s rule is gone entirely, its box properties now Tailwind on `<main>`. That also drops `--docs-page-inline`, a custom property whose only reader was the `padding-inline` in the same rule.
- `#nd-docs-layout` becomes `.docs-layout`, with both breakpoints' `grid-template-columns` in one block so a single owner holds the property. The `:has()` stays because the toc renders null when a page has no headings, so the layout cannot know from props alone.
- `#nd-toc` becomes `.docs-toc`, still needed as the `:has()` hook, and `#nd-toc > .sticky` moves onto the component as `top-[calc(var(--docs-header-height)_+_1rem)]`. This preserves the current 64px and removes the `top-14` that the stylesheet had been silently overriding.

`styles/fumadocs.css` is renamed `styles/docs.css`.

Left for follow-ups on #6576, to keep this diff free of rendering changes: the `prose-<element>:` variants on the grok and perplexity example pages are inert, and the toc panel's `max-h-[calc(100vh-3.5rem)]` still assumes the 56px offset it no longer has.

## Verification

Built the docs css with the tailwind cli before and after, then diffed computed styles in a browser.

- After stripping the `!important`s: 61 elements (h1 through h4, p, ul, ol, li, blockquote, pre, code, span with real shiki dual-theme inline styles, hr, img, figure, dl/dt/dd, kbd, table/thead/th/td/tr, not-prose, tabpanel, prose-no-margin, prose-blog, preview-code-preview, sidebar-tree-content, and the layout elements) across 44 properties in light and dark. The only elements that moved were the ones synthesized to exercise the two deleted classes.
- After the token sweep and the `#nd-*` move, rendering both markup variants against their own stylesheet: 24 nodes across 36 properties at 1400px, 900px and 600px in light and dark, six runs, zero diff.
- All 16 `--color-fd-x` definitions compared against their `--color-x` counterparts: identical.
- Text diff of the two builds: differs only by the removed `!important`s and the two dead blocks.
- `oxfmt --check` clean. `oxlint` reports nothing new. `tsc --noEmit` reports nothing in the nine changed files.

## Public surface

None. `@assistant-ui/docs` is private, so no changeset.

<a href="https://rupic.dev/s/s2.gPaoWMhrygFKibIvMLUDRrL0UIvOBrGvUK4OWW3p1Uthj5A5sL4ETPOR-d4?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>


okisdev:
fixed in 2a8f127ce. both claims in that line were stale: the path, and the `!important` that no longer exists. the trap itself survives, so the sentence now names the mechanism that actually carries it (unlayered beats layered utilities) rather than the one that was removed.

confirmed it was the only remaining reference to the old path in the repo.

okisdev:
@rupic-app the deferral is the design, not an oversight, so this merges as is. taking the three points in turn.

**the two rendering defects.** they are deferred on purpose and the PR body says so. this PR's review property is that its computed-style diff is zero; folding a behavior change into it destroys the only thing that makes 395 deleted lines cheap to review. they are also not one fix: the toc offset is two lines, while making the `prose-<element>:` variants win is a cascade decision (give the docs `.prose` its own layer after `utilities`, or stop the example pages inheriting the global `.prose`) that deserves its own review. `@layer components` is the obvious answer and it is wrong, since it sits below the `utilities` layer the typography plugin writes into. both are tracked in #6576 as follow-ups.

**the cubic marker.** correct, and removed. it was appended to the body by the cubic app after the PR was opened, not written into the description.

**"not verified: removing the !important declarations preserves rendered behavior".** that is an accurate description of what this review did, since it ran no browser. it was verified: both stylesheets built with the tailwind cli, then computed styles diffed in chrome over 61 elements and 44 properties in light and dark, including `pre` and `span` carrying real shiki dual-theme inline styles, which are the only realistic competitor an unlayered normal declaration still loses to. zero diff outside the two synthesized nodes that exercise the deleted classes. the method and counts are in the verification section.

PR #6653, opened 2026-08-31
chore(ui): stop teaching the deprecated Suggestion props

samdickson22:
## Problem

The shadcn kit's `ThreadFollowupSuggestions` component still passes the deprecated `method="replace"`/`autoSend` props to `ThreadPrimitive.Suggestion`, and the suggestions guide teaches the same deprecated form. Since the kit is the canonical source copied into user projects via the registry, every new install propagates the deprecated API.

## Change

Replace `method="replace" autoSend` with the current `send` prop. Behavior is identical: `send` resolves the same as `autoSend` did (`send ?? autoSend ?? false` in `useThreadSuggestion`), `method` was already ignored by the hook, and `clearComposer` defaults to `true`. The colocated test's mock and assertions now reflect the new props, and the two docs pages that teach/describe this pattern (`guides/suggestions.mdx`, `elements/follow-up-suggestions.mdx`) are updated to match — including fixing the prose that claimed the prompt is placed into the composer before sending (it is sent directly and the composer is cleared).

The deprecation documentation in `primitives/suggestion.mdx` and `primitives/index.mdx` is intentionally left in place.

## Verification

- `pnpm turbo run test --filter=@assistant-ui/ui`: 15 files, 408 passed, 15 skipped
- `tsc --noEmit` in `packages/ui`: 7 pre-existing errors on clean main (unrelated files: `image.test.tsx`, `thread-list.aui.test.tsx`, vue tests), same 7 with this change — nothing new
- `pnpm lint` clean

## Public surface

None. `@assistant-ui/ui` and the docs app are private, so no changeset. No template copies of this component exist (`sync-templates` unaffected).

<a href="https://rupic.dev/s/s2.TxtmZSeKRdWvav_zEZtHaXgxk_099n_25NkLZI1SVN6kKd73IRq-gRDN5Ao?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/apps/docs/content/docs/guides/suggestions.mdx
+++ b/apps/docs/content/docs/guides/suggestions.mdx
@@ -277,8 +277,7 @@ function FollowUps() {
           <ThreadPrimitive.Suggestion
             key={i}
             prompt={s.prompt}
-            method="replace"
-            autoSend
+            send
           >
             {s.title ?? s.prompt}
             {s.label && (
--- a/apps/docs/content/docs/guides/suggestions.mdx
+++ b/apps/docs/content/docs/guides/suggestions.mdx
@@ -274,11 +274,7 @@ function FollowUps() {
     <AuiIf condition={(s) => !s.thread.isEmpty && !s.thread.isRunning}>
       <div className="flex gap-2">
         {suggestions.map((s, i) => (
-          <ThreadPrimitive.Suggestion
-            key={i}
-            prompt={s.prompt}
-            send
-          >
+          <ThreadPrimitive.Suggestion key={i} prompt={s.prompt} send>
             {s.title ?? s.prompt}
             {s.label && (
               <span className="text-muted-foreground ms-1">{s.label}</span>
--- a/apps/docs/content/elements/follow-up-suggestions.mdx
+++ b/apps/docs/content/elements/follow-up-suggestions.mdx
@@ -100,7 +100,7 @@ The first chip reads "Weather" with "in SF" trailing it, but sends the full weat
 
 ### How a suggestion sends
 
-Each chip uses `ThreadPrimitive.Suggestion` with `method="replace"` and `autoSend`, so clicking one replaces whatever's in the composer with the suggestion's prompt and sends it immediately, with no confirmation step.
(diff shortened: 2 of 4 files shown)


okisdev (approved):
approved. the equivalence claim is the whole PR and it holds: `method` is destructured to `_method` in `ThreadSuggestion.ts:15` and never referenced again in the file, `send ?? autoSend ?? false` at `:42` makes bare `send` identical to bare `autoSend`, and `clearComposer` defaults to `true` in `useSuggestionTrigger.ts:14`. that is a read of both files rather than a runtime check, but the read is conclusive because `_method` has no second occurrence. `send` is also already in the `createActionButton` forward list at `ThreadSuggestion.ts:61`, which is the part that would have failed quietly: a prop missing from that list falls through to `Primitive.button` and lands on the DOM as an unknown attribute.

the prose fix in `follow-up-suggestions.mdx` is the better half of this change, since the old sentence described a code path that does not exist on the send branch. worth knowing why the new sentence is safe to state so flatly: `useSuggestionTrigger` skips `composer.setText("")` when a send is queued during a run, so "clears the composer" is unconditionally true only because `ThreadFollowupSuggestions` gates on `!s.thread.isRunning`. copied to a surface that renders during a run, that sentence becomes wrong.

i pushed one commit on top. the guide sample's opening tag was still split across five lines, a shape that only made sense while it carried four props. oxfmt collapses `<ThreadPrimitive.Suggestion key={i} prompt={s.prompt} send>` onto one line at that indent, which i confirmed by running oxfmt on the equivalent tsx. `pnpm lint` reports clean formatting either way, because oxfmt does not reach inside mdx fences, so nothing in CI would have caught it.

`pnpm turbo run test --filter=@assistant-ui/ui` on the current head: 15 files, 408 passed, 15 skipped.

one thing left standing, and not for this PR: `primitives/suggestion.mdx:142` labels `ThreadPrimitive.Suggestion` legacy and points at `ThreadPrimitive.Suggestions` plus `SuggestionPrimitive` parts instead, while `thread.aui.tsx:257` already uses the newer form. the kit teaches both. moving `follow-up-suggestions` across is a behavior-shaped change on a different data path and does not belong in a prop swap.

PR #6655, opened 2026-08-31
fix(docs): surface WebMCP docs-tool errors instead of Chrome's generic UnknownError

samdickson22:
## Problem

Two issues in the docs-site WebMCP tools (`apps/docs/lib/webmcp-tools.ts`), both found while dogfooding the tools against Chrome's native WebMCP implementation (Chrome 151.0.7922.174 stable, launched with `--enable-blink-features=WebMCP`):

1. The intentional error messages never reach the agent. The tools throw on failure (missing argument, transport failure, page not found), but Chrome's native WebMCP masks page-side `execute` rejections behind a generic `UnknownError: Tool was executed but the invocation failed. For example, the script function threw an error` — the original message is lost. An agent calling `getDoc` on a missing page gets no usable signal about what went wrong, while a resolved result arrives with its text intact (verified: the route-level "Page not found" text survives only when the tool resolves rather than throws).
2. `getDoc`'s description and input schema cite `/docs/getting-started` as their example path, which 404s on the live site — the content directory `(getting-started)` is a Next.js route group, so its pages resolve at `/docs/architecture`, `/docs/installation`, etc. An agent that copies the example path gets a not-found error on its first call.

## Root cause

The tools were written to the spec draft's contract, where a rejected `execute` promise is the failure channel. Chrome's implementation drops the rejection value at the Blink boundary by construction, not as a bug: `ToolFunctionFinishedCallback::HandleFailure` in `third_party/blink/renderer/core/script_tools/model_context.cc` logs the real message to the console and forwards the rejection value only to the DevTools probe, then `ModelContext::OnToolExecuted` builds a `ScriptToolError(kToolInvocationFailed)` with no message, so `GetToolErrorMessage` falls back to the generic literal.

Resolved values take the other path: `React` JSON-serializes a resolved object whole and delivers it as a successful result. Blink has no `isError` concept, so under Chrome the flag rides along inside that serialized payload rather than marking the call as failed. That is still what keeps the message reachable, and hosts that implement the spec's `CallToolResult` shape (the WebMCP polyfill's `normalizeToolResponse`, and `@assistant-ui/react`'s own provider) honor the flag directly, so the result stays spec-shaped instead of Chrome-specific.

## Change

- Wrap each tool's `execute` at the registration boundary with `withErrorResults`, which catches non-abort failures and resolves them as `{ isError: true, content: [{ type: "text", text }] }` — the same pattern `@assistant-ui/react`'s WebMCP provider uses in `convertTools.ts`. `AbortError` rejections still propagate so a cancellation the caller requested stays a rejection rather than a tool error.
- Route-level `isError` results from `/api/mcp` now pass through unchanged (previously they were re-thrown with only the first text item, which the wrapper would have reconstructed anyway; the pass-through also preserves the full content array).
- Replace `/docs/getting-started` with `/docs/installation` in `getDoc`'s description and schema, and in the tests that mirrored it. That is the example `read_page`'s own schema in `lib/mcp-tool-definitions.ts` already uses, so an agent reading both descriptions sees one canonical path.

The wrapper is applied only at WebMCP registration; the module has no other consumers (`registerWebMcpTools` is called from `WebMcpTools` and tests only).

## Verification

- `apps/docs`: full vitest suite 58 files / 471 tests green, including updated `lib/webmcp-tools.test.ts` (failure paths now pinned as `isError` results; abort paths pinned as rejections, covering `DOMException`, non-Error `{name:"AbortError"}` values, and aborts surfacing from `response.json()` mid-stream) and the route transport integration tests in `app/api/mcp/route.test.ts`.
- `oxlint` and `oxfmt --check` clean on the changed files.
- Masking behavior observed directly against Chrome 151.0.7922.174 with `--enable-blink-features=WebMCP`: a throwing docs tool surfaces as a generic error with the message dropped, while a resolved result arrives with its text intact. Confirmed against `model_context.cc` at Chromium HEAD, which discards the rejection value rather than propagating it.
- `/docs/getting-started` returns 404 on the live site and `/docs/installation` returns 200; `content/docs/(getting-started)` is a Next.js route group, so it never appears in a URL.

## Public surface

None. `apps/docs` is private; no changeset needed. Diff: +61/−32 across the module (+40 impl) and its test file (+21 net).

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6655?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.fomGtU5hucXdIGgZiHm1IerksaKceVVsMmq3mQCoJCA6CXae1SWhnqsUnok?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/apps/docs/lib/webmcp-tools.test.ts
+++ b/apps/docs/lib/webmcp-tools.test.ts
@@ -123,12 +123,12 @@ describe("registered tools", () => {
   it("getDoc calls read_page with the given path", async () => {
     const fetchImpl = fetchReturning({ result: okResult });
     await toolByName(fetchImpl, "getDoc").execute({
-      path: "/docs/architecture",
+      path: "/docs/installation",
     });
 
     expect(sentRequest(fetchImpl).body.params).toEqual({
       name: "read_page",
-      arguments: { path: "/docs/architecture" },
+      arguments: { path: "/docs/installation" },
     });
   });
 
@@ -283,7 +283,7 @@ describe("registered tools", () => {
       },
     }));
     const rejection = await toolByName(abortingBody, "getDoc")
-      .execute({ path: "/docs/architecture" })
+      .execute({ path: "/docs/installation" })
       .then(
         () => {
           throw new Error("expected rejection");
--- a/apps/docs/lib/webmcp-tools.test.ts
+++ b/apps/docs/lib/webmcp-tools.test.ts
@@ -11,6 +11,10 @@ const okResult = {
   content: [{ type: "text", text: "hello" }],
 };
 
+function errorResult(text: string) {
+  return { isError: true, content: [{ type: "text", text }] };
+}
+
 function fetchReturning(payload: unknown, ok = true, status = 200) {
   return vi.fn(async () => ({
     ok,
@@ -119,12 +123,12 @@ describe("registered tools", () => {
   it("getDoc calls read_page with the given path", async () => {
     const fetchImpl = fetchReturning({ result: okResult });
     await toolByName(fetchImpl, "getDoc").execute({
-      path: "/docs/getting-started",
+      path: "/docs/architecture",
     });
 
     expect(sentRequest(fetchImpl).body.params).toEqual({
       name: "read_page",
-      arguments: { path: "/docs/getting-started" },
+      arguments: { path: "/docs/architecture" },
     });
   });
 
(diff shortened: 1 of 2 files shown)


okisdev (approved):
approving. the premise holds, and it holds structurally rather than just on the build you tested against.

i pushed two things on top. the first is the comment on `withErrorResults`: it said Chrome passes isError results through with their text intact, and that isn't what happens, because Chrome has no isError channel at all. `ToolFunctionFinishedCallback::React` in `third_party/blink/renderer/core/script_tools/model_context.cc` JSON-serializes a resolved object whole and hands it up as a plain success string, and `ScriptToolExecutedCallback` is a `base::expected<String, ScriptToolError>`, so there is nowhere for the flag to live. what your change actually buys under Chrome is that the text survives; the call still reads as successful and the agent has to read the JSON to see that it failed. that is a reading of Chromium at HEAD, not something i ran, but it changes what someone debugging this should expect to see.

the other half of your claim came out stronger than the description put it. the masking is not a Chrome 151 quirk that might get patched: `HandleFailure` logs the real message to the console and forwards the rejection value only to the DevTools probe, then `OnToolExecuted` builds a `ScriptToolError(kToolInvocationFailed)` with no message, so `GetToolErrorMessage` falls back to the literal you saw. the value is discarded on purpose at the boundary, which is why the fix is durable.

the trap here, and the reason i rewrote the comment instead of dropping it: grep the spec at the ref the file header pins and there is no `isError` anywhere, which makes the comment you removed look correct and this change look like it contradicts the spec. the answer is not in the spec, it is that the implementations disagree. the polyfill's `normalizeToolResponse` returns any object with a `content` array verbatim, so the flag is honored there and in `convertTools.ts`, while Chrome ignores it. resolving a spec-shaped `CallToolResult` is right for both, so the shape stays as you wrote it.

second change is smaller: `/docs/architecture` becomes `/docs/installation`, since `read_page`'s own schema in `lib/mcp-tool-definitions.ts` already uses that path and an agent reads both descriptions. i checked both against the live site, `/docs/getting-started` 404s and `/docs/installation` returns 200.

on the tests, i ran the head suite against your pre-fix implementation and got 4 failed / 15 passed, all four being the failure-path assertions, then 28 passed on head including the route tests. so they pin the new contract rather than passing either way.

one thing i deliberately left alone: abort still propagates as a rejection here while `convertTools.ts` turns cancellation into an isError result. Chrome flattens a rejected AbortError into that same generic message anyway, so the agent cannot tell it apart, but the caller is the one who aborted, so keeping it a rejection is the better of the two. your reasoning in the description stands.

PR #6661, opened 2026-09-01
fix(core): report run events from background thread runtimes

### LINKED ISSUE #6643: bug(core): thread.runEnd is dropped when its thread is switched away
## Checklist

- [x] I searched the existing issues and did not find a duplicate.
- [x] This is reproducible on the latest `main`; it is not a usage question.

## Environment info

- assistant-ui commit: `48e12ab0692f5f252c8ae57993fecb3520e2bae3`
- `@assistant-ui/core`: `0.3.16` workspace
- React: `19.2.8`
- Node.js: `v24.19.0`
- pnpm: `12.0.0`
- macOS arm64

## Affected areas

Runtime / Core, Thread Management, Store / Tap

## Runtime adapter in use

`useRemoteThreadListRuntime` with a `useLocalRuntime` thread runtime.

## Steps to reproduce

1. Mount a remote thread-list runtime with two threads.
2. Subscribe to `thread.runEnd` at root scope:
   ```tsx
   useAuiEvent({ scope: "*", event: "thread.runEnd" }, onRunEnd);
   ```
3. Start a delayed run on thread A.
4. Switch to thread B while thread A is still running.
5. Resolve the run on thread A.

A focused regression test on current `main` attaches both an internal runtime listener and the store listener:

```tsx
const threadAId = runtime.threads.mainItem.getState().id;
const threadA = runtime.threads.getById(threadAId);
const coreRunEnd = vi.fn();
const storeRunEnd = vi.fn();

threadA.unstable_on("runEnd", coreRunEnd);
// Mounted inside AssistantRuntimeProvider:
useAuiEvent({ scope: "*", event: "thread.runEnd" }, storeRunEnd);

threadA.startRun({ parentId: null });
await waitFor(() => expect(threadA.getState().isRunning).toBe(true));

await runtime.threads.switchToThread("thread-b");
run.resolve({ content: [] });

await waitFor(() => expect(coreRunEnd).toHaveBeenCalledTimes(1));
expect(storeRunEnd).toHaveBeenCalledExactlyOnceWith({ threadId: threadAId });
```

The final assertion fails with zero calls. The internal `runEnd` listener is called once, confirming that the background run finishes and the thread runtime emits the event.

## Expected vs actual behavior

Expected: the root-scope listener receives `{ threadId: threadAId }` when thread A finishes, even though thread B is currently selected. This matches the `scope: "*"` documentation, which says it receives emissions from any descendant scope regardless of the current React context.

Actual: `thread.runEnd` is not delivered after thread A is switched away. The internal thread runtime still emits `runEnd`.

A default-scope `useAuiEvent("thread.runEnd", ...)` following the selected thread may reasonably stop listening to thread A. The bug is that the explicit root-scope subscription also misses the event.

## Root cause

`ThreadListClient` creates a `ThreadClient` only for `runtime.main`. The effect in `ThreadClient` bridges `runStart`, `runEnd`, `initialize`, and `modelContextUpdate` from the runtime into store events. When selection moves to thread B, that resource is replaced and its cleanup unsubscribes from thread A. Thread A remains alive and completes in the background, but no event bridge is attached when it emits `runEnd`.

The fix likely needs the event bridge to stay attached to each retained thread runtime, or equivalent list-level forwarding, while avoiding duplicate events for the selected thread.

<a href="https://rupic.dev/s/s2.jgHFkBfbWfe_TdOqqrm3dQpvliz-eil-EAYeV39Z4ZmlPBwrICOj63y25CE?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

okisdev:
## Problem

`thread.runEnd` never reaches store listeners when a run outlives its thread's selection. Reproduces on current `main`: start a delayed run on thread A of a `useRemoteThreadListRuntime` list, subscribe with `useAuiEvent({ scope: "*", event: "thread.runEnd" }, cb)`, switch to thread B, then settle A. The retained runtime emits its internal `runEnd`, and the store listener receives nothing.

A run that starts on a thread that is not selected is invisible the same way, in both directions.

Fixes #6643.

## Root cause

The selected `ThreadClient` owns the runtime-to-store event bridge (`packages/core/src/store/runtime-clients/thread-runtime-client.ts`). It subscribes to `runtime.main`, whose `EventSubscriptionSubject` rebinds on every selection change, so switching threads unsubscribes the bridge from the thread that is still running. A remote thread list keeps that runtime alive, so the run finishes with no bridge attached.

## Change

`RemoteThreadListHookInstanceManager` already subscribes to every retained runtime to track `isRunning`. It now forwards `runStart` and `runEnd` from that same subscription, tagged with the thread id, through `__internal_subscribeRunEvents`. `RemoteThreadListThreadListRuntimeCore` exposes it as the optional core capability `unstable_subscribeThreadRunEvents`, alongside the existing `unstable_isThreadRunning`, and `ThreadListRuntimeImpl` surfaces it (returning a no-op unsubscribe for lists that do not implement it). `ThreadListClient` subscribes once and emits `thread.runStart` / `thread.runEnd` for any thread that is not the selected one, leaving the selected thread's events to `ThreadClient`.

Cleanup rides on the manager's existing `unsubscribeRunning`, so delete, detach, restart and runtime replacement all stop the forwarding without a second lifecycle to keep in sync. Thread lists that mount only the main thread (external store, local) leave the capability undefined: they retain no background runtime, so there is nothing to forward and no gate is needed.

The manager is the layer that already owns retained-runtime identity, so no captured core, generation stamp, pending map or synthesized event is needed to keep the reports honest. Every emitted event is one the runtime actually raised, at the moment it raised it, which matters because the devtools run timeline (`packages/react-devtools/src/views/runs/parse.ts`) pairs runs by these events.

Background events are emitted on the thread list's client stack, so they reach `scope: "*"` listeners. Default-scope listeners still follow the selected thread, unchanged.

This supersedes #6645 and #6644, which fix the same issue in the store client layer.

## Verification

- The three new failing-first regressions were confirmed against the unmodified checkout: switched-away delivery, repeated switching, and a never-selected run all fail without the change and pass with it.
- `pnpm --filter @assistant-ui/core test` (153 files, 1573 tests)
- `pnpm --filter @assistant-ui/react test` (56 files, 507 tests)
- `pnpm --filter @assistant-ui/store test` (24 files, 186 tests), `pnpm --filter @assistant-ui/tap test` (51 files, 580 tests), `pnpm --filter @assistant-ui/vue test` (18 files, 106 tests, 1 skipped)
- `pnpm lint`, `oxfmt --check` on the changed files, `pnpm changesets:check`
- `tsc --noEmit` reports no diagnostics on any changed file; the remaining errors are the pre-existing test-file ones on `main`.

## Public surface

`@assistant-ui/core`: adds the optional `ThreadListRuntimeCore.unstable_subscribeThreadRunEvents`, the `ThreadRunEvent` type, and `ThreadListRuntime.unstable_subscribeThreadRunEvents`. Additive only. No documentation, API-reference or template changes. Includes a patch changeset.

<a href="https://rupic.dev/s/s2.Mn1d_bePlo2B1UG8gMJXEfgAJKWuq2QQ5DVrp6izd4uWZcD_UDXkKc-fZRA?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/background-thread-run-events.md
+++ b/.changeset/background-thread-run-events.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/core": patch
+---
+
+fix: report run start and end for threads a remote thread list keeps alive in the background
--- a/api-surface/assistant-ui__ai-sdk.ts
+++ b/api-surface/assistant-ui__ai-sdk.ts
@@ -1722,6 +1722,7 @@ type ThreadListRuntime = {
     unarchive?: boolean;
   }): Promise<void>;
   switchToNewThread(): Promise<void>;
+  unstable_subscribeThreadRunEvents(callback: (event: ThreadRunEvent) => void): Unsubscribe;
   getLoadThreadsPromise(): Promise<void>;
   reload(): Promise<void>;
   reloadMainThread(): Promise<void>;
@@ -1789,6 +1790,11 @@ type ThreadMessageLike = {
     readonly isOptimistic?: boolean | undefined;
     readonly custom?: Record<string, unknown> | undefined;
   } | undefined;
+};
+
+type ThreadRunEvent = {
+  readonly threadId: string;
+  readonly type: "runEnd" | "runStart";
 };
 
 type ThreadRuntime = {
--- a/api-surface/assistant-ui__core.ts
+++ b/api-surface/assistant-ui__core.ts
@@ -3734,6 +3734,7 @@ declare class RemoteThreadListHookInstanceManager extends BaseSubscribable {
   }> | undefined;
   __internal_isThreadRunning(threadId: string): boolean;
   __internal_subscribeRunningChanged(callback: () => void): Unsubscribe$1;
+  __internal_subscribeRunEvents(callback: (event: ThreadRunEvent) => void): Unsubscribe$1;
   __internal_subscribeRuntimeReplaced(callback: () => void): Unsubscribe$1;
   stopThreadRuntime(threadId: string): void;
   setRuntimeHook(newRuntimeHook: RemoteThreadListHook): void;
@@ -3910,6 +3911,7 @@ declare class RemoteThreadListThreadListRuntimeCore extends BaseSubscribable imp
     unstable_on<E extends ThreadRuntimeEventType>(event: E, callback: ThreadRuntimeEventCallback<E>): Unsubscribe$1;
   }>;
   unstable_isThreadRunning(threadIdOrRemoteId: string): boolean;
+  unstable_subscribeThreadRunEvents(callback: (event: ThreadRunEvent) => void): Unsubscribe$1;
   getItemById(threadIdOrRemoteId: string): RemoteThreadData | undefined;
   switchToThread(threadIdOrRemoteId: string, options?: {
     unarchive?: boolean;
@@ -4718,6 +4720,7 @@ type ThreadListRuntime = {
     unarchive?: boolean;
   }): Promise<void>;
   switchToNewThread(): Promise<void>;
+  unstable_subscribeThreadRunEvents(callback: (event: ThreadRunEvent) => void): Unsubscribe$1;
   getLoadThreadsPromise(): Promise<void>;
   reload(): Promise<void>;
   reloadMainThread(): Promise<void>;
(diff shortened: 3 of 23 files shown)


okisdev:
closing in favor of #6645, which now carries this implementation on its own branch.

Issue #6311, opened 2026-08-25
lint: fix the react/set-state-in-effect warnings and promote the rule to error


okisdev:
`oxlint` 1.79 promoted a batch of React Compiler rules into the `correctness` category, which `.oxlintrc.json` sets to `error`, so a clean `pnpm lint` became 567 errors overnight. #6303 held the linter at `~1.78.0` to keep CI green; the unpin classifies those rules instead of fixing 567 sites, and `react/set-state-in-effect` is the one that survives as a real finding.

### why the rest were turned off rather than fixed

The split is not close. Of the 567:

| rule | count | what it is here |
| --- | ---: | --- |
| `react/refs` | 252 | reading `ref.current` during render, which is the mechanism the tap runtime and the adapter runtime hooks are built on |
| `react/globals` + `react/immutability` in test files | 163 | the standard way a test captures a render-time value into an outer binding |
| `react/set-state-in-effect` | 70 | ordinary component code, and a real defect class |
| everything else | 82 | mixed |

`packages/tap/src/hooks/utils/useRenderMemo.ts` makes the point on its own: it draws 16 errors for reading and writing a ref during render, which is exactly what implementing a memo primitive means. Changing it to satisfy the rule would break it.

The repo also already owns a React Compiler gate with far better precision on this architecture: `pnpm check:resource-memo` compiles every `useResource` / `useResources` / `useClientLookup` host and fails when the compiler bails, which is the property that actually matters here. The turned-off oxlint rules duplicate that concern and lose on precision.

Turning them off follows what `.oxlintrc.json` already does with `react/exhaustive-deps` and `react/rules-of-hooks` (both `warn`, plus three path-scoped `off` overrides) for the same reason: rules that assume idiomatic component code, applied to a codebase whose product is headless hook machinery.

### what is left to do

`react/set-state-in-effect` is at `warn`, and it is the only one of the batch that is not architectural. 72 warnings across 63 files, concentrated where you would expect real instances rather than false positives:

- `apps/docs`, 38
- `packages/ui`, 11
- the rest spread one or two at a time across `packages/react`, `packages/core`, `packages/cloud-ai-sdk`, `packages/react-mcp`, `packages/react-devtools`, `packages/react-langgraph`, `packages/react-langchain`, `templates/minimal`

[...]

PR #6703, opened 2026-09-01
refactor(react): collapse the three trigger-item query predicates onto one helper

### LINKED ISSUE #6423: react: collapse the three matchesQuery copies onto one trigger-item helper
Three near-copies of the same trigger-item predicate live in `packages/react/src`:

- `packages/react/src/unstable/useMentionAdapter.ts:211-217`
- `packages/react/src/unstable/useSlashCommandAdapter.ts:121-127` (byte-identical to the above as of #6406)
- `packages/react/src/primitives/composer/trigger/triggerNavigationResource.ts:9-15` (same result, written without the empty-query early return)

All three answer the same question: does an `Unstable_TriggerItem` match a lowercased query across `id`, `label`, and `description`. The navigation resource's copy is the fallback path for adapters that do not implement `search`, so the three are exercised on different code paths but agree on semantics.

Collapsing them to one exported helper would make a future change to match rules (fuzzy matching, matching `metadata`, locale-aware casing) land once instead of three times, and would stop the two `unstable/` copies from silently diverging.

Out of scope for #6406, which only retyped the slash copy to take an item; filing so the cleanup is not lost.

<a href="https://rupic.dev/s/s2.yM76Fk6BZ-1c5vc1gtlC2oWX93TPbZw3ShaIvrzHtcdWtOdTmGGD11QzX9Q?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

Kinfe123:
Closes #6423.

## Problem

Three near-copies of the same trigger-item predicate live in `packages/react/src`: `useMentionAdapter.ts`, `useSlashCommandAdapter.ts` (byte-identical since #6406), and `triggerNavigationResource.ts` (same result written without the empty-query early return). A future change to match rules would have to land three times, and the copies can silently diverge.

## Change

One shared `matchesTriggerItemQuery(item, lowerQuery)` in `primitives/composer/trigger/`, imported by all three call sites; the local copies are deleted. The helper keeps the empty-query early return — behavior-neutral for the navigation resource's copy too, since `id.toLowerCase().includes("")` is already `true` for every item, so all three copies agreed on the empty-query case. Not exported from the package barrel; the public surface is unchanged.

A colocated test pins the contract: case-insensitive matching across `id`, `label`, and `description`, match-all on empty query, and no throw on items without a description.

## Verification

- `pnpm vitest run` over the new helper test plus the existing `useSlashCommandAdapter` and `triggerSelectionResource` suites: 14 tests pass, no type errors.
- oxlint clean on all touched files (the three pre-existing `triggerNavigationResource` warnings are #6311's scope and untouched).

## Public surface

`@assistant-ui/react` internals only; no new exports. Changeset: patch.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6703?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.7O2Bpcw5l8_BPA70c5pxLx1NQ05-zwjSNI2HPcQIo6XqQ03OV-wkQYcOjJY?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/trigger-item-query-helper.md
+++ b/.changeset/trigger-item-query-helper.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react": patch
+---
+
+refactor: collapse the three trigger-item query predicates onto one shared helper
--- a/packages/react/src/primitives/composer/trigger/matchesTriggerItemQuery.test.ts
+++ b/packages/react/src/primitives/composer/trigger/matchesTriggerItemQuery.test.ts
@@ -17,9 +17,9 @@ describe("matchesTriggerItemQuery", () => {
   });
 
   it("matches case-insensitively against the id", () => {
-    expect(matchesTriggerItemQuery(item({ id: "SummArize" }), "marize")).toBe(
-      true,
-    );
+    expect(
+      matchesTriggerItemQuery(item({ id: "SummArize", label: "y" }), "marize"),
+    ).toBe(true);
   });
 
   it("matches case-insensitively against the label", () => {
--- a/packages/react/src/primitives/composer/trigger/matchesTriggerItemQuery.test.ts
+++ b/packages/react/src/primitives/composer/trigger/matchesTriggerItemQuery.test.ts
@@ -1,0 +1,52 @@
+import { describe, it, expect } from "vitest";
+import { matchesTriggerItemQuery } from "./matchesTriggerItemQuery";
+import type { Unstable_TriggerItem } from "@assistant-ui/core";
+
+const item = (
+  overrides: Partial<Unstable_TriggerItem>,
+): Unstable_TriggerItem => ({
+  id: "summarize",
+  type: "command",
+  label: "Summarize",
+  ...overrides,
+});
+
+describe("matchesTriggerItemQuery", () => {
+  it("matches every item on an empty query", () => {
+    expect(matchesTriggerItemQuery(item({}), "")).toBe(true);
+  });
+
+  it("matches case-insensitively against the id", () => {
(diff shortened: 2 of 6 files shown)


okisdev (approved):
approved. the equivalence claim in the description holds: i ran the three old predicates and the new helper against 5000 combinations of id, label and description (including undefined, empty string and non-ascii) and got zero mismatches, so the navigation copy's missing empty-query early return really was a no-op. full react suite passes in a clean worktree, 514 tests, no type errors, and the lint warnings under the trigger directory are all pre-existing on main.

also confirmed there is no fourth copy hiding anywhere; the three named in #6423 were the whole set.

one thing left standing that i am not asking you to change: `query.toLowerCase()` is still at six call sites, so of the future changes #6423 names, fuzzy matching and matching metadata now land in one place but locale-aware casing would not. hoisting the lowercase out of the filter loop is the normal shape and nothing in the repo is asking for locale matching yet, so it stays as is.

PR #6707, opened 2026-09-01
chore: name the lib lane in sync-templates output and provenance doc

Kinfe123:
Follow-up to #6704, which merged before its review round landed. Addresses both reviewer notes:

- The per-lane provenance comment block now names where minimal's `lib` copies sync from (claude's nit).
- The success line now says "components, hooks, and lib files", matching what the check actually validates (cubic's P3).

Verified: `bash scripts/sync-templates.sh` exits green with the updated output.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6707?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.JntsvwO8D7jN-D6fx7WVl3_RlmkuZ2JVTIK2SPoMG2-jnjHLM1ZzcRzSHZ4?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/scripts/sync-templates.sh
+++ b/scripts/sync-templates.sh
@@ -34,8 +34,9 @@ NUXT_DIR="$TEMPLATES_ROOT/nuxt/app/components/assistant-ui"
 # is already the Base UI source, so minimal's assistant-ui element copies sync
 # from it directly, and `components/ui` copies sync from the vendored
 # `ui/base` stand-ins. Base sources already use the scaffold import shape.
-# Minimal's `hooks` copies sync from packages/ui/src/hooks. The Nuxt template's
-# vue kit copies sync verbatim from packages/ui/src/components/vue/assistant-ui.
+# Minimal's `hooks` copies sync from packages/ui/src/hooks, and its `lib`
+# copies from packages/ui/src/lib. The Nuxt template's vue kit copies sync
+# verbatim from packages/ui/src/components/vue/assistant-ui.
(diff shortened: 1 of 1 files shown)


okisdev (approved):
ran the script both ways on 330cf2f instead of just reading it: green on the head with the new success line, and with a junk line appended to `templates/minimal/lib/utils.ts` it reports `drift detected in 1 minimal lib file(s)` and exits 1. so the wording now names a lane that actually fires, which is the part a docs-only diff can't show on its own.

the provenance sentence lines up with `LIB_SOURCE_DIR` / `MINIMAL_LIB_DIR` at the top of the file, and mirroring the hooks sentence is the right call given the lib lane is structurally one-to-one with hooks (`rendered_lib`, candidate collection, `same_normalized` compare, `--write` copy, failure block). grepped the tree for the old "components and hooks are in sync" wording and there's no second copy left. script only, nothing published, so no changeset needed.

PR #6764, opened 2026-09-02
fix(core): keep a pending human() interrupt when execution starts

### LINKED ISSUE #6763: core: a frontend tool that awaits human() before its first real await never renders its approval prompt
## Summary

A frontend tool whose `execute` starts with `await human(...)` never shows its approval prompt on the AI SDK runtime. The tool part stays in its "running" state, `hasExecutingTools` keeps the thread running, the composer stays disabled, and the parked `human()` promise is never resumed. A tool that awaits something else first (a `fetch`, even `await Promise.resolve()`) works, which is why this is easy to miss.

Found while driving `examples/with-webmcp` end to end with a real model (`upstream/main` @ `39c99b5d0`). Its `clear_completed_tasks` tool is exactly this shape.

## Reproduction

Any `useChatRuntime` app with a frontend tool like this:

```ts
clear_completed_tasks: {
  parameters: z.object({}),
  execute: async (_args, { human }) => {
    const approved = (await human({})) === true;
    if (!approved) return { cleared: 0, message: "The user declined." };
    return { cleared: taskStore.clearCompleted() };
  },
  render: ({ result, interrupt, resume }) => {
    if (result) return <p>{result.cleared} cleared</p>;
    if (interrupt) return <button onClick={() => resume(true)}>Approve</button>;
    return <p>Clearing completed tasks…</p>;
  },
}
```

Or concretely: run `examples/with-webmcp`, send "Clear my completed tasks" (answer "yes" if the model asks first). The part shows "Clearing completed tasks…" forever; Approve/Decline never render, no request is in flight, and the thread stays running.

Moving the `await human({})` behind any other `await` makes the prompt appear.

## Mechanism

Instrumenting the tracker gives this order for the call:

```
capture     call_…   (execute wrapper claims the execution)
humanInput  call_…   (tool body ran synchronously up to `await human({})`)
setStatus   call_…   interrupt
execStart   call_…
setStatus   call_…   executing   <- overwrites the interrupt
```

1. `packages/assistant-stream/src/core/tool/ToolExecutionStream.ts:233` calls `internalOptions.execute(...)`. That runs the tool body synchronously up to its first `await`. When that first await is `human()`, the request fires inside this synchronous slice.
2. `human()` lands in `ToolInvocationTracker._onHumanInput` (`packages/core/src/runtimes/tool-invocations/ToolInvocationTracker.ts:491`), which parks the resolver and sets status `interrupt`.
3. Only then does `ToolExecutionStream.ts:244` fire `onExecutionStart`. `ToolInvocationTracker._onExecutionStart` (`ToolInvocationTracker.ts:507`) unconditionally sets status `executing`, replacing the interrupt.
4. `packages/ai-sdk/src/converters/convertMessage.ts:159` only emits `part.interrupt` when the status is `interrupt`, so `render` never sees the interrupt and the fallback branch shows forever. The status never leaves `executing`, so the thread never settles.

The stream cannot easily fire `onExecutionStart` before `execute`: it only learns the tool is frontend-executed once `execute` returns something other than `undefined`. So the ordering is inherent, and the tracker has to tolerate a status that `execute` already set.

## Why the existing test misses it

The only `human()` coverage in `ToolInvocationTracker.test.ts` (around line 176) requests human input from `streamCall`. `streamCall` never goes through `onExecutionStart`, so the clobbering write never happens in that test. There is no test that requests human input synchronously from `execute`.

## Proposed fix

In `_onExecutionStart`, keep the execution registered (so `_onExecutionEnd` and the settled-resolver bookkeeping still see it) but do not overwrite a pending interrupt:

```ts
this._executing.add(executionId!);
if (this._humanInput.get(toolCallId)?.executionId === executionId) return;
this._setStatus(toolCallId, { type: "executing" });
```

Keying on `_humanInput`'s execution id rather than on the status map matters: a pipeline restart clears `_executing` but leaves `_statuses` and `_humanInput` alone, so a status-keyed check would also suppress the `executing` write for the fresh execution that follows.

With this one line, in the same live run, the prompt renders standalone as intended and both the Decline and Approve round trips complete (`resume(false)` yields "The user declined.", `resume(true)` clears the task). A unit test that drives the `execute` path with a synchronous `human()` call, asserts the status stays `interrupt`, then resumes and asserts the result, pins it.

Happy to send the PR.

<a href="https://rupic.dev/s/s2.pp__KA5APxly7nhbqyTnIwULhmfmUhx3AjG7c2BaWq5H5hYcP4Sh0RYKKmM?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

samdickson22:
Closes #6763

A frontend tool whose `execute` starts with `await human(...)` never shows its approval prompt on the AI SDK runtime: the part sits in its running state, the thread never settles, and the parked `human()` promise is never resumed. Found by driving `examples/with-webmcp` with a real model; its `clear_completed_tasks` tool is exactly this shape.

## Mechanism

`ToolExecutionStream` calls `execute` first and fires `onExecutionStart` only after it returns, because it cannot know the tool is frontend-executed until then. The tool body runs synchronously up to its first `await`, so when that await is `human()`, the tracker has already parked the resolver and set status `interrupt` by the time `_onExecutionStart` runs. That handler then wrote `executing` unconditionally, replacing the interrupt. `convertMessage` only emits `part.interrupt` for status `interrupt`, so `render` never sees it.

## Fix

`_onExecutionStart` still registers the execution (so `_onExecutionEnd` and the settled-resolver bookkeeping are unchanged) but skips the `executing` write when this execution's own human-input request is already pending. The check is keyed on `_humanInput`'s execution id rather than on the status map: a pipeline restart clears `_executing` but leaves `_statuses` and `_humanInput` alone, and a status-keyed check would have suppressed the write for the fresh execution that follows.

## Tests

- `keeps a human-input interrupt that execute requests before its first await`: drives the `execute` path (the existing `human()` test uses `streamCall`, which never reaches `onExecutionStart`) with a synchronous `human()`, asserts the status stays `interrupt`, then resumes and asserts the result lands and the status clears. Fails on main with `executing`.
- `marks a fresh execution as executing when an earlier one left a human-input request behind`: kills the pipeline while a request is pending, promotes a new execution with changed args, asserts it reports `executing`. Fails with a status-keyed guard.

The second commit rewords the example's `clear_completed_tasks` description. With "Requires the user's approval" the model asked in prose first on every run and only called the tool after a "yes", so the real approval UI took two turns to appear. The gate is enforced by `human()`, not by the model.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6764?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.HwQyUxcC19Y_u9W_QJ3e7zC9e4-IsprtEbEosa-CbdHnmjmRoxfoGQiyHFs?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/human-interrupt-before-execution-start.md
+++ b/.changeset/human-interrupt-before-execution-start.md
@@ -1,0 +1,7 @@
+---
+"@assistant-ui/core": patch
+---
+
+fix: keep a frontend tool's pending `human()` interrupt when execution starts
+
+a tool whose `execute` awaits `human()` before any other await had its `interrupt` status overwritten by the `executing` status the execution-start callback writes, so the approval prompt never rendered and the run never settled.
--- a/examples/with-webmcp/app/toolkit.tsx
+++ b/examples/with-webmcp/app/toolkit.tsx
@@ -36,7 +36,7 @@ export default defineToolkit({
     // the approval prompt behind a disclosure the user has no reason to open.
     display: "standalone",
     description:
-      "Remove every completed task from the user's task list. Requires the user's approval.",
+      "Remove every completed task from the user's task list. The UI asks the user to confirm before anything is removed, so call this directly instead of asking first.",
     parameters: z.object({}),
     execute: async (_args, { human }) => {
       "use client";
--- a/packages/core/src/runtimes/tool-invocations/EDGE_CASES.md
+++ b/packages/core/src/runtimes/tool-invocations/EDGE_CASES.md
@@ -137,6 +137,17 @@ produces: every entry pending in a snapshot closes in the same
 every call, and an execution parked on `human()` reports `interrupt` rather
 than `executing`, which `_hasExecutingTools` does not count.
 
+That holds even when `execute` opens with `await human()`. The stream
+fires `onExecutionStart` only after `execute` returns, because until then
+it cannot tell a frontend tool from one without an `execute`, so by the
+time the callback runs the body has already parked its resolver and set
+`interrupt`. `_onExecutionStart` still registers the execution for
+`_onExecutionEnd` but leaves the status alone when this execution's own
+request is pending. The check is keyed on the request's execution id,
+not on the status map: a pipeline restart (F.4) clears `_executing` but
+keeps the status map, and a stale `interrupt` must not stop the fresh
+execution from reporting `executing` (#6763).
+
 A call the adapter reports as client-owned (A.9) closes as soon as its
 arguments parse, because the adapter has already said the provider will
 not answer it. Ownership says who answers a call, not whether it is
--- a/packages/core/src/runtimes/tool-invocations/ToolInvocationTracker.test.ts
+++ b/packages/core/src/runtimes/tool-invocations/ToolInvocationTracker.test.ts
@@ -247,6 +247,96 @@ describe("ToolInvocationTracker", () => {
     }
   });
 
+  it("keeps a human-input interrupt that execute requests before its first await", async () => {
+    const execute = vi.fn(async (_args, { human }) => ({
+      approved: (await human({ request: "approve" })) === true,
+    }));
+    const getTools = () => ({
+      weatherSearch: {
+        parameters: { type: "object", properties: {} },
+        execute,
+      } satisfies Tool,
+    });
+    const onResult = vi.fn();
+    let statuses: Record<string, ToolExecutionStatus> = {};
+    const onStatusesChange = (s: ReadonlyMap<string, ToolExecutionStatus>) => {
+      statuses = Object.fromEntries(s);
+    };
+
+    const tracker = new ToolInvocationTracker(getTools, {
+      onResult,
+      onStatusesChange,
+    });
+    tracker.setState(createState([], false));
+    tracker.setState(
+      createState(
+        [createAssistantMessage('{"query":"London"}', { query: "London" })],
+        false,
+      ),
+    );
+
+    await waitFor(() => {
+      expect(statuses["tool-1"]?.type).toBe("interrupt");
+    });
+
+    expect(tracker.resume("tool-1", true)).toBe(true);
+
+    await waitFor(() => {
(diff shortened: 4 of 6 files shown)


okisdev (approved):
approving. the ordering is real: `ToolExecutionStream` calls `execute` at :233 and only fires `onExecutionStart` at :244, once it has a return value to tell a frontend tool apart from one without an execute, so a body that opens on `await human(...)` has already parked its resolver and set `interrupt` before the callback runs. keying the guard on `_humanInput`'s execution id rather than on the status map is the part that makes it correct rather than merely green, and the second test pins exactly that.

i ran the suites in a worktree at b7bcc66: core 1643 passing across 157 files, eve 161. then i took the guard back out and both new tests fail with `expected 'executing' to be 'interrupt'`, and swapped in the status-keyed version instead, where the fresh-execution test fails with `expected 'interrupt' to be 'executing'`. so both claims in the description hold literally. the with-webmcp end to end run i took on your word; i only read that part.

the eve test swap is worth flagging for anyone reading this later: it isn't papering over a behavior change. the old `isRunning` assertion was leaning on the very `executing` status this fixes, and `humanRequests === 2` is the thing it actually meant to wait for.

the toolkit description reword is a second concern strictly speaking, but it's one line on the repro example and i'd rather it ship next to the fix.

PR #6794, opened 2026-09-03
refactor(assistant-stream): drop the redis client members the resumable store never calls

### LINKED ISSUE #6786: RedisLikeClient carries members the resumable store never calls
`RedisResumableStreamStore` calls only `setNX`, `get`, `del`, `xRange`, `pipeline` and `finalizeIfUnchanged` on its client, but the exported `RedisLikeClient` in `packages/assistant-stream/src/resumable/stores/redis-impl.ts:43` also declares `set`, `expire`, `exists` and `xAdd`. grepping `this.client.` in that file returns one hit, `del`, plus the six above; the other four have no caller and predate #6737.

`PipelineCommand`'s `"set"` variant joins them after #6737: the fallback branch that built it was the only construction site, so the variant is now unconstructed while both adapters still translate it (`redis.ts:126`, `ioredis.ts:101`) and the test fake still handles it.

`RedisLikeClient` is reachable from the public `assistant-stream/resumable` entry as a type, but no public entry accepts one: `RedisResumableStreamStore` is not exported, `createRedisResumableStreamStore` takes `NodeRedisLike`, and `createIoredisResumableStreamStore` takes `IoRedisLike`. narrowing the interface therefore cannot break a runtime consumer, only code that typed a value against it directly.

worth deciding as one change rather than trimming a member at a time: either drop the unused members and the `"set"` variant, or keep the interface as a deliberately complete Redis subset and say so in its doc comment.

<a href="https://rupic.dev/s/s2.OBqeGCJLzNSzX1Krm91scr1S6mCzQqgcElzWYuXTpcSJKIH2owDflqIqfps?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

okisdev:
## Problem

`RedisResumableStreamStore` talks to redis through `RedisLikeClient`, but the interface declares four methods the store never calls. `rg 'this\.client\.' redis-impl.ts` returns ten hits covering `setNX`, `get`, `del`, `xRange`, `pipeline` and `finalizeIfUnchanged`; `set`, `expire`, `exists` and `xAdd` have no caller.

`PipelineCommand`'s `"set"` variant joins them as of #6737. that PR replaced the non-atomic finalize fallback with `finalizeIfUnchanged`, and the fallback was the variant's only construction site:

```diff
-      { type: "set", key: metaKey, value: meta, ttlSec },
-      { type: "xAdd", key: dataKey, fields },
-      { type: "expire", key: dataKey, ttlSec },
```

both adapters still translated it, and so did both test fakes.

closes #6786.

## Root cause

the contract was written as a redis facade rather than as the port the store actually needs, so members outlived their callers instead of being removed with them.

## Change

`RedisLikeClient` keeps the six methods the store calls; `PipelineCommand` keeps `xAdd` and `expire`.

that cascades into the node-redis adapter, whose exported `NodeRedisLike` documents itself as the "structural subset of node-redis v5 used by the adapter": `expire`, `exists`, `xAdd` and the non-NX `set` overload are no longer used by it, and `NodeRedisMultiCommand` loses `set` plus `exec`, which was already dead (only `execAsPipeline` is ever awaited). narrowing a parameter type only widens what callers may pass, and the comment goes back to being true.

the python port carried the same drift: its `RedisLikeClient` protocol declared `exists` with no caller, and `_RedisAsyncioAdapter.pipeline` still had the `"set"` branch. both are gone, keeping the two ports on one contract.

kept the interface narrow rather than documenting it as a deliberate redis subset, because it is not one: there is no `ttl`, `xLen`, `unlink` or `scan`, and the shapes are store-shaped anyway (`setNX(key, value, ttlSec)` collapses `SET NX EX`; `finalizeIfUnchanged` is a bespoke lua call, not a redis command).

## Verification

`packages/assistant-stream` run against a real redis so the adapter translation is actually exercised rather than skipped:

```
REDIS_URL=redis://127.0.0.1:6399 pnpm --filter assistant-stream test
  test:v6         531 passed (531)
  test:peer-v5    531 passed (531)
  test:types:peer-v5  clean
```

both legs include the 22 redis integration tests that skip without a server, so node-redis v5/v6 and ioredis v5/v6 all round-trip through the narrowed pipeline.

python, same server:

```
REDIS_URL=redis://127.0.0.1:6399 uv run --extra dev --extra langgraph pytest -q
  182 passed
```

that turns on the 7 redis adapter tests that otherwise skip.

`pnpm lint` clean on the touched files, `pnpm changesets:check` passes, `pnpm api-surface` regenerated.

## Public surface

- `assistant-stream`: `RedisLikeClient` loses `set`, `expire`, `exists`, `xAdd`; `PipelineCommand` loses its `"set"` variant; `NodeRedisLike` loses `expire`, `exists`, `xAdd` and the non-NX `set` overload. no export is removed, and no public entry accepts a `RedisLikeClient`: `RedisResumableStreamStore` is not exported, `createRedisResumableStreamStore` takes `NodeRedisLike`, `createIoredisResumableStreamStore` takes `IoRedisLike`. runtime behavior is unchanged.
- `assistant-stream` (python): `RedisLikeClient` protocol loses `exists`.
- `api-surface/assistant-stream.ts` regenerated.
- changeset: `assistant-stream` patch.
- docs: none. no page references `RedisLikeClient`; the custom-store guide points at `ResumableStreamStore`.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6794?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.vPGnSk7VwL1ljclsR1KIpuY4hqDj2_nILzSdMeSH3jg7arRzxZhmDOX2fhE?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/redis-like-client-narrow.md
+++ b/.changeset/redis-like-client-narrow.md
@@ -3,3 +3,5 @@
 ---
 
 refactor: drop the redis client members the resumable store never calls
+
+`RedisLikeClient` no longer declares `set`, `expire`, `exists` or `xAdd`, `PipelineCommand` no longer has a `"set"` variant, and `NodeRedisLike` no longer requires `expire`, `exists`, `xAdd` or the non-NX `set` overload; no public API accepts a `RedisLikeClient`, so this only affects code that typed a value against the interface directly.
--- a/.changeset/redis-like-client-narrow.md
+++ b/.changeset/redis-like-client-narrow.md
@@ -1,0 +1,5 @@
+---
+"assistant-stream": patch
+---
+
+refactor: drop the redis client members the resumable store never calls
--- a/api-surface/assistant-stream.ts
+++ b/api-surface/assistant-stream.ts
@@ -924,14 +924,8 @@ interface NodeRedisLike {
     NX: true;
     EX: number;
   }): Promise<string | null>;
-  set(key: string, value: string, options: {
-    EX: number;
-  }): Promise<string | null>;
   get(key: string): Promise<string | null>;
-  expire(key: string, seconds: number): Promise<unknown>;
-  exists(key: string): Promise<number>;
   del(keys: string | string[]): Promise<unknown>;
-  xAdd(key: string, id: string, fields: NodeRedisFields): Promise<string>;
   sendCommand<T = unknown>(args: ReadonlyArray<string | Buffer>, options?: {
     typeMapping?: Record<number, unknown>;
   }): Promise<T>;
@@ -941,11 +935,7 @@ interface NodeRedisLike {
 interface NodeRedisMultiCommand {
   xAdd(key: string, id: string, fields: NodeRedisFields): NodeRedisMultiCommand;
   expire(key: string, seconds: number): NodeRedisMultiCommand;
-  set(key: string, value: string, options: {
-    EX: number;
-  }): NodeRedisMultiCommand;
   execAsPipeline(): Promise<unknown>;
-  exec(): Promise<unknown>;
 }
 
 type NodeRole = "all" | "master" | "slave";
@@ -1019,11 +1009,6 @@ type PipelineCommand = {
 } | {
   readonly type: "expire";
   readonly key: string;
-  readonly ttlSec: number;
-} | {
-  readonly type: "set";
-  readonly key: string;
-  readonly value: string;
   readonly ttlSec: number;
 };
 
@@ -24155,12 +24140,8 @@ type RedisKey = string | Buffer;
 
 interface RedisLikeClient {
   setNX(key: string, value: string, ttlSec: number): Promise<boolean>;
-  set(key: string, value: string, ttlSec: number): Promise<void>;
   get(key: string): Promise<string | null>;
-  expire(key: string, ttlSec: number): Promise<void>;
-  exists(key: string): Promise<boolean>;
   del(keys: string[]): Promise<void>;
-  xAdd(key: string, fields: Record<string, string | Uint8Array>): Promise<string>;
   xRange(key: string, start: string, end: string): Promise<Array<{
     id: string;
     fields: Record<string, string | Uint8Array>;
--- a/packages/assistant-stream/src/resumable/stores/ioredis.ts
+++ b/packages/assistant-stream/src/resumable/stores/ioredis.ts
@@ -33,26 +33,12 @@ function adapt(client: IoRedisLike): RedisLikeClient {
       const result = await client.set(key, value, "EX", ttlSec, "NX");
       return result === "OK";
     },
-    async set(key, value, ttlSec) {
-      await client.set(key, value, "EX", ttlSec);
-    },
     async get(key) {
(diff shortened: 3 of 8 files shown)


okisdev:
not taking the compatibility split. the precedent this repo already set on this exact interface settles it.

published `assistant-stream@0.3.40` declares `RedisLikeClient` with nine members and no `finalizeIfUnchanged`:

```ts
// npm pack assistant-stream@latest → dist/resumable/stores/redis-impl.d.ts:22
interface RedisLikeClient {
  setNX(...); set(...); get(...); expire(...); exists(...);
  del(...); xAdd(...); xRange(...); pipeline(...);
}
```

#6737 added `finalizeIfUnchanged` as a required member. that breaks every implementer of the shipped interface, which is strictly more disruptive than removing members that break only callers, and it shipped as a patch with the reasoning written into `.changeset/calm-streams-finalize.md`:

> `RedisLikeClient` now requires a `finalizeIfUnchanged` method; no public API accepts a `RedisLikeClient`, so this only affects code that typed a value against the interface directly.

that changeset is still pending, so 0.3.40 is the last published shape and both changes land in the same release. no npm version ever sees one without the other.

#6737 also deleted "custom or proxied clients can implement it directly" from the interface's doc comment. the invitation to implement `RedisLikeClient` was already withdrawn; keeping a fat exported copy would re-publish a contract we have stopped offering, and it would be a false one, since nothing would require or call the four members.

on `AGENTS.md:40`: it forbids removing an export that has shipped. no export is removed here. `RedisLikeClient`, `NodeRedisLike`, `PipelineCommand` and `RedisFinalizeOptions` all still resolve.

on the claim you marked as not holding: `NodeRedisLike` is a parameter type (`createRedisResumableStreamStore(client: NodeRedisLike, …)`), so dropping members only weakens what a caller's client must satisfy. a real `RedisClientType` stays assignable, which the `test:types:peer-v5` leg pins. the python `RedisLikeClient` is a `Protocol`, so dropping `exists` widens the set of accepted implementations rather than narrowing it.

adopted your changeset point in 60cb46ae5: it now names the removed members and the consumer-visible consequence, matching the #6737 wording.

verification ran against a live redis rather than the default skip: 531/531 on both the v6 and peer-v5 legs including the 22 integration tests, and 182 python tests including the 7 redis adapter tests.

PR #6796, opened 2026-09-03
fix(react-a2a): preserve raw zero-byte files

Kinfe123:
## Problem

`contentPartsToA2AParts()` drops raw zero-byte file parts because an empty base64 string is treated like missing data. A file such as `{ type: "file", data: "", mimeType: "text/plain" }` therefore disappears from the outgoing A2A message on current main.

## Root cause

The file converter used a truthiness check after confirming the data type. Empty strings are valid raw base64 for zero-byte files, so the check rejected a valid payload.

## Change

Accept every string file payload, including the empty string, while continuing to skip missing and non-string data.

## Verification

- Confirmed the focused regression fails before the change and passes afterward.
- `node_modules/.bin/vitest run src/conversions.test.ts` (90 tests)
- `node_modules/.bin/aui-build`
- `git diff --check`

## Public surface

`@assistant-ui/react-a2a` behavior changes for zero-byte file conversion. No exports or API types change. Includes a patch changeset.

<a href="https://rupic.dev/s/s2.R7evxWakCudZ37yCogHtZzV3TQ_GNIMvjunRcq45ciiEOsmX-yNnIGw_TNI?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/calm-files-smile.md
+++ b/.changeset/calm-files-smile.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react-a2a": patch
+---
+
+fix: preserve zero-byte files when converting messages to A2A parts
--- a/packages/react-a2a/src/conversions.test.ts
+++ b/packages/react-a2a/src/conversions.test.ts
@@ -514,6 +514,20 @@ describe("contentPartsToA2AParts", () => {
     expect(result).toEqual([{ raw: "ZmlsZQ==", mediaType: "text/csv" }]);
   });
 
+  it("preserves raw zero-byte file data", () => {
+    const result = contentPartsToA2AParts([
+      {
+        type: "file",
+        data: "",
+        mimeType: "text/plain",
+        filename: "empty.txt",
+      },
+    ]);
+    expect(result).toEqual([
+      { raw: "", mediaType: "text/plain", filename: "empty.txt" },
+    ]);
+  });
+
   it("falls back to the attachment MIME type when the file part MIME is empty", () => {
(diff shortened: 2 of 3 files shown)


okisdev (approved):
approving. the guard this removes was never a decision to drop zero-byte files: `!part.data` arrived in #5036 when `data` was still `string | undefined`, and #5220 only prefixed the `typeof` narrowing in front of it, at which point the leftover truthiness started rejecting a valid empty payload. #5220 also widened the data URL regex to `(.*)` precisely so zero-byte payloads survive, so this is that same decision finally reaching the bare base64 leg.

executed rather than read: `vitest run src/conversions.test.ts` at 90 passed, and with the truthiness put back the new case is the only failure, so the repro holds in both directions. i also probed the neighbouring legs, and `data:text/plain;base64,` and `audio.data: ""` already yielded `raw: ""` on main, which means the function now preserves zero bytes on every path instead of two out of three. read only: the sibling converters (ai-sdk, eve, google-adk, langchain, langgraph, assistant-stream), none of which drop empty file data, leaving react-a2a as the last holdout.

one thing i noticed and am not asking you to change: `{ sourceType: "url", data: "" }` now emits `{ url: "" }` where it used to be dropped. that input is malformed either way and the langchain and google-adk converters pass it straight through as well, so guarding it here would be the sort of defensive branch this repo deliberately avoids.

Issue #6801, opened 2026-09-03
pnpm test:types fails in 5 packages and no CI job runs it


okisdev:
## problem

`pnpm test:types` fails in 5 packages on a clean `main` checkout: `@assistant-ui/mcp-docs-server`, `@assistant-ui/tap`, `assistant-ui`, `@assistant-ui/next`, `@assistant-ui/cloud`.

```
$ pnpm test:types
...
Error: ERR_PNPM_RECURSIVE_FAIL
  × "pnpm recursive exec" failed in 5 packages
```

## mechanism

no CI job runs it. `code-quality.yaml:361` runs `pnpm turbo test`, which is vitest, and vitest never typechecks. the root script

```json
"test:types": "pnpm -r --workspace-concurrency=4 --no-bail exec sh -c '! test -f tsconfig.json || tsc --noEmit'"
```

is the only thing that runs `tsc --noEmit` per package, so nothing gates it and it has drifted.

two distinct causes:

**test files outside the tsconfig's type environment.** most hits are colocated `.test.ts` files that `tsc --noEmit` compiles without vitest globals or `@types/node`:

- `packages/tap/src/standalone-shim/standalone-shim.graph.test.ts` and `.parity.test.ts`: `TS2591` on `node:fs` / `node:path`, `TS2552` on `__dirname`
- `packages/mcp-docs-server/src/proxy.test.ts`: `TS2304: Cannot find name 'expect'`
- `packages/cloud/src/tests/AssistantCloudAuthStrategy.test.ts`: `TS2591` on `Buffer`

**genuine type drift in source:**

- `packages/mcp-docs-server/src/proxy.ts:47` `TS2339`: `id` does not exist on the `JSONRPCMessage & { method: "initialize" }` intersection
- `packages/cli/src/codemods/v0-12/assistant-api-to-aui.ts:117-121` `TS2339`: `id` does not exist on `VariableDeclarator | IdentifierKind`; `:257` `TS2551`: `exportKind` does not exist on `ExportSpecifier`
- `packages/next/src/with-aui.test.ts:41,90` `TS2339` on `turbopack`; `:101` `TS2554` argument count
- `packages/cloud/src/tests/AssistantCloudAuthStrategy.test.ts:261,601` `TS2352` on `Response` and `Storage` casts

## repro

```
pnpm install --frozen-lockfile
pnpm build
pnpm test:types
```

verified in a clean worktree install at `800494dc0`.

## options

[...]

PR #6802, opened 2026-09-03
feat(cli): scaffold lib/utils from the assistant-ui registry so new projects get cn

### LINKED ISSUE #6769: scaffolded projects still get clsx + tailwind-merge instead of cn
## problem

`packages/ui/src/lib/utils.ts` now re-exports `cn` from the `cn` package, but every project scaffolded by `create-assistant-ui` still lands on `clsx` + `tailwind-merge`. the switch stops at the monorepo boundary.

## mechanism

the registry treats `lib/utils` as an ambient import (`AMBIENT_LOCAL_IMPORT_PATHS` in `apps/registry/scripts/build-registry.ts`), so no registry item ships the file. `create-project.ts` compensates by appending shadcn's own `utils` item to the component install:

```ts
const allShadcn = shadcnUI.includes("utils") ? shadcnUI : [...shadcnUI, "utils"];
```

`https://ui.shadcn.com/r/styles/base-nova/utils.json` ships `twMerge(clsx(inputs))` and declares `clsx` + `tailwind-merge`, so every aliasing template and example has to keep declaring that pair. only `packages/ui`, `templates/minimal` (which ships its own `lib/utils.ts` because `hasLocalComponents` skips the registry install) and `apps/docs` (whose demo zip ships the packages/ui file) declare `cn`.

## options

1. leave it. `lib/utils` stays the consumer's file, which matches shadcn's convention, and users adopt `cn` on their own.
2. ship an assistant-ui `utils` registry item and stop requesting shadcn's. scaffolds land on `cn`, at the cost of `shadcn add` overwriting a `lib/utils.ts` the user may have customized.

option 2 is a change to the registry install contract, so it needs a deliberate decision rather than riding along with the dependency swap.

<a href="https://rupic.dev/s/s2.kCPPHvuSh8amlF7SF57Rw8biuhrvOAWb1Q8AYBUcZxJZur9ayQsbLRglTdk?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

okisdev:
closes #6769

## Problem

`packages/ui/src/lib/utils.ts` re-exports `cn` from the `cn` package, but every project scaffolded by `create-assistant-ui` still landed on `clsx` + `tailwind-merge`. the switch stopped at the monorepo boundary: only `packages/ui`, `templates/minimal`, and `apps/docs` declared `cn`, while 39 templates and examples kept declaring the pair they no longer used.

## Root cause

the registry treats `lib/utils` as an ambient import (`AMBIENT_LOCAL_IMPORT_PATHS` in `apps/registry/scripts/build-registry.ts`), so no assistant-ui item ships the file. `create-project.ts` compensated by appending shadcn's own `utils` item to the component install:

```ts
const allShadcn = shadcnUI.includes("utils") ? shadcnUI : [...shadcnUI, "utils"];
```

`https://ui.shadcn.com/r/styles/base-nova/utils.json` ships `twMerge(clsx(inputs))` and declares `clsx` + `tailwind-merge`, so that one line decided what every scaffolded project got.

## Change

add an `@assistant-ui/utils` registry item (`registry:lib`, `target: lib/utils.ts`, `dependencies: ["cn"]`) whose content is the `packages/ui/src/lib/utils.ts` the monorepo already uses, and request it from the CLI instead of shadcn's:

```ts
const components = ["@assistant-ui/utils", ...shadcnUI, ...auiComponents];
```

shadcn's base-nova items do not declare `utils` as a registryDependency (checked `button` and `dialog`), so dropping the append is sufficient. nothing else pulls the old pair into the install closure.

`lib/utils` stays ambient in the registry's install validation, so no item gains a `utils` registryDependency and an existing project running `shadcn add @assistant-ui/thread` is unaffected. this changes what a fresh scaffold gets, not what a shipped project already has.

the rest of the sweep:

- 39 templates and examples swap `clsx` + `tailwind-merge` for `cn`
- `examples/with-cloud-standalone/lib/utils.ts`, the one example carrying its own copy, becomes `export { cn } from "cn"`
- `apps/registry` drops both devDependencies: it aliases `@/lib/*` into `packages/ui` and has no direct import of either
- `cn` moves to `^0.2.4` across the repo

`packages/react-devtools` deliberately keeps `clsx`. it imports it directly in ten files rather than through `@/lib/utils`, and declares no `tailwind-merge`, so swapping it would add Tailwind conflict merging it never had. that is a separate change if it is wanted at all.

## Verification

- end to end: served `apps/registry/dist` locally and ran `shadcn@4.20.1 add <url>/utils.json --yes` into a clean base-nova project. it wrote `lib/utils.ts` as `export { cn } from "cn";` and added `cn` to dependencies, with no `clsx` or `tailwind-merge`.
- `pnpm -C packages/cli test` 315 passed. the existing single-shadcn-add test now pins `@assistant-ui/utils` present and bare `utils` absent.
- `pnpm -C apps/registry test` 77 passed, `pnpm -C apps/registry build` emits the expected `utils.json`.
- `pnpm build` 95/95, `pnpm lint`, `pnpm sync-templates`, `changesets:check`, `workspace-ranges:check`, `unmanaged-pins:check` all clean.
- `cn@0.2.4` resolves under the repo's `minimumReleaseAge: 1440`, and `import('cn')` from `packages/ui` exports `cn`.

`pnpm test:types` fails in five packages on this branch and on `main` alike; it is not a CI gate and the failures are unrelated. filed as #6801, and proven pre-existing for `packages/cli` by byte-identical `tsc --noEmit` output with and without this change.

## Public surface

- `assistant-ui` (CLI): behavior change in `transformProject`, patch changeset included.
- registry: new public `utils` item at `https://r.assistant-ui.com/styles/{style}/utils.json`. additive; no existing item changes.
- no exports added or removed. api-reference unaffected.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6802?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.dQb6J9-aK-MzkJEOEHEe2YZZRobWECfisASOLp25RbSV8M3WhyNEUcpe4M8?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>


okisdev:
the maintainer decision is recorded on #6769: option 2, scaffolds land on `cn` from our own registry item. `cn` is shadcn's own package, so this is the `lib/utils` convention moving rather than a divergence from it.

on the trade-off you named: `shadcn add` does not overwrite a customized `lib/utils.ts`. checked against shadcn 4.20.1, `packages/shadcn/src/utils/updaters/update-files.ts:227-251` — with `--overwrite` false, which is the default and which `create-project.ts` never passes, an existing file is skipped non-interactively, or prompted with `initial: false` interactively. `--yes` only skips the install confirmation and is a separate flag. the cost you are pointing at does not exist in that form.

the real one is the inverse, and it is why `lib/utils` stays in `AMBIENT_LOCAL_IMPORT_PATHS` and no item gains a `utils` registryDependency: in an existing project the file would be silently skipped while `cn` still landed in `package.json`. this PR changes what a fresh scaffold gets and nothing else.

your two unverified claims were both checked before this was opened. a clean base-nova project fed `apps/registry/dist` over a local server and run through `shadcn@4.20.1 add <url>/utils.json --yes` received `lib/utils.ts` as `export { cn } from "cn";` with `cn` in dependencies and neither `clsx` nor `tailwind-merge`. the existing-project path is unchanged by construction, since no item references the new one.

CI is green: 25 pass, 3 skipping, 0 failing.

Issue #6745, opened 2026-09-02
ai-sdk: a stopped run leaves the assistant message status complete instead of incomplete/cancelled


okisdev:
## problem

on `useChatRuntime` (and `useAISDKRuntime`), pressing stop mid-stream ends the run with the last assistant message reporting `status: { type: "complete" }`. a UI cannot tell a stopped reply from a finished one, so a "you stopped this response" affordance (the ChatGPT pattern) has nothing to key off. the local runtime records `{ type: "incomplete", reason: "cancelled" }` for the same action, so this is a cross-runtime parity gap.

## mechanism

`packages/ai-sdk/src/runtime/useAISDKRuntime.ts` derives message status through `getAutoStatus(isLast, isRunning, ..., chatHelpers.error)`. after `chatHelpers.stop()` the abort error is swallowed in `onCancel`, `isRunning` flips to false, and no cancellation signal remains, so the last message falls through to `AUTO_STATUS_COMPLETE`. text parts keep whatever `state` the AI SDK left on them.

## repro

any `useChatRuntime` app (the docs home demo works): send a long prompt, press stop after a few tokens, read `useAuiState((s) => s.thread.messages.at(-1)?.status)`. observed on main at 8c33c29b8 with ai 7.0.x.

## proposal

record the cancelled run in the runtime (the abort in `onCancel` is the moment) and feed it into the status derivation so the last assistant message reports `incomplete` / `cancelled`, matching the local runtime. thread UIs and the kit `stopped-run` element can then render the stopped state.

<a href="https://rupic.dev/s/s2.HUQa8ZH6DgIXAVDlkHzC8rrZSPHkD0SuA6hoY8XM7SGHw7AKlI6T4TTNLOI?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg"><img alt="Track in Rupic" src="https://rupic.dev/badges/track-light-v1.svg"></picture></a>

PR #6832, opened 2026-09-04
fix(ai-sdk): leave a cancelled assistant message when a run is stopped before the first chunk

### LINKED ISSUE #6772: ai-sdk: stopping before the first chunk leaves no cancelled message
## problem

on `useAISDKRuntime`, pressing stop while `chatHelpers.status` is still `submitted` (the request is out, no chunk has arrived) leaves the thread ending on the user message with no indication that a run was started and stopped. #6761 reports a stopped run as `incomplete` / `cancelled`, but only for an assistant message that already exists.

LocalRuntime appends the assistant message when the run starts, so the same action there leaves a message with `{ type: "incomplete", reason: "cancelled" }` and an empty body. this is the remaining half of the parity gap from #6745.

## mechanism

`onCancel` marks the tail message only when it is an assistant message:

```ts
const cancelledId =
  isRunning && message?.role === "assistant" ? message.id : undefined;
```

the AI SDK creates the assistant `UIMessage` when the stream produces its first part, so during `submitted` the tail is the user message and nothing is marked.

## repro

measured against #6761 at `fcd7906ab` with a probe on `useAISDKRuntime`: set `chat.status = "submitted"` with a single user message, call `thread.cancelRun()`, then flip to `ready`. the converted thread is:

```
tail: user:undefined
```

no assistant message, no cancelled status.

## proposal

synthesize the assistant message the way LocalRuntime does, so a stopped run always leaves something to render. that is a larger change than threading the existing message's status, which is why it was left out of #6761.

<a href="https://rupic.dev/s/s2.UuSAxShjVQ16Tu3TixiyXOAB0DfTb-qe7FSlEuiB0sjIjd31dA9XMqO5Onk?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

Kinfe123:
Closes #6772.

## Problem

On `useAISDKRuntime`, pressing stop while `chatHelpers.status` is still `submitted` — the request is out, no chunk has arrived — leaves the thread ending on the user message with no indication a run was started and stopped. #6761 reports a stopped run as `incomplete`/`cancelled`, but only for an assistant message that already exists; the AI SDK creates the assistant `UIMessage` with the stream's first part, so during `submitted` there is nothing to mark. LocalRuntime appends its assistant message at run start, so the same action there leaves `{ type: "incomplete", reason: "cancelled" }` with an empty body — this is the remaining half of the #6745 parity gap.

## Change

Takes the issue's proposal: `onCancel` synthesizes the assistant message LocalRuntime would have appended. When the cancel began during `submitted` and no assistant tail existed, then after `chatHelpers.stop()` settles it appends an empty assistant `UIMessage` via `setMessages` and marks its id cancelled, so the shared conversion renders it as `incomplete`/`cancelled` — same shape as LocalRuntime.

Ordering safeguards:

- The synthesis happens **after** the stop settles, so a first chunk racing the abort cannot leave two assistant messages: if the tail is an assistant message by then, that raced-in message takes the mark instead and nothing is appended.
- A non-`AbortError` rejection from `stop()` rethrows before any synthesis, so a failed stop appends nothing (the existing retraction path for pre-marked ids is unchanged).
- A later provider run that resumes into the synthesized tail retracts its cancellation through the existing `resumedMessageId` effect, unchanged.

The pre-existing mark path for an existing assistant tail is untouched; its marking block is extracted as a local `markCancelled` shared by both paths.

## Verification

- The issue's repro (single user message, `status: "submitted"`, `thread.cancelRun()`, flip to `ready`) fails on main — the thread still ends `user:undefined` — and with this change ends on an empty assistant message with `{ type: "incomplete", reason: "cancelled" }`. The new test pins it and fails without the fix.
- Two more tests pin the safeguards: the raced-first-chunk case marks the real message and appends nothing, and a `TypeError` from `stop()` synthesizes nothing.
- Full `@assistant-ui/ai-sdk` suite: 268 tests pass.

## Public surface

No API changes. Changeset: patch for `@assistant-ui/ai-sdk`.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6832?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.UdXkFnEjyaplNbtjqbDzT2xOhQ3iEU4iC2AlhvB96geUScxu4lu6dzDmkF4?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/ai-sdk-pre-chunk-cancel.md
+++ b/.changeset/ai-sdk-pre-chunk-cancel.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/ai-sdk": patch
+---
+
+fix: leave a cancelled assistant message when a run is stopped before the first chunk
--- a/packages/ai-sdk/src/runtime/useAISDKRuntime.test.ts
+++ b/packages/ai-sdk/src/runtime/useAISDKRuntime.test.ts
@@ -352,6 +352,104 @@ describe("useAISDKRuntime", () => {
     });
   });
 
+  it("synthesizes a cancelled assistant message when stopped before the first chunk", async () => {
+    const chat = createChatHelpers([
+      { id: "u1", role: "user", parts: [{ type: "text", text: "hi" }] },
+    ]);
+    chat.status = "submitted";
+    chat.stop = vi.fn(async () => {});
+
+    const { result, rerender } = renderHook(() => useAISDKRuntime(chat));
+
+    await act(async () => {
+      result.current.thread.cancelRun();
+    });
+    act(() => {
+      chat.status = "ready";
+      rerender();
+    });
+
+    await waitFor(() => {
+      const tail = result.current.thread.getState().messages.at(-1);
+      expect(tail?.role).toBe("assistant");
+      expect(tail?.content).toEqual([]);
+      expect(tail?.status).toMatchObject({
+        type: "incomplete",
+        reason: "cancelled",
+      });
+    });
+    expect(chat.messages).toHaveLength(2);
+  });
+
+  it("marks the assistant message that raced the abort instead of synthesizing one", async () => {
+    const chat = createChatHelpers([
+      { id: "u1", role: "user", parts: [{ type: "text", text: "hi" }] },
+    ]);
+    chat.status = "submitted";
+    let resolveStop!: () => void;
+    chat.stop = vi.fn(
+      () =>
+        new Promise<void>((resolve) => {
+          resolveStop = resolve;
+        }),
+    );
+
+    const { result, rerender } = renderHook(() => useAISDKRuntime(chat));
+
+    await act(async () => {
+      result.current.thread.cancelRun();
+      chat.setMessages([
+        ...chat.messages,
+        {
+          id: "assistant-raced",
+          role: "assistant",
+          parts: [{ type: "text", text: "par" }],
+        },
+      ]);
+      resolveStop();
+      await Promise.resolve();
+    });
+    act(() => {
+      chat.status = "ready";
+      rerender();
+    });
+
+    await waitFor(() => {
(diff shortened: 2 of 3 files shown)


Kinfe123:
Closing — the review is right and the premise was wrong.

I verified claude's claim empirically on clean main with the resync flushed (my original repro asserted before the `setTimeout(0)` resync ran, which is exactly why it appeared to show "no indication"): `cancelRun` during `submitted` already rolls the trailing user message back into the composer and empties the thread — prompt restored, ready to edit and resend. Probe result: `{"threadMsgs":[],"composerText":"my prompt","chatMsgs":[]}`.

So the state #6772 reports only occurs when the rollback is refused (composer already holding text, or a user message carrying non-text content), and this PR's synthesis would have replaced the deliberate restore-the-prompt affordance with an empty cancelled bubble and a cleared composer on every pre-first-chunk stop — a UX regression masquerading as a parity fix, with the stale-snapshot race on top.

Posting the corrected analysis on #6772; whether the refused-rollback cases should get the empty-bubble treatment (and in core rather than per-adapter) is a maintainer product call.

PR #6868, opened 2026-09-05
fix(assistant-stream): preserve positive exponent arguments

ephraimduncan:
## Problem

In assistant-stream 0.3.41, a streamed `{"amount":1e+2` followed by `3}` resolves the awaited amount as `1`, not `1e23`.

## Root cause

The numeric scanner treats `+` as the end of a number and marks the field complete too early.

## Change

Keep `+` inside the number scan so the argument reader waits for the remaining digits.

## Verification

`ToolCallReader.test.ts` includes the reproduction, which failed before the correction and passed after it. `./node_modules/.bin/vitest run src/core/tool/ToolCallReader.test.ts src/utils/json/parse-partial-json-object.test.ts` passed all 53 tests from `packages/assistant-stream`.

## Public surface

`assistant-stream`, with a patch changeset. No export or documentation changes.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6868?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.OYvoHXtTIfej2AmmKYlz5EK22mlFa5aCL3Zak-JucwCqvAqlf6U8tUo0uRE?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/stream-positive-exponent-fields.md
+++ b/.changeset/stream-positive-exponent-fields.md
@@ -1,0 +1,5 @@
+---
+"assistant-stream": patch
+---
+
+fix: Keep positive exponents incomplete until all argument digits arrive.
--- a/.changeset/stream-positive-exponent-fields.md
+++ b/.changeset/stream-positive-exponent-fields.md
@@ -2,4 +2,4 @@
 "assistant-stream": patch
 ---
 
-fix: Keep positive exponents incomplete until all argument digits arrive.
+fix: keep positive exponents incomplete until all argument digits arrive
--- a/packages/assistant-stream/src/core/tool/ToolCallReader.test.ts
+++ b/packages/assistant-stream/src/core/tool/ToolCallReader.test.ts
@@ -10,6 +10,17 @@ type Args = {
 const createReader = () => new ToolCallReaderImpl<Args, string>();
 
 describe("ToolCallArgsReader.get", () => {
+  it("waits for all digits of a positive exponent", async () => {
+    const reader = new ToolCallReaderImpl<{ amount: number }, string>();
+    const amount = reader.args.get("amount");
+
+    await reader.appendArgsTextDelta('{"amount":1e+2');
+    await reader.appendArgsTextDelta("3}");
+    await reader.finishArgsText();
+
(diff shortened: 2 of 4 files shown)


okisdev (approved):
approving. `+` was the last character missing from the number grammar in the scanner, so this closes the class rather than patching an instance.

worth spelling out why it bit harder than a truncated value would suggest. the `default:` arm in `INSIDE_NUMBER` does not just leave the number state, it also runs `path.pop()`, and that pop is what marks the field complete. so `{"amount":1e+2` did not merely read back as `1`, it resolved the awaited promise at `1`, and everything waiting on `args.get` had already moved on by the time the exponent digits arrived.

i swept every prefix of `{"amount":1e+23,"b":2}` against the same run for `1e-23` and they now agree at every cut: `1e` and `1e+` hold at partial 1, `1e+2` at partial 100, `1e+23` at partial 1e+23, and the field only goes complete on the comma. same for `[1e+23,4]` and `1E+5`.

i pushed one commit on top. the repro you wrote sits in `ToolCallReader.test.ts`, two layers above the defect, so i added the equivalent pair to the table in `parse-partial-json-object.test.ts`, next to the `// negative number whose digits have not arrived yet` block covering the sibling case. deleting `case "+"` from `fix-json.ts` fails test cases #41 and #42 there and passes with it restored; package suite is 512 passed, 22 skipped. also lowercased the changeset line to match the 38 of 39 that already are.

one for elsewhere: this is inherited. `vercel/ai` still has the same gap at `packages/ai/src/util/fix-json.ts:313-318` on its default branch (read, not run). worth an issue upstream.

PR #6869, opened 2026-09-05
fix(python): keep setdefault containers attached to state

ephraimduncan:
## Problem

In Python assistant-stream 0.0.36, writes through a new `draft.setdefault("items", [])` result change server state without stream updates.

## Root cause

The missing-key branch returns the plain default container instead of a state proxy.

## Change

The result uses the same live lookup as the existing-key branch, so later writes reach the client.

## Verification

`uv run --frozen --extra dev --extra langgraph pytest tests/test_state.py -q` passes 29 tests from `python/assistant-stream`.

The new `test_draft_setdefault_container_emits_later_writes` regression fails before the correction and passes after it. It inserts an empty list, clears the initial operations, and appends through the returned value.

## Public surface

Python `assistant-stream` state proxies. No export or documentation contract changes. The Python package is outside the npm changeset release system.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6869?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.znSdh2YIijFxmqzmajBhtxpKqn3XlTLtrfXjd1KxkdMXE62iz-e_rDoEhpY?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/python/assistant-stream/src/assistant_stream/state.py
+++ b/python/assistant-stream/src/assistant_stream/state.py
@@ -486,7 +486,7 @@ class StateProxy:
             return self[key]
 
         self[key] = default
-        return default
+        return self[key]
 
     # Unsupported operations that would be inefficient
     def insert(self, index: int, item: Any) -> None:
--- a/python/assistant-stream/tests/test_state.py
+++ b/python/assistant-stream/tests/test_state.py
@@ -155,6 +155,19 @@ def test_draft_list_append_emits_index_set() -> None:
     assert state.state["items"] == ["a", "b"]
 
 
+def test_draft_setdefault_container_emits_later_writes() -> None:
+    ops: list[dict[str, Any]] = []
+    state = AssistantState({})
+    draft = state.draft(ops.extend)
+    items = draft.setdefault("items", [])
+    ops.clear()
+
+    items.append("x")
+
+    assert ops == [{"type": "set", "path": ["items", "0"], "value": "x"}]
+    assert state.state == {"items": ["x"]}
+
+
 def test_draft_mutating_list_methods_raise() -> None:
     state = AssistantState({"items": ["a", "b"]})
     draft = state.draft(lambda _ops: None)


okisdev (approved):
approving. worth saying that this was worse than a lost update: `deep_apply` returns `op["value"]` by reference for a terminal set, so the plain `[]` handed back by `setdefault` was the object living in `AssistantState._state`, not a copy of it. appending to it mutated server state in place while emitting zero operations, so the server and the client diverged silently rather than the client merely falling behind. routing the return through `__getitem__` makes the missing-key branch identical to the existing-key branch two lines above, which is the right shape for it.

verified by running rather than reading. reverted `return self[key]` to `return default` and confirmed the new test fails there and passes with the fix, and confirmed the aliasing directly: pre-fix, `ops` was empty, `state.state` was `{'items': ['x']}`, and `state.state["items"] is items` was True. full python suite is 176 passed, 7 skipped. also probed the branches the regression test does not cover, all clean: primitive and `None` defaults still come back as raw values, an existing key holding `None` is not overwritten, `_ensure_no_proxy` still rejects a proxy passed as the default, nested `setdefault().setdefault()` emits the right path, and `type(self)` in `__getitem__` keeps the `state_proxy.StateProxy` subclass so `StateProxyJSONEncoder` still resolves it.

one thing still standing, which does not block: the aliasing itself is general, so `draft["cfg"] = cfg` followed by `cfg["a"] = 1` has the same silent shape. `setdefault` was just the one place the API handed you the reference, so this closes the entry point that invites it. the rest is a `deep_apply` change and does not belong in this diff; filed as #6910.

PR #6874, opened 2026-09-05
fix(langgraph): preserve streamed message snapshots

ephraimduncan:
## Problem

In `@assistant-ui/react-langgraph` 0.14.26, a new text chunk changes text in a message that the runtime already emitted. Consumers cannot retain an unchanged snapshot.

## Root cause

The merger copies the content array but writes into the shared text block.

## Change

Each text append replaces the last block instead of changing the previous block.

## Verification

The regression in `appendLangChainChunk.test.ts` appends `B` to an array-backed `A` message through string and array chunks. The new message must contain `AB`, and the previous message must contain `A`.

The two cases failed before the correction. All 14 tests passed after it with `node node_modules/vitest/vitest.mjs run src/appendLangChainChunk.test.ts --config <local-source-config>`. The temporary configuration resolves copied package dependencies to this worktree's source because their built exports were stale. No live provider test was necessary for this pure merger.

## Public surface

The behavior changes in `@assistant-ui/react-langgraph`. Exports and signatures stay unchanged. A patch changeset is included.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6874?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.B7f5dbK5wXRFBIsdJLNxYjIO264gCn2tGO_VSx4eIsbO9tl6zkX3aBGlh6s?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/quiet-text-snapshots.md
+++ b/.changeset/quiet-text-snapshots.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react-langgraph": patch
+---
+
+fix: keep previous message text unchanged when new stream chunks arrive
--- a/packages/react-langgraph/src/appendLangChainChunk.test.ts
+++ b/packages/react-langgraph/src/appendLangChainChunk.test.ts
@@ -351,3 +351,24 @@ describe("appendLangChainChunk updates-event partial_json", () => {
     expect(final).toBe(toolMessage);
   });
 });
+
+describe("appendLangChainChunk content snapshots", () => {
+  it.each(["B", [{ type: "text" as const, text: "B" }]])(
+    "keeps the previous text block unchanged for %j",
+    (content) => {
+      const before: AiMessage = {
+        id: "ai-1",
+        type: "ai",
+        content: [{ type: "text", text: "A" }],
+      };
+      const after = append(before, {
+        id: "ai-1",
+        type: "AIMessageChunk",
+        content,
+      });
+
+      expect(after.content).toEqual([{ type: "text", text: "AB" }]);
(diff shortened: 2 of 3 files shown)


ephraimduncan:
The accumulator reproduction confirms the caller-input mutation. Start with an AI message whose content is `[{ type: "text", text: "A" }]`. Retain `accumulator.getMessages()`, then append an `AIMessageChunk` with the same ID and content `"B"`. On the base branch, the caller input and retained message change to `AB`. On this branch, they remain `A`, and the current message contains `AB`. The array-chunk case gives the same results. All 44 focused merger and accumulator tests pass with workspace source exports. The image/text ordering issue also occurs on the base branch and remains outside this snapshot correction. The release note refers to retained snapshots, not the current message text.

okisdev:
superseded. #6896 already replaced both in-place text mutations with a fresh block object, and #6914 then rewrote the array text branch to merge by index through `mergeDefined`. the diff here is byte for byte what #6896 landed, which is also why this branch is conflicting.

i ran this pr's regression test against current main without the source change: both cases pass, `after.content` is `AB` and `before.content` stays `A`.

PR #6877, opened 2026-09-05
fix(assistant-stream): isolate replay byte buffers

ephraimduncan:
## Problem

In assistant-stream 0.3.41, producer or consumer buffer mutations change the bytes returned by later stream replays.

## Root cause

The in-memory store retains the input array and returns that same array to each reader.

## Change

Copy bytes on append and read. `new Uint8Array` also copies Node.js Buffer inputs, whose `slice` method shares memory.

## Verification

The new mutation regression in `InMemoryResumableStreamStore.test.ts` failed before the correction and passed after it. From `packages/assistant-stream`, `./node_modules/.bin/vitest run src/resumable/stores/InMemoryResumableStreamStore.test.ts src/resumable/ResumableStreamContext.test.ts` passed 46 tests. A direct `requireResume` scenario returned the original bytes after producer and consumer mutations.

## Public surface

`assistant-stream/resumable`, with a patch changeset. No exports changed.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6877?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.MrYpJRuQesh_g3cDvwPOF7-O48iZ5Ec7QuMi0BGS7UmJ6lud10583_zRgcg?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/resumable-stream-owned-bytes.md
+++ b/.changeset/resumable-stream-owned-bytes.md
@@ -1,0 +1,5 @@
+---
+"assistant-stream": patch
+---
+
+fix: Keep stored replay bytes separate from producer and consumer buffers.
--- a/apps/docs/content/docs/guides/resumable-stream-stores.mdx
+++ b/apps/docs/content/docs/guides/resumable-stream-stores.mdx
@@ -36,9 +36,13 @@ export interface ResumableStreamStore {
 
 `append(streamId, chunk)` adds a `Uint8Array` to the log under a fresh, monotonically increasing cursor. Callers expect the chunk to be observable to `read` before the promise resolves. Implementations should refresh the TTL on each call so a stream that is still actively producing does not expire mid-flight, and should reject when the stream is missing or already finalized.
 
+After `append` resolves, changes to the input buffer must not change stored bytes.
+
 `finalize(streamId, status, error?)` flips the stream into a terminal state. Pending and future `read` iterables drain buffered entries and then either complete (`"done"`) or throw with `error` (`"error"`). Implementations must make `finalize` idempotent: a duplicate call with the same status is a no-op, and the producer task may retry on transient errors.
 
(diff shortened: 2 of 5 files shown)


ephraimduncan:
The repair documents buffer ownership on `append` and `read`. It also corrects the custom-store example, which reproduced the same corruption. The copied example preserves the original bytes after producer and consumer mutations. All 57 focused tests passed.

okisdev (approved):
approving. the invariant is the right one: a replay log has to own its bytes, and the redis adapter already behaved this way (`read` rebuilds through `toBytes`, `append` awaits the pipeline before it resolves), so the in-memory store was the odd one out and swapping adapters silently changed aliasing semantics. that parity argument is what carries it for me more than the producer-side hazard, since the consumer side is reachable from our own shipped path: two readers of the same stream used to get the identical entry object out of `readFromStore`.

i checked out the head and ran `src/resumable/` (65 passed, 22 skipped), then reverted each copy on its own to confirm both are load bearing. with neither, the new test reports `['b']`, the producer mutation landing in the log; with only the append copy it reports `['c']`, the consumer writing straight into the stored entry; with only the read copy it is `['b']` again. so the test discriminates both halves instead of passing on one of them. the rest i read rather than ran: `redis-impl.ts` already satisfies the new `types.ts` wording, python's `append(chunk: bytes)` is immutable by type so there is no parity gap to chase, and api-surface strips JSDoc so the interface comment needs no snapshot regen.

one thing standing, not blocking: the changeset line reads `fix: Keep stored replay bytes...`, while every other entry in `.changeset/` starts lowercase and carries no trailing period.

PR #6878, opened 2026-09-05
fix(examples): release already-aborted pi event streams

ephraimduncan:
## Problem

A disconnected Pi events request could leave its subscription and heartbeat active if the request aborted before its route parameters resolved.

## Root cause

Abort signals do not replay an earlier abort event to a new listener.

## Change

The route checks the signal after listener registration and uses the existing cleanup. Disconnect still does not cancel the Pi run.

## Verification

A focused Bun check called the real route with deferred parameters, aborted the request, then resolved the parameters. Before the correction, one subscription and heartbeat remained. After the correction, neither remained. The Pi client boundary and timers were isolated. The commit hook passed Oxlint and Oxfmt.

Minimal reproduction sequence:

```ts
const params = Promise.withResolvers<{ threadId: string }>();
const response = GET({
  signal: controller.signal,
  nextUrl: new URL("http://localhost/api/pi/threads/thread-1/events"),
} as NextRequest, { params: params.promise });
controller.abort();
params.resolve({ threadId: "thread-1" });
await (await response).body?.cancel();
```

## Public surface

Only the private Pi example changes. No published package or API changes.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6878?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.-brBKsLYUr-uh86qYurf2B_J5kBsWW53P-b-BufQWz-CtbpEFWxdrKqgymc?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/examples/with-pi/app/api/pi/threads/[threadId]/events/route.ts
+++ b/examples/with-pi/app/api/pi/threads/[threadId]/events/route.ts
@@ -47,6 +47,7 @@ export async function GET(req: NextRequest, { params }: Context) {
     { includeSnapshot: req.nextUrl.searchParams.get("snapshot") !== "false" },
   );
   req.signal.addEventListener("abort", cleanup, { once: true });
+  if (req.signal.aborted) cleanup();
 
   return new Response(stream.readable, {
     headers: {


okisdev (approved):
approved. the leak is real and i reproduced it against the actual route rather than reading it: with the pi client stubbed and setInterval instrumented, an abort that lands while `await params` is still pending leaves unsubscribes at 0 and one live interval, and with this line both go to zero. the heartbeat is the worse half, since a 20s interval that never clears keeps the writer and its closure resident for the life of the process.

the `.aborted` recheck is also what the rest of the repo already does at every registration point (host.ts, voice.ts, useDataStreamRuntime.ts, McpServerResource.ts, extensionUi.ts, toolResultStream.ts, InMemoryResumableStreamStore.ts), and voice.ts uses this same register-then-check ordering, so this was the last unguarded one. no changeset is right, with-pi is private.

the branch is behind main, but nothing has touched this file since and the required checks are not strict, so it goes in as is.

PR #6880, opened 2026-09-05
fix(core): match complete attachment extensions

ephraimduncan:
## Problem

Attachment adapters reject `backup.tar.gz` for `.tar.gz` and accept an extensionless file named `png` for `.png`. This affects `@assistant-ui/core` 0.3.17.

Minimal reproduction from the repository root with Bun:

```sh
bun --eval '
import { fileMatchesAccept } from "./packages/core/src/adapters/attachment.ts";
assert.equal(fileMatchesAccept({ name: "backup.tar.gz", type: "" }, ".tar.gz"), true);
assert.equal(fileMatchesAccept({ name: "png", type: "" }, ".png"), false);
'
```

## Root cause

The matcher keeps only the final dot-separated segment and adds a dot, even when the filename has no extension.

## Change

Extension rules compare against the complete filename suffix without case sensitivity. Composer acceptance and composite adapter selection use the same matcher. MIME rules remain unchanged.

## Verification

The two regression cases failed before the correction. All eight attachment tests pass after the correction:

```sh
cd packages/core
./node_modules/.bin/vitest run src/tests/attachment-adapters.test.ts
```

Bun assertions also passed for suffix boundaries, MIME rules, wildcards, and real composite adapter add/send/remove operations. Changed-file `oxlint` and `oxfmt` completed successfully.

## Public surface

The PR includes a patch changeset for `@assistant-ui/core`. Exports and signatures remain unchanged. No documentation, API-reference, or template changes are necessary.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6880?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.2CvGcEZusM4ZgKp1hc85zDR7QagAKA49gdaUyPIe4ZcSmChez9e76U0BDhQ?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/match-complete-attachment-extensions.md
+++ b/.changeset/match-complete-attachment-extensions.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/core": patch
+---
+
+fix: match attachment extensions against the complete filename suffix, so `.tar.gz` accepts `backup.tar.gz` and `.png` rejects extensionless `png`.
--- a/packages/core/src/adapters/attachment.ts
+++ b/packages/core/src/adapters/attachment.ts
@@ -142,11 +142,11 @@ export function fileMatchesAccept(
     .split(",")
     .map((type) => type.trim().toLowerCase());
 
-  const fileExtension = `.${file.name.split(".").pop()!.toLowerCase()}`;
+  const fileName = file.name.toLowerCase();
   const fileMimeType = file.type.split(";", 1)[0]!.trim().toLowerCase();
 
   for (const type of allowedTypes) {
-    if (type.startsWith(".") && type === fileExtension) {
+    if (type.startsWith(".") && fileName.endsWith(type)) {
       return true;
     }
 
--- a/packages/core/src/tests/attachment-adapters.test.ts
+++ b/packages/core/src/tests/attachment-adapters.test.ts
@@ -9,10 +9,17 @@ import {
 
 describe("fileMatchesAccept", () => {
   it.each([
-    ["BACKUP.TAR.GZ", ".tar.gz", true],
-    ["png", ".png", false],
-  ])("matches the full suffix of %s", (name, accept, expected) => {
-    expect(fileMatchesAccept({ name, type: "" }, accept)).toBe(expected);
+    ["BACKUP.TAR.GZ", ".tar.gz"],
+    [".env", ".env"],
+  ])("accepts %s for %s", (name, accept) => {
+    expect(fileMatchesAccept({ name, type: "" }, accept)).toBe(true);
+  });
+
+  it.each([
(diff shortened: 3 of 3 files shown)


okisdev (approved):
approving. the one-line change is right because it is not a new rule, it is the standard one: `fileName.toLowerCase().endsWith(validType)` is byte-for-byte what `attr-accept` does, and that package is already in this lockfile at 2.2.5 under react-dropzone. so core's matcher was the odd one out, rather than `.tar.gz` being an exotic ask. that is a reading of `node_modules/.pnpm/attr-accept@2.2.5/node_modules/attr-accept/dist/es/index.js`, not something i ran.

the objection a reviewer reaches for first is that `endsWith` must over-match, that `.js` would swallow `foo.mjs`. it cannot, because the branch is guarded by `type.startsWith(".")`, so every accept entry carries its own left boundary. i ran the old and new matchers side by side over 25 cases and exactly four diverge:

```
backup.tar.gz  vs .tar.gz      old=false new=true
BACKUP.TAR.GZ  vs .tar.gz      old=false new=true
png            vs .png         old=true  new=false
Dockerfile     vs .dockerfile  old=true  new=false
```

`foo.mjs` vs `.js`, `foo.mp4` vs `.p4` and `.4`, and `archive.tgz` vs `.gz` stay false on both sides.

i pushed one commit on top, tests only. your `it.each` title rendered as `matches the full suffix of png` over a row asserting a non-match, so it is now split into `accepts %s for %s` and `rejects %s for %s`, which is also how `normalizePartStatus.test.ts` and `data-url.test.ts` phrase theirs. and i pinned the two properties that make `endsWith` safe instead of leaving them to the argument above: `.env` against `.env` still accepts, `foo.mjs` against `.js` still rejects. the dotfile one earns its test because `.env` sits in react-opencode's `DEFAULT_ACCEPT`, so a regression there would have been real; it does not regress, but nothing was holding it.

one thing to know rather than act on: the extensionless half of this is a class, not one filename. anything whose whole name equals the extension now flips to rejected, so `Dockerfile` for `.dockerfile` and `env` for `.env` travel with `png`. browsers and attr-accept agree, so patch is the right size, but it is the one user-visible tightening in here.

PR #6882, opened 2026-09-05
fix(langgraph): keep mixed content blocks in order

ephraimduncan:
## Problem

In `@assistant-ui/react-langgraph` 0.14.26, text after an image can move before that image when the two blocks arrive in one chunk. The corrected result keeps the text after the image.

Minimal reproduction:

```ts
import { appendLangChainChunk } from "@assistant-ui/react-langgraph";

const image = {
  type: "image_url" as const,
  image_url: "data:image/png;base64,AA==",
};
const result = appendLangChainChunk(
  { id: "a", type: "ai", content: "Before" },
  {
    id: "a",
    type: "AIMessageChunk",
    content: [image, { type: "text", text: "After" }],
  },
);
```

Before the correction, `result.content` contains text `BeforeAfter`, then the image. The expected result is text `Before`, the image, then text `After`.

## Root cause

The loop uses one tail index for the entire chunk. After it appends an image, that index still points to the text before the image.

## Change

The shared append function calculates the tail index for each block. No runtime API changes are necessary.

## Verification

The regression failed before the correction and passed after it. The append and accumulator suites passed all 43 tests with:

```sh
pnpm --filter @assistant-ui/react-langgraph test src/appendLangChainChunk.test.ts src/LangGraphMessageAccumulator.test.ts --config /tmp/langgraph-content-order.vitest.mjs
```

The temporary configuration used the package test settings and mapped `@assistant-ui/react-langchain/converter` to its workspace source because its local build was missing. It is not part of this PR.

A direct Bun call also passed assertions for image order and adjacent text merging. Changed-file `oxfmt --check` and `oxlint` checks passed.

## Public surface

The PR includes a patch changeset for `@assistant-ui/react-langgraph`. Exports, documentation, API reference, and templates are unchanged.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6882?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.ZgfgpvVaYnWwYJCYm3vvCN8n9liXuLS6-BEEK4-ao-s6XIBm8oN5sTEE2Yc?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/langgraph-mixed-content-order.md
+++ b/.changeset/langgraph-mixed-content-order.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react-langgraph": patch
+---
+
+fix: keep text and images in order when a streamed chunk contains multiple content blocks
--- a/packages/react-langgraph/src/appendLangChainChunk.test.ts
+++ b/packages/react-langgraph/src/appendLangChainChunk.test.ts
@@ -351,3 +351,26 @@ describe("appendLangChainChunk updates-event partial_json", () => {
     expect(final).toBe(toolMessage);
   });
 });
+
+describe("appendLangChainChunk content order", () => {
+  it("keeps text after an image in the same chunk", () => {
+    const image = {
+      type: "image_url" as const,
+      image_url: "data:image/png;base64,AA==",
+    };
+    const result = append(
+      { id: "ai-1", type: "ai", content: "Before" },
+      {
+        id: "ai-1",
+        type: "AIMessageChunk",
+        content: [image, { type: "text", text: "After" }],
+      },
+    );
+
+    expect(result.content).toEqual([
+      { type: "text", text: "Before" },
+      image,
+      { type: "text", text: "After" },
+    ]);
+  });
+});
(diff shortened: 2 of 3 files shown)


okisdev:
closing: superseded by #6914, which moved the `lastIndex` computation inside the loop as part of the text-block-by-index rewrite. main already passes this PR's regression test with no source change.

merging as-is would break the build. git auto-merges it without a conflict, but the result declares `const lastIndex` twice in the same block scope:

```
[PARSE_ERROR] Identifier `lastIndex` has already been declared
   src/appendLangChainChunk.ts:149:13
```

the ordering behavior itself is covered on main by "keeps text after intervening content and joins adjacent text deltas", which asserts the same thing with thinking and file as the intervening blocks.

the diagnosis was correct, it just landed twice.

PR #6886, opened 2026-09-05
fix(opencode): count patch files by full path

ephraimduncan:
## Problem

In the `with-opencode` example (version `0.0.0`), a patch for `src/index.ts` and `tests/index.ts` shows `index.ts` instead of `2 files`.

This input to `getPatchInfo` reproduces the incorrect count:

```ts
getPatchInfo(`*** Begin Patch
*** Update File: src/index.ts
-old
+new
*** Update File: tests/index.ts
-before
+after
*** End Patch`).files.length;
```

The result was `1` instead of `2`.

## Root cause

The parser removed directory names before it removed duplicate paths. Distinct files with the same name became one entry.

## Change

The parser removes duplicate full paths before it shortens display names. Repeated paths still count once, and the summary shows `2 files` for the reproduction.

## Verification

A Bun assertion against the real parser failed before the correction and passed after it. Focused assertions also passed for repeated trimmed paths, Add/Delete headers, empty input, and line counts.

Server renders of the actual `ApplyPatchDiff` and `ApplyPatchInline` components each contained `2 files`. This was not a live browser check. The example has no existing test suite, so the checks used temporary scripts without module mocks.

Changed-file `oxlint` and `oxfmt` completed successfully.

## Public surface

Only the private example changes. Published packages, exports, documentation, and templates remain unchanged. No changeset is necessary.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6886?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.CbQrrdkk3jmd9VF1IplmS6hqSvQfGT-3OXr0NNdd_3uabOeHyFNzVxqv4Ro?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/examples/with-opencode/components/tools/tool-ui-apply-patch.tsx
+++ b/examples/with-opencode/components/tools/tool-ui-apply-patch.tsx
@@ -12,6 +12,7 @@ import { cn } from "@/lib/utils";
 import { PatchDiff } from "@pierre/diffs/react";
 import {
   ToolStatusIcon,
+  formatPatchFiles,
   getPatchInfo,
   isCancelledToolStatus,
   str,
@@ -178,9 +179,7 @@ export const ApplyPatchDiff: ToolCallMessagePartComponent = memo(
               <span className="font-medium">{toolName}</span>
               {patchInfo.files.length > 0 && (
                 <span className="opacity-60">
-                  {patchInfo.files.length === 1
-                    ? patchInfo.files[0]
-                    : `${patchInfo.files.length} files`}
+                  {formatPatchFiles(patchInfo.files)}
                 </span>
               )}
               {(patchInfo.added > 0 || patchInfo.removed > 0) && !isRunning && (
--- a/examples/with-opencode/components/tools/tool-ui-inline.tsx
+++ b/examples/with-opencode/components/tools/tool-ui-inline.tsx
@@ -6,6 +6,7 @@ import { cn } from "@/lib/utils";
 import {
   ToolStatusIcon,
   basename,
+  formatPatchFiles,
   getPatchInfo,
   isCancelledToolStatus,
   str,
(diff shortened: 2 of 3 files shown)


okisdev (approved):
approving. the root cause read is right: `basename` running before `unique` collapses `src/index.ts` and `tests/index.ts` into one entry, and moving the dedupe onto full paths is the correct place to fix it.

i pushed one commit on top. your version leaves `files` holding duplicate basenames (`["index.ts", "index.ts"]` for the repro), which is harmless today because both call sites only read `.length` and `[0]`, but the label ternary was copy pasted into `ApplyPatchDiff` and `ApplyPatchInline`, and that duplication is why a single ordering mistake was able to surface in two renderers at once. so `getPatchInfo` now returns unique full paths and `formatPatchFiles` owns the label, called from both.

what i executed: the pure parser extracted from `main` and from head into a node script, comparing the rendered label across six inputs. the repro moves from `main="index.ts"` to `head="2 files"`, and every other case is byte identical, including the example patch out of upstream's own `apply_patch.txt` (`main="3 files"`, `head="3 files"`, `+2/-1`). the two renderers themselves i only read; there was no browser pass.

one thing worth not touching later: the header regex deliberately does not match `*** Move to:`. reading `sst/opencode@v1.18.25` `packages/opencode/src/tool/apply_patch.ts`, a rename is a single hunk of `type: "move"` carrying both `filePath` and `movePath`, so matching the `Update File:` line alone is what keeps a rename counted as one file. folding `Move to:` into that alternation would quietly double the count for every rename, and it looks like an omission until you check upstream.

separately, and not for this PR: the comment at `tool-ui-apply-patch.tsx:21` and the local `parseClaudePatchBlocks` name credit the wrong tool for this format. it is opencode's `apply_patch`, and the misattribution will follow anyone who copies the example.

PR #6888, opened 2026-09-05
fix(react): preserve nested viewport scroll behavior

ephraimduncan:
## Problem

a scroll control rendered outside `ThreadPrimitive.Viewport` jumps to the bottom instantly even when it asks for smooth scrolling. reproduced on `@assistant-ui/react` 0.15.18:

```tsx
  <ThreadPrimitive.Viewport
    autoScroll={false}
    scrollToBottomOnInitialize={false}
    style={{ height: 240, overflowY: "auto" }}
  >
    <div style={{ height: 1800 }}>Thread content</div>
  </ThreadPrimitive.Viewport>
  <ThreadPrimitive.ScrollToBottom behavior="smooth">
    Scroll to bottom
  </ThreadPrimitive.ScrollToBottom>
</AssistantRuntimeProvider>
```

## Root cause

`AssistantRuntimeProvider` mounts a viewport store, and `ThreadPrimitive.Viewport` mounts a second one nested inside it. a control outside the viewport therefore reaches the outer store, which forwards the request down to the inner store in `ThreadViewportProvider`. that forwarding callback dropped its config argument, so the inner `scrollToBottom` fell back to its `"auto"` default. the forwarding has been there since #1551.

## Change

forward the config through. outer and inner both default `behavior` to `"auto"`, so a call that passes no behavior is unchanged and only an explicit behavior is newly preserved. a control inside the viewport already talked to the inner store directly, so nothing about that path moves. subscription cleanup and public APIs are untouched.

## Verification

```sh
pnpm --filter @assistant-ui/react exec vitest run src/primitives/thread/useThreadViewportAutoScroll.test.tsx
```

10 passed. the new regression asserts the viewport's `scrollTo` receives `behavior: "smooth"`; reverting the one-line change makes exactly that test fail with `"auto"` and leaves the other 9 green.

## Public surface

`@assistant-ui/react` gets a patch changeset. no exports, API signatures, documentation claims, or templates change.

<a href="https://rupic.dev/s/s2.9Abn3qpv5RAMXAsrJ8wllGTcVNBwBAE9QXivbMD_bhwAZgdiFah2MIP2wxI?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/nested-viewport-scroll-behavior.md
+++ b/.changeset/nested-viewport-scroll-behavior.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react": patch
+---
+
+fix: preserve the requested scroll behavior across nested viewport providers
--- a/packages/react/src/context/providers/ThreadViewportProvider.tsx
+++ b/packages/react/src/context/providers/ThreadViewportProvider.tsx
@@ -25,8 +25,8 @@ const useThreadViewportStoreValue = (options: ThreadViewportStoreOptions) => {
 
   // Forward scrollToBottom from outer viewport to inner viewport
   useEffect(() => {
-    return outerViewport?.getState().onScrollToBottom(() => {
-      store.getState().scrollToBottom();
+    return outerViewport?.getState().onScrollToBottom((config) => {
+      store.getState().scrollToBottom(config);
     });
   }, [outerViewport, store]);
 
--- a/packages/react/src/primitives/thread/useThreadViewportAutoScroll.test.tsx
+++ b/packages/react/src/primitives/thread/useThreadViewportAutoScroll.test.tsx
@@ -1,6 +1,13 @@
 // @vitest-environment jsdom
 
-import { act, cleanup, render, screen, waitFor } from "@testing-library/react";
+import {
+  act,
+  cleanup,
+  fireEvent,
+  render,
+  screen,
+  waitFor,
+} from "@testing-library/react";
 import {
   afterAll,
   afterEach,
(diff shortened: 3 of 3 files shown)


okisdev (approved):
approving. the diagnosis lands on the right seam. the nesting is easy to miss: `AssistantRuntimeProvider` already mounts a viewport store, so a control rendered outside `ThreadPrimitive.Viewport` only ever talks to that outer store and reaches the real viewport through the forwarding callback in `ThreadViewportProvider`, which was swallowing its config. fixing it there instead of at the button is what makes this a one-liner.

i verified rather than read it. reverting just that line makes the new regression fail with `behavior: "auto"` and leaves the other nine green, so the test is pinned to the mechanism and not to its surroundings. i also confirmed this is the only `onScrollToBottom` forwarding site in the repo (no parallel copy in react-native or react-ink), and that the change is inert for callers that pass no behavior, since outer and inner both default to `"auto"`.

two small things pushed onto the branch: dropped `expect(viewport.scrollTop).toBe(getMaxScrollTop(viewport))`, which the file's own `scrollTo` mock makes true by construction once the spy assertion passes, and put the new import back in alphabetical order.

i also rewrote the description. it claimed the repo needs a temporary vitest config aliasing `zustand` because of a missing dependency link, and that is not the case. `vitest.config.ts` imports `@assistant-ui/vite`, which is a workspace package, so it just needs to be built first: `pnpm --filter @assistant-ui/react... build` and then the plain vitest command runs the file as-is. worth knowing for the next one.

PR #6890, opened 2026-09-05
fix(agent-launcher): keep selected skill with an empty prompt

ephraimduncan:
## Problem

A direct `launch` call with an empty prompt ignores the selected skill. Claude starts without the requested skill command.

Minimal reproduction at base `928c580f4132496ee6ae9dc5a64fe44ca4bfd1b7` (`@assistant-ui/agent-launcher` version `0.1.14`), run with Bun from the repository root:

```sh
bun --eval 'import { launch } from "./packages/agent-launcher/src/launch.ts"; launch({ pluginDir: "/tmp/plugin", skillName: "assistant-ui", prompt: "", dry: true });'
```

The output omits `/assistant-ui`. The corrected output is `claude /assistant-ui --plugin-dir /tmp/plugin`.

The shipped CLI reaches the same path. `agent` declares a required variadic `<prompt...>`, so `aui agent ""` parses as `[""]` and joins to an empty prompt before calling `launch`.

## Root cause

The skill condition requires a nonempty prompt, although the public API accepts an empty string.

## Change

The selected skill becomes the initial command even without prompt text. Nonempty prompts keep their existing behavior.

## Verification

The reproduction and regression failed before the correction and passed after it. A second regression pins the nonempty branch of the same condition, so both command forms the skill can produce are now asserted. All four tests passed with `./node_modules/.bin/vitest run src/launch.test.ts` from `packages/agent-launcher`.

A temporary local executable confirmed subprocess arguments for skill-only, skill-plus-prompt, prompt-only, and empty inputs. No model calls occurred. Scoped Oxlint and formatting completed successfully.

## Public surface

The PR includes a patch changeset for `@assistant-ui/agent-launcher`. Exports and types remain unchanged. Documentation and templates require no changes.

<a href="https://rupic.dev/s/s2.rh_ZKEPmeHAFh1cs56VovzrRbKdoDDaatP3rIodqIxRmci2FAMKMX7oca-k?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6890?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/launcher-empty-prompt-skill.md
+++ b/.changeset/launcher-empty-prompt-skill.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/agent-launcher": patch
+---
+
+fix: keep the selected skill when the prompt is empty
--- a/packages/agent-launcher/src/launch.test.ts
+++ b/packages/agent-launcher/src/launch.test.ts
@@ -27,6 +27,21 @@ describe("launch", () => {
     }
   });
 
+  it("invokes the selected skill when the prompt is empty", () => {
+    const log = vi.spyOn(console, "log").mockImplementation(() => {});
+
+    launch({
+      pluginDir: "/tmp/plugin",
+      skillName: "assistant-ui",
+      prompt: "",
+      dry: true,
+    });
+
+    expect(log).toHaveBeenCalledWith(
+      "claude /assistant-ui --plugin-dir /tmp/plugin",
+    );
+  });
+
   it("falls back to a signal exit code if re-raising does not terminate", () => {
     mocks.sync.mockReturnValue({
       pid: 123,
--- a/packages/agent-launcher/src/launch.test.ts
+++ b/packages/agent-launcher/src/launch.test.ts
@@ -27,6 +27,21 @@ describe("launch", () => {
     }
   });
 
+  it("passes the prompt to the selected skill", () => {
+    const log = vi.spyOn(console, "log").mockImplementation(() => {});
+
+    launch({
+      pluginDir: "/tmp/plugin",
+      skillName: "assistant-ui",
+      prompt: "add a thread",
+      dry: true,
+    });
+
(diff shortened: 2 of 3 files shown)


okisdev (approved):
approved. `prompt` is a required `string` in `LaunchOptions`, so the empty string is the only way to express "skill, no prompt", and dropping the skill on it was a hole in the public surface rather than an edge case nobody meant to reach.

i checked both directions instead of trusting the added test: with `launch.ts` reverted to `928c580f4` the new case fails on `Number of calls: 1`, `/assistant-ui` missing from the printed command, and passes on your change. reachability is also wider than the description claimed, so i extended it there: `agent` declares `<prompt...>` as a required variadic and commander parses `aui agent ""` to `[""]`, which joins to an empty prompt before `launch` ever sees it. i ran that parse rather than inferring it, so the shipped CLI hits this too, not only direct API consumers.

one commit from me on top. the condition you touched now has two branches and only the empty one was pinned, and nothing in the file asserted `/skill prompt` before or after. the added case covers it and pins the dry-run quoting along the way, since `/assistant-ui` prints bare while `/assistant-ui add a thread` comes back double-quoted from the `escaped` map; rewriting that quoting now breaks a test.

not in scope here: `aui agent` with no arguments is still rejected by commander before `launch` runs, so the bare command remains the one path that cannot open the skill promptless.

PR #6892, opened 2026-09-05
fix(python): use default subgraph state for null tool artifacts

ephraimduncan:
## Problem

A tool message with `artifact=None` stops subgraph processing with `TypeError` when the caller requests `subgraph_state`. Without a field name, the helper returns `None` instead of the supplied default.

Reproduction on `assistant-stream 0.0.36`:

```sh
cd assistant-ui
git checkout 928c580f4132496ee6ae9dc5a64fe44ca4bfd1b7
cd python/assistant-stream
uv sync --all-extras --frozen
uv run --frozen python - <<'PY'
from types import SimpleNamespace
from langchain_core.messages import ToolMessage
from assistant_stream import get_tool_call_subgraph_state

controller = SimpleNamespace(state={"messages": [
    ToolMessage(content="", tool_call_id="c1", artifact=None).model_dump()
]})
state = get_tool_call_subgraph_state(
    controller, ("tools:task1",), "tools", {},
    artifact_field_name="subgraph_state",
)
assert state == {}
PY
```

## Root cause

The helper initializes an absent artifact key but leaves a null value unchanged.

## Change

Null artifacts use the existing default-state assignment. Non-null artifacts retain their behavior.

## Verification

The regression uses real `ToolMessage` and `RunController` objects. The two field-selection cases raise `TypeError` before the fix and pass after it. The tests also confirm that subsequent events update the stored artifact.

From `python/assistant-stream`:

```sh
PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 uv run --frozen pytest -p anyio.pytest_plugin tests/test_langgraph.py -k null_tool_artifact -q -o faulthandler_timeout=15
PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 uv run --frozen pytest -p anyio.pytest_plugin tests/test_langgraph.py -q
```

The focused regression passes twice, and all 12 module tests pass. The direct public-export call returns `{}` after the fix. The default pytest invocation timed out, so these runs load only the AnyIO plugin.

## Public surface

Only the Python `assistant-stream` behavior changes. Exports, signatures, and documentation remain unchanged. No JavaScript package or changeset is affected.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6892?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.1ZYGmUStbLTC-7QWT9XglXFapGu_eNg-wYCIeFMhnJiA6LMRqch4X2R7AUk?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/python/assistant-stream/src/assistant_stream/modules/langgraph.py
+++ b/python/assistant-stream/src/assistant_stream/modules/langgraph.py
@@ -267,7 +267,7 @@ def get_tool_call_subgraph_state(
                 artifact = last_message["artifact"]
 
                 if artifact_field_name:
-                    if artifact_field_name not in artifact:
+                    if artifact.get(artifact_field_name) is None:
                         artifact[artifact_field_name] = default_state
                     return artifact[artifact_field_name]
                 else:
--- a/python/assistant-stream/src/assistant_stream/modules/langgraph.py
+++ b/python/assistant-stream/src/assistant_stream/modules/langgraph.py
@@ -262,7 +262,7 @@ def get_tool_call_subgraph_state(
             # Check if last message is already a ToolMessage
             if last_message["type"] == "tool":
                 # Last message is already a ToolMessage, extract and return artifact field
-                if "artifact" not in last_message:
+                if last_message.get("artifact") is None:
                     last_message["artifact"] = {} if artifact_field_name else default_state
                 artifact = last_message["artifact"]
 
--- a/python/assistant-stream/tests/test_langgraph.py
+++ b/python/assistant-stream/tests/test_langgraph.py
@@ -256,3 +256,31 @@ async def test_null_tool_artifact_uses_default_subgraph_state(
     if artifact_field_name:
         artifact = artifact[artifact_field_name]
     assert artifact["answer"] == "42"
+
+
+@pytest.mark.anyio
+@pytest.mark.parametrize(
+    ("artifact", "artifact_field_name", "expected"),
+    [
+        ({"answer": "kept"}, None, {"answer": "kept"}),
+        ({"subgraph_state": {"answer": "kept"}}, "subgraph_state", {"answer": "kept"}),
+        ({"subgraph_state": {}}, "subgraph_state", {}),
+    ],
+)
+async def test_existing_tool_artifact_survives_default_state(
+    artifact, artifact_field_name, expected
+) -> None:
+    controller = RunController(
+        asyncio.Queue(),
+        {"messages": [ToolMessage(content="", tool_call_id="c1", artifact=artifact).model_dump()]},
+    )
+
+    state = get_tool_call_subgraph_state(
+        controller,
+        ("tools:task1",),
+        "tools",
+        {"answer": "default"},
+        artifact_field_name=artifact_field_name,
+    )
+
+    assert state == expected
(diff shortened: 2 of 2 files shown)


ephraimduncan:
The null named-field case also needs the default state. I corrected it and added the case to the existing regression test. The case raised TypeError before the repair. All 13 LangGraph tests pass after the repair. Seven public streaming scenarios also pass, including empty and populated states. The tool-selection behavior remains unchanged.

okisdev:
pushed one more test commit on top.

the three parametrized null cases all pass under a truthiness check too (`if not artifact.get(field)`), so nothing pinned the "non-null artifacts retain their behavior" half of the change. added a sibling parametrize covering an existing nested value, an existing artifact with no field name, and `{"subgraph_state": {}}`. that last one is the load bearing case: it fails on a truthiness mutant and passes on `is None`.

separately, worth dropping the plugin autoload note from the description. `uv run pytest`, which is exactly what the python-tests workflow runs, finishes the full suite here in 3.2s (181 passed, 7 skipped), so the timeout looks local to your setup rather than something the repo needs to work around.

on the fix itself: `ToolMessage.model_dump()` always emits the `artifact` key, so `"artifact" not in last_message` never fired for a pydantic serialized message. that guard was dead for every real tool result, which is why `python/assistant-transport-backend-langgraph/main.py:562` hit it. verified both directions against the merge base: the `subgraph_state` path raised `TypeError: argument of type 'NoneType' is not a container or iterable`, and the field-less path returned `None` and then blew up downstream in `append_langgraph_event` with `TypeError: 'NoneType' object does not support item assignment`.

okisdev (approved):
approving. the fix is at the right layer, and the reason is worth writing down because the diff hides it: `"artifact" not in last_message` was dead code. i ran `ToolMessage(content="", tool_call_id="c1").model_dump()` against langchain-core 1.6.1 and `artifact` comes back in the key set with value `None`, so that guard never fired for any pydantic serialized tool message. every real tool result took the unguarded path, which is why `python/assistant-transport-backend-langgraph/main.py:562` walked straight into it.

i executed both directions against the merge base rather than reading them. on `928c580f4` the `subgraph_state` path raises `TypeError: argument of type 'NoneType' is not a container or iterable`, and the `artifact_field_name=None` path is worse than it looks: it returns `None` quietly and then `append_langgraph_event` dies with `TypeError: 'NoneType' object does not support item assignment`. both return `{'answer': 'pending'}` on your head.

i pushed one test commit. your three null cases all still pass if `is None` is swapped for `if not artifact.get(field)`, so nothing pinned the half of the change that keeps a populated artifact intact. the case that does the work is `{"subgraph_state": {}}`, the only one that fails that truthiness mutant, which is the exact edit a future reader makes while simplifying this. separately, `uv run pytest` is what the python-tests workflow actually runs and it finishes the suite here in 3.2s (181 passed, 7 skipped), so the plugin autoload workaround reads as local to your machine and is worth dropping from the description.

one thing i am deliberately leaving: the helper still cannot tell the synthetic streaming `ToolMessage` it creates (the one carrying `additional_kwargs={"streaming": True}`) from a genuine completed tool result, so it now claims either one as the subgraph state carrier. that only became visible because your fix stopped it crashing there. nothing is lost, since a null artifact holds nothing, but it belongs in its own change rather than this one.

PR #6895, opened 2026-09-05
fix(core): remove pending attachments without contentType

ephraimduncan:
## Problem

A pending image attachment can remain in the composer with an error when the custom adapter omits `contentType`. Removal must use the same adapter that accepted the file.

This occurs in `@assistant-ui/core` 0.3.17 at base `928c580f4132496ee6ae9dc5a64fe44ca4bfd1b7`.

<details>

Clone the repository and install its dependencies:

```sh
cd assistant-ui
git checkout 928c580f4132496ee6ae9dc5a64fe44ca4bfd1b7
```

Save this as `repro.mjs` in the repository root. Run `bun repro.mjs`:

```js
import assert from 'node:assert/strict';
import { CompositeAttachmentAdapter } from './packages/core/src/adapters/attachment.ts';
let removed;
const imageAdapter = {
  accept: 'image/*',
  async add({ file }) {
    return {
      id: 'image-1', type: 'image', name: file.name, file,
      status: { type: 'requires-action', reason: 'composer-send' }
    };
  },
  async send(attachment) {
    return { ...attachment, status: { type: 'complete' }, content: [] };
  },
  async remove(attachment) { removed = attachment.id; }
};
const composite = new CompositeAttachmentAdapter([imageAdapter]);
const pending = await composite.add({
  file: new File(['image bytes'], 'photo.png', { type: 'image/png' })
});
await composite.remove(pending);
assert.equal(removed, 'image-1');
```

Before the correction, removal throws `No matching adapter found for attachment`. After the correction, the assertion passes.

</details>

## Root cause

Add and send select an adapter from the original `File`. Removal instead uses attachment metadata, where `contentType` is optional. An empty MIME type does not match `image/*`.

## Change

Removal uses the original `File` when present. Completed attachments without a `File` retain the name and content-type fallback.

## Verification

The reproduction and pending-removal regression failed before the correction and passed after it. The completed-attachment fallback also passes.

Focused checks passed:

```sh
pnpm --filter @assistant-ui/core test src/adapters/attachment.test.ts src/tests/attachment-adapters.test.ts
pnpm exec oxlint packages/core/src/adapters/attachment.ts packages/core/src/adapters/attachment.test.ts
pnpm exec oxfmt --check packages/core/src/adapters/attachment.ts packages/core/src/adapters/attachment.test.ts
```

All 15 tests passed. Verification covers the adapter at runtime, not a browser interface.

## Public surface

The PR includes a patch changeset for `@assistant-ui/core`. Exports, signatures, documentation, API-reference output, and templates remain unchanged.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6895?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.8T30wVR_aAOpkvvJIg6Mu9kqVeq4KzgJzJ5-1UZUkEYA1pHcgM7UlOVBWjA?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/fix-pending-attachment-removal-file.md
+++ b/.changeset/fix-pending-attachment-removal-file.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/core": patch
+---
+
+fix: use the original file to select the adapter for attachment removal when contentType is absent
--- a/packages/core/src/adapters/attachment.test.ts
+++ b/packages/core/src/adapters/attachment.test.ts
@@ -1,5 +1,10 @@
 import { afterEach, describe, expect, it } from "vitest";
-import { getFileDataURL, SimpleImageAttachmentAdapter } from "./attachment";
+import {
+  type AttachmentAdapter,
+  CompositeAttachmentAdapter,
+  getFileDataURL,
+  SimpleImageAttachmentAdapter,
+} from "./attachment";
 
 const originalFileReader = globalThis.FileReader;
 const originalBuffer = globalThis.Buffer;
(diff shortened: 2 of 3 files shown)


okisdev (approved):
approving. the three composite methods now agree on how they pick an adapter: `add` selects from `state.file`, `send` from `attachment.file`, and `remove` was the odd one out reading `contentType`, which is optional on `BaseAttachment` and so empty for any adapter that does not fill it in. since `PendingAttachment.file` is required, preferring the file makes the whole pending lifecycle route through one adapter by construction instead of by convention.

i checked out the branch and ran it both ways: the pending case throws exactly `No matching adapter found for attachment` on the merge base and passes on the head, and the two focused suites are 15/15. the call sites i read rather than exercised: the legacy composer only reaches `remove` for non-complete attachments (base-composer-runtime-core.ts:557), so what a user saw was the stuck `incomplete/error` status set a few lines below, and external-thread.ts:519 is the path that can hand `remove` a complete attachment, which the `?? { name, contentType }` fallback and the second test leg cover.

one thing still standing, and it predates this: a CompleteAttachment carrying neither `file` nor `contentType` still throws, and `partToCompleteAttachment` produces exactly that shape for image parts. follow-up, not a blocker. no overlap with #6880 either, that one changes how `fileMatchesAccept` compares extensions while this changes what is passed to it.

PR #6921, opened 2026-09-06
test(react): cover MCP app size forwarding

### LINKED ISSUE #6907: test(react): assert McpAppFrame forwards size changes to the sandbox host
`McpAppFrame` wires the widget's `notifications/size_changed` to `SandboxHostApi.setHeight`, and forwards the same event to the host's own `onSizeChange` handler. neither is asserted anywhere.

every call site in `packages/react/src/mcp-apps/app-frame.test.tsx` passes `{ setHeight: vi.fn() }` purely to satisfy the `createBridge` signature, and no test reads it back. the frame could stop calling `setHeight` entirely, or stop forwarding to the host handler, and the suite would stay green.

worth an `it` that mounts the frame, drives `handlers.onSizeChange({ height })` through the captured `createBridge`, and asserts both that `setHeight` received the height and that the host's live `onSizeChange` received the event, including after a rerender swaps the handler object.

<a href="https://rupic.dev/s/s2.QpuQARo3hoJl5NcqUW9f3YPWsjjoWuurzELNBtSOgbpOGjxvxfyqTgVThUg?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

Gujiassh:
## Problem

`McpAppFrame` already forwards widget size changes to both the sandbox host and the host callback, but the existing tests only supplied `setHeight` as a signature placeholder. Either forwarding path could regress without a test failure.

Closes #6907.

## Root cause

The app-frame suite did not drive the `onSizeChange` handler created by `McpAppFrame`, and it did not exercise handler replacement through the live ref.

## Change

Add a focused regression test that:

- verifies a size event calls `SandboxHostApi.setHeight` with the reported height;
- verifies the complete size event reaches the host `onSizeChange` handler;
- rerenders with a replacement handler and confirms the existing bridge dispatches to the new callback;
- disposes the sandbox bridge and confirms the underlying MCP app bridge is disposed.

## Verification

- `packages/react/node_modules/.bin/vitest.cmd run src/mcp-apps/app-frame.test.tsx` — 11 tests passed; no type errors
- scoped `oxfmt --check`
- scoped `oxlint`
- repository `lint-staged` tasks for the changed file
- `git diff --check`

## Public surface

None. This changes test coverage only; runtime code, exports, documentation, templates, and release metadata are unchanged.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6921?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.yHQ2UWU0f4HUzRh8mfydF7ufrQJtmpx7BYS5NFYrRPZm0uOHgVRfxSAFtH0?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/packages/react/src/mcp-apps/app-frame.test.tsx
+++ b/packages/react/src/mcp-apps/app-frame.test.tsx
@@ -171,6 +171,67 @@ describe("McpAppFrame", () => {
     },
   );
 
+  it("forwards size changes to the sandbox host and live handler", () => {
+    let createBridge: SandboxHostProps["createBridge"] | null = null;
+    sandboxHostMock.mockImplementation((props: SandboxHostProps) => {
+      createBridge ??= props.createBridge;
+      return null;
+    });
+    const bridge: McpAppBridge = {
+      onMessage: vi.fn(),
+      dispose: vi.fn(),
+      notifyToolInput: vi.fn(),
+      notifyToolResult: vi.fn(),
+      notifyHostContextChanged: vi.fn(),
+    };
+    createMcpAppBridgeMock.mockReturnValue(bridge);
+    const initialOnSizeChange = vi.fn();
+    const replacementOnSizeChange = vi.fn();
+    const view = (onSizeChange: typeof initialOnSizeChange) => (
+      <McpAppFrame
+        app={{ resourceUri: "ui://example/widget" }}
+        resource={{
+          uri: "ui://example/widget",
+          mimeType: MCP_APP_MIME_TYPE,
+          html: "",
+        }}
+        handlers={{ onSizeChange }}
+      />
+    );
+    const rendered = render(view(initialOnSizeChange));
+
+    const setHeight = vi.fn();
+    const sandboxBridge = createBridge!(
+      {
+        iframe: document.createElement("iframe"),
+        origin: "https://widget.example",
+        sendMessage: vi.fn(),
+      },
+      { setHeight },
+    );
+    const options = createMcpAppBridgeMock.mock
+      .calls[0]![0] as CreateMcpAppBridgeOptions;
+
+    try {
+      const initialSize = { width: 640, height: 360 };
+      options.handlers?.onSizeChange?.(initialSize);
+      expect(setHeight).toHaveBeenCalledWith(360);
+      expect(initialOnSizeChange).toHaveBeenCalledWith(initialSize);
+
+      rendered.rerender(view(replacementOnSizeChange));
+      const replacementSize = { width: 800, height: 480 };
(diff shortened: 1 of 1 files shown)


okisdev (approved):
approving. the test earns its place, and i mutation-tested it rather than trusting that it covers the paths it names: the pre-existing suite caught none of what it now catches.

concretely, i patched `app-frame.tsx` five ways and reran `vitest run src/mcp-apps/app-frame.test.tsx` each time. dropping `host.setHeight(p.height)`, dropping `live()?.onSizeChange?.(p)`, swapping the live ref for an `initial?.onSizeChange` snapshot, and dropping `bridge.dispose()` each printed `1 failed | 11 passed`, the failure always landing on one of the two cases here. against your original commit the same four printed `1 failed | 10 passed`, so the ten tests that existed before this PR caught exactly zero of them. that is the issue's premise measured rather than assumed.

i pushed one commit on top of yours. the `p.height != null` guard in `createBridge` had no coverage, so it now takes a width-only event and asserts `setHeight` stays at two calls; removing that guard was the one mutation of the five that kept your version green. i also lifted `expect(bridge.dispose).toHaveBeenCalledOnce()` into its own `it`, because as written a disposal regression failed a case named "forwards size changes to the sandbox host and live handler", and the next person reading that failure goes looking in the wrong place. the assertion was worth keeping; nothing else in the file covers it.

one trap for whoever extends this next: `options.handlers?.onSizeChange?.(size)` optional-chains twice, so if the frame ever stopped passing the handler into `createMcpAppBridge`, that line degrades to a silent no-op instead of a failure. it is safe here only because a `setHeight` or handler assertion follows every call. a case added there without one would pass vacuously.

on the missing changeset the checks flagged: not needed, and please don't add one. the seven most recent test-only PRs (#6294, #6062, #5266, #5265, #5264, #5255, #5254) all merged without one, and `scripts/check-changesets.mjs` validates changesets that exist rather than requiring that one exist. adding it here would cut a `@assistant-ui/react` release whose dist is byte-identical.

PR #6936, opened 2026-09-07
test(core): pin title-state cleanup on successful thread deletion

### LINKED ISSUE #6929: legacy remote thread delete has no coverage for title-state cleanup
`RemoteThreadListThreadListRuntimeCore.delete()` clears the deleted thread's title state on the success path, and nothing tests it. deleting the line keeps the whole suite green:

```diff
-    clearThreadTitleState(this._titleStates, data.id);
```

```
pnpm -C packages/core exec vitest run
Test Files  161 passed (161)
     Tests  1688 passed (1688)
```

surfaced while reviewing #6864. an assertion that looked like coverage went in there and came back out in f9b77cbc7, because `__internal_setOptions` clears `_titleStates` synchronously on an adapter change, so the assertion already held before the deletion resolved.

the tap client's equivalent, `clearThreadTitleState(session.titleStates, data.id)` in `RemoteThreadList.ts`, is worth checking the same way.

a covering test seeds a title state for a regular thread, deletes it against an adapter whose `delete()` resolves, and asserts the entry is gone, with no adapter swap anywhere in the sequence.

<a href="https://rupic.dev/s/s2.Btn3h2NdALw-bf9Wq99EsInmhh_tBVTgbij2d8lCIVpA8VQUT8WHVserZpk?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

Kinfe123:
Closes #6929.

## Problem

`RemoteThreadListThreadListRuntimeCore.delete()` clears the deleted thread's title state on the success path, and nothing tests it — deleting the `clearThreadTitleState` line kept all 1688 core tests green. The issue also asked for the tap client's equivalent in `RemoteThreadList.ts` to be checked the same way; mutating that line kept the suite green too, so both cleanups were uncovered.

## Change

Two covering tests, one per client, both verified by mutation (each fails with its `clearThreadTitleState` call removed and passes with it restored):

- **Legacy runtime** (`RemoteThreadListThreadListRuntimeCore.switch-delete.test.ts`): exactly the sequence the issue specifies — seed a title state for a regular thread, delete against an adapter whose `delete()` resolves, assert the entry is gone, with no adapter swap anywhere (the earlier attempt in #6864 came out because `__internal_setOptions` clears `_titleStates` on adapter change, making the assertion vacuous). Mirrors the existing failure-path test ("preserves thread cleanup state when remote deletion fails") from the other side.
- **Tap client** (`RemoteThreadList.test.ts`): the tap client's session is not reachable from outside, so this one pins the observable consequence instead: rename a thread manually (recording the manual-title state), delete it, reload against a server that still lists the same remote id (the slot is reborn under the same id), and assert an `automatic: true` `generateTitle` on the reborn thread still reaches the adapter. With the cleanup removed, the leaked manual-rename entry suppresses the auto-title and the test fails.

## Verification

- Both target files green on this branch: 57 tests.
- Mutation check per test as above; the tap pin also fails when `reload()` clears every title state unconditionally, so the control cannot go vacuous.
- Full `@assistant-ui/core` suite green: 163 files, 1733 tests.

## Public surface

None; tests only, no changeset per repo precedent for test-only core PRs.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6936?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.MohokfHm3uNUK-iMVO7UZbp9GjUyFoAqjJN4KjYxa_hG9u3iIhxmQ-ojUqo?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/packages/core/src/react/client/RemoteThreadList.test.ts
+++ b/packages/core/src/react/client/RemoteThreadList.test.ts
@@ -404,6 +404,41 @@ describe("RemoteThreadList", () => {
     handle.destroy();
   });
 
+  it("clears the deleted thread's title state so a reborn slot can auto-title", async () => {
+    const adapter = makeAdapter({
+      list: vi.fn(async () => ({
+        threads: [{ status: "regular" as const, remoteId: "t1", title: "One" }],
+      })),
+    });
+    const { handle } = mountList(adapter);
+    const aui = handle.getClient();
+    await aui.threads.getLoadThreadsPromise();
+    await vi.waitFor(() => {
+      expect(aui.threads.getState().threadIds).toEqual(["t1"]);
+    });
+
+    await aui.threads.item({ id: "t1" }).rename("Manual title");
+    await aui.threads.item({ id: "t1" }).delete();
+    await vi.waitFor(() => {
+      expect(aui.threads.getState().threadIds).toEqual([]);
+    });
+
+    // The server still lists t1, so a reload rebirths the slot under the same
+    // remote id; a leaked manual-rename entry would suppress its auto-title.
+    await aui.threads.reload();
+    await vi.waitFor(() => {
+      expect(aui.threads.getState().threadIds).toEqual(["t1"]);
+    });
+    flushTapSync(() => aui.threads.switchToThread("t1"));
+    await vi.waitFor(() => {
+      expect(aui.threads.getState().mainThreadId).toBe("t1");
+    });
+
+    await aui.threads.item({ id: "t1" }).generateTitle({ automatic: true });
+    expect(adapter.generateTitle).toHaveBeenCalledOnce();
+    handle.destroy();
+  });
+
   it("keeps a manual rename made during automatic title generation", async () => {
     const generatedTitle = deferred<ReadableStream>();
     const adapter = makeAdapter({
--- a/packages/core/src/react/client/RemoteThreadList.test.ts
+++ b/packages/core/src/react/client/RemoteThreadList.test.ts
@@ -407,27 +407,31 @@ describe("RemoteThreadList", () => {
   it("clears the deleted thread's title state so a reborn slot can auto-title", async () => {
     const adapter = makeAdapter({
       list: vi.fn(async () => ({
-        threads: [{ status: "regular" as const, remoteId: "t1", title: "One" }],
+        threads: [
+          { status: "regular" as const, remoteId: "t1", title: "One" },
+          { status: "regular" as const, remoteId: "t2", title: "Two" },
+        ],
       })),
     });
     const { handle } = mountList(adapter);
     const aui = handle.getClient();
     await aui.threads.getLoadThreadsPromise();
     await vi.waitFor(() => {
-      expect(aui.threads.getState().threadIds).toEqual(["t1"]);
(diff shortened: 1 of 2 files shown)


packages/core/src/react/client/RemoteThreadList.test.ts
```diff
@@ -404,6 +404,55 @@ describe("RemoteThreadList", () => {
...
+    await vi.waitFor(() => {
+      expect(aui.threads.getState().threadIds).toEqual(["t1", "t2"]);
+    });
+
+    await aui.threads.item({ id: "t1" }).rename("Manual title");
+    await aui.threads.item({ id: "t2" }).rename("Kept manual title");
+    await aui.threads.item({ id: "t1" }).delete();
+    await vi.waitFor(() => {
+      expect(aui.threads.getState().threadIds).toEqual(["t2"]);
+    });
```
Kinfe123, line 428: Added via 1cc03d265: a second listed thread renamed but not deleted rides in the same test as the control — after the reload its automatic generateTitle stays suppressed, so a reload that cleared every title state fails the test instead of hollowing the pin.

okisdev (approved):
approving. the two pins hold under mutation, which is the whole point of the PR: dropping `clearThreadTitleState` at `RemoteThreadListThreadListRuntimeCore.tsx:1139` fails only the new legacy test (1 failed | 6 passed), and dropping it at `RemoteThreadList.ts:1168` fails only the new tap test (1 failed | 49 passed). i also ran the mutation the thread was actually worried about, hoisting `session.titleStates.clear()` out of the `adapterChanged` branch in `reload()` so every title state gets cleared: the tap test fails there too, through t2. the control is load bearing, not decoration.

worth writing down for whoever reads this next: the tap test pins a consequence rather than the cleanup itself, and it stays meaningful only because `reload()` leaves title states alone when the adapter is identical. that dependency is now a failing test instead of a silent vacuum, so the control has to survive any future edit to this test.

i could not reproduce the two failures the description attributed to main. fresh worktree on your branch, `pnpm -C packages/core exec vitest run` reports 163 files and 1733 tests all passing, both named tests pass under a `-t` filter, and CI's Test Changed Packages is green on the same tree. from the setup i hit myself, my guess (a guess, not a finding) is an unbuilt workspace dependency: with `@assistant-ui/store` missing its dist, those files fail at import rather than at an assertion. i dropped that bullet and put the full suite result in its place.

the one commit i pushed only trims the control comment to two lines; the three line version explained the review exchange rather than the invariant, and the default here is no comment at all.

PR #6937, opened 2026-09-07
fix(python): derive with_parent_id controllers without rebuilding a StateManager

### LINKED ISSUE #6924: python: with_parent_id builds and discards a whole StateManager
## problem

`RunController.with_parent_id` (`python/assistant-stream/src/assistant_stream/create_run.py:54-63`) constructs a full `RunController` and then overwrites almost everything it just built:

```python
def with_parent_id(self, parent_id: str) -> 'RunController':
    controller = RunController(self._queue, self._state_manager.state_data, parent_id)
    controller._loop = self._loop
    controller._dispose_callbacks = self._dispose_callbacks
    controller._stream_tasks = self._stream_tasks
    controller._state_manager = self._state_manager
    ...
```

the constructor runs `StateManager(self._put_chunk_nowait, state_data)`, which builds an `AssistantState`, a `Flusher`, a `StateDraft` and a `StateProxy`, and calls `asyncio.get_running_loop()`. the very next lines throw all of it away in favour of the parent's `_state_manager`, so the only thing the constructor contributes is `_queue`, `_parent_id` and the two cancellation fields.

it also means `with_parent_id` cannot be called outside a running loop, purely because of the constructor path it does not use.

## why now

after #6923, `AssistantState.__init__` detaches its input, so that discarded `StateManager` now deep-copies the entire state on every `with_parent_id` call before dropping it. the waste predates that change and is not caused by it, but it stops being free.

the blast radius is small today: `add_reasoning_part` is the only production caller, and the affected path is one deep copy per sub-run rather than anything in a hot loop.

## suggested fix

build the derived controller without going through `__init__`, or give `RunController` a private constructor path that takes an existing `StateManager` instead of `state_data`. the fields `with_parent_id` actually wants are `_queue`, `_parent_id`, `_cancelled_event` and `_cancelled_signal`.

<a href="https://rupic.dev/s/s2.iMsxHEvansDO9fEXZIOvTg96jNk9Wk3JkbyYNZl15cFKVPUWIFnqfUIuSt4?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

Kinfe123:
Closes #6924.

## Problem

`RunController.with_parent_id` constructs a full `RunController` and then overwrites almost everything it just built. The constructor runs `StateManager(...)` — building an `AssistantState`, a `Flusher`, a `StateDraft` and a `StateProxy` — and calls `asyncio.get_running_loop()`; the next lines throw all of it away in favour of the parent's fields. Since #6923 detaches state on `AssistantState.__init__`, that discarded `StateManager` deep-copies the entire state on every call. It also means `with_parent_id` cannot be called outside a running loop, purely because of the constructor path it does not use.

## Change

The issue's suggested fix: the derived controller is built with `RunController.__new__` and explicit field assignments — every field `__init__` sets is either shared from the parent (queue, loop, dispose callbacks, stream tasks, state manager, cancellation event and signal) or is the new `parent_id`. No `StateManager` is constructed, no state is copied, and no running loop is required.

## Verification

- New `tests/test_with_parent_id.py`, three tests. The first pins that every shared field is identical between parent and derived controller and that a chunk emitted through the derived controller carries the parent id. The second pins `vars(derived).keys() == vars(controller).keys()`, so a future `__init__` field that `with_parent_id` forgets to carry over surfaces as a red test rather than a distant `AttributeError`. The third pins that `with_parent_id` works off-loop: it derives *and* appends from a worker thread (`asyncio.to_thread`, asserting no running loop first) and asserts the chunk arrives carrying the derived parent id. That third test fails on main with `RuntimeError: no running event loop` and passes with the fix.
- Full suite: 176 passed, 7 skipped (`tests/test_langgraph.py` fails collection on a clean checkout because `langchain_core` is not in the `dev` extra — pre-existing, untouched).

## Public surface

No API change; `with_parent_id`'s behavior for existing callers is identical. Python package, so no npm changeset.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6937?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.PDBB5OI6oA5g6K_SD9YSWV1Ln0SijUTENhxlx6tJ_8KlrE8d_scj0SSJlTA?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/python/assistant-stream/src/assistant_stream/create_run.py
+++ b/python/assistant-stream/src/assistant_stream/create_run.py
@@ -52,12 +52,20 @@ class RunController:
         self._cancelled_signal = ReadOnlyCancellationSignal(self._cancelled_event)
 
     def with_parent_id(self, parent_id: str) -> 'RunController':
-        """Create a new RunController instance with the specified parent_id."""
-        controller = RunController(self._queue, self._state_manager.state_data, parent_id)
+        """Create a new RunController instance with the specified parent_id.
+
+        The derived controller is a view over this one — every field but the
+        parent id is shared — so it is built without running ``__init__``,
+        which would construct (and immediately discard) a whole StateManager
+        and require a running event loop.
+        """
+        controller = RunController.__new__(RunController)
+        controller._queue = self._queue
         controller._loop = self._loop
         controller._dispose_callbacks = self._dispose_callbacks
         controller._stream_tasks = self._stream_tasks
         controller._state_manager = self._state_manager
+        controller._parent_id = parent_id
         controller._cancelled_event = self._cancelled_event
         controller._cancelled_signal = self._cancelled_signal
         return controller
--- a/python/assistant-stream/src/assistant_stream/create_run.py
+++ b/python/assistant-stream/src/assistant_stream/create_run.py
@@ -54,10 +54,9 @@ class RunController:
     def with_parent_id(self, parent_id: str) -> 'RunController':
         """Create a new RunController instance with the specified parent_id.
 
-        The derived controller is a view over this one — every field but the
-        parent id is shared — so it is built without running ``__init__``,
-        which would construct (and immediately discard) a whole StateManager
-        and require a running event loop.
+        Every field but the parent id is shared with this controller, so the
+        derived one skips ``__init__``, which would build and discard a whole
+        StateManager and require a running event loop.
         """
         controller = RunController.__new__(RunController)
         controller._queue = self._queue
--- a/python/assistant-stream/tests/test_with_parent_id.py
+++ b/python/assistant-stream/tests/test_with_parent_id.py
@@ -40,20 +40,34 @@ async def test_with_parent_id_shares_everything_but_the_parent_id():
 
 
 @pytest.mark.anyio
-async def test_with_parent_id_works_outside_a_running_loop():
+async def test_with_parent_id_sets_every_field_init_sets():
     captured: dict[str, RunController] = {}
 
     async def run_callback(controller: RunController):
+        captured["derived"] = controller.with_parent_id("p1")
         captured["controller"] = controller
-        controller.append_text("hello")
 
     [chunk async for chunk in create_run(run_callback)]
 
-    def derive() -> RunController:
-        with pytest.raises(RuntimeError):
-            asyncio.get_running_loop()
-        return captured["controller"].with_parent_id("p2")
+    assert vars(captured["derived"]).keys() == vars(captured["controller"]).keys()
 
-    derived = await asyncio.to_thread(derive)
-    assert derived._parent_id == "p2"
-    assert derived._state_manager is captured["controller"]._state_manager
+
+@pytest.mark.anyio
+async def test_with_parent_id_works_outside_a_running_loop():
+    async def run_callback(controller: RunController):
+        def derive_and_use() -> RunController:
+            with pytest.raises(RuntimeError):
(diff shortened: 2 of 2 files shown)


python/assistant-stream/src/assistant_stream/create_run.py
```diff
@@ -52,12 +52,20 @@ def __init__(self, queue, state_data, parent_id: Optional[str] = None):
...
+        controller = RunController.__new__(RunController)
+        controller._queue = self._queue
         controller._loop = self._loop
         controller._dispose_callbacks = self._dispose_callbacks
         controller._stream_tasks = self._stream_tasks
         controller._state_manager = self._state_manager
+        controller._parent_id = parent_id
         controller._cancelled_event = self._cancelled_event
         controller._cancelled_signal = self._cancelled_signal
         return controller
```
Kinfe123, line 70: Took option 2: the sharing test now pins `vars(derived).keys() == vars(controller).keys()`, so a ninth `__init__` field turns into a red test instead of a distant AttributeError, and the explicit what-is-shared list stays readable at the call site. Also folded in cubic's note — the off-loop test now derives *and uses* the controller from the worker thread (`append_text` crosses back via `enqueue_threadsafe`) and asserts the chunk arrives carrying the derived parent id.

okisdev (approved):
approving. the change is exactly the mechanism the issue named, and both sides of it hold up under execution rather than just reading right.

i ran the off-loop test against the merge base and against the head under the same command. on base it dies at `create_run.py:46` with `RuntimeError: no running event loop`; on head it passes and the chunk comes back carrying `p2`. full suite on the head is 176 passed, 7 skipped (`test_langgraph.py` still fails collection on a clean checkout because `langchain_core` is not in the `dev` extra, pre-existing and untouched here).

the thing worth pinning down for whoever reads this next: `__new__` plus a hand-written field list is only safe because of `test_with_parent_id_sets_every_field_init_sets`. i checked that it actually bites rather than assuming it, by adding a ninth field to `__init__` and leaving `with_parent_id` alone, which turns that test red immediately. so it is load-bearing, not decoration, and deleting it later as a redundant-looking assertion puts the `AttributeError` hazard straight back. `copy.copy(self)` would drop the coupling entirely but trades this failure mode for the inverse one (a future field that should not be shared gets shared silently), and the explicit list keeps what-is-shared readable at the call site, so option 2 was the right pick.

one commit from me on top: the docstring carried the only two em-dashes in the whole `python/` tree, so i reworded it. no behavior change.

and a correction to my own issue: #6924 claims `add_reasoning_part` is the only production caller, which is wrong. there is no production caller at all; the only in-repo call site is `tests/test_data_stream.py:332`. it does not change anything about the fix, since `with_parent_id` is public API and external callers reach it through `PartsGroupedByParentId`, but it means the deep copy this removes was costing user code rather than ours.

Issue #6690, opened 2026-09-01
with-nuxt: pnpm typecheck fails on two vue kit test files, and nothing runs it


okisdev:
found while running the dependency update in #6639, reproduced on clean `main`.

## problem

`pnpm typecheck` in `examples/with-nuxt` exits 2:

```
../../packages/ui/src/components/vue/assistant-ui/thread-list.test.ts(68,42): error TS2554: Expected 0 arguments, but got 1.
../../packages/ui/src/components/vue/assistant-ui/thread.test.ts(153,11): error TS2322: Type 'Error' is not assignable to type 'ReadonlyJSONValue | undefined'.
  Type 'Error' is not assignable to type 'ReadonlyJSONObject'.
    Index signature for type 'string' is missing in type 'Error'.
```

both files landed in #6618. `templates/nuxt` passes; `examples/with-nuxt` fails because its tsconfig reaches into `packages/ui`'s vue sources, single-sourced in #6566.

reproduced by building a worktree at `main`, running `pnpm build`, then `pnpm typecheck` in `examples/with-nuxt`. the two errors are byte-identical there, so this predates every dependency change in #6639.

## why nothing caught it

`typecheck` is a script in exactly three packages (`examples/with-react-router`, `examples/with-nuxt`, `templates/nuxt`), is not a turbo task, and is referenced by no workflow. there is no root `pnpm typecheck` either. #6557 already lists this as the reason the nuxt `typescript` outlier can only be caught by hand.

the errors themselves are about assistant-ui's own types (`RemoteThreadList` arity, `ReadonlyJSONValue` rejecting an `Error` instance), not about vue or any third-party bump.

## the decision this needs

fixing the two call sites is small. the question worth answering first is whether `typecheck` should become a turbo task so the vue surface is type-checked in CI at all — right now the entire vue kit is checked by nothing, and `packages/ui` carries no typecheck of its own because `test:types` skips packages without a `tsconfig.json` at their root.

<a href="https://rupic.dev/s/s2.6TQYIVxWnqhyzyCkjooClj0EqpcIXUiEEbJktnfKn6uBKRuQZo5UFjIsh-o?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg"><img alt="Track in Rupic" src="https://rupic.dev/badges/track-light-v1.svg"></picture></a>

PR #6939, opened 2026-09-07
fix: repair test:types drift across five packages

Kinfe123:
Addresses #6801's five reported packages (the CI-lane half — whether a job should run `test:types` — is a maintainer scope decision and stays out, same split as #6690).

## Problem

`pnpm test:types` fails in the five packages the issue names on a clean checkout: `@assistant-ui/mcp-docs-server`, `@assistant-ui/tap`, `assistant-ui` (the CLI), `@assistant-ui/next`, and `@assistant-ui/cloud`. Nothing gates the script, so it drifted — two distinct causes, per the issue.

## Change

**Test files outside their tsconfig's type environment** (no behavior change):

- `mcp-docs-server/proxy.test.ts` used bare `describe`/`it`/`expect`/`vi`; they are now imported from vitest explicitly, the repo's convention everywhere else.
- `tap`'s two standalone-shim tests and `cloud`'s auth-strategy test use node builtins (`node:fs`, `__dirname`, `Buffer`) but both packages extend the browser-neutral `ts/base`, which deliberately omits node globals. Each file now carries `/// <reference types="node" />`, pulling node types for that file alone via the shared typeRoots without changing the packages' neutral typing.
- `cloud`'s test also had one implicitly-any parameter and two casts `@tsconfig/strictest` rejects (`as Response`, `as Storage` on partial mocks), now explicit (`(key: string)`, `as unknown as`).
- `next/with-aui.test.ts` called `withAui` with literal configs, so the generic inferred types without `turbopack`/`webpack`; the calls now pass `Parameters<typeof withAui>[0]` explicitly (a pattern the file's own `rulesFor` helper already used).

**Genuine drift in source** (both behavior-neutral, changeset attached):

- `mcp-docs-server/proxy.ts`: `isInitializeRequest` narrows to a shape without `id`; the read is now guarded by the already-imported `isJSONRPCRequest` too, which carries `id` in its narrowing — an initialize request is always a JSON-RPC request, so the runtime path is identical.
- `cli` codemod: `VariableDeclaration.declarations` is typed `(VariableDeclarator | IdentifierKind)[]`, so declarators are now guarded with `j.VariableDeclarator.check` before `.id` access, and the `exportKind` read on `ExportSpecifier` (a runtime-real Babel property missing from ast-types' typings) goes through a narrow structural type.

## Newer drift, out of scope

While verifying, `@assistant-ui/store` now also fails `tsc --noEmit` (10 errors in 4 test files) on clean main — that landed after this issue was filed and sits in the actively-developed custom-scope test area (#6511's territory), so it is deliberately not bundled here; noting it on the issue.

## Verification

- Before: `pnpm test:types` fails in the five named packages with the issue's exact errors. After: all five pass `tsc --noEmit`.
- All five packages' vitest suites pass unchanged: tap 581, cli 339, cloud 119, next 9, mcp-docs-server 1.

## Public surface

No API changes. Patch changesets for the two source-touched published packages (`assistant-ui`, `@assistant-ui/mcp-docs-server`); the rest are test-only.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6939?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.VHrseGLncdKVGhy4LNxAOIKRGEHr-y0TN4OzDGzfYSbAKtx7Ob1tjevWuQI?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/test-types-drift.md
+++ b/.changeset/test-types-drift.md
@@ -1,0 +1,6 @@
+---
+"assistant-ui": patch
+"@assistant-ui/mcp-docs-server": patch
+---
+
+fix: repair the type drift tsc --noEmit catches in the codemod and proxy sources
--- a/packages/cli/src/codemods/v0-12/assistant-api-to-aui.ts
+++ b/packages/cli/src/codemods/v0-12/assistant-api-to-aui.ts
@@ -113,6 +113,7 @@ const migrateAssistantApiToAui = createTransformer(
         }
         if (j.VariableDeclaration.check(statement)) {
           for (const declarator of statement.declarations) {
+            if (!j.VariableDeclarator.check(declarator)) continue;
             if (
               j.Identifier.check(declarator.id) &&
               declarator.id.name === "api"
@@ -254,7 +255,7 @@ const migrateAssistantApiToAui = createTransformer(
             return;
           if (
             grandparent?.exportKind === "type" ||
-            parent.exportKind === "type"
+            (parent as { exportKind?: string }).exportKind === "type"
           )
             return;
           if (parent.exported === path.value && parent.local !== path.value)
--- a/packages/cli/src/codemods/v0-12/assistant-api-to-aui.ts
+++ b/packages/cli/src/codemods/v0-12/assistant-api-to-aui.ts
@@ -253,6 +253,8 @@ const migrateAssistantApiToAui = createTransformer(
             grandparent.source != null
           )
             return;
+          // Babel emits exportKind on ExportSpecifier for inline
+          // `export { type api }`; ast-types' typings omit it.
           if (
             grandparent?.exportKind === "type" ||
             (parent as { exportKind?: string }).exportKind === "type"
--- a/packages/cloud/src/tests/AssistantCloudAuthStrategy.test.ts
+++ b/packages/cloud/src/tests/AssistantCloudAuthStrategy.test.ts
@@ -1,3 +1,4 @@
+/// <reference types="node" />
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import {
   AssistantCloudAnonymousAuthStrategy,
@@ -60,7 +61,7 @@ describe("AssistantCloudAnonymousAuthStrategy", () => {
   it("reads the stored anonymous refresh token", () => {
     const values = new Map([[refreshTokenKey, JSON.stringify(refreshToken)]]);
     installLocalStorage({
-      getItem: (key) => values.get(key) ?? null,
+      getItem: (key: string) => values.get(key) ?? null,
       setItem: (key, value) => {
         values.set(key, value);
       },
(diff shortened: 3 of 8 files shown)


Kinfe123:
Review round addressed in d2e2526ce. Verified empirically: tsc flags exactly one `(key)` site, and the difference is that it is the file's only `as unknown as Storage` — the two-step cast erases the contextual typing every plain `as Storage` sibling gets; the annotation now carries that sentence so it doesn't read as deletable noise. The cli cast got its why-comment too (Babel emits `exportKind` on `ExportSpecifier` for inline `export { type api }`; ast-types' typings omit it). On the `withAui` signature: agreed `T` under-reports the synthesized `turbopack`/`webpack`, and `T & Required<Pick<NextConfigLike, "turbopack" | "webpack">>` reads right — but changing a published signature belongs in its own PR with its own changeset, not this drift repair, so the tests keep the explicit type argument for now.

okisdev (approved):
approving, with one commit pushed on top: `packages/next` still failed `tsc --noEmit` at d2e2526, with 7 errors instead of the 3 it has on main.

the trap is that `nextConfig: T = {} as T` carries a default, which makes the parameter optional, so `Parameters<typeof withAui>[0]` is `NextConfigLike | undefined`. `rulesFor` gets away with it because inference drops `undefined` from the candidate set; an explicit `withAui<AuiNextConfig>(...)` bypasses inference, hits `T extends NextConfigLike` head on, and then carries the `undefined` into the return type. `NonNullable<...>` on line 4 is the whole fix, and the three explicit type arguments were already right.

worth knowing for the next round: green CI is not evidence on this one. `code-quality.yaml:370` runs `pnpm turbo test` plus `test:react-compiler` at :380, and no workflow calls `test:types` at all, so all 25 checks passed over the regression. I ran it by hand in a worktree after `pnpm install --frozen-lockfile` and `pnpm build`: before, `mcp-docs-server`, `tap`, `cli` and `cloud` exit 0 while `next` exits 1; after, all five exit 0 and `packages/next` vitest stays at 9 passed. the `proxy.ts` change I read rather than exercised, and `isJSONRPCRequest` is typed `(value: unknown) => value is JSONRPCRequest`, so the added guard is what supplies `id` and the pair is behavior neutral.

`@assistant-ui/store`'s 10 errors are real and correctly left out. the CI lane in #6801 is the half that keeps this from drifting again, so that issue stays open on it.

PR #6949, opened 2026-09-07
fix(react-ag-ui): ignore superseded thread loads

### LINKED ISSUE #6855: react-ag-ui: a superseded thread load still replaces the selected thread
`packages/react-ag-ui/src/useAgUiRuntime.ts` (lines 190 to 206) wraps `onSwitchToThread` the same way `react-a2a` did before #6853, with no check for whether a newer thread selection has superseded the in-flight load. two rapid thread switches that resolve out of order leave the selected thread showing the older thread's conversation.

it is worse in ag-ui than it was in a2a, because the wrapper calls `core.applyExternalMessages([])` before awaiting: a superseded load first blanks the selected thread and then repopulates it with the stale thread's messages, plus the stale `result.state` through `loadExternalState`, and `result.unstable_resume` can start a run against the wrong thread.

AGENTS.md asks for cross-runtime parity or a written reason for the divergence. #6853 deliberately kept its scope to a2a under one concern per PR, so this is the follow-up.

## suggested fix

mirror the guard that landed in `packages/react-a2a/src/useA2ARuntime.ts`: assign a monotonic generation in both `onSwitchToNewThread` and `onSwitchToThread`, and skip `applyExternalMessages`, `loadExternalState`, `resetState`, and `resumeInFlightRun` when that generation is no longer current. the idiom already exists in core at `packages/core/src/react/runtimes/RemoteThreadListThreadListRuntimeCore.tsx:604,632`.

## verification

colocated vitest cases under `packages/react-ag-ui/src/`, mirroring the two added in `packages/react-a2a/src/useA2ARuntime.test.tsx`: one where an older `onSwitchToThread` resolves after a newer one, and one where `onSwitchToNewThread` supersedes a pending load. both must fail without the guard and pass with it. plus a patch changeset for `@assistant-ui/react-ag-ui`.

## known limitation to document rather than fix

the guard cannot cancel the adapter's own `setCurrentThreadId`. with the shape `apps/docs/content/docs/runtimes/ag-ui/runtime-options.mdx:123` teaches (await the fetch, then flip the id), the superseded call's id flip still lands last, so the runtime selects thread a while showing thread b's messages, and re-clicking a is a no-op because `ExternalStoreThreadListRuntimeCore.switchToThread` early-returns on `_mainThreadId === threadId`. the runtime has no cancellation signal to hand the adapter, so the fix is one line in the ag-ui thread-list docs saying the id flip belongs before the fetch.

sicauzxl:
## Problem

Rapid AG-UI thread switches can resolve out of order. Select A and then B, let B finish loading first, and finally resolve A: the UI still selects B but displays A's messages. The late result also replaces or resets state and can resume the wrong run. Creating a new thread has the same race in either direction.

Reproduced on main at `96f011a916de97ea73e25c307d9b2cb2f4758a85` (`@assistant-ui/react-ag-ui` 0.0.58). The included deferred-promise tests are a minimal offline reproduction.

Fixes #6855.

## Root cause

Both adapter wrappers apply their post-await effects without checking whether a newer switch has superseded them.

## Change

Use one monotonically increasing generation shared by existing-thread and new-thread switches, following the A2A runtime's existing approach. Only the current generation applies messages, state, or resume requests. Preserve the existing clear-before-ID-change ordering.

The runtime-options example now selects the thread before fetching history and explains that adapter-owned state updates need their own ordering guard.

## Verification

- On the unmodified source, the focused regression suite has 4 failures and 1 passing control; with the fix all 5 pass.
- `pnpm --filter @assistant-ui/react-ag-ui test`: 561 tests pass across 18 files.
- `pnpm exec turbo build --filter=@assistant-ui/react-ag-ui... --concurrency=4`: 11 successful tasks.
- `pnpm --filter @assistant-ui/react-ag-ui exec tsc --noEmit`: passes.
- `pnpm lint`: passes, with existing hook warnings.
- `pnpm changesets:check`: passes.
- `git diff --check`: passes.

## Public surface

`@assistant-ui/react-ag-ui` behavior fix with a patch changeset and runtime-options documentation update. No exports, signatures, templates, or generated API-reference pages change.

Implementation and validation were completed with AI assistance.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6949?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.AJvpCd56_VWur7KraNbmoz61ypbq9Rnwtv_rMILO_72j-51GIYioavvZwPY?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/agui-superseded-thread-loads.md
+++ b/.changeset/agui-superseded-thread-loads.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react-ag-ui": patch
+---
+
+Ignore superseded thread switches so late history loads cannot overwrite the active thread's messages or state, resume a stale run, or clear a newer selection.
--- a/apps/docs/content/docs/runtimes/ag-ui/runtime-options.mdx
+++ b/apps/docs/content/docs/runtimes/ag-ui/runtime-options.mdx
@@ -121,14 +121,19 @@ const runtime = useAgUiRuntime({
         setCurrentThreadId(await createThread());
       },
       onSwitchToThread: async (id) => {
-        const { messages, state } = await loadThread(id);
         setCurrentThreadId(id);
+        const { messages, state } = await loadThread(id);
         return { messages, state };
       },
     },
   },
 });
 ```
+
+Set the selected thread ID before awaiting its history. The runtime ignores messages,
+state, and resume requests returned by a superseded thread switch, but it cannot
+undo state updates inside your adapter. If creating a thread is asynchronous, also
+guard your adapter's thread ID update against a newer selection.
 
 ## Interrupts (experimental)
 
--- a/packages/react-ag-ui/src/useAgUiRuntime.thread-switch.test.tsx
+++ b/packages/react-ag-ui/src/useAgUiRuntime.thread-switch.test.tsx
@@ -1,0 +1,199 @@
+// @vitest-environment jsdom
+
+import { useState } from "react";
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { act, cleanup, renderHook } from "@testing-library/react";
+import type { HttpAgent } from "@ag-ui/client";
+import type { ThreadMessage } from "@assistant-ui/core";
+import { AgUiThreadRuntimeCore } from "./runtime/AgUiThreadRuntimeCore";
+import type { UseAgUiThreadListAdapter } from "./runtime/types";
+import { useAgUiRuntime } from "./useAgUiRuntime";
+
+type ThreadLoad = Awaited<
+  ReturnType<NonNullable<UseAgUiThreadListAdapter["onSwitchToThread"]>>
+>;
+
+function deferred<T>() {
+  let resolve!: (value: T) => void;
+  const promise = new Promise<T>((res) => {
+    resolve = res;
+  });
+  return { promise, resolve };
+}
+
+function message(id: string): ThreadMessage {
+  return {
+    id,
+    role: "user",
+    content: [{ type: "text", text: id }],
+    attachments: [],
+    createdAt: new Date(0),
+    metadata: { custom: {} },
+  };
+}
+
+function renderRuntime(
+  load: (id: string) => Promise<ThreadLoad>,
+  create: () => Promise<void> = async () => {},
+) {
+  const agent = {
+    runAgent: vi.fn(),
+    abortRun: vi.fn(),
+  } as unknown as HttpAgent;
+  return renderHook(() => {
+    const [threadId, setThreadId] = useState("initial");
+    return useAgUiRuntime({
+      agent,
+      adapters: {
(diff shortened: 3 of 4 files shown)


sicauzxl:
Confirmed the GitHub Actions results for the unchanged PR head `6252176fcf2a90d58ee5f531c12f958eb3a547bd`:

- [Code Quality](https://github.com/assistant-ui/assistant-ui/actions/runs/34084610438): all 7 jobs succeeded, including Test Changed Packages, Build Changed Packages, Oxlint + Oxfmt, Workspace Ranges, Template Sync, Build Changed Apps, and API Reference Drift.
- [Changeset Semver Check](https://github.com/assistant-ui/assistant-ui/actions/runs/34084908104): succeeded (earlier superseded runs were cancelled).
- [Source Snapshot Budget](https://github.com/assistant-ui/assistant-ui/actions/runs/34084610402) and [autofix.ci](https://github.com/assistant-ui/assistant-ui/actions/runs/34084610430): succeeded.

These are the remote CI results checked today; the test counts in the PR description describe the original local validation. Vercel and CodeRabbit also currently report success.

The in-flight run behavior mentioned in the review is separate from this ordering fix. I will investigate it with a deterministic reproduction before proposing a follow-up.

okisdev (approved):
approved. the guard is the right shape: it is the same monotonic generation `useA2ARuntime` already runs, so this closes the parity gap #6855 named instead of inventing an ag-ui specific mechanism for it.

i ran it rather than only reading it. with `useAgUiRuntime.ts` checked back out to main the new suite is 4 failed and 1 passing control, and 5 of 5 with the guard restored, so the regression cases genuinely bite; the full package suite is 561 passing on your head. your checks predate #6962, which rewrote abort handling in the class these tests spy on, so i also merged current main into your head locally and reran: 565 passing, no interaction. that merge stayed local, there is nothing here that needs the branch refreshed.

two things i noticed that do not block. the `onSwitchToNewThread` snippet in runtime-options.mdx is still `setCurrentThreadId(await createThread())`, which is the await-then-set shape the paragraph you added warns about, and showing the remedy inline would take a generation ref in userland that the page does not have room for. and spying on `AgUiThreadRuntimeCore.prototype.resumeInFlightRun` is heavier than the mock client the a2a tests go through, though `afterEach` puts it back.

PR #6953, opened 2026-09-07
fix(react-mcp): keep static oauth clients out of the dynamic cache

ephraimduncan:
## Problem

After an app removes `clientId`, a dynamic OAuth provider can return the former static client instead of an independent registration.

The reproduction applies to `@assistant-ui/react-mcp` 0.1.16 at base `7fd03e375b41101f9ff57874d11e4f8eb0e80c4a`. Save this as `repro.ts` in the repository root and run `bun repro.ts`:

```ts
import { createOAuthProvider } from "./packages/react-mcp/src/auth/createOAuthProvider";

let saved = null;
const storage = {
  loadCustomServers: async () => [],
  loadAuthState: async () => saved,
  saveAuthState: async (_id, state) => { saved = structuredClone(state); },
  clearAuthState: async () => { saved = null; },
};
const options = {
  serverId: "docs",
  serverUrl: "https://example.invalid/mcp",
  storage,
  redirectUri: "https://app.invalid/callback",
  onAuthorizationUrl: () => {},
};
const configured = createOAuthProvider({
  ...options, config: { type: "oauth", clientId: "configured-client" },
});
const info = await configured.clientInformation();
assert.ok(info);
await configured.saveClientInformation({
  ...info, issuer: "https://issuer.invalid",
}, { issuer: "https://issuer.invalid" });
const dynamic = createOAuthProvider({ ...options, config: { type: "oauth" } });
assert.equal(await dynamic.clientInformation(), undefined);
```

## Root cause

The SDK adds an issuer to static client information and calls `saveClientInformation`. That method wrote the static client into the shared dynamic cache and storage.

## Change

The statically configured client becomes a provider-owned overlay that the SDK's write-backs replace wholesale, so the issuer stamp lands locally and dynamic registrations stay in the shared record. `invalidateCredentials("client")` and `("all")` reset the overlay to the configured client, keeping its lifecycle in step with the cache it shadows.

Replacing rather than merging matters when a changed issuer makes the SDK discard the static client and register a new one: a merge would leave the configured `client_secret` on the freshly issued `client_id`, and `selectClientAuthMethod` would then send that secret to the new authorization server.

This prevents new cache pollution without a migration of saved records.

## Verification

- The API reproduction failed before the correction and passed afterward.
- The regressions call real SDK `auth` with cached discovery data. All five cases failed before the correction and passed afterward: the two write-back isolation cases, the re-registration at a new issuer, and the two invalidation scopes.
- `pnpm --filter @assistant-ui/react-mcp test`: 192 tests passed.
- `pnpm exec turbo run build --filter=@assistant-ui/react-mcp --force`: the package and its dependencies built successfully.
- Scoped `oxlint` and `pnpm changesets:check` passed.

The checks used local storage data and did not contact a live OAuth server.

## Public surface

`@assistant-ui/react-mcp` has a patch changeset. Exports, signatures, documentation, API-reference output, and templates stay unchanged.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6953?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.sKbOgqMb65s41EYx6TqJO5CO-V0noJjPaJunfzN8SWj8gVI5x33G1F5vGgo?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/oauth-static-client-cache.md
+++ b/.changeset/oauth-static-client-cache.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react-mcp": patch
+---
+
+fix: keep static OAuth client updates out of the dynamic registration cache
--- a/packages/react-mcp/src/auth/createOAuthProvider.test.ts
+++ b/packages/react-mcp/src/auth/createOAuthProvider.test.ts
@@ -68,6 +68,41 @@ const createProvider = (storage: MCPStorage) =>
     onAuthorizationUrl: () => {},
   });
 
+const createStaticProvider = (storage: MCPStorage, clientSecret?: string) =>
+  createOAuthProvider({
+    serverId: "docs",
+    serverUrl,
+    config: {
+      type: "oauth",
+      clientId: "client-a",
+      ...(clientSecret ? { clientSecret } : {}),
+    },
+    storage,
+    redirectUri: "http://localhost/callback",
+    onAuthorizationUrl: () => {},
+  });
+
+const discoveryStateFor = (issuer: string): OAuthDiscoveryState => ({
+  ...discoveryState,
+  authorizationServerUrl: issuer,
+  authorizationServerMetadata: {
+    issuer,
+    authorization_endpoint: `${issuer}/authorize`,
+    token_endpoint: `${issuer}/token`,
+    registration_endpoint: `${issuer}/register`,
+    response_types_supported: ["code"],
+    code_challenge_methods_supported: ["S256"],
+  },
+  resourceMetadata: {
+    resource: "https://mcp.example.com",
+    authorization_servers: [issuer],
+  },
+});
+
+const rejectFetch = async () => {
+  throw new Error("Unexpected OAuth request");
+};
+
 describe("createOAuthProvider callback state", () => {
   it("persists the generated state with the PKCE verifier", async () => {
     const { storage, getState } = createStorage();
@@ -587,13 +622,7 @@ describe("createOAuthProvider persistence across provider instances", () => {
     async (clientId) => {
       const { storage, getState } = createStorage({
         serverUrl,
-        discoveryState: {
-          ...discoveryState,
-          authorizationServerMetadata: {
-            ...discoveryState.authorizationServerMetadata,
-            code_challenge_methods_supported: ["S256"],
-          },
-        },
+        discoveryState: discoveryStateFor("https://auth.example.com"),
       });
       const dynamicProvider = createProvider(storage);
       if (clientId) {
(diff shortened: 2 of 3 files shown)


okisdev (approved):
approving. the diagnosis is right: the SDK stamps an issuer onto whatever `clientInformation()` returns and writes it straight back, and the old `saveClientInformation` put that into the shared cache, so a provider built later without a `clientId` read the static client out of storage. your test fails on the base for exactly the reason it claims, `client_id: "client-a"` where `undefined` or `registered-client` belongs.

i pushed one change on top, because `Object.assign` merges where it needs to replace. once a changed issuer makes the SDK discard the static client, `auth()` falls through to dynamic registration and calls `saveClientInformation` with the DCR result, and the merge leaves the configured `client_secret` sitting on the freshly issued `client_id`. i ran that path against real `auth()` with a moved discovery state:

```
after issuer A: { client_id: "client-a",   client_secret: "s3cret", issuer: "https://a.example.com" }
after issuer B: { client_id: "dcr-client", client_secret: "s3cret", issuer: "https://b.example.com" }
```

`selectClientAuthMethod` (`@modelcontextprotocol/client@2.0.0`, dist/index.mjs:457-463, read not run) picks `client_secret_basic` or `client_secret_post` purely on `client_secret !== undefined`, so the next token request would carry the configured secret to the authorization server SEP-2352 had just refused to trust the static client at. replacing the overlay wholesale keeps the registration result clean. the same commit resets the overlay in `invalidateCredentials("client")` and `("all")` so it dies with the record it shadows rather than outliving it with a stale stamp.

worth knowing before someone reaches for the cheap check: `pnpm --filter @assistant-ui/react-mcp test` is green on your commit. the suite alone never reaches this, because it only appears once the discovery state moves between two `auth()` calls on the same provider.

smaller: spreading `discoveryState.authorizationServerMetadata` and overriding fields yields a union tsc will not assign to `AuthorizationServerMetadata`, so the new case reported an error under `tsc --noEmit -p tsconfig.json`. the package has no `test:types` script, which is why CI stayed green on it. writing the metadata literal out clears it, 14 errors down to 13.

the two gaps this deliberately leaves, records polluted before it lands and tokens outliving the client that minted them, are #6966.

PR #6955, opened 2026-09-07
fix(store): detect changes in sparse array positions

ephraimduncan:
## Problem

Shallow selectors can keep an empty array position after a new value replaces it. The result also depends on the argument order.

The reproduction uses `@assistant-ui/store` 0.3.12 source at base `7fd03e375b41101f9ff57874d11e4f8eb0e80c4a`. This command runs from the repository root with Bun:

```sh
bun -e '
import { shallowEqual } from "./packages/store/src/utils/shallow-equal.ts";
const before = new Array(1);
const after = [42];
assert.equal(shallowEqual(before, after), false);
assert.equal(shallowEqual(after, before), false);
'
```

## Root cause

`Array.prototype.every` skips empty positions in the first array. Thus, `new Array(1)` compares equal to `[42]`, but the reverse comparison returns `false`.

## Change

The shared comparator visits each array index. Shallow selectors accept values that replace empty positions. Empty positions still compare as `undefined`. Length checks and `Object.is` comparisons stay unchanged.

## Verification

The reproduction and the regression test returned `true` instead of `false` before the correction. The same checks pass after the correction.

- `pnpm --filter @assistant-ui/store test src/utils/shallow-equal.test.ts` passes all four tests.
- `pnpm --filter @assistant-ui/tap build && pnpm --filter @assistant-ui/store build` completes.
- Scoped `oxlint` and `oxfmt --check` pass for the changed TypeScript files.
- Direct API checks pass for the freshly built client export. Real hook checks cover filled positions, cleared positions, and equal-value reference reuse.

## Public surface

`@assistant-ui/store` receives a patch changeset. Public exports and signatures stay unchanged. Documentation, API-reference output, and templates need no changes.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6955?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.Yg-CIUC7_VtpeCJZSOaF3BRR2QHujDZSvPbTE-ar585g9WN6a_duBiDNkWM?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/shallow-sparse-array-positions.md
+++ b/.changeset/shallow-sparse-array-positions.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/store": patch
+---
+
+fix: detect values that replace empty positions in sparse arrays during shallow comparison
--- a/packages/store/src/utils/shallow-equal.test.ts
+++ b/packages/store/src/utils/shallow-equal.test.ts
@@ -1,0 +1,37 @@
+import { describe, expect, it } from "vitest";
+import { shallowEqual } from "./shallow-equal";
+
+describe("shallowEqual", () => {
+  it("detects a value in a previously empty array position", () => {
+    const sparse = new Array<number>(1);
+    const filled = [42];
+
+    expect(shallowEqual(sparse, filled)).toBe(false);
+    expect(shallowEqual(filled, sparse)).toBe(false);
+  });
+
+  it("compares sparse values at their array positions", () => {
+    const left = new Array<number>(2);
+    left[0] = 42;
+    const right = new Array<number>(2);
+    right[1] = 42;
+
+    expect(shallowEqual(left, right)).toBe(false);
+    expect(shallowEqual(right, left)).toBe(false);
+  });
+
+  it("treats empty array positions as undefined", () => {
+    const sparse = new Array<undefined>(1);
+
+    expect(shallowEqual(sparse, new Array<undefined>(1))).toBe(true);
+    expect(shallowEqual(sparse, [undefined])).toBe(true);
+    expect(shallowEqual([undefined], sparse)).toBe(true);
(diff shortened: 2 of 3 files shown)


okisdev (approved):
approving. `Array.prototype.every` skips holes, so the array branch was asymmetric: `shallowEqual(new Array(1), [42])` returned true while the reverse returned false. #5067 already closed the object dimension of this same comparator on the same reasoning. i ran this rather than read it: reverting `shallow-equal.ts` to the merge base while keeping your test file gives 1 failed, 3 passed, with the failure landing exactly on `shallowEqual(sparse, filled)`; on your head all four pass and the package suite is 198 green. treating a hole as `undefined` is also the right call, since a consumer reading by index cannot tell the two apart, and the `i in a` alternative would add two checks per element to a comparator that runs on every publish.

worth stating in the body rather than leaving implicit: no path inside this repo produces a sparse array, since every array reaching this comparator comes from map, filter, or a literal. the fix rests on `shallowEqual` being exported from `@assistant-ui/store/client`, which makes it the one instance a third party can feed a hole. the same `.every` pattern is still live in eight other comparators, tracked in #6967; those are module private with repo controlled inputs, so they do not block this and belong in their own sweep.

PR #6956, opened 2026-09-07
fix(react): keep empty mention categories empty

ephraimduncan:
## Problem

- An empty `categories` list exposes flat mention items instead of an empty result. The same error occurs after the last category disappears.
- Reproduced on base `7fd03e375b41101f9ff57874d11e4f8eb0e80c4a` (`@assistant-ui/react` 0.15.18). This component displays Alice before the correction and `[]` afterward:

```tsx
import { unstable_useMentionAdapter } from "@assistant-ui/react";

export function Repro() {
  const { adapter } = unstable_useMentionAdapter({
    items: [{ id: "alice", type: "person", label: "Alice" }],
    categories: [],
  });
  return <pre>{JSON.stringify(adapter.search?.(""))}</pre>;
}
```

## Root cause

- The hook selects categorized mode only when the category list contains entries. An empty list therefore activates the flat-item path.

## Change

- A supplied `categories` list takes precedence over `items`, including an empty list, as the existing options contract specifies.
- The existing categorized adapter handles the empty result. Tool-category selection and flat mode without categories remain unchanged.

## Verification

- `pnpm --filter @assistant-ui/react exec vitest run src/unstable/useMentionAdapter.test.tsx`: the two new regression cases failed before the correction. All nine tests passed afterward.
- An unmocked React DOM smoke check in JSDOM confirmed initial empty categories, populated categories, category removal, and omitted categories.
- Scoped Oxlint and formatting checks passed. `node scripts/check-changesets.mjs` passed.

## Public surface

- `@assistant-ui/react` receives a patch changeset. Public exports and signatures remain unchanged.
- Existing documentation already specifies category precedence. API-reference output and templates remain unchanged.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6956?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.1gcO5D4NUB3r88XImC9cBlMgBQjbvx7yGHAoVYRDEVm0oYSHDysQS559cd8?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/empty-mention-categories.md
+++ b/.changeset/empty-mention-categories.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react": patch
+---
+
+fix: ignore flat mention items when categories is an empty list
--- a/packages/react/src/unstable/useMentionAdapter.test.tsx
+++ b/packages/react/src/unstable/useMentionAdapter.test.tsx
@@ -11,7 +11,10 @@ import type { Unstable_TriggerAdapter } from "@assistant-ui/core";
 import { useState } from "react";
 import { beforeEach, describe, expect, it, vi } from "vitest";
 import { TriggerNavigationResource } from "../primitives/composer/trigger/triggerNavigationResource";
-import { unstable_useMentionAdapter } from "./useMentionAdapter";
+import {
+  unstable_useMentionAdapter,
+  type Unstable_UseMentionAdapterOptions,
+} from "./useMentionAdapter";
 
 const runtime = vi.hoisted(() => {
   const state = {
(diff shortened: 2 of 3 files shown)


okisdev (approved):
approved. what convinced me is that this isn't a new rule, it's removing a contradiction the file already carried: `includeTools` on line 148 computes `!items && !categories`, and `![]` is `false`, so an empty `categories` array already counted as supplied for the tools default while `isCategorized` on line 155 said it did not. blame puts that `.length > 0` in #6706 as a mechanical carry of the older `if (categories && categories.length > 0)` truthiness guard, so nothing intentional is being undone.

i ran it rather than only reading it. 9/9 in a worktree with the fix, and reverting line 155 alone fails exactly your two new cases, so the repro holds in both directions. i also probed the one behavior that moves and the description does not mention: `categories: []` together with `includeModelContextTools: true` now yields a `Tools` drill-down group where it used to yield a flat list. that reads as the right side of the change, since it is what the options table in mentions.mdx and the `Unstable_ModelContextToolsOptions` JSDoc already promise, and it removes the mode flip a popover would do while async categories load. `useSlashCommandAdapter` is flat only, so nothing else carries the pattern.

one nit i am not asking you to change: line 148 asks the same question as line 155, just as `!categories` rather than `categories !== undefined`.

PR #6957, opened 2026-09-07
fix(store): expire falsy values in the last-valid cache

ephraimduncan:
## Problem

The last-valid cache retains `0`, `false`, `""`, and `NaN` after the stale window. Cached `null` and `undefined` bypass that window entirely.

The reproduction applies to base `7fd03e375b41101f9ff57874d11e4f8eb0e80c4a`, where `@assistant-ui/store` is version `0.3.12`. From the repository root, `bun repro.mjs` runs this snippet:

```js
import { createLastValidCache } from "./packages/store/src/utils/last-valid-cache.ts";

const pending = [];
let reports = 0;
const cache = createLastValidCache(
  () => { reports++; },
  callback => pending.push(callback),
);
assert.equal(cache.resolve(true, () => 0), 0);
assert.equal(cache.resolve(false, () => 99), 0);
for (const callback of pending.splice(0)) callback();
assert.equal(reports, 1);
assert.equal(cache.resolve(false, () => 99), 99);
```

## Root cause

The expiry condition uses truthiness, but the return expression uses nullish fallback. Neither condition distinguishes an empty cache from all permitted values.

## Change

All cached values remain available during the stale window and expire afterward. A private symbol marks an empty cache without a wrapper object for each update. Generation-based expiry cancellation stays unchanged.

## Verification

- The reproduction fails before the correction because `reports` remains `0`, and passes afterward.
- `pnpm --filter @assistant-ui/store test src/utils/last-valid-cache.test.ts`: six regression cases fail before the correction, and all 12 tests pass afterward.
- `pnpm exec turbo build --filter=@assistant-ui/store --force`: the store and tap builds pass without cached build output.
- A Node smoke check through the rebuilt `@assistant-ui/store/client` entry passes for nine values with `queueMicrotask`, including expiry and subsequent resolver errors.
- Scoped Oxlint checks and `pnpm changesets:check` pass.

## Public surface

`@assistant-ui/store/client` retains its exports and public signatures. The PR includes a store patch changeset. Documentation, API-reference output, and templates remain unchanged because their contracts do not change.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6957?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.dubXz6RlXsok2O9w4eJZXYQambrfRQGZ9U4TJ4_TTF_JGIfz3kgLuWga70c?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/last-valid-cache-falsy-values.md
+++ b/.changeset/last-valid-cache-falsy-values.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/store": patch
+---
+
+fix: expire falsy values in the last-valid cache and preserve cached null and undefined during the stale window.
--- a/packages/store/src/utils/last-valid-cache.test.ts
+++ b/packages/store/src/utils/last-valid-cache.test.ts
@@ -26,20 +26,27 @@ describe("createLastValidCache", () => {
     expect(cache.resolve(false, () => "fresh")).toBe("b");
   });
 
-  it("drops the cache and reports once the stale window settles", () => {
-    const scheduler = createScheduler();
-    const reportStale = vi.fn();
-    const cache = createLastValidCache<string>(reportStale, scheduler.schedule);
+  it.each([0, false, "", null, undefined, NaN, "a"])(
+    "drops cached %o and reports once the stale window settles",
+    (value) => {
+      const scheduler = createScheduler();
+      const reportStale = vi.fn();
+      const cache = createLastValidCache<unknown>(
+        reportStale,
+        scheduler.schedule,
+      );
 
-    cache.resolve(true, () => "a");
-    expect(cache.resolve(false, () => "fresh")).toBe("a");
-    scheduler.flush();
-    expect(reportStale).toHaveBeenCalledTimes(1);
+      expect(cache.resolve(true, () => value)).toBe(value);
+      expect(cache.resolve(false, () => "fresh")).toBe(value);
+      expect(cache.resolve(false, () => "fresh")).toBe(value);
+      scheduler.flush();
+      expect(reportStale).toHaveBeenCalledTimes(1);
 
-    const resolveItem = vi.fn(() => "fallthrough");
-    expect(cache.resolve(false, resolveItem)).toBe("fallthrough");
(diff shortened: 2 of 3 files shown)


okisdev (approved):
approving. the condition swap is the right fix and it lands on the cause: `if (last) expire()` and `last ?? resolveItem()` were both standing in for "is the cache populated", which neither can express for a generic `T`, so falsy entries never scheduled an expiry and cached `null` / `undefined` skipped the stale window entirely.

i ran the diff rather than reading it. a matrix over `0`, `false`, `""`, `NaN`, `-0`, `0n`, `null`, `undefined` reproduces both defects against `7fd03e375` and shows them gone after, and reverting `last-valid-cache.ts` to main while keeping your tests fails 6 of the 7 `it.each` cases, so that table is a real regression net rather than decoration. full store suite is 200/200 locally, oxlint, oxfmt and `changesets:check` clean. the `EMPTY` sentinel is also what the package already reaches for (`UNAPPLIED` in `tap-assistant-context.ts`), so it introduces no new pattern.

one thing worth recording rather than fixing: nothing in this repo can reach the bug today. every call site instantiates `T` as an object handle, and both consumers (`@assistant-ui/vue`, `@assistant-ui/svelte`) are still private at `0.0.0`. it earns the patch anyway, since `createLastValidCache` is a published export of `@assistant-ui/store/client`, hoisted there in #5913 precisely so the bridges could not drift, and a public generic that misbehaves for primitive `T` is worth closing before someone outside the repo instantiates it that way. just not a user-visible fix.

PR #6959, opened 2026-09-07
fix(mcp-docs-server): exit the proxy when stdin closes

ephraimduncan:
## Problem

the docs proxy stays alive after its client goes away. `runProxy` settles only when one of its two transports reports a close, and stdin reaching EOF is not one of those events, so the process keeps running behind an open HTTP event stream.

affected base: `7fd03e375b41101f9ff57874d11e4f8eb0e80c4a` (`@assistant-ui/mcp-docs-server` 0.3.0).

at the process level, against a local MCP server so no network is involved: start `src/stdio.ts` with `ASSISTANT_UI_MCP_URL` pointing at it, complete the initialize handshake, then close the child's stdin. on `main` the child is still alive 15 seconds later; with this change it exits with code 0.

the same thing reproduces in-process without a server, because `runProxy` itself never settles. after `pnpm install --frozen-lockfile`, from `packages/mcp-docs-server`:

```sh
node --import tsx --input-type=module <<'JS'
import { PassThrough } from 'node:stream';
import { runProxy } from './src/proxy.ts';
const stdin = new PassThrough();
const proxy = runProxy({
  url: new URL('https://example.invalid/mcp'),
  stdin,
  stdout: new PassThrough(),
});
const timeout = setTimeout(() => {
  throw new Error('Proxy did not close after EOF');
}, 2000);
stdin.end();
await proxy;
clearTimeout(timeout);
console.log('Proxy closed after EOF');
JS
```

## Root cause

`StdioServerTransport.start()` subscribes to `data` and `error` on stdin and nothing else, and its `onclose` fires only from an explicit `close()`. nothing turns "stdin is finished" into a transport close, so `runProxy` waits on `closed` forever.

## Change

`runProxy` hands stdin to `stream.finished(input, { writable: false })` and closes the stdio transport from that callback, through the existing shutdown path that closes the HTTP transport with it. reaching for the platform primitive instead of a hand-rolled `end` listener is what makes this cover every way stdin can stop rather than one of them: a clean EOF, an EOF on a stream that never emits `close`, a stream destroyed without EOF, and a stream that had already finished before the proxy started. the cleanup function `finished` returns runs in `finally`, so the proxy leaves no listeners behind on a `process.stdin` that outlives it.

## Verification

- `pnpm --filter @assistant-ui/mcp-docs-server test`: 5 passed. each of the four new cases times out against `main`'s `proxy.ts` and passes with the change.
- process level: with a local MCP server and a completed handshake, the CLI is still alive 15 seconds after stdin EOF on `main`, and exits with code 0 with this change.
- `tsc --noEmit` is clean, scoped oxlint and oxfmt pass, and `pnpm changesets:check` and `pnpm changeset-semver:check` pass.

known and left out of this PR: closing the HTTP transport aborts its in-flight request, so a clean shutdown now prints one `HTTP transport error [AbortError]` line on stderr. that is shutdown log policy rather than the hang, and pinning it needs a streaming SSE test harness this file does not have yet, so it is tracked in #6968.

## Public surface

`@assistant-ui/mcp-docs-server` gets a patch changeset. exports and signatures are unchanged. docs, api-reference output, and templates are unchanged.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6959?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.BTGAn59_WiNtrbw39SL6y_fN98E9CQJ-CHroox2xxWxI09b7vVnIxUA7hNg?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/docs-proxy-stdin-eof.md
+++ b/.changeset/docs-proxy-stdin-eof.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/mcp-docs-server": patch
+---
+
+fix: close the docs proxy and its HTTP transport once stdin can no longer deliver messages, whether it reaches EOF or is destroyed
--- a/packages/mcp-docs-server/src/proxy.test.ts
+++ b/packages/mcp-docs-server/src/proxy.test.ts
@@ -1,4 +1,5 @@
 import { describe, expect, it, vi } from "vitest";
+import { once } from "node:events";
 import {
   createServer,
   type IncomingMessage,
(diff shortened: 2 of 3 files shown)


okisdev (approved):
approving. the diagnosis was right: `StdioServerTransport` subscribes to `data` and `error` on stdin and nothing else, so nothing turns EOF into a transport close and `runProxy` waits on `closed` forever. i rewrote the shutdown to `stream.finished(input, { writable: false })` and rebased onto main.

the reason for swapping out the `end` listener is that `"end"` is one of three ways a readable stops. i ran your version against a `PassThrough` once per termination: `stdin.end()` closed, while `stdin.destroy()` and `stdin.destroy(new Error(...))` both still had `runProxy` pending when the probe gave up, which is the same failure this PR set out to fix. `finished` also invokes its callback for a stream that already finished, which is what the `readableEnded` probe was doing by hand, and it returns its own cleanup, which is what the `finally` `off` was doing by hand, so the general form came out smaller than the specific one. the four table cases in `proxy.test.ts` each time out against main's `proxy.ts` and pass on this head.

worth knowing for whoever reads this next: the repro in your description does not actually demonstrate the process hang. `StreamableHTTPClientTransport.start()` only allocates an `AbortController`, it never opens a socket, so with `example.invalid` nothing holds the event loop open and node exits on its own. spawning `src/stdio.ts` with stdin closed immediately returns 0 on main too. what your repro does show is that `runProxy` never settles, which is the real defect. to watch the process survive you need a live stream, so i pointed the CLI at a local MCP server, completed the initialize handshake, then closed the child's stdin: still alive 15 seconds later on main, exit code 0 with this change. that version is in the description now.

the conflict was #6939 landing the vitest globals import and the `isJSONRPCRequest` guard on main, which is also why the type-check caveat you noted is gone. `tsc --noEmit` is clean on the rebased head.

one thing left over: closing the HTTP transport aborts its in-flight request, so a clean exit now prints an `AbortError` line on stderr. pinning that needs the local-server harness in `proxy.test.ts` to stream its responses instead of buffering them through `arrayBuffer()`, which is more than this PR should carry, so it is #6968.

PR #6975, opened 2026-09-07
fix: harden the langchain converters against non-spec content blocks

### LINKED ISSUE #6768: harden the shared langchain content block converter against non-spec blocks
## problem

#6753 closed the container-level holes in `@assistant-ui/react-langchain`'s converter (`null` content, non-array content, `null` entries inside a content array). two holes in the same class are still open, both reachable from the same non-spec provider payloads the "defensive converters" rule in `AGENTS.md` targets.

### 1. a text block with no `text` produces a part that violates its own contract

`convertLangChainContentBlock` passes the block's payload field through unchecked at `packages/react-langchain/src/converter.ts:62`:

```ts
case "text":
case "text_delta":
  return { type: "text" as const, text: part.text };
```

so a block that is a well-formed object but carries no `text` survives #6753's `contentBlocks` guard (it only rejects non-objects) and yields a part violating the `text: string` contract:

```ts
convertLangChainBaseMessage(
  { _getType: () => "human", id: "m", content: [{ type: "text" }] },
  {},
);
// → [{ type: "text", text: undefined }]
```

that is not inert. `actionBarCopyDisabled` reads `part.text.length` over `s.message.parts` (`packages/core/src/store/primitive-predicates.ts:31`), so a thread carrying one such block throws `TypeError: Cannot read properties of undefined (reading 'length')` as soon as the copy action's disabled state is computed.

the guarding is inconsistent within the same switch: `image_url` returns `null` when its url is missing and `reasoning` coalesces with `part.reasoning ?? ""`, while `text`, `text_delta`, and `thinking` pass theirs through raw. `@assistant-ui/react-langgraph` imports this converter (`convertLangChainMessages.ts:18`), so one fix covers both packages.

### 2. `react-langgraph`'s `contentToParts` still has the array-entry gap

`packages/react-langgraph/src/convertLangChainMessages.ts:122` guards `content == null` and the string case, then maps without a per-entry guard, and the first thing the callback reads is `part.type` (line 132). a `null` or `undefined` entry throws `TypeError: Cannot read properties of null (reading 'type')`, the exact failure #6753 closed on the langchain side. that PR named this as a deliberate follow-up; the cross-runtime parity rule wants the two converters agreeing rather than diverging.

## suggested change

for (1), coalesce the missing field inside the shared `convertLangChainContentBlock` so both packages inherit it, following what `reasoning` already does.

for (2), lift #6753's `contentBlocks` filter into the langgraph converter, or share it, so both converters agree on what counts as a usable content block.

## verification

both cases reproduce on `main` today. each should get a colocated test that fails without the guard and passes with it, matching the shape #6753 used.

<a href="https://rupic.dev/s/s2.Adpi-Kl4g_kOzcIaKjP2SxoV-vjn_mpo5f9fgh4_K29ZqkQa9F6RwpvK6FQ?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

Kinfe123:
Closes #6768.

## Problem

Two holes remained in the class #6753 closed, both reachable from the same non-spec provider payloads the "defensive converters" rule targets, and both reproducing on main:

1. A well-formed text block with no `text` field survives #6753's container guard (which only rejects non-objects) and `convertLangChainContentBlock` passes the missing field through, yielding `{ type: "text", text: undefined }` — a part violating its own `text: string` contract, inherited by both `react-langchain` and `react-langgraph` through the shared converter.
2. `react-langgraph`'s `contentToParts` maps a content array without a per-entry guard and reads `part.type` first, so a `null` entry throws `TypeError: Cannot read properties of null (reading 'type')` — the exact failure #6753 fixed on the langchain side, named there as a deliberate follow-up.

## Change

Per the issue's suggested shape:

- The shared `convertLangChainContentBlock` coalesces a missing or non-string `text` to `""`, so both packages inherit the guard. Coalescing (rather than dropping, as the `reasoning` case does) preserves the existing behavior for streaming `text_delta` frames that legitimately carry empty text.
- `react-langgraph`'s `contentToParts` filters non-object entries before mapping — #6753's `contentBlocks` filter lifted into the langgraph converter rather than exported, keeping `@assistant-ui/react-langchain`'s published surface unchanged. This also aligns the two converters on tolerating stray string entries, per the cross-runtime parity rule.

One extra test pins a neighboring seam that turned out to be accidentally safe: `getStringContent` maps a textless block to `undefined`, but `Array.prototype.join` renders that as an empty string, so system text was never corrupted — pinned so a future refactor away from `join` cannot regress it silently.

## Verification

- Both repros fail on main exactly as filed: the coalesce test sees `text: undefined`, the langgraph null-entry test throws the issue's `TypeError`. Both pass with the fix.
- Full suites: `react-langchain` 140 passed, `react-langgraph` 315 passed. (An initial 127-test failure storm in langgraph was ruled out as stale local dists: the identical failures reproduce on a clean main checkout and disappear after `turbo build` of the dependency graph, with zero source changes involved.)

## Public surface

No API changes; no new exports. Patch changesets for `@assistant-ui/react-langchain` and `@assistant-ui/react-langgraph`.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6975?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.dWQiudeOhEqlQZR7Igd4KvYVKgnVIjeH9QD5ZSi1WHqmJOs0u73lDkkVRu0?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/langchain-converter-hardening.md
+++ b/.changeset/langchain-converter-hardening.md
@@ -1,0 +1,6 @@
+---
+"@assistant-ui/react-langchain": patch
+"@assistant-ui/react-langgraph": patch
+---
+
+fix: harden the shared content block converter against non-spec blocks
--- a/packages/react-langchain/src/convertMessages.test.ts
+++ b/packages/react-langchain/src/convertMessages.test.ts
@@ -6,6 +6,7 @@ import {
   getMessageContent,
   getMessageType,
 } from "./convertMessages";
+import { createLangChainStreamingTimingAccessors } from "./converter";
 import type { LangChainBaseMessage, UIMessage } from "./types";
 
 const humanMessage = (content: unknown): LangChainBaseMessage => ({
@@ -1148,7 +1149,28 @@ describe("convertLangChainBaseMessage malformed messages", () => {
     expect(contentOf(result)).toEqual([{ type: "text", text: "" }]);
   });
 
-  it("skips a textless block when collecting system text", () => {
+  it("skips null entries when measuring streamed text length", () => {
+    const { getTextLength } = createLangChainStreamingTimingAccessors<{
+      id?: string | undefined;
+      content?: unknown;
+      _getType: () => string;
+    }>((message) => message._getType());
+
+    expect(
+      getTextLength(
+        [
+          {
+            _getType: () => "ai",
+            id: "ai-1",
+            content: [null, { type: "text", text: "abc" }, undefined],
+          },
+        ],
+        "ai-1",
+      ),
+    ).toBe(3);
+  });
+
+  it("renders a textless block as empty system text", () => {
     const result = convertLangChainBaseMessage(
       {
         _getType: () => "system",
--- a/packages/react-langchain/src/convertMessages.test.ts
+++ b/packages/react-langchain/src/convertMessages.test.ts
@@ -1139,6 +1139,28 @@ describe("convertLangChainBaseMessage malformed messages", () => {
     expect(contentOf(result)).toEqual([{ type: "text", text: "kept" }]);
   });
 
+  it("coalesces a text block without a text field to empty text", () => {
+    const result = convertLangChainBaseMessage(
+      humanMessage([{ type: "text" }]),
+      {},
+    );
+
+    expect(contentOf(result)).toEqual([{ type: "text", text: "" }]);
+  });
+
+  it("skips a textless block when collecting system text", () => {
+    const result = convertLangChainBaseMessage(
+      {
+        _getType: () => "system",
+        id: "msg-7",
+        content: [{ type: "text" }, { type: "text", text: "kept" }],
+      },
+      {},
+    );
+
+    expect(contentOf(result)).toEqual([{ type: "text", text: "kept" }]);
+  });
+
   it("skips null entries when collecting system text", () => {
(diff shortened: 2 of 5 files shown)


Kinfe123:
Both review findings addressed in bcaf2c688: `getTextLength` in the streaming timing accessor now skips non-object entries before reading `part.type` — verified the new test throws the exact `TypeError` without the guard and passes with it (141 langchain + 315 langgraph tests green). And the system-text test is retitled to "renders a textless block as empty system text", since cubic is right that nothing is skipped there — `join` coalesces the undefined, which is exactly the accidental safety the pin exists to protect.

PR #6977, opened 2026-09-07
fix(python): let state demo streams complete

ephraimduncan:
## Problem

the string, list, and dictionary demos in `python/state-test` emit a `NoneType` error instead of state updates. all three fail on their first statement, so those routes have never run to completion.

reproduction on `98010f17b`, against the workspace `assistant-stream`:

```sh
cd python/state-test
uv sync --frozen
uv run --frozen python - <<'PY'
import asyncio
from server import string_test, list_test, dict_test

async def main():
    for handler in (string_test, list_test, dict_test):
        response = await handler()
        async for chunk in response.body_iterator:
            print(chunk, end="")

asyncio.run(main())
PY
```

each handler emits `3:"'NoneType' object does not support item assignment"` and raises `TypeError`.

## Root cause

`create_run` defaults to `state=None`, and `StateManager.state` returns `None` rather than a proxy while the root is `None`. these three callbacks assign dictionary entries before initializing the root, so the first `controller.state[...] = ...` indexes into `None`.

## Change

the three routes pass `state={}` to `create_run`, the same entry point `python/assistant-transport-backend` already uses (`create_run(run_callback, state=request.state)`). `simple-test` and `complex-test` keep their in-callback root assignment, since `simple-test` asserts `controller.state is None` as part of what it demonstrates.

the library default stays `None`. that default is contract (`test_create_run_detaches_initial_state`, plus the assertion in `simple-test`), so the fix belongs at the call site rather than in `create_run`.

an initial state emits no root `set` operation, so the first operation on the wire is `set /message`. that is safe: `GorpStreamAccumulator.updatePath` does `state ??= {}` before descending, so the client materializes the root on the first write.

## Verification

before the change, `string`, `list`, and `dict` each raise `TypeError` and emit a single error frame. after it, all five routes complete with no error frame (`simple` 2 frames, `complex` 7, `string` 6, `list` 11, `dict` 8).

accumulating the emitted operations client-side gives the expected final state for each fixed route:

```
[string] {"message": "Hello world!", "uppercase": "HELLO WORLD!", "contains_world": true, "length": 12}
[list]   {"items": [], "first_item": "item_0", "count": 7, "contains_batch": true}
[dict]   {"config": {}, "theme_unchanged": true, "missing_with_default": "default_value", "all_keys": ["theme", "language", "notifications"], "all_values": ["light", "en", true]}
```

## Public surface

none. `python/state-test` is a private demo (`package = false`) and no published package changes, so no changeset.


okisdev (approved):
approved. ran both sides in a worktree: on `98010f17b` the string, list, and dict handlers each raise `TypeError` and emit a single `3:` frame; on `543eceb19` all five routes complete clean. accumulating the emitted operations client-side also gives the right final state for the three fixed routes (`length: 12`, `count: 7`, `theme_unchanged: true`, both containers empty after their `clear()`), so the append-text path, `setdefault`, and `clear()` all propagate rather than merely not crashing. that part matters more than the crash going away, since these routes had never run to completion before.

not a regression, for the record. `StateManager` has set `_state_data = None` and returned `None` instead of a proxy since #1912, the same PR that added `state-test`, and `create_run(state=)` only arrived in #1967. those three routes have been broken their whole life, which is presumably why nobody noticed.

`state={}` over an in-callback `controller.state = {}` is the right call: it matches `python/assistant-transport-backend/main.py:145`, and it leaves `simple-test`'s `assert controller.state is None` intact as the documented default. dropping the changeset and `test_server.py` in the second commit was right too, since the changeset named `assistant-stream` for a diff that touches none of it, and `python/state-test` has no pytest wiring so that file would never have run. the README's browser UI at :8000 has never existed either, but that predates this and is filed as #7004.

PR #6978, opened 2026-09-07
fix(core): keep nested branch selection through ancestors

ephraimduncan:
## Problem

A selected nested branch disappears when an ancestor is selected again. The thread returns to the previously selected sibling branch.

Reproduction on `@assistant-ui/core` 0.3.17 at base `98010f17b4a8d4bc506a6084007ecdaf4a823503`: this snippet runs with `bun repro.ts` from the repository root.

```ts
import {
  MessageRepository,
  ExportedMessageRepository,
} from "./packages/core/src/runtime/utils/message-repository.ts";

const repo = new MessageRepository();
const add = (parentId: string | null, id: string) => {
  for (const { message } of ExportedMessageRepository.fromArray([
    { id, role: "user", content: [{ type: "text", text: id }] },
  ]).messages) {
    repo.addOrUpdateMessage(parentId, message);
  }
};
add(null, "r");
add("r", "a");
add("a", "x");
add("r", "b");
add("b", "y");
repo.switchToBranch("y");
assert.deepEqual(repo.getMessages().map(m => m.id), ["r", "b", "y"]);
repo.switchToBranch("r");
assert.deepEqual(repo.getMessages().map(m => m.id), ["r", "b", "y"]);
```

The last assertion receives `["r", "a", "x"]` before the correction.

## Root cause

Branch selection updates only the immediate parent's `next` pointer. Ancestor selection follows unchanged pointers to a different branch.

## Change

Branch selection updates each ancestor pointer, including the root pointer. The selected branch stays visible after ancestor selection. Stored descendants and sibling branches remain available.

## Verification

- The same API reproduction fails before the correction and passes afterward.
- Two regression cases fail before the correction and pass afterward. They cover ancestor selection and root selection after message deletion.
- `pnpm --filter @assistant-ui/core test src/tests/MessageRepository.test.ts src/runtime/utils/message-repository-session.test.ts`: 52 tests pass.
- Scoped Oxlint and TypeScript diagnostics report no errors. `pnpm changesets:check` passes.

## Public surface

The correction affects `@assistant-ui/core` and includes a patch changeset. Public exports and signatures remain unchanged. No documentation, API-reference, or template changes are necessary.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6978?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.6Am-y1NSJh9N8YKHLJQXAgIwuAGskLsg8cSNjbSRpVJzciiW9khAlnGmWV8?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/nested-branch-selection.md
+++ b/.changeset/nested-branch-selection.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/core": patch
+---
+
+fix: keep the selected nested branch when an ancestor is selected again
--- a/packages/core/src/runtime/utils/message-repository.ts
+++ b/packages/core/src/runtime/utils/message-repository.ts
@@ -381,8 +381,14 @@ export class MessageRepository {
       );
 
     const previousHead = this.head;
-    const prevOrRoot = message.prev ?? this.root;
-    prevOrRoot.next = message;
+    for (
+      let current: RepositoryMessage | null = message;
+      current;
+      current = current.prev
+    ) {
+      const prevOrRoot = current.prev ?? this.root;
+      prevOrRoot.next = current;
+    }
 
     this.head = findHead(message);
 
--- a/packages/core/src/tests/MessageRepository.test.ts
+++ b/packages/core/src/tests/MessageRepository.test.ts
@@ -240,6 +240,48 @@ describe("MessageRepository", () => {
       expect(messages2.map((m) => m.id)).toEqual(["parent-id", "branch2-id"]);
     });
 
+    it("keeps a nested branch selected when revisiting its ancestor", () => {
+      repository.addOrUpdateMessage(null, createTestMessage({ id: "r" }));
+      repository.addOrUpdateMessage("r", createTestMessage({ id: "a" }));
+      repository.addOrUpdateMessage("a", createTestMessage({ id: "x" }));
+      repository.addOrUpdateMessage("r", createTestMessage({ id: "b" }));
+      repository.addOrUpdateMessage("b", createTestMessage({ id: "y" }));
+
+      repository.switchToBranch("y");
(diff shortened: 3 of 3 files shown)


okisdev (approved):
approved. the change restores the invariant that the chain of `next` pointers from the root down to the head is exactly the head path, which is what `findHead` reads whenever the head is recomputed from an ancestor. updating only the immediate parent left every higher pointer aimed at whatever branch was selected last, so `getMessages()` still looked right (it walks `prev` upward) while any later read from above fell back into the stale subtree.

i ran both new cases against main and they fail there (`['r','a','x']` and `['a','x']`) and pass with the change, plus the full `@assistant-ui/core` suite, 164 files and 1752 tests green. the deletion case is the one that bites hardest in practice: `deleteMessage` recomputes the head through `findHead(this.root)`, so a stale root pointer moved the thread to an unrelated top-level branch after removing a single message. good call covering it.

one thing standing, and it does not block this: `performOp` has the same single-pointer update in its `relink` branch, so re-parenting an existing message through `addOrUpdateMessage` still leaves the pointers above the new parent stale. filed as #7001, along with a note that `resetHead` already carries a loop identical to the one added here, so the three sites are worth collapsing into one private helper.

PR #6981, opened 2026-09-07
fix(core): keep discarded text out of dictation after reset

ephraimduncan:
## Problem

Dictation can restore discarded composer text after `reset()`. The next transcript must start from the empty draft, not the discarded draft.

Reproduction on base `98010f17b4a8d4bc506a6084007ecdaf4a823503` (`@assistant-ui/core` version `0.3.17`): after dependency installation and the core build, save this snippet as `repro.ts` at the repository root. Run `bun repro.ts`.

```ts
import { LocalRuntimeCore } from "./packages/core/src/runtimes/local/local-runtime-core.ts";

let emit;
const session = {
  status: { type: "running" },
  stop: async () => {},
  cancel: () => {},
  onSpeech: (callback) => { emit = callback; return () => {}; },
  onSpeechStart: () => () => {},
  onSpeechEnd: () => () => {},
};
const core = new LocalRuntimeCore({ adapters: {
  chatModel: { run: async () => ({ content: [] }) },
  dictation: { listen: () => session },
}}, undefined);
const composer = core.threads.getMainThreadRuntimeCore().composer;
try {
  composer.setText("old");
  composer.startDictation();
  await composer.reset();
  emit({ transcript: "new", isFinal: true });
  assert.equal(composer.text, "new");
} finally {
  composer.stopDictation();
  await Promise.resolve();
}
```

The base produces `"old new"`. The correction produces `"new"`. The adapter supplies speech results without a microphone.

## Root cause

Reset clears the composer text directly but retains the dictation baseline. The next speech result appends to that discarded baseline.

## Change

The reset helper uses the existing dictation rebase helper to clear the baseline and interim transcript. The dictation session stays active.

## Verification

- The runtime reproduction and regression failed before the correction and passed afterward.
- The regression also covers reset during an interim transcript, followed by interim and final speech results.
- `pnpm --filter @assistant-ui/core test src/tests/base-composer-runtime-core.test.ts src/tests/base-composer-runtime-core-send.test.ts src/tests/base-composer-runtime-core-addAttachment.test.ts src/tests/default-edit-composer-runtime-core.test.ts`: 93 tests passed.
- The core build and scoped oxlint check passed.
- Existing `QuoteInfo` type errors remain in the unchanged `retractDraft` test. This PR does not correct those errors.

## Public surface

`@assistant-ui/core` receives a patch changeset. Public exports and signatures, documentation, API-reference output, and templates stay unchanged.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6981?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.sIIg9oXcCyEcexrQ68WUdWJnkbD5BodqhDdae1Y82CwUSPSJvjsJ7qG_VLY?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/clear-dictation-draft.md
+++ b/.changeset/clear-dictation-draft.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/core": patch
+---
+
+fix: keep discarded composer text out of dictation results after reset
--- a/packages/core/src/runtime/base/base-composer-runtime-core.ts
+++ b/packages/core/src/runtime/base/base-composer-runtime-core.ts
@@ -159,6 +159,7 @@ export abstract class BaseComposerRuntimeCore
   private _emptyTextAndAttachments() {
     this._attachments = [];
     this._text = "";
+    this._rebaseDictation("");
     this._notifySubscribers();
   }
 
--- a/packages/core/src/tests/base-composer-runtime-core.test.ts
+++ b/packages/core/src/tests/base-composer-runtime-core.test.ts
@@ -103,6 +103,45 @@ describe("BaseComposerRuntimeCore", () => {
     expect(listener).not.toHaveBeenCalled();
   });
 
+  it("reset keeps discarded text out of later dictation results", async () => {
+    let emitSpeech!: (result: DictationAdapter.Result) => void;
+    composer.setDictationAdapter({
+      listen: () => ({
+        status: { type: "running" },
+        stop: async () => {},
+        cancel: () => {},
+        onSpeechStart: () => () => {},
+        onSpeechEnd: () => () => {},
+        onSpeech: (callback) => {
+          emitSpeech = callback;
+          return () => {};
+        },
+      }),
+    });
+
+    composer.setText("old");
+    composer.startDictation();
+    try {
+      await composer.reset();
+      expect(composer.text).toBe("");
+      emitSpeech({ transcript: "new", isFinal: true });
+      expect(composer.text).toBe("new");
+
+      emitSpeech({ transcript: "pending", isFinal: false });
+      expect(composer.text).toBe("new pending");
+      await composer.reset();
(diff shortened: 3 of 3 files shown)


okisdev (approved):
approving. the fix is at the cause: `_dictationBaseText` is the live session's anchor, and the file already carries the invariant that any write bypassing `setText` has to move it. `reset()` was the last direct write that didn't, and every other one already holds the line, `send()` cancels the session before clearing, `restoreDraft` and `retractDraft` both rebase. putting the call in `_emptyTextAndAttachments` next to the `_text = ""` it guards is the right place for it, and that helper has no other caller.

verified in a worktree rather than read: dropping the one line makes the new test fail with `expected 'old new' to be 'new'`, which is exactly the repro in the description, and restoring it gives 93 passed across the four composer suites and 1751 passed for the whole core package. both changed files are clean under tsc, oxlint and oxfmt. i also went through the early return in `reset()`, since it skips the rebase when the text is already empty, but `_text` is always `_dictationBaseText + _currentInterimText`, so an empty text means both are empty and there is nothing left to rebase.

one thing in the description no longer holds: the `QuoteInfo` errors you mention in the `retractDraft` test are not on this head. tsc reports nothing in either file you touched; core's 103 pre-existing errors are all in `model-context` and other untouched files.

PR #6983, opened 2026-09-07
fix: forward every part renderer option through MessagePrimitive.Parts

ephraimduncan:
## Problem

`MessagePrimitive.Parts` silently drops part renderer options the caller configured. Each of the three distributions drops a different subset, so the same `components` object renders differently depending on which package you import from.

| | `generativeUI` | `data` under `ChainOfThought` |
|---|---|---|
| `@assistant-ui/react` | forwarded | **dropped** |
| `@assistant-ui/react-native` | **dropped** | **dropped** |
| `@assistant-ui/react-ink` | **dropped** | forwarded |

Reproduction, against an assistant message carrying a `generative-ui` part and two `data` parts:

```tsx
<MessagePrimitive.Parts
  components={{
    generativeUI: { components: { Card } },
    data: { by_name: { chart: Chart } },
    // and, for the second column, ChainOfThought
  }}
/>
```

On `@assistant-ui/react` before this change, turning on `ChainOfThought` renders `AnswerThought groupResult` where `AnswerThought groupResultChart: 42` is expected: the chart renderer is gone. On `@assistant-ui/react-native` the same message renders `AnswerChart: 42Data: 7` with the generative UI missing in both grouping modes. On `@assistant-ui/react-ink` the wrapper never puts `generativeUI` on the map at all.

## Root cause

Each platform wrapper rebuilds the component map key by key before handing it to the shared `MessagePrimitiveParts`, so a key that nobody remembered to re-add is dropped. `generativeUI` was added to `BaseComponents` in #3967 and only `@assistant-ui/react` was updated. The `ChainOfThought` branch made it worse by enumerating a second, shorter key list, which is where `data` fell out on web and native even though `data` lives on `BaseComponents` and is valid in both grouping modes.

## Change

The wrappers now spread the caller's `components` and only layer the platform defaults on top, so the grouping branch decides *which defaults get injected* rather than *which keys survive*. Anything core adds to `BaseComponents` later reaches the shared renderer without a wrapper edit.

That also removes a pile of no-op injections: `messagePartsDefaultComponents` is core's own `defaultComponents` (aliased at `core/src/react/index.ts:284`), and core already destructures those same fallbacks itself (`MessageParts.tsx:407-412`) plus `components?.ToolGroup ?? defaultComponents.ToolGroup` and the `ReasoningGroup` equivalent. Re-supplying identical references from each wrapper did nothing. What is left is genuinely platform-specific:

- `@assistant-ui/react`: `Text`, `Image`
- `@assistant-ui/react-native`: `Text`
- `@assistant-ui/react-ink`: `Text`, `Image`, `Source`, `File`, the terminal-safe `data.Fallback`, and `Reasoning` in the standard branch

Ink keeps its early return so `Reasoning` stays off the map under `ChainOfThought`, which is what its existing contract test asserts and what core's `ChainOfThoughtComponents` type requires.

## Verification

- `packages/react/src/tests/MessageParts.rendererOptions.test.tsx` (new) and `packages/react-native/src/primitives/message/MessageParts.test.tsx` (new) render a message with `generative-ui` and `data` parts through both grouping modes and back. `packages/react-ink/src/tests/MessageParts.test.tsx` gains a `generativeUI` forwarding case in both modes.
- Each new test fails on the pre-change wrapper and passes after: web `AnswerThought groupResultUnavailable: Unknown` vs the expected string, native `AnswerChart: 42Data: 7` vs the expected string, ink `expected undefined to be { Object (components) }`.
- Full suites pass: `@assistant-ui/react` 570, `@assistant-ui/react-native` 256, `@assistant-ui/react-ink` 282.
- `pnpm lint`, `pnpm changesets:check`, `pnpm build`, and `pnpm size:check` pass. `tsc --noEmit` on the three packages reports only errors that are pre-existing on `main` in unrelated test files.

## Public surface

No export, signature, or type changes. Patch changesets for `@assistant-ui/react`, `@assistant-ui/react-native`, and `@assistant-ui/react-ink`. Every size budget stays within tolerance, so `size-budgets.json` is untouched. No documentation or API-reference impact.

### DIFF
--- a/.changeset/ink-part-renderer-options.md
+++ b/.changeset/ink-part-renderer-options.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react-ink": patch
+---
+
+fix: forward every part renderer option through `MessagePrimitive.Parts`, so configured generative UI renders
--- a/.changeset/native-part-renderer-options.md
+++ b/.changeset/native-part-renderer-options.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react-native": patch
+---
+
+fix: render configured generative UI in MessagePrimitive.Parts and keep data renderers when ChainOfThought is active.
--- a/.changeset/native-part-renderer-options.md
+++ b/.changeset/native-part-renderer-options.md
@@ -2,4 +2,4 @@
 "@assistant-ui/react-native": patch
 ---
 
-fix: render configured generative UI in MessagePrimitive.Parts and keep data renderers when ChainOfThought is active.
+fix: forward every part renderer option through `MessagePrimitive.Parts`, so configured generative UI renders and data renderers survive `ChainOfThought`
--- a/.changeset/web-part-renderer-options.md
+++ b/.changeset/web-part-renderer-options.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react": patch
+---
+
+fix: forward every part renderer option through `MessagePrimitive.Parts`, so data renderers survive `ChainOfThought`
--- a/packages/react-ink/src/primitives/message/MessageParts.tsx
+++ b/packages/react-ink/src/primitives/message/MessageParts.tsx
@@ -33,35 +33,28 @@ export namespace MessagePrimitiveParts {
 const mergeWithInkDefaults = (
   components: NonNullable<MessagePrimitiveParts.Props["components"]>,
 ): NonNullable<MessagePrimitiveParts.Props["components"]> => {
-  const shared = {
+  const inkOverrides = {
     Text: components.Text ?? inkDefaultComponents.Text,
     Image: components.Image ?? inkDefaultComponents.Image,
     Source: components.Source ?? inkDefaultComponents.Source,
     File: components.File ?? inkDefaultComponents.File,
-    Unstable_Audio:
-      components.Unstable_Audio ?? messagePartsDefaultComponents.Unstable_Audio,
     data: components.data
       ? {
-          by_name: components.data.by_name,
+          ...components.data,
           Fallback:
             components.data.Fallback ?? inkDefaultComponents.data.Fallback,
         }
       : inkDefaultComponents.data,
-    Quote: components.Quote,
-    Empty: components.Empty,
   };
 
   if ("ChainOfThought" in components) {
-    return { ...shared, ChainOfThought: components.ChainOfThought };
+    return { ...components, ...inkOverrides };
   }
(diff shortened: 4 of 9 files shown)


ephraimduncan:
Confirmed the web data-renderer defect. PR #6988 contains the separate correction and links to this PR. The regression reproduces the missing content before the correction and passes afterward.

okisdev (approved):
approving. the native fix was right in shape but only a third of the bug: the same key-dropping wrapper is in `@assistant-ui/react` and `@assistant-ui/react-ink`, so i pushed it across all three rather than leave two distributions rendering the same `components` object differently.

what i ran, per wrapper: revert to the pre-change merge, run the new test, restore. web printed `AnswerThought groupResultUnavailable: Unknown` where `...Chart: 42Data: 7` was expected, so `data` was gone the moment `ChainOfThought` was set. native printed `AnswerChart: 42Data: 7`, missing generative UI in both grouping modes. ink returned `expected undefined to be { Object (components) }`, since `generativeUI` never reached the map at all. all three pass after. full suites 570 / 256 / 282, plus `pnpm lint`, `changesets:check`, `build`, and `size:check` (every entry within tolerance, so `size-budgets.json` is untouched).

the part worth knowing for next time: everything the wrappers injected below `Text` was already core's own work. `messagePartsDefaultComponents` is the same object as core's `defaultComponents` (aliased at `core/src/react/index.ts:284`), `MessagePartComponent` destructures those exact fallbacks itself, and the grouped path does `components?.ToolGroup ?? defaultComponents.ToolGroup` plus the `ReasoningGroup` equivalent. that much is a reading of core; the suites passing with the injections removed is the behavioral half. so the grouping branch now decides which defaults get injected, never which keys survive, and only genuinely platform-specific slots remain: `Text` and `Image` on web, `Text` on native, and ink's larger set.

one trap in ink. i first wrote its merge as a single object with `...("ChainOfThought" in components ? {} : { Reasoning })`, which typechecks on web and native only because those two cast with `as any`. ink annotates its return, and TypeScript flattens the conditional spread into `Reasoning?: ReasoningMessagePartComponent` across the whole result, which cannot satisfy `Reasoning?: never` on the `ChainOfThoughtComponents` arm. the early return is what narrows `components` before the spread, so it is load-bearing rather than style.

#6988 is superseded by this. it moves `data` out of the standard branch on web while leaving the enumeration in place, and it edits the same file, so whichever lands second conflicts.

PR #6985, opened 2026-09-07
fix(react): keep parent ids distinct from ungrouped parts

ephraimduncan:
## Problem

`Unstable_PartsGroupedByParentId` can combine unrelated text parts into one group. It also removes parent IDs that start with `__ungrouped_`.

This reproduces on `@assistant-ui/react` 0.15.18 at base `98010f17b4a8d4bc506a6084007ecdaf4a823503`:

```tsx
import {
  AssistantRuntimeProvider,
  MessagePrimitive,
  ThreadPrimitive,
  useExternalStoreRuntime,
  type ThreadMessageLike,
} from "@assistant-ui/react";

const messages: ThreadMessageLike[] = [{
  id: "message",
  role: "assistant",
  content: [
    { type: "text", text: "standalone" },
    { type: "text", text: "child", parentId: "__ungrouped_0" },
  ],
}];

function Message() {
  return <MessagePrimitive.Unstable_PartsGroupedByParentId components={{
    Text: ({ text }) => <span>{text}</span>,
    Group: ({ groupKey, indices, children }) => (
      <section>
        <header>{JSON.stringify({ groupKey, indices })}</header>
        {children}
      </section>
    ),
  }} />;
}

export default function Example() {
    messages,
    convertMessage: (message) => message,
    onNew: async () => {},
  });
  return <AssistantRuntimeProvider runtime={runtime}>
    <ThreadPrimitive.MessageByIndex index={0} components={{ Message }} />
  </AssistantRuntimeProvider>;
}
```

Before: one group with indices `[0, 1]` and no parent ID. After: separate groups with indices `[0]` and `[1]`, with the second parent ID intact.

## Root cause

Generated ungrouped keys share the string namespace with parent IDs. A prefix check also treats valid parent IDs as ungrouped keys.

## Change

Ungrouped parts use numeric indices as internal keys. Parent IDs remain strings. Group order and public signatures stay unchanged.

## Verification

- The real-runtime regression failed before the correction and passed afterward with the same command: `pnpm --filter @assistant-ui/react test src/primitives/message/MessagePartsGrouped.test.tsx`.
- The regression and adjacent `MessagePartText.test.tsx` passed together: three tests.
- Browser checks covered the collision and content updates with non-adjacent groups, prefix IDs, numeric-string IDs, and empty IDs.
- The React package build, scoped Oxlint check, and `pnpm changesets:check` passed.
- API-reference generation completed with no tracked changes.

## Public surface

`@assistant-ui/react` receives a patch changeset. The grouping guide and source example use the corrected keys. Exports and public types stay unchanged.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6985?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.c9PJ3dBiotiQoEaS8DRKnWi-320QSfdwma3QnXOOd3FTmXo_TJEW-8AYRyc?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/distinct-parent-groups.md
+++ b/.changeset/distinct-parent-groups.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react": patch
+---
+
+fix: keep parent IDs separate from ungrouped message parts
--- a/apps/docs/content/docs/guides/part-grouping.mdx
+++ b/apps/docs/content/docs/guides/part-grouping.mdx
@@ -496,17 +496,17 @@ Use `MessagePrimitive.Unstable_PartsGrouped` only when you need to collect non-a
 ```tsx
 <MessagePrimitive.Unstable_PartsGrouped
   groupingFunction={(parts) => {
-    const groups = new Map<string, number[]>();
+    const groups = new Map<string | number, number[]>();
 
     parts.forEach((part, index) => {
-      const key = part.parentId ?? `__ungrouped_${index}`;
+      const key = part.parentId ?? index;
       const indices = groups.get(key) ?? [];
       indices.push(index);
       groups.set(key, indices);
     });
 
     return Array.from(groups.entries()).map(([key, indices]) => ({
-      groupKey: key.startsWith("__ungrouped_") ? undefined : key,
+      groupKey: typeof key === "string" ? key : undefined,
       indices,
     }));
   }}
--- a/packages/react/src/primitives/message/MessagePartsGrouped.test.tsx
+++ b/packages/react/src/primitives/message/MessagePartsGrouped.test.tsx
@@ -1,0 +1,93 @@
+// @vitest-environment jsdom
+
+import { render, screen } from "@testing-library/react";
+import { describe, expect, it } from "vitest";
+import type { ThreadMessageLike } from "@assistant-ui/core";
+import {
+  AssistantRuntimeProvider,
+  useExternalStoreRuntime,
+} from "@assistant-ui/core/react";
+import { ThreadPrimitiveMessageByIndex } from "../thread/ThreadMessages";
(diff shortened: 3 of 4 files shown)


okisdev (approved):
approved. the collision is real and both halves of it reproduce: with `parentId: "__ungrouped_0"` the old keying folds index 0 and index 1 into one group and erases the parent id, and any id starting with `__ungrouped_` loses its `groupKey` to the prefix check.

what i actually ran, at the refreshed head `5b2909d3`: your regression test against a reverse applied source hunk, which fails with `indices: "0,1"` where it expects `"0"` and `"1"`, then passes once the fix is back. also the wider `src/primitives/message`, `src/primitives/messagePart` and `src/tests` suites, 87 tests green. the number versus string split is the right key because `Map` compares with SameValueZero, so `parentId: "1"` and positional key `1` stay distinct; i checked that case and `parentId: ""` produce exactly the grouping they did before.

read but not executed: `MessageGroupedParts` carries no equivalent sentinel, so the primitive this one is deprecated in favour of has no parallel bug. one thing still standing that does not block: the tool name example just below the one you corrected still destructures `key` for a prop named `groupKey`, and imports a `groupMessagePartsByToolName` that nothing exports. both predate this change.

PR #6986, opened 2026-09-07
fix(react-native): supply status to registered part renderers

ephraimduncan:
## Problem

Registered tool and data renderers can crash when they read `status.type`. They receive no status, even when the runtime has one.

The affected checkout is `98010f17b4a8d4bc506a6084007ecdaf4a823503`, where `@assistant-ui/react-native` is version `0.1.40`.

<details>

In that checkout, mount this component from `packages/react-native`. The tool renderer throws instead of displaying `running`.

```tsx
import { Text } from "react-native";
import type { ThreadMessageLike } from "@assistant-ui/core";
import {
  AssistantRuntimeProvider,
  MessageByIndexProvider,
  useAssistantToolUI,
  useExternalStoreRuntime,
} from "@assistant-ui/core/react";
import { MessageContent } from "./src/primitives/message/MessageContent";

const message: ThreadMessageLike = {
  id: "m1",
  role: "assistant",
  status: { type: "running" },
  content: [{
    type: "tool-call",
    toolName: "search",
    toolCallId: "c1",
    args: {},
    argsText: "{}",
  }],
};

function Content() {
  useAssistantToolUI({
    toolName: "search",
    render: ({ status }) => <Text>{status.type}</Text>,
  });
  return <MessageContent />;
}

export default function Repro() {
    messages: [message],
    convertMessage: (value) => value,
    onNew: async () => {},
  });
  return (
      <MessageByIndexProvider index={0}>
        <Content />
      </MessageByIndexProvider>
    </AssistantRuntimeProvider>
  );
}
```

</details>

## Root cause

`MessageContent` reads raw `message.content`. Registered renderer props require the derived status that the runtime supplies through `message.parts`.

## Change

Select `message.parts` so registered tool and data renderers receive their status. Keep renderer selection, fallback renderers, and tool callbacks unchanged.

## Verification

- The real-runtime regression produced two `status.type` crashes before the correction. The same command passes all 16 tests afterward: `pnpm --filter @assistant-ui/react-native test src/primitives/message/MessageContent.test.tsx`.
- A browser preview reproduced the crashes and then displayed tool `running`, `requires-action`, and `complete` states, plus data `running` and `complete` states. This check used React Native Web, not a device simulator.
- The native package build, scoped Oxlint checks, TypeScript diagnostics, and `pnpm changesets:check` pass.

## Public surface

`@assistant-ui/react-native` receives a patch changeset. Public exports and signatures stay unchanged. Documentation and generated API output need no changes.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6986?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.5Qa793-FhJviMj2VDwoOvNSozpq8ma5bFVBrf4f0nua2wwdVvEMmeFUasmA?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/native-renderer-part-status.md
+++ b/.changeset/native-renderer-part-status.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/react-native": patch
+---
+
+fix: supply derived status to registered tool and data renderers
--- a/packages/react-native/src/primitives/message/MessageContent.test.tsx
+++ b/packages/react-native/src/primitives/message/MessageContent.test.tsx
@@ -1,6 +1,7 @@
 import { act, type ReactElement } from "react";
 import { createRoot, type Root } from "react-dom/client";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import type { MessagePartState, ThreadMessageLike } from "@assistant-ui/core";
 import { MessageContent } from "./MessageContent";
 
 type AnyPart = { type: string; [key: string]: unknown };
@@ -10,7 +11,12 @@ const h = vi.hoisted(() => ({
   resumeToolCall: vi.fn(),
   respondToToolApproval: vi.fn(),
   state: {
-    message: { content: [] as AnyPart[] },
+    message: {
+      content: [] as AnyPart[],
+      get parts() {
+        return this.content;
+      },
+    },
     tools: { toolUIs: {} as Record<string, unknown> },
     dataRenderers: { renderers: {} as Record<string, unknown> },
   },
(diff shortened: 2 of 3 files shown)


okisdev (approved):
approving. right root cause: registered renderer props are `MessagePartState & ...`, so anything reading `status.type` needs the derived parts, and react-native was the last distribution still selecting raw `content` while react-ink and react's grouped parts had already moved.

verified in a worktree rather than off the diff. reverting the selector to `s.message.content` fails 2 of the file's tests on `status.type`, and the whole package is 256/256 with it in place, so the new test guards the thing it names. the index alignment the tool callbacks rely on holds too: `parts` is `runtimeState.content.map(...)` in message-runtime-client.ts:71, so `aui.message.part({ index })` still resolves the same part. that last one is a reading, not something i ran.

i pushed one commit on top. the trap is that the internal display props declared `part: ToolCallMessagePart` / `DataMessagePart` and only reached status through the `as ToolCallMessagePartProps` casts, so they looked like they carried status while nothing enforced it: with your commit alone, `tsc --noEmit` passed clean after i reverted the selector back to `content`. 00afb1ef5 types those props as `Extract<MessageContentStatePart, {...}>` the way react-ink does at MessageContent.tsx:22,83,117, and the same revert now errors with TS2375 and TS2322. public `MessageContentProps` is untouched; the fallback prop just moves to the `Extract<MessageContentPart, ...>` form `renderToolCall` already declares.

one thing i left alone: the new describe block calls `vi.doUnmock` and `vi.resetModules()` inside the test body, so it quietly depends on running after the mocked blocks in the same file. fine as the file stands, worth knowing if the suite ever gets shuffled.

PR #6987, opened 2026-09-07
fix: compare arrays with indexed loops in the remaining comparators

### LINKED ISSUE #6967: array shallow comparators skip holes via Array.prototype.every
## problem

`Array.prototype.every` skips empty positions, so any comparator built on `a.every((v, i) => ... b[i])` reads a hole in `a` as "no element to check" while reading the matching hole in `b` as `undefined`. the result is an asymmetric equality: `cmp(new Array(1), [42])` returns true, `cmp([42], new Array(1))` returns false.

#6955 fixed this in `shallowEqual` (`packages/store/src/utils/shallow-equal.ts`) by replacing `.every` with an indexed loop. that was the instance that mattered most, since `shallowEqual` is exported from `@assistant-ui/store/client` and `/internal` and a third party can feed it anything.

## remaining sites

the same pattern is still live in eight comparators:

- `packages/core/src/store/clients/model-context-client.ts:15`
- `packages/core/src/store/clients/suggestions.ts:28`
- `packages/react-markdown/src/primitives/MarkdownText.tsx:91`
- `packages/react-streamdown/src/primitives/StreamdownText.tsx:68`
- `packages/react/src/mcp-apps/app-frame.tsx:48`
- `packages/react/src/unstable/webmcp/useWebMcpProvider.ts:68`
- `packages/react-mcp/src/resources/McpServerResource.ts:91`
- `packages/ui/src/components/react/assistant-ui/elements/conversation-map.aui.tsx:186`

(`packages/ui/src/components/react/assistant-ui/elements/markdown-text.tsx:33` iterates `Object.keys(value)`, which never contains holes, so it is not affected.)

## severity

low, and deliberately so. all eight are module private, and their inputs are built from `map`, `filter`, `sort`, or literals, so none of them can receive a sparse array as the code stands. this is a latent contract defect rather than a live bug, which is why it is filed rather than folded into #6955.

## fix

replace each `.every` with an indexed loop, matching what #6955 landed. patch changesets for `@assistant-ui/core`, `@assistant-ui/react`, `@assistant-ui/react-markdown`, `@assistant-ui/react-streamdown`, and `@assistant-ui/react-mcp`; `@assistant-ui/ui` is private and takes none.

<a href="https://rupic.dev/s/s2.ME59L_xiK0KjitM48-q65SmVXG6kpaGFC53vzlfYnmaoACirMhNSERwcLqg?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

Kinfe123:
Closes #6967.

## Problem

`Array.prototype.every` skips empty positions, so a comparator built on `a.every((v, i) => ... b[i])` reads a hole in `a` as nothing-to-check while the matching position in `b` compares as `undefined`, an asymmetric equality where `cmp(new Array(1), [42])` is true but `cmp([42], new Array(1))` is false. #6955 fixed the instance that mattered (the exported `shallowEqual`); this closes the same pattern everywhere else it is still live.

The issue enumerated eight sites and called all eight latent. Both halves of that turned out to be off.

The inventory missed one: `attachmentsEqual` (`packages/core/src/adapters/attachment.ts`), which is the worst of the set, because its `b[i]!.id` non-null assertion makes a hole in `b` throw rather than compare unequal.

And three of the sites take a caller-provided array straight into the comparator, so they are reachable, not latent: the `ThreadSuggestions` / `Suggestions` config path, `McpAppFrame`'s `hostContext` prop, and the `MarkdownText` / `StreamdownText` props comparison.

At the Suggestions seam the comparator fix alone turns a silent wrong answer into a crash. `map` preserves holes, so a sparse caller array survives into the per-suggestion resource lookup, which indexes every slot and dies on `getElementKey(undefined)`; on main the `.every` hole-skip hides that by reading the sparse update as unchanged and reusing the previous dense state. So this PR also normalizes there: sparse caller arrays compact to dense lists at the single construction point, with identity reuse probed at the compacted destination index so repeated equivalent sparse updates stay stable.

## Change

Each `.every`-based comparison becomes an indexed loop in #6955's shape, or delegates to `shallowEqual`, whose array branch already is that loop:

- `core`: `toolNamesEqual` (model-context-client) and the suggestions identity check both delegate to `shallowEqual`; `attachmentsEqual` (adapters/attachment) takes an indexed loop and loses the non-null assertion that made a hole throw
- `react`: `isSameHostContext`'s array branch (mcp-apps app-frame) keeps its own recursive call inside an indexed loop; `useStableNames` (webmcp) delegates to `shallowEqual`
- `react-markdown` / `react-streamdown`: both `isShallowEqual` array branches keep their recursive call inside an indexed loop
- `react-mcp`: `areConnectionDependenciesEqual` was a line-for-line reimplementation of `shallowEqual`'s array branch, so it is deleted and its one call site calls `shallowEqual`
- `ui`: the conversation-map visible-ids check, extracted as a local `sameIds` rather than an import from `@assistant-ui/store/internal`, so the registry item stays self-contained

The sites that moved to `shallowEqual` trade `===` for `Object.is`, which is the same relation on the strings and identity-compared values they hold. No behavior change for dense arrays anywhere.

Regression tests cover the two seams where a hole changes an observable outcome: a sparse Suggestions update compacts, does not crash the lookup, and keeps identities stable across an equivalent repeat; and `McpAppFrame` treats a hole and a value as unequal in both operand orders (sparse-to-dense is the direction main reads as equal). The rest is comparator hygiene with no downstream consumer that could observe it, so it carries no test.

## Verification

Touched packages' suites pass: core 1751, ui 728 (16 skipped), react 568, react-streamdown 223, react-mcp 203, react-markdown 126.

The Suggestions claim was checked against its counterfactual: with only the issue's prescribed comparator fix applied and the compaction left out, the new test fails with `TypeError: Cannot read properties of undefined (reading 'key')`. Both halves are load-bearing.

## Public surface

No API changes. `attachmentsEqual` is exported from its module but reaches no barrel, and `areConnectionDependenciesEqual` was module-private. Patch changesets for the five published packages per the issue; `@assistant-ui/ui` is private and takes none.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/6987?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.cXOrmbpYL8ti0Sm1PoUx33pU2KTW3CzKavSnRrvhTwsdTNsf5e_ggEkHoSc?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/comparator-array-holes.md
+++ b/.changeset/comparator-array-holes.md
@@ -1,0 +1,9 @@
+---
+"@assistant-ui/core": patch
+"@assistant-ui/react": patch
+"@assistant-ui/react-markdown": patch
+"@assistant-ui/react-streamdown": patch
+"@assistant-ui/react-mcp": patch
+---
+
+fix: compare arrays with indexed loops so sparse-array holes cannot read as equal
--- a/.changeset/comparator-array-holes.md
+++ b/.changeset/comparator-array-holes.md
@@ -6,4 +6,4 @@
 "@assistant-ui/react-mcp": patch
 ---
 
-fix: compare arrays with indexed loops so sparse-array holes cannot read as equal
+fix: compare arrays with indexed loops so sparse-array holes cannot read as equal; a sparse suggestions list now compacts to a dense one before it reaches the per-suggestion lookup
--- a/packages/core/src/adapters/attachment.ts
+++ b/packages/core/src/adapters/attachment.ts
@@ -170,7 +170,10 @@ export function attachmentsEqual(
   b: readonly CompleteAttachment[],
 ): boolean {
   if (a.length !== b.length) return false;
-  return a.every((att, i) => att.id === b[i]!.id);
+  for (let i = 0; i < a.length; i++) {
+    if (a[i]?.id !== b[i]?.id) return false;
+  }
+  return true;
 }
 
 export function partToCompleteAttachment(
--- a/packages/core/src/store/clients/model-context-client.ts
+++ b/packages/core/src/store/clients/model-context-client.ts
@@ -3,6 +3,7 @@ import { resource } from "@assistant-ui/tap";
 import type { ClientOutput } from "@assistant-ui/store";
 import { CompositeContextProvider } from "../../utils/composite-context-provider";
 import type { ModelContextState } from "../scopes/model-context";
+import { shallowEqual } from "@assistant-ui/store/internal";
 
 const EMPTY_TOOL_NAMES: readonly string[] = [];
 
@@ -11,17 +12,8 @@ const INITIAL_STATE: ModelContextState = {
   toolNames: EMPTY_TOOL_NAMES,
 };
 
-const toolNamesEqual = (
-  a: readonly string[],
-  b: readonly string[],
-): boolean => {
-  if (a === b) return true;
-  if (a.length !== b.length) return false;
-  for (let i = 0; i < a.length; i++) {
-    if (a[i] !== b[i]) return false;
-  }
-  return true;
-};
+const toolNamesEqual = (a: readonly string[], b: readonly string[]): boolean =>
+  a === b || shallowEqual(a, b);
 
 const deriveState = (
   composite: CompositeContextProvider,
--- a/packages/core/src/store/clients/model-context-client.ts
+++ b/packages/core/src/store/clients/model-context-client.ts
@@ -11,8 +11,17 @@ const INITIAL_STATE: ModelContextState = {
   toolNames: EMPTY_TOOL_NAMES,
 };
 
-const toolNamesEqual = (a: readonly string[], b: readonly string[]): boolean =>
-  a === b || (a.length === b.length && a.every((v, i) => v === b[i]));
+const toolNamesEqual = (
+  a: readonly string[],
+  b: readonly string[],
+): boolean => {
+  if (a === b) return true;
+  if (a.length !== b.length) return false;
+  for (let i = 0; i < a.length; i++) {
+    if (a[i] !== b[i]) return false;
+  }
+  return true;
+};
 
 const deriveState = (
   composite: CompositeContextProvider,
--- a/packages/core/src/store/clients/suggestions.test.ts
+++ b/packages/core/src/store/clients/suggestions.test.ts
@@ -171,8 +171,21 @@ describe("ThreadSuggestions", () => {
       const expected = { title: "Prompt B", label: "", prompt: "Prompt B" };
       expect(state.suggestions).toEqual([expected]);
       expect(Object.keys(state.suggestions)).toEqual(["0"]);
-      expect(root.getValue().suggestion({ index: 0 }).getState()).toEqual(
-        expected,
+      const firstSuggestion = root
+        .getValue()
+        .suggestion({ index: 0 })
+        .getState();
+      expect(firstSuggestion).toEqual(expected);
+
+      // An equivalent sparse update keeps both identities stable: the reuse
+      // probe reads the compacted destination index, not the sparse source one.
+      const sparseAgain: ThreadSuggestion[] = new Array(2);
+      sparseAgain[1] = { prompt: "Prompt B" };
+      flushTapSync(() => setSuggestions(sparseAgain));
+
+      expect(root.getValue().getState()).toBe(state);
+      expect(root.getValue().suggestion({ index: 0 }).getState()).toBe(
+        firstSuggestion,
       );
     } finally {
       root.unmount();
--- a/packages/core/src/store/clients/suggestions.test.ts
+++ b/packages/core/src/store/clients/suggestions.test.ts
@@ -151,6 +151,34 @@ describe("ThreadSuggestions", () => {
     }
   });
 
+  it("drops sparse holes before the per-suggestion lookup", () => {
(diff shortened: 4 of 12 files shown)


Kinfe123:
All three review notes addressed in 1c7878917: suggestions.ts drops the local helper for the already-imported shallowEqual (whose array branch is exactly the #6955 loop) and the import block is whole again; toolNamesEqual in model-context-client reuses shallowEqual behind its identity fast path; and sameIds moved above the reading-line JSDoc so that comment documents readingLine again. Store-client tests (36) and the conversation-map suite (14) pass.

Kinfe123:
rupic's round 2 addressed in 9e96dc7b7 — the blocking finding was real and is now fixed at the source:

- **Suggestions sparse crash (blocking)**: confirmed empirically — with the comparator fix alone, a dense-to-sparse update flows a sparse list into the per-suggestion lookup and `useResources` throws `Cannot read properties of undefined (reading 'key')`, exactly as described. The chosen contract is normalize: `useStableSuggestionsState` now builds the stable list with `forEach`, which skips holes, so a sparse caller array becomes a dense one at the single construction point before anything indexes it. The new regression drives a sparse update through the public `ThreadSuggestions` seam, asserts the resulting state is dense with the hole dropped and the lookup works, and crashes without the normalization.
- **McpAppFrame hole coverage**: the existing host-context notify test now exercises a hole against a value in both operand orders. The sparse-to-dense direction is the one `.every` read as equal by skipping the hole — verified by swapping in main's comparator, which fails the new assertion (no notify), while this branch's loop passes.
- **webmcp `namesEqual` (confirmed by both reviewers)**: dropped for the shared `shallowEqual` from `@assistant-ui/store/internal`, which the package already depends on.
- **`attachmentsEqual`** stays as rupic marked it: pre-existing and untouched by this PR.
- **`is-json-equal` / `useSlashCommandAdapter`**: real instances of the same class, but outside the eight sites #6967 enumerates — keeping this PR to the issue's scope; happy to note them on the issue as follow-up sites so the enumeration stays the source of truth.

Full suites re-verified: core 1751, react 495.

Kinfe123:
Round 3's blocking finding fixed in aeb5310d6: the identity-reuse probe now reads `previous?.suggestions[suggestions.length]` — the compacted destination index — instead of the sparse source index, and the sparse test sends an equivalent second update asserting both the state and the per-suggestion identity stay stable (`toBe`). Verified the new assertion fails with the source-index probe and passes with the fix; store suite 37/37. The PR description is also refreshed — it no longer claims the sites are unreachable or untested, and now describes the two reachable seams and the normalization contract.

okisdev (approved):
approving. the thing worth saying out loud is that the compaction is not scope creep, it is the other half of the fix, and the obvious cheap check reports the opposite: on main a sparse suggestions update does not crash, so "does this reproduce today" answers no. i ran the counterfactual instead. with only the comparator fix the issue asked for, and the compaction left out, the new test fails:

```
× drops sparse holes before the per-suggestion lookup
TypeError: Cannot read properties of undefined (reading 'key')
```

`map` preserves holes, so the sparse array survives into `useClientLookup` and `getElementKey(undefined)` throws; main's `.every` hole-skip hides that by reading the update as unchanged and reusing the previous dense state. both halves are load bearing.

four things landed on top of your head:

1. the issue's inventory was one short. `attachmentsEqual` (`packages/core/src/adapters/attachment.ts:173`) is the same shape and the worst of the set, because `b[i]!.id` makes a hole in `b` throw rather than compare unequal. it is currently dead, though: a repo-wide grep finds no caller outside its own test. that is a reading, not something i executed, and deleting it is a separate concern.

2. `areConnectionDependenciesEqual` was `shallowEqual`'s array branch written out again, and react-mcp already carries `@assistant-ui/store` in `dependencies`, so it was the odd one out next to the reuse you did in core and react. `sameIds` staying local in `ui` is the right call and stays, since that file ships through the registry and should not grow an `@assistant-ui/store/internal` import.

3. two test comments described the old implementation rather than the contract they assert.

4. the changeset named only the comparator loops. the compaction is the one user-observable change in the release, since `suggestions.length` and the `suggestion({ index })` key space are the post-compaction ones and a hole shifts every later entry down a slot.

on the red `Test Changed Packages` you may have seen: that was a runner flake, not you. it failed in `packages/react-native/src/primitives/message/MessageContent.test.tsx`, which this branch does not touch and which only exists in that form on main. i ran it on plain main and on this branch merged with main, 16/16 both times, and the rerun is green.

suites at head: core 1751, ui 728 (16 skipped), react 568, react-streamdown 223, react-mcp 203, react-markdown 126. tsc adds no new errors; the leftovers in core and react-mcp reproduce on your head under `git stash`.

PR #7008, opened 2026-09-08
test(react-native): give the derived-status cases a real timeout

### LINKED ISSUE #7007: flaky: react-native MessageContent derived-status cases sit on the 5s vitest timeout
a `pull_request` CI job fails non-deterministically on `packages/react-native/src/primitives/message/MessageContent.test.tsx:281` ("renders derived status for registered %s UIs").

## problem

three runs of the same job produced three different outcomes:

- run [34183051897](https://github.com/assistant-ui/assistant-ui/actions/runs/34183051897) on head `6e2408ba`: 2 failures, `expected 'running' to be 'complete'` at 5008ms and 2356ms
- rerun of that same head: 1 failure
- run [34183730771](https://github.com/assistant-ui/assistant-ui/actions/runs/34183730771) on head `91fe9c8c`: one `Error: Test timed out in 5000ms` plus one assertion failure

it does not reproduce locally: 10/10 green on `main`, and 10/10 green on a branch merged with main.

## root cause

the case sits right on vitest's default 5000ms timeout. the file takes 5826ms on the CI runner because each of the two `it.each` cases calls `vi.resetModules()` and then dynamically re-imports `@assistant-ui/core/react`, re-evaluating the full module graph, before rendering three sequential states through a real `useExternalStoreRuntime`. on a loaded runner that crosses 5s, and the pending React update has not flushed when the assertion runs, which surfaces either as a timeout or as the last state still reading `running` instead of `complete`.

## suggested fix

hoist the `vi.resetModules()` and the dynamic imports out of the `it.each` body into a `beforeAll`, so the module graph is re-evaluated once instead of once per case, or give the case an explicit timeout.

## context

surfaced while shipping #6980. that PR is not on this test's code path: `export()` is reached only through the `store.messageRepository` branch (`packages/core/src/runtimes/external-store/external-store-thread-runtime-core.ts:331`), and this test uses `store.messages`.

<a href="https://rupic.dev/s/s2.6i4pvmfS1kNagA0U91kbmW_w4q3rb4bdeGi_68FQL2Npjh3BMlw1uWXM87U?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

okisdev:
## Problem

`packages/react-native/src/primitives/message/MessageContent.test.tsx:281` ("renders derived status for registered %s UIs") fails non-deterministically on CI. Four consecutive runs across two heads produced four different outcomes: two assertion failures, then one, then a `Test timed out in 5000ms` plus one assertion failure.

It does not reproduce locally. On this machine the file takes 929ms; on the CI runner it takes 5826ms.

Closes #7007.

## Root cause

The case has no headroom against vitest's default 5000ms per-test budget. Each of the two `it.each` cases calls `vi.resetModules()` and then dynamically re-imports `@assistant-ui/core/react`, re-evaluating the full module graph, before rendering three sequential states through a live `useExternalStoreRuntime`. At 5826ms on a loaded runner it crosses the budget, and the pending React update has not flushed when the assertion runs, so it surfaces either as a timeout or as the last state still reading `running` instead of `complete`.

The work the case does is legitimate, so the budget is what is wrong, not the test.

## Change

Pass an explicit 30s timeout to the `it.each` case. Restructuring the case to re-evaluate the module graph once instead of once per case would also help, but it would trade away the per-case module isolation the current shape deliberately buys, so it is left alone.

## Verification

`pnpm --filter @assistant-ui/react-native test src/primitives/message/MessageContent.test.tsx`: 16 passed, and 10/10 green on repeat runs both before and after, since the failure only reproduces under CI load.

## Public surface

None. Test-only, so no changeset.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/7008?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.K9Za-dY6JWQ8oEJTZKcUPy7l_9xm9qMyKZm8vVObzXODkKq4Su-0E5gjfmc?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/packages/react-native/src/primitives/message/MessageContent.test.tsx
+++ b/packages/react-native/src/primitives/message/MessageContent.test.tsx
@@ -354,5 +354,6 @@ describe("MessageContent with a runtime", () => {
         await act(async () => root.unmount());
       }
     },
+    30_000,
   );
 });


okisdev:
superseded by #7009, closing.

raising the timeout to 30s widens the window without removing what fills it. each `it.each` case still calls `vi.resetModules()` and re-imports `@assistant-ui/core/react`, so the file still costs ~5.8s on a runner, and a case that does blow the budget still skips its `finally` unmount and leaves an act queue behind for its sibling, which is where the `expected 'running' to be 'complete'` half of the failure comes from.

#7009 moves the two cases out of the mocked file, which was the only reason the module-registry reset existed. measured on a `node:24 --cpus=2` container that reproduces the slow path: 719 to 725ms before, 21 to 22ms after; 568ms to 18ms on macOS.

PR #7039, opened 2026-09-08
fix(docs): tell MCP clients search_docs reads page content

### LINKED ISSUE #6721: searchDocs matches page metadata only, not docs content
the browser-facing `searchDocs` tool from #6467 proxies the docs route's `search_docs`, which is a substring filter over title, URL, and description (`apps/docs/app/api/mcp/route.ts:274-284`):

```ts
return allPages()
  .map(({ page }) => pageSummary(page))
  .filter((page) =>
    [page.title, page.url, page.description ?? ""].some((value) =>
      value.toLowerCase().includes(normalized),
    ),
  )
  .slice(0, SEARCH_DOCS_RESULT_LIMIT);
```

so a query whose terms appear only in a page body returns nothing. `packages/mcp-docs-server/src/utils/search.ts` scores page content, and that is the search the proposal in #6396 asked the browser tool to reuse.

the result is two surfaces with the same purpose and different recall, with nothing in the repo recording the difference as deliberate. either lift the content-ranked scoring behind the route (it needs an index the route can read without node `fs`), or write the divergence down in `apps/docs/lib/mcp-tool-definitions.ts` so the stdio server and the browser tools are not silently different.

<a href="https://rupic.dev/s/s2.ALtOcynGosHUI4odyWJkRiOEgR6z_V0ZqtybNT49cQxXOq0BYknOhQ74w1Y?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

okisdev:
## Problem

`search_docs` tells every MCP client it matches "by title, description, or URL". That was true when #6721 was filed, but #6807 replaced the substring filter with `searchContent` over `buildContentIndex()`, so the tool now ranks title, headings, URL, description and body text and returns the matching paragraphs as an excerpt. The description was never swept, and the browser-facing `searchDocs` (WebMCP, #6467) interpolates the same string, so both surfaces still advertise a metadata-only search and never mention the `headings` and `excerpt` they return.

The recall itself was also only covered at the unit level. `route.test.ts` mocks `@/lib/source` with an empty `getPages()`, so every `search_docs` assertion in it was `[]`; pointing `searchDocs()` back at the old `pageSummary` filter kept the whole suite green.

## Root cause

#6807 changed the implementation and the demo tool. `apps/docs/lib/mcp-tool-definitions.ts` was outside its diff, and it is the file that carries the model-facing contract for the route and the browser tool alike.

## Change

- `search_docs` now describes what it does: ranks each page over title, headings, URL, description and body text, and returns the paragraphs that matched as an excerpt.
- the `query` hint matches the one the landing-page demo already uses ("a short phrase of distinctive documentation terms, with filler words omitted"). that guidance is load-bearing here: `rank()` requires every non-stop-word token on one page before the per-token fallback applies.
- `route.test.ts` gets a real two-page corpus with `structuredData()`, plus a body-only query case at the route and at the WebMCP adapter. the metering tests keep their concern and now assert against that corpus instead of `[]`.
- the `searchContent` tests move from `lib/ai/search-docs.test.ts` to `lib/search/content-search.test.ts`, beside the module they exercise. `search-docs.test.ts` keeps the `createSearchDocsTool` case.
- the tool table in `llm.mdx` picks up `elements` and the content search.

No behavior change: `route.ts` is untouched.

## Verification

- `vitest run app/api/mcp/route.test.ts lib/search lib/ai` in `apps/docs`: 68 passed.
- reproduction: swapping `searchDocs()` in `route.ts` back to the pre-#6807 metadata filter fails exactly the two new cases ("ranks search_docs over page content, not metadata alone" and "executes the WebMCP adapter through the route transport", both `expected [] to deeply equal [ '/docs/guides/keyboard' ]`). restoring the current implementation passes them.
- `oxlint` and `oxfmt --check` clean at the repo root.
- `tsc --noEmit` for `apps/docs` reports nothing in the touched files (a worktree without built package dists reports the usual `TS2307` cascade for every workspace import).

Closes #6721.

## Public surface

None. `@assistant-ui/docs` is private, so no changeset. The `search_docs` tool description and query hint change is what MCP clients and the WebMCP `searchDocs` tool see.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/7039?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.kizjwgMWpmeyUaBbUIWdL-pHe-9KXgD7Jf7hXfCyEsE6WN87qIW_hhqoIL4?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/apps/docs/app/api/mcp/route.test.ts
+++ b/apps/docs/app/api/mcp/route.test.ts
@@ -32,6 +32,7 @@ vi.mock("@/lib/source", () => {
     url: string,
     title: string,
     description: string,
+    headings: string[],
     contents: string[],
   ) => ({
     url,
@@ -39,7 +40,10 @@ vi.mock("@/lib/source", () => {
       title,
       description,
       structuredData: () => ({
-        headings: [],
+        headings: headings.map((content) => ({
+          id: content.toLowerCase(),
+          content,
+        })),
         contents: contents.map((content) => ({ content })),
       }),
     },
@@ -49,12 +53,22 @@ vi.mock("@/lib/source", () => {
     source: {
       ...makeSource(),
       getPages: vi.fn(() => [
-        docsPage("/docs/ui/thread", "Thread", "Render a conversation.", [
-          "An unrelated opening paragraph.",
-        ]),
-        docsPage("/docs/guides/keyboard", "Keyboard", "Shortcuts.", [
-          "Press the escape key to dismiss the composer autocomplete popover.",
-        ]),
+        docsPage(
+          "/docs/ui/thread",
+          "Thread",
+          "Render a conversation.",
+          [],
+          ["An unrelated opening paragraph."],
+        ),
+        docsPage(
+          "/docs/guides/keyboard",
+          "Keyboard",
+          "Shortcuts.",
+          ["Bindings"],
+          [
+            "Press the escape key to dismiss the composer autocomplete popover.",
+          ],
+        ),
       ]),
     },
     examples: makeSource(),
@@ -320,6 +334,17 @@ describe("POST /api/mcp", () => {
 
     expect(pages.map((page) => page.url)).toEqual(["/docs/guides/keyboard"]);
     expect(pages[0]?.excerpt).toContain("autocomplete popover");
+  });
+
+  it("ranks a heading-only match and returns the headings with it", async () => {
+    const response = await requestMcp("tools/call", {
+      name: "search_docs",
+      arguments: { query: "bindings" },
+    });
+    const pages = searchedPages(getToolCallResult(response));
+
+    expect(pages.map((page) => page.url)).toEqual(["/docs/guides/keyboard"]);
+    expect(pages[0]?.headings).toEqual(["Bindings"]);
   });
 
   it("still matches search_docs on page metadata", async () => {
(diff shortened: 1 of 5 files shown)


okisdev:
addressed the cubic P3: the route fixture now carries a heading, and a new case asserts a heading-only query ranks and that `headings` comes back with the match. the tool description advertises that field, so it should be covered at the route and WebMCP surfaces rather than only in `searchContent`'s own tests.

okisdev:
good catch, fixed. `rank()` falls back to `page.contents.slice(0, 1)` when nothing in the body scored, so the description now says it excerpts the paragraphs that matched and falls back to the page's opening paragraph when only metadata matched. `llm.mdx` no longer claims a matching excerpt on every result either. added a route case pinning that fallback, so the claim is covered where it is made.

PR #7058, opened 2026-09-08
fix(assistant-stream): stop cancelled tools after validation

ephraimduncan:
## Problem

A cancelled run can still start tool side effects when asynchronous schema validation completes. This occurs in `assistant-stream` 0.3.41 at base `55c34a62512f901194470c6ad6fc561d69be207b`.

The following reproduction runs with Bun from that repository checkout after `pnpm install --frozen-lockfile`. The assertion fails because the tool executes once.

```js
import { unstable_runPendingTools } from "./packages/assistant-stream/src/core/tool/toolResultStream.ts";
import { createInitialMessage } from "./packages/assistant-stream/src/core/accumulators/assistant-message-accumulator.ts";

let finishValidation;
const validation = new Promise(resolve => { finishValidation = resolve; });
let executions = 0;
const part = {
  type: "tool-call", toolCallId: "t", toolName: "save", state: "call",
  status: { type: "running", isArgsComplete: true }, argsText: "{}", args: {},
};
const pending = unstable_runPendingTools(
  { ...createInitialMessage(), parts: [part], content: [part] },
  { save: {
    parameters: { "~standard": {
      version: 1, vendor: "example", validate: () => validation,
    } },
    execute: () => { executions++; return "saved"; },
  } },
  controller.signal,
  async () => null,
);
controller.abort();
finishValidation({ value: {} });
await pending;
assert.equal(executions, 0);
```

## Root cause

The cancellation check preceded the validation await. After validation, the cancellation promise did not prevent the immediate call to the tool.

## Change

Cancelled runs return the existing cancellation response before tool execution. The shared correction covers pending and streamed tool calls. Cancellation behavior for tools that already started stays unchanged.

## Verification

The reproduction and new regression failed before the correction and passed afterward. All 28 tests in `toolResultStream.test.ts` passed. The command was `pnpm --filter assistant-stream exec vitest run src/core/tool/toolResultStream.test.ts src/core/tool/ToolExecutionStream.test.ts`. Only `toolResultStream.test.ts` matched.

A Node smoke test against the new build covered pending and streamed calls with valid and invalid arguments. Each case returned cancellation without side effects.

The package build, scoped Oxlint and formatting checks, and `pnpm changesets:check` passed.

## Public surface

The change includes an `assistant-stream` patch changeset. Public exports and signatures stay unchanged. Documentation, API-reference output, and templates need no changes. Validation that never completes remains outside this correction.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/7058?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.jkpQMH_2jHzqO4bNDp2OeeWehNQtNY_WH3mLpObF2uMEwMOPUU62NhkxKGE?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/cancel-tools-after-validation.md
+++ b/.changeset/cancel-tools-after-validation.md
@@ -1,0 +1,5 @@
+---
+"assistant-stream": patch
+---
+
+fix: prevent tool execution after cancellation during asynchronous schema validation
--- a/packages/assistant-stream/src/core/tool/toolResultStream.test.ts
+++ b/packages/assistant-stream/src/core/tool/toolResultStream.test.ts
@@ -348,6 +348,7 @@ describe("unstable_runPendingTools", () => {
     expect(execute).not.toHaveBeenCalled();
     expect(settled.parts[0]).toMatchObject({
       state: "result",
+      result: "Tool execution was cancelled.",
       isError: true,
     });
   });
--- a/packages/assistant-stream/src/core/tool/toolResultStream.test.ts
+++ b/packages/assistant-stream/src/core/tool/toolResultStream.test.ts
@@ -317,6 +317,41 @@ describe("unstable_runPendingTools", () => {
     });
   });
 
+  it("does not execute a tool cancelled during async validation", async () => {
+    const abortController = new AbortController();
+    const validation = promiseWithResolvers<{
+      value: Record<string, unknown>;
+    }>();
+    const execute = vi.fn(() => "executed");
+    const pending = unstable_runPendingTools(
+      createPendingToolMessage("async-validation"),
+      {
+        tool: {
+          parameters: {
+            "~standard": {
+              version: 1,
+              vendor: "test",
+              validate: () => validation.promise,
+            },
+          },
+          execute,
+        },
+      },
+      abortController.signal,
+      async () => {},
+    );
+
+    abortController.abort();
+    validation.resolve({ value: {} });
+    const settled = await pending;
+
+    expect(execute).not.toHaveBeenCalled();
+    expect(settled.parts[0]).toMatchObject({
+      state: "result",
+      isError: true,
+    });
+  });
+
   it.each(["resolves", "rejects"] as const)(
     "does not enqueue pending tool output after cancellation when execution %s",
     async (settlement) => {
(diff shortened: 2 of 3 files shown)


okisdev (approved):
approving. the guard lands in the right place and it isn't a new shape here: packages/react/src/unstable/webmcp/convertTools.ts:198 already does exactly this, same position, after the `experimental_onSchemaValidationError` substitution and before execution. so this closes a parity gap between the two tool paths rather than adding a pattern that only lives in one of them.

one correction to the root cause as written, because it changes what the next reader looks for. the `Promise.race` never had a chance to prevent the call: `executePromise` is an async IIFE (toolResultStream.ts:137), so `executeFn` is invoked synchronously the moment it is constructed at :145. the race only ever chose which value came back, and it chose the cancellation response, which is why the old code looks correct from the outside. that is also the trap in the test. assert on the returned response and the bug is invisible, since `state: "result", isError: true` is what you already got; asserting `execute` was never called is the only thing that sees it.

i ran it both ways. reverting only the source hunk while keeping your test gives `expect(execute).not.toHaveBeenCalled()` failing with `Number of calls: 1`; restoring it gives 28/28 in `toolResultStream.test.ts` and 520 passed / 22 skipped across 37 files for the package, `pnpm lint` clean. the hole also predates #6656: `git show f0d0aa2f8` has `if (result instanceof Promise) result = await result`, so any validator returning a native promise (async zod, async valibot) has had this window all along and #6656 only widened it to thenables.

i pushed one commit on top. the third copy of the cancellation `ToolResponse` is what made it worth pulling into `cancelledToolResponse()`, and i pinned the message string in your new assertion so the response is nailed down rather than just its error flag. `preserves cancellation ordering for synchronous validation` still passes untouched, which was the check i wanted: with a sync validator `executeFn` runs before the abort lands, so the new guard never fires on that path.

the never-settling `validate()` case stays open and scoping it out is right. the guard cannot reach it; that needs the validation itself raced against the signal, which is a different change.

PR #7059, opened 2026-09-08
fix(assistant-stream): remove nested partial schema requirements

ephraimduncan:
## Problem

Partial update schemas still require nested fields when the nested object has no `properties` keyword. For example, `{ settings: {} }` still requires `settings.name`.

This reproduction fails at base `55c34a62512f901194470c6ad6fc561d69be207b` (`assistant-stream` 0.3.41). It passes after the correction:

```js
import { toPartialJSONSchema } from "./packages/assistant-stream/src/core/tool/schema-utils.ts";

const partial = toPartialJSONSchema({
  type: "object",
  properties: {
    settings: {
      type: "object",
      required: ["name"],
      additionalProperties: { type: "string" },
    },
  },
  required: ["settings"],
});

assert.equal(partial.properties.settings.required, undefined);
assert.deepEqual(partial.properties.settings.additionalProperties, { type: "string" });
```

The snippet runs with `bun` from the repository root.

## Root cause

Nested conversion depended on the presence of `properties`, but JSON Schema does not require that keyword beside `required`.

## Change

The shared utility removes `required` from each nested property schema. It retains `additionalProperties` and leaves array-item schemas unchanged.

## Verification

The original reproduction and the new regression failed before the correction and passed afterward. All 38 tests passed with `pnpm --filter assistant-stream exec vitest run src/core/tool/schema-utils.test.ts`.

`pnpm --filter assistant-stream build`, scoped Oxlint and format checks, and `pnpm changesets:check` passed. The reproduction also passed through the built `assistant-stream` export in Node.

## Public surface

The correction affects `assistant-stream` and includes a patch changeset. Public exports and signatures are unchanged. No documentation, API-reference output, or template changes are necessary.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/7059?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.PKYuz9EjdqFIEpgDW2sr0i5pBp6YrsrXezLGUfgpNqYePeIUSINBgs74cz4?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/partial-schema-nested-required.md
+++ b/.changeset/partial-schema-nested-required.md
@@ -1,0 +1,5 @@
+---
+"assistant-stream": patch
+---
+
+fix: remove nested required constraints from partial schemas without properties
--- a/packages/assistant-stream/src/core/tool/schema-utils.test.ts
+++ b/packages/assistant-stream/src/core/tool/schema-utils.test.ts
@@ -175,6 +175,31 @@ describe("toPartialJSONSchema", () => {
     expect(address.required).toBeUndefined();
   });
 
+  it("removes nested required without properties and preserves value constraints", () => {
+    const schema = {
+      type: "object" as const,
+      properties: {
+        settings: {
+          type: "object" as const,
+          required: ["name"],
+          additionalProperties: { type: "string" as const },
+        },
+      },
+      required: ["settings"],
+    };
+
+    expect(toPartialJSONSchema(schema)).toEqual({
+      type: "object",
+      properties: {
+        settings: {
+          type: "object",
+          additionalProperties: { type: "string" },
+        },
+      },
+    });
+    expect(schema.properties.settings.required).toEqual(["name"]);
+  });
+
   it("leaves array item schemas unchanged", () => {
     const schema = {
       type: "object" as const,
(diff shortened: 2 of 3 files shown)


okisdev (approved):
the `properties != null` gate was the actual mechanism, so dropping it is the right cut instead of special casing the `additionalProperties` shape.

verified rather than read: swapped the base `55c34a625` copy of `schema-utils.ts` back into a worktree and reran the suite, your new case fails at line 191 and the other 37 pass, then 38/38 with the patch. and the case is reachable rather than synthetic, since `Tool["parameters"]` is typed `StandardSchemaV1 | JSONSchema7`, so a hand-authored interactable `stateSchema` really can carry `required` with no `properties` beside it.

one residual that does not block this: the recursion still only walks `properties`, so nested `required` survives inside `anyOf` / `oneOf`. measured against zod 4.5.4 with your patch applied, both `z.object({...}).nullable()` and `z.discriminatedUnion(...)` leave `required` on the branch schemas, which is the same class of gap for a shape zod emits far more often than the one fixed here. filed separately as #7097. array `items` staying excluded is correct and should stay that way.

PR #7060, opened 2026-09-08
fix(assistant-stream): prevent standard schema conversion crashes

ephraimduncan:
## Problem

Tool parameter conversion throws when a Standard JSON Schema converter reads its mandatory `target` option. With this correction, the converter receives `draft-07` and returns the tool schema.

This reproduction throws on base `55c34a62512f901194470c6ad6fc561d69be207b` (`assistant-stream` 0.3.41). It runs with Bun from the repository root:

```js
import { toToolsJSONSchema } from "./packages/assistant-stream/src/core/tool/schema-utils.ts";

const convert = ({ target }) => {
  assert.equal(target, "draft-07");
  return { type: "object", properties: { name: { type: "string" } } };
};
const parameters = {
  "~standard": {
    version: 1,
    vendor: "local-example",
    validate: (value) => ({ value }),
    jsonSchema: { input: convert, output: convert },
  },
};
const converted = toToolsJSONSchema({ example: { parameters } });
assert.equal(converted.example.parameters.type, "object");
```

## Root cause

The shared `toJSONSchema` function called `jsonSchema.input()` without options. Its private type also described a zero-argument converter, contrary to the Standard JSON Schema contract.

## Change

The shared converter supplies `{ target: "draft-07" }` to match its existing `JSONSchema7` return contract. Its private type uses the installed standard definition. Other conversion methods and their priority remain unchanged.

## Verification

The regression in the existing schema test file throws before the correction. After the correction, `pnpm --filter assistant-stream exec vitest run src/core/tool/schema-utils.test.ts` passes all 37 tests.

The original API reproduction passes against the source and the fresh public build. `pnpm --filter assistant-stream build`, scoped Oxlint, and `pnpm changesets:check` also pass.

## Public surface

`assistant-stream` includes a patch changeset. Public exports and signatures remain unchanged. No documentation claims, generated API pages, or templates change.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/7060?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.HnJFxEtQ0ON1rhdDJjIMS3ZbnU92DmqhdRwPFS9l41arxdIm2V2AbY2RmME?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/standard-schema-draft-target.md
+++ b/.changeset/standard-schema-draft-target.md
@@ -1,0 +1,5 @@
+---
+"assistant-stream": patch
+---
+
+fix: supply the draft-07 target to Standard JSON Schema converters so tool parameter conversion does not throw on missing options
--- a/packages/assistant-stream/src/core/tool/schema-utils.test.ts
+++ b/packages/assistant-stream/src/core/tool/schema-utils.test.ts
@@ -1,4 +1,5 @@
 import { describe, expect, it } from "vitest";
+import type { StandardJSONSchemaV1 } from "@standard-schema/spec";
 import {
   toJSONSchema,
   toPartialJSONSchema,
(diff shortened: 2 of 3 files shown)


okisdev (approved):
verified and merging. the premise checks out against the pinned spec: `@standard-schema/spec@1.1.0` types the converter as `input: (options: StandardJSONSchemaV1.Options) => Record<string, unknown>` with `target` non-optional, so the bare call was contract-violating rather than merely under-specified.

i ran it rather than read it. in a worktree i reverted just the one src line and the new test fails exactly as you describe (`TypeError: Cannot destructure property 'target' of 'undefined'` at schema-utils.ts:87), restored it and got 37/37, then confirmed scoped tsc, oxlint and oxfmt clean. also grepped the repo for other `jsonSchema.input(` call sites and this is the only one, so the sweep is complete.

one thing survives that is not yours to fix here: the `schema.toJSONSchema()` branch still passes no options and zod defaults that to `draft-2020-12`, so the same function keeps emitting two dialects under a `JSONSchema7` return type depending on which schema library the caller brought. same for the JSDoc, which attributes both `~standard` branches to zod v4 when pinned zod implements neither. tracked in #7098.

PR #7062, opened 2026-09-08
fix(core): complete final-only voice transcripts

ephraimduncan:
## Problem

An assistant voice reply that arrives as a single final transcript stays in `running` status. The thread keeps showing the streaming affordances for a reply that has already finished, and when the agent sends two replies in a row the first one never leaves `running` at all, because `_finishVoiceAssistantMessage` only repairs `_voiceMessages.at(-1)` on the next user turn.

`examples/with-elevenlabs-conversational` hits this on every reply: its adapter maps each `onMessage` to `emitTranscript({ ..., isFinal: true })` and never emits a partial. The behavior also contradicts the documented contract in `apps/docs/content/docs/guides/voice.mdx`: "The message stays in a running status until `isFinal: true`".

## Root cause

`_handleVoiceTranscript` expressed the `isFinal` rule in only one of its two branches. The update branch applied `complete/stop`; the create branch hard-coded `running`. A final transcript then cleared `_currentAssistantMsg`, so a first-and-only transcript never reached the branch that would have completed it.

## Change

The `isFinal` rule is derived once, before the branch, and both the created message and the updated message take it. That removes the duplication the gap grew out of, rather than copying the rule into the second branch. A partial transcript still yields `running`, and because a final transcript clears `_currentAssistantMsg`, the update branch only ever sees a message that was already `running`.

## Verification

`pnpm --filter @assistant-ui/core test src/runtime/base/base-thread-runtime-core.test.ts src/adapters/voice.test.ts`: 16 passed.

Both directions on the new test: with the source reverted to the base commit it fails with `status: { type: "running" }` where `{ type: "complete", reason: "stop" }` is expected, and it passes on this branch. The full `@assistant-ui/core` suite reports the same 11 pre-existing failures before and after the change (all under `src/react/**`, from an unbuilt `@assistant-ui/tap`).

The test runs on the `TestRuntime` and `createVoiceAdapter` harness this file already carries, with `emitTranscript` added next to the existing `emitVolume`, so a base-class test does not have to construct a sibling runtime. It covers a final-only reply followed by a separate partial-to-final reply.

Oxlint and oxfmt pass on both touched files.

## Public surface

`@assistant-ui/core` gets a patch changeset. No export, signature, template, or api-reference change. The fix restores the documented `isFinal` behavior.

<a href="https://rupic.dev/s/s2.aNseM1EflVSNeE_UQnIAnrCjcxVFq0y_HE9l4AHG_qy4SxvI6RjswMBsQXk?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

### DIFF
--- a/.changeset/final-only-voice-transcript.md
+++ b/.changeset/final-only-voice-transcript.md
@@ -1,0 +1,5 @@
+---
+"@assistant-ui/core": patch
+---
+
+fix: complete assistant voice messages when the first transcript is final
--- a/packages/core/src/runtime/base/base-thread-runtime-core.test.ts
+++ b/packages/core/src/runtime/base/base-thread-runtime-core.test.ts
@@ -1,11 +1,6 @@
 import { afterEach, describe, expect, it, vi } from "vitest";
-import {
-  createVoiceSession,
-  type RealtimeVoiceAdapter,
-  type VoiceSessionHelpers,
-} from "../../adapters/voice";
+import type { RealtimeVoiceAdapter } from "../../adapters/voice";
 import type { ModelContextProvider } from "../../model-context/types";
-import { LocalRuntimeCore } from "../../runtimes/local/local-runtime-core";
 import type { AppendMessage } from "../../types/message";
 import { CompositeContextProvider } from "../../utils/composite-context-provider";
 import type {
@@ -21,6 +16,9 @@ import { BaseThreadRuntimeCore } from "./base-thread-runtime-core";
 
 const createVoiceAdapter = () => {
   let volumeCallback: ((volume: number) => void) | undefined;
+  let transcriptCallback:
+    | ((transcript: RealtimeVoiceAdapter.TranscriptItem) => void)
+    | undefined;
   const session: RealtimeVoiceAdapter.Session = {
     status: { type: "running" },
     isMuted: false,
@@ -28,7 +26,12 @@ const createVoiceAdapter = () => {
     mute: vi.fn(),
     unmute: vi.fn(),
     onStatusChange: () => () => {},
-    onTranscript: () => () => {},
+    onTranscript: (callback) => {
+      transcriptCallback = callback;
+      return () => {
+        transcriptCallback = undefined;
+      };
+    },
     onModeChange: () => () => {},
     onVolumeChange: (callback) => {
       volumeCallback = callback;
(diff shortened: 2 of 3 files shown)


okisdev (approved):
approving. the diagnosis is right and the failing case is real: `_handleVoiceTranscript` only ever applied `isFinal` in the update branch, so a first-and-only final transcript created a `running` message and then had its reference cleared before anything could complete it.

i pushed two commits on top rather than merging it as it stood. the first hoists the `isFinal` rule above the branch instead of adding a second copy of it inside the create branch. the duplication is the mechanism here: the rule lived in one of two branches, and copying it into the other leaves the same trap for whoever adds a third. what makes the shared `status` safe in the update branch is an invariant worth knowing before you read that line as a downgrade: a final transcript clears `_currentAssistantMsg` at the end of the handler, and both `_finishVoiceAssistantMessage` callers null it immediately after, so a non-null `_currentAssistantMsg` is always `running` and assigning `running` to it is a no-op.

the second moves the test onto the `TestRuntime` and `createVoiceAdapter` harness the file already carries, with `emitTranscript` added next to the existing `emitVolume`. a base-class test that constructs `LocalRuntimeCore` and a stub chat model to reach `BaseThreadRuntimeCore` behavior is getting its coverage through a sibling runtime it does not need.

what i ran: `vitest run src/runtime/base/base-thread-runtime-core.test.ts src/adapters/voice.test.ts` gives 16 passed on the branch, and with `base-thread-runtime-core.ts` checked out back to `55c34a625` the new test fails with `"type": "running"` where `{ type: "complete", reason: "stop" }` is expected. one warning for the next person: the full `@assistant-ui/core` suite reports 61 failures on this branch, which looks like the change broke something and does not. the count is identical before and after, every one is under `src/react/**`, and a clean `main` checkout reproduces them from an unbuilt `@assistant-ui/tap`.

i also swapped the reproduction in the description. `examples/with-elevenlabs-conversational` maps every `onMessage` to `emitTranscript({ ..., isFinal: true })` and never emits a partial, so every assistant reply in that example is affected, and when the agent sends two replies in a row the first never leaves `running` at all because `_finishVoiceAssistantMessage` only repairs `at(-1)`. that is a reading of the adapter rather than something i drove end to end, but it is a stronger claim on the fix than the synthetic script, and it matches what `apps/docs/content/docs/guides/voice.mdx` already documents.

PR #7085, opened 2026-09-08
fix: preserve Expo-managed dependency versions

Solaris-star:
## Problem

When the dependency updater cannot complete the fresh install or `expo install --fix`, its fallback must restore Expo-managed dependency versions from the committed manifest without reverting unrelated dependency updates.

## Change

- Preserve the fallback coverage for Expo and React Native package families.
- Add focused tests for failures during both the preliminary install and Expo repin stages.
- Keep the test fixture aligned with the current Expo example manifest.
- Run the dependency-updater test from the existing Code Quality workflow rather than a standalone workflow.
- Remove stale documentation and comments about a direct `@react-native/metro-config` dependency that no longer exists.

## Verification

- `node --test scripts/update-deps.test.mjs`
- `pnpm exec oxfmt --check scripts/update-deps.test.mjs`
- `pnpm exec oxlint scripts/update-deps.test.mjs`
- `git diff --check`

All focused checks pass locally.

## Public surface

None. This changes repository maintenance automation, its focused regression tests, workflow wiring, and maintainer documentation.


Kinfe123 (commented):
good one but can we remove or escape the apostrophes inside this single-quoted node -e block? react-native's closes the shell string early; on the current head the command fails with bash: //: is a directory, so the fallback never restores the dependencies. A test that executes this fallback would catch it.

Solaris-star:
Maintenance follow-up pushed in bbd0d342f (on top of af0c0a2f7).

- Restored the scoped fallback predicate for @react-native/* and documented why @react-native/metro-config must track react-native; the note distinguishes fallback restoration from successful Expo repinning.
- Added live scoped and SDK-matrix fixture entries (@react-native/metro-config and @shopify/flash-list) plus assertions for install, Expo-repin failure, and success paths.
- Verified: node --test scripts/update-deps.test.mjs (3/3), shellcheck -S error scripts/update-deps.sh, oxfmt, oxlint, and git diff --check.

Issue #7032, opened 2026-09-08
flaky: react-google-adk 'ignores superseded run events after cancellation' fails with Entry not available in the store


okisdev:
`packages/react-google-adk/src/useAdkRuntime.replacement.test.tsx` failed once on the Test Changed Packages job:

```
❯ src/useAdkRuntime.replacement.test.tsx (3 tests | 1 failed) 1111ms
  × ignores superseded run 'events after cancellation' 614ms

FAIL src/useAdkRuntime.replacement.test.tsx > useAdkRuntime replacement runs > ignores superseded run 'events after cancellation'
Error: Entry not available in the store
```

seen on https://github.com/assistant-ui/assistant-ui/actions/runs/34205504054 (#7028, a change confined to `apps/docs`, which invalidated this package's test cache and made it re-run). the rerun of the same job on the same commit passed, and the file passes 5 out of 5 locally, so it is intermittent rather than a break.

the sibling case, `ignores superseded run 'errors without cancellation'`, passed in the same run, so whatever the race is, it is specific to the cancellation path.

`Entry not available in the store` is the remote thread list's own error and has been a real bug before (#3653, #3971, #6050, #6177, #6645, all closed), so this is worth a look rather than a retry annotation.

<a href="https://rupic.dev/s/s2.prKaVuU-uvyjy8HiFgm8ws5serpMmIhvs4LTQYe9dzF0OMiLen02Ih1cuaw?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg"><img alt="Track in Rupic" src="https://rupic.dev/badges/track-light-v1.svg"></picture></a>

PR #7368, opened 2026-09-14
fix(assistant-stream): cancel idle children when transforms stop

Kinfe123:
## Problem

Cancelling an `AssistantTransformStream` can leave its merged child streams running and locked when they are idle. The same missing cleanup occurs if the input is aborted or a transformer callback rejects.

Reproduced on main at `b443165bb9f2c6a17347607d4436bc95c9126c12` (`assistant-stream` source version `0.3.42`). This minimal example never calls the child's `cancel()` on that commit:

```ts
import type { AssistantStreamChunk } from "assistant-stream";
import { AssistantTransformStream } from "assistant-stream/utils";

let cancelled = false;
const child = new ReadableStream<AssistantStreamChunk>({
  cancel() {
    cancelled = true;
  },
});
const transform = new AssistantTransformStream({
  start(controller) {
    controller.merge(child);
  },
});

await transform.readable.cancel();
console.log(cancelled, child.locked); // expected: true, false
```

## Root cause

The controller's internal stream is piped into the transform, but cancellation and errors on the transform do not stop that pipe. Cleanup only happens incidentally if a later chunk tries to enqueue into the closed output. An idle child never produces that chunk.

## Change

Give the internal pipe an abort signal and expose a non-prefetching readable wrapper that forwards reads, cancels the pipe, and waits for child cleanup. Observe the underlying reader's errors so input aborts and rejected callbacks also clean up without waiting for another read. A synchronous `start` exception gets the same cleanup before being rethrown.

The wrapper follows the existing `PipeableTransformStream` pattern. It does not rely on `Transformer.cancel`, which is [not implemented across the supported browsers](https://raw.githubusercontent.com/mdn/browser-compat-data/main/api/TransformStream.json). The original writable/readable strategies remain on the underlying transform; the wrapper has a zero high-water mark and does not prefetch.

## Verification

- Nine focused cases cover cancellation before the first read, cancellation after output, a pending read with delayed child cleanup, input abort, synchronous start failure, three rejected callback phases, and normal flush/output ordering.
- The idle-child cancellation, input-abort, and three asynchronous callback-error tests fail against the unchanged base implementation and pass with the fix.
- `pnpm --filter assistant-stream exec vitest run`: 588 passed, 22 skipped across 39 files. The local run invoked installed Vitest directly with `--configLoader native` to reuse dependencies in the isolated checkout.
- The peer-v5 config also completed with 588 passed and 22 skipped, using `--configLoader runner`.
- `aui-build`, including declarations, passed.
- Focused oxlint, oxfmt, changeset validation, and `git diff --check` passed.
- Measured all five published entries. The utils entry changes from 4,604 to 4,953 gzipped bytes (+349 bytes); its budget update is included. Other entries stay within their existing budgets.

## Public surface

`assistant-stream`: lifecycle fix, regression tests, patch changeset, and one measured size-budget update. No new exports or constructor options. No documentation, API-reference, or template changes.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/7368?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.-C67bWlI7IrRIP96j3Wdg173yDNqI7TD2OLT_kJRKXz3dIgGPmhcgfMpp8I?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>


Kinfe123:
Checked the CI failure and reran only the failed job on the unchanged commit `8f4181e68`.

The failure was in the Google ADK replacement-run test (`ignores superseded run events after cancellation`), with `Entry not available in the store`. This matches the intermittent failure tracked in #7032. The same test file also passes locally, all 3 tests.

The rerun is green: 88/88 build and test tasks passed, followed by 11/11 React Compiler tasks. The performance contract suite ran this time and passed all 92 tests across 23 files. All PR checks are now passing or explicitly skipped. [Successful rerun](https://github.com/assistant-ui/assistant-ui/actions/runs/34871338141/job/104453052543).

For the performance review, the existing [comparison report](https://github.com/assistant-ui/assistant-ui/pull/7368#issuecomment-5667841956) flags the 100-delta decode case at +16.1% (608.11 to 706.18 microseconds). The 1,000- and 4,000-delta decode cases are +5.4% and +6.8%, both within the report's noise thresholds. The encode cases are also within noise. The passing counter contracts do not dismiss the flagged wall-time result.

No source changes were needed to clear this CI failure. The underlying intermittent ADK issue is still open; a passing rerun does not resolve it.

PR #7382, opened 2026-09-14
fix(react-google-adk): normalize nested media field aliases

Kinfe123:
## Problem

ADK events using snake_case media fields can crash image rendering or lose a file URL. This affects custom streams and stored sessions that use Python-style field names. The normal camelCase payload remains supported.

Reproduced on main `b443165bb9f2c6a17347607d4436bc95c9126c12`, package version `@assistant-ui/react-google-adk@0.0.29`:

```ts
import { AdkEventAccumulator } from "@assistant-ui/react-google-adk";

const event = JSON.parse(JSON.stringify({
  id: "image-1",
  author: "agent",
  content: {
    parts: [{ inline_data: { mime_type: "image/png", data: "aGVsbG8=" } }],
  },
}));

new AdkEventAccumulator().processEvent(event);
```

Before the fix, this throws when the converter calls `startsWith` on the missing camelCase MIME field. Similarly, `file_data: { mime_type: "application/pdf", file_uri: "https://example.test/report.pdf" }` produces an image URL part with an undefined URL instead of the PDF file part.

## Root cause

The accumulator already recognizes `inline_data` and `file_data`, but only renames those outer fields. Nested `mime_type` and `file_uri` never reach the camelCase fields used by the content converters.

## Change

Extend the existing event normalization boundary to map the known nested media aliases. Apply the mapping to both camelCase and snake_case containers, preserving camelCase values when both forms are present.

The input event is not mutated, and opaque tool arguments are not recursively renamed. User and assistant events use the same normalization path, so the fix covers both live streams and session replay without separate conversion logic.

## Verification

- Before the source change, **14 regression cases failed** with the reported exception or incorrect file output. Existing and camelCase control cases passed.
- `vitest run --configLoader native --no-fsModuleCache` in `packages/react-google-adk`: **375 tests passed across 15 files**.
- Added 20 cases covering image and file media, user and assistant messages, mixed casing, camelCase precedence, input immutability, opaque tool arguments, SSE consumption, and session loading.
- Scoped oxlint and oxfmt checks passed for all four changed TypeScript files.
- `aui-build` passed. Both built ADK entries pass the existing size budgets; unrelated unbuilt packages were skipped.
- `node scripts/check-changesets.mjs` and `git diff --check` passed.
- The package type check is not green: `tsc6 --noEmit` reports the same five pre-existing test errors on both unchanged main and this branch, in `useAdkMessages.test.ts` and `useAdkRuntime.replacement.test.tsx`. No new type diagnostics were introduced.

## Public surface

`@assistant-ui/react-google-adk` behavior fix with a patch changeset. No new exports, dependencies, documentation pages, API-reference changes, or template changes.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/7382?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.EipA4mJOcYDIoVsQTsWksT1WX6Bz3IxJrrO6B_IbABleI77Qkaes2dnqXkY?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>


Kinfe123:
Synced with current main in 286670442 and resolved the two test conflicts from #7381. Kept both the media alias tests and the tool-error tests. All 416 ADK tests pass, along with scoped lint, formatting, the package build, and size checks. The same five pre-existing test type errors remain unchanged. No feature changes were added during the conflict resolution.

okisdev (approved):
approving. the fix completes the existing normalization boundary instead of adding a conversion path, so live SSE events and session replay both pick up the nested `mime_type` and `file_uri` aliases through the same `normalizeEventPart` call.

ran it in a worktree rather than reading it: on the merge base `c3cfdafbd` your new cases fail with the `startsWith` TypeError, on `286670442` the full ADK suite passes (416 tests), and the merge tree with main at `622f6a3e4` passes as well. the 5 `tsc` errors in `useAdkMessages.test.ts` and `useAdkRuntime.replacement.test.tsx` are identical on main, so they are not from this change.

two related gaps stay outside this PR and are tracked separately: a media part with no mime type still throws in the accumulator (#7431), and `artifacts.load` still rejects a snake_case `inline_data` response (#7432).

PR #7493, opened 2026-09-15
fix(core): preserve tool-result modelContent and artifact

Kinfe123:
## Problem

A client-side tool that returns `new ToolResponse({ result, modelContent: [...] })` — the documented way to attach model-facing content distinct from the UI result — had its `modelContent` silently dropped on two paths:

1. `MessagePartRuntime.addToolResult` (the public API a tool UI calls) forwarded `result`, `artifact`, and `isError` from the `ToolResponse` but not `modelContent`. So for **every** runtime reached through the public API, the model-facing content was lost, even though `AddToolResultOptions.modelContent` is a documented field ("Forwarded through `adapter.onAddToolResult`") and the external-store and assistant-transport paths already forward it. A cross-runtime parity break.
2. `LocalThreadRuntimeCore.addToolResult` never read `modelContent`, so it was absent from the stored tool-call part (and from the payload the resumed adapter sees). It also spread `artifact` unconditionally, so a later `addToolResult` for the same call that omitted `artifact` overwrote a stored one with `undefined`.

## Root cause

`modelContent` was omitted from the field list in both `MessagePartRuntimeImpl.addToolResult` (never copied off the `ToolResponse`) and `LocalThreadRuntimeCore.addToolResult` (never destructured or stored). The artifact clobber is the same shape: an optional field spread unconditionally overrides a previously stored value with `undefined`.

## Change

- `MessagePartRuntimeImpl.addToolResult` forwards `response.modelContent` into `AddToolResultOptions` (conditionally, per `exactOptionalPropertyTypes`), alongside the `artifact`/`isError` it already forwarded.
- `LocalThreadRuntimeCore.addToolResult` destructures `modelContent` and stores it on the part, and only overrides `artifact`/`modelContent` when supplied — so a later result that omits them preserves the stored value, while `result`/`isError` are still written every call.

## Verification

- `packages/core`: `pnpm vitest run` — 2010 passing.
- New tests, each with regression teeth:
  - `LocalThreadRuntimeCore.addToolResult content` — stores `modelContent` and forwards it to the resumed adapter (fails without the local-runtime fix); keeps a stored artifact when a later result omits it (fails against the unconditional spread).
  - `useAssistantTransportRuntime` — a `ToolResponse` `modelContent` reaches the outbound `add-tool-result` command through the public `MessagePartRuntime.addToolResult` API (fails without the `MessagePartRuntimeImpl` fix; verified by reverting the hunk).
- `tsc --noEmit` adds no new errors in the changed files (the pre-existing test-file type errors on `main`, tracked in #6801, are unrelated); `pnpm lint` clean.

## Public surface

- `@assistant-ui/core`: behavior fix only, no API changes. Changeset included (patch).

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/7493?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.HddOryTr_5vRIjAhMyfIC8Ajxy7b9j6o5rThK72W_iaQJMu9kK8kqX88R2E?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>


Kinfe123:
Thanks — both notes addressed:

- **Inline nit (message-part-runtime.ts):** applied in 24e32a6 — `artifact` is now spread conditionally too, so the API layer no longer hands an adapter `artifact: undefined` explicitly. The falsy-artifact test still passes (false/0/""/null are `!== undefined`).
- **react-ag-ui parity gap:** confirmed and fixed in a focused follow-up, #7494. `AgUiThreadRuntimeCore.addToolResult` had the same two defects, and as you noted `emitToolResult` already prefers `part.modelContent` — so post-merge the value reaches ag-ui's core and was dropped one line before use. Kept it separate to leave this PR core-scoped (both reviewers approved it as-is) and because that file also has an unrelated PR in flight. #7494 stores `modelContent` and makes `artifact`/`modelContent` non-clobbering, mirroring this PR, with a regression test that fails on main.

PR #7494, opened 2026-09-15
fix(react-ag-ui): preserve tool-result modelContent and artifact

Kinfe123:
## Problem

`AgUiThreadRuntimeCore.addToolResult` has the same two defects that #7493 fixes in core, one runtime over:

1. It never stored the result's `modelContent`. ag-ui's outbound conversion already prefers it — `emitToolResult` (`packages/react-ag-ui/src/runtime/adapter/conversions.ts:1138`) sends `extractText(part.modelContent)` when present and falls back to `JSON.stringify(part.result)` otherwise. So a frontend tool's model-facing content was dropped one line before the code that would have sent it to the agent, and the agent received the UI result instead.
2. It spread `artifact` unconditionally, so a later `addToolResult` for the same call that omitted `artifact` overwrote a stored one with `undefined`.

This is the cross-runtime parity gap flagged in the #7493 review: once core forwards `modelContent` through the public `MessagePartRuntime.addToolResult` API, the value now reaches ag-ui's core and was being dropped here.

## Change

`AgUiThreadRuntimeCore.addToolResult` stores `modelContent` on the tool-call part and only overrides `artifact`/`modelContent` when they are supplied, so a later result that omits them preserves the stored value while `result`/`isError` are still written every call. This matches `LocalThreadRuntimeCore.addToolResult` (#7493).

## Verification

- `packages/react-ag-ui`: `pnpm vitest run` — 581 passing. The one unrelated failure (`useAgUiRuntime.toolDeferral` — "does not run a frontend execute on a call the agent answers itself") fails identically on unmodified `main`, independent of this change.
- New test `stores modelContent and keeps a stored artifact on addToolResult` — asserts a frontend `addToolResult` stores `modelContent` and `artifact`, and that a later result omitting them preserves both. Fails on `main` (verified by reverting the fix).
- `tsc --noEmit` clean for react-ag-ui; `pnpm lint` clean.

## Public surface

- `@assistant-ui/react-ag-ui`: behavior fix only, no API changes. Changeset included (patch).

## Related

Cross-runtime parity follow-up to #7493 (core). Independent of it: this stores what the runtime is given, and its end-to-end effect (model content reaching the agent) activates once #7493 delivers `modelContent` through the public API.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/7494?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.nk_UGSC9kOLYZ1SbRq0NGGKtVAfqDAwMvcqBg60uicijX3N6ndyixA1uPmk?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>


Kinfe123:
Good catch — fixed in 30cf04e. You're right that storing `modelContent` in `addToolResult` wasn't enough: `preserveToolResults` (~L1590) rebuilds the part after the RUN_FINISHED aggregator snapshot and copied only `result`/`artifact`/`isError`, so `modelContent` was stripped before `emitToolResult` ran and it fell back to `JSON.stringify(part.result)`.

`preserveToolResults` now carries `modelContent` too (conditionally, same shape as `artifact`/`isError`). And I extended the end-to-end test you pointed at ("auto-resumes when a tool result is added before RUN_FINISHED") — the mid-run `addToolResult` now carries `modelContent`, and it asserts both that the stored part keeps it across the snapshot rebuild and that `runInputs[1]`'s tool message content is the model-facing text (`"22C and sunny"`), not the result JSON. Verified it fails with only the `addToolResult` fix and passes with the `preserveToolResults` fix too.

react-ag-ui suite: 581 passing (the one unrelated `toolDeferral` failure fails identically on main); tsc clean.

Kinfe123:
Thanks @cubic-dev-ai — looked at both:

**@883 (preserveToolResults drops modelContent):** already fixed in 30cf04e (pushed before this review ran) — `preserveToolResults` now carries `prior.modelContent` alongside `result`/`artifact`/`isError`, and the extended end-to-end test asserts the follow-up run receives the model-facing text.

**@882 (file-only modelContent → empty tool result):** this is intentional, not a regression. `test/conversions.spec.ts:334` ("does not serialize Host-only data when modelContent is empty") locks the contract: when `modelContent` is present but yields no text, `emitToolResult` deliberately sends empty content rather than falling back to `JSON.stringify(part.result)`. That's a privacy guarantee — the raw result can carry host-only data (base64 images, `structuredContent`) that must not leak to the agent. Falling back to `result` on empty extraction, as suggested, would defeat that test and leak exactly that data. AG-UI tool content is a text-only string, so file-only model content has no wire representation here by design; a tool that wants text to reach the agent includes a text part in `modelContent`. So I'm leaving this as-is — I tried the fallback and it broke the host-only-data contract, which is the correct behavior to keep.

PR #7498, opened 2026-09-15
fix(react-pi): cancel a send during session open

Kinfe123:
## Problem

Send a message, then immediately press Stop while the Pi session is still opening (a cold thread with no live `AgentSession` yet). The cancel returns successfully (HTTP 204), but the session finishes opening and the prompt launches anyway — the run the user tried to stop starts.

## Root cause

`PiThreadSupervisor.cancelRun` (`packages/react-pi/src/node/ThreadSupervisor.ts`) only aborts a live record's session:

```ts
async cancelRun(threadId) {
  await this.records.get(threadId)?.session.abort();
}
```

During a cold open the thread lives in `pendingOpens`, not `records`. So `records.get(threadId)` is `undefined`, the optional chain no-ops, and `cancelRun` resolves having done nothing. Meanwhile `sendMessage` (`await this.send(await this.ensureOpen(threadId), input)`) has no cancellation check between the open resolving and `send` → `session.prompt(...)`, so the prompt fires. `deleteThread`/`dispose` already abort pending opens; `cancelRun` never got the same treatment.

## Change

`sendMessage` now registers a per-send cancellation token in a `startingSends` map for the window between starting the cold open and launching the prompt. `cancelRun` flips that token (in addition to aborting a live session). After `ensureOpen` resolves, `sendMessage` checks the token and returns without calling `send` if it was cancelled. Unlike the delete path, this does not dispose the opened session — the thread stays live and usable; only the run is skipped, matching "cancel pending startup work before it can launch the prompt." The token is keyed by object identity and cleared in a `finally`, so a cancel after the send has launched only aborts the live session (the existing behavior) and never leaks to a later send.

## Verification

- `packages/react-pi`: `pnpm vitest run` — 266 passing.
- New test "cancels a send whose session is still opening without launching the prompt": holds `createAgentSession` open, starts a send, calls `cancelRun` while the open is pending, resolves the session, and asserts `prompt` is never called. Fails on `main` (verified — the prompt fires) and passes with the fix.
- `tsc --noEmit` adds no new errors (the pre-existing test-file type errors, tracked in #6801, are unrelated); `pnpm lint` clean.

## Public surface

- `@assistant-ui/react-pi`: behavior fix only, no API changes. Changeset included (patch).

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/7498?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.qy4aLC4oop742Vcs2IP9A3UMoGPwPeLYgO-_S6kM1vQuVuCTkYHei2Op2Cs?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>


Kinfe123:
Addressed both issues in 9e9be23.

**Stuck UI after cancel (@claude, @cubic-dev-ai — P1):** you're right — returning silently left the controller's optimistic message and `running` mark with no event to settle them, so the spinner and ghost message stuck. The cancelled cold send now **rejects** (`"Pi run was cancelled before it started"`) instead of resolving. That drives `PiThreadController.sendMessage`'s existing send-path catch (`ThreadController.ts:446-461`), which removes the optimistic user message and clears `running` — the same rollback a failed send already uses. I chose reject over emitting a settle event because an `agent_end`-style event would clear the spinner but not remove the optimistic message (nothing reconciles it without a real message arriving), whereas the catch does both.

**Concurrent sends overwriting the token (rupic, @claude, @cubic-dev-ai — P2):** `startingSends` is now `Map<string, Set<token>>`; each send adds its own token and `cancelRun` flips every token for the thread, so a second send sharing the same in-flight cold open is cancelled too. Added a regression test with two sends sharing one cold open — both reject, the prompt never fires, and only one `createAgentSession` runs.

PR #7499, opened 2026-09-15
fix(react-pi): ignore stale queue-clear responses

Kinfe123:
## Problem

Clear the queue, then queue another message. If the clear request's response is slow, the old response arrives and empties the UI queue — even though the newer message is still queued on the server. The queue looks empty locally while the server holds the message.

## Root cause

`PiThreadController.clearQueue` (`packages/react-pi/src/runtime/ThreadController.ts`) empties the local mirror *after* its request resolves, reading `this.state.queue` fresh at continuation time:

```ts
public async clearQueue() {
  const cleared = await this.client.clearQueue(this.threadId);
  if (this.state.queue.steering.length > 0 || this.state.queue.followUp.length > 0) {
    this.setState({ ...this.state, queue: { steering: [], followUp: [] } });
  }
  return cleared;
}
```

Any queue mutation that lands during the await — a later `sendQueued` optimistic append, or a server `queue_update` that reflects a newer message — is clobbered by the blind empty. The clear response carries no sequence/version, and this write bypasses the reducer's seq guard (it goes straight through `setState`), so it is a last-writer race whose outcome depends purely on whether the newer `queue_update` lands before or after the clear response.

## Change

Snapshot the queue reference before awaiting and only empty when it is unchanged. Both mutation paths allocate a fresh `queue` object on every change (`sendQueued` spreads a new array; the `queue_update` reducer replaces the arrays wholesale), so reference identity is a reliable generation check: if anything newer landed during the await, `this.state.queue !== queueBefore` and the stale empty is skipped, leaving the server-authoritative state intact.

## Verification

- `packages/react-pi`: `pnpm vitest run` — 266 passing (the existing happy-path clear test still passes: no intervening mutation ⇒ reference unchanged ⇒ empty still applies).
- New test "does not empty a queue that a newer message repopulated before the clear response": defers the clear response, emits `queue_update` clearing then repopulating the queue while the clear is in flight, resolves the clear, and asserts the newer message survives. Fails on `main` (verified — the queue is wiped) and passes with the fix.
- `tsc --noEmit` adds no new errors (pre-existing test-file type errors, #6801, are unrelated); `pnpm lint` clean.

## Public surface

- `@assistant-ui/react-pi`: behavior fix only, no API changes. Changeset included (patch).

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/7499?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.uN1Ed-RKNbi5IyAYJt7n9uir0v30q5_mr_GOMlWE3UG-P4rYsEeDlpq0ukc?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>


Kinfe123:
Addressed the review notes in 3b80449.

**Comment accuracy (@cubic-dev-ai P3, @claude):** you're right — `applySnapshot` is a third queue writer and it allocates a fresh queue object even when the contents are unchanged, so a changed reference can be a reconnect/refresh, not a newer message. Rewrote the comment to say exactly that, and that the guard deliberately biases toward skipping the empty (a stale entry self-heals on the next `queue_update`, whereas emptying could drop a message the server still holds).

**sendQueued-path coverage (@cubic-dev-ai P3):** added a second test that repopulates the queue via an optimistic `sendMessage` (not a `queue_update`) while the clear is in flight, and asserts the entry survives. Fails on `main`, passes with the fix.

**Changeset (@cubic-dev-ai P3):** trimmed to the user-observable behavior, dropped the internal reference-snapshot detail.

**@cubic-dev-ai P2 (@514 — a queued send that began during the clear later rejects and its rollback rebuilds a pre-clear queue):** this lands in the benign direction — the local mirror keeps a stale entry until Pi's next `queue_update` heals it, rather than dropping a live message. It's the same trade-off @claude flagged as acceptable here, so I'm leaving it rather than adding cross-operation reconciliation. (Note #7501 separately hardens the `sendQueued` rollback itself to gate on queue identity.)

PR #7501, opened 2026-09-15
fix(react-pi): roll back only the failed queued send

Kinfe123:
## Problem

Queue the same text twice while a run is active. If the first request fails *after* the second succeeds, the first send's failure cleanup removes the **successful** message from the local queue instead of its own entry — the genuinely-queued message vanishes from the UI.

## Root cause

`PiThreadController.sendQueued` (`packages/react-pi/src/runtime/ThreadController.ts`) rolls back a failed optimistic enqueue by content and position:

```ts
const entries = this.state.queue[mode];
const index = entries.lastIndexOf(input.content);
// ...filter out index
```

Two flaws combine:

1. `lastIndexOf(input.content)` can't distinguish two entries with identical text.
2. It reads `this.state.queue[mode]` as it stands at failure time. A server `queue_update` — or a snapshot on (re)connect/refresh — replaces the queue wholesale (`threadState.ts` / `applySnapshot`), so by the time the first send fails, the queue can already be reconciled down to the single genuinely-queued entry (the second send's). `lastIndexOf` then matches and deletes that surviving entry.

Because a running thread is subscribed to live events, the `queue_update` self-heal is the normal path, so the failure is reachable in practice.

Note: the public `PiThreadState.queue` is exported as `{ steering: readonly string[]; followUp: readonly string[] }`. Reshaping it to carry per-entry ids (the obvious "match by id" fix) would be a breaking public change, so this fix keeps the string-array shape.

## Change

Guard the rollback on **optimistic-queue object identity**. `sendQueued` builds a fresh `optimisticQueue` object when it enqueues and stores it as `this.state.queue`. On failure it compares the current `this.state.queue` against that captured reference:

```ts
const reconciled = this.state.queue !== optimisticQueue;
const index = reconciled ? -1 : entries.lastIndexOf(input.content);
```

Every path that rewrites the queue — a `queue_update`, a snapshot via `applySnapshot`, a `clearQueue()`, or a sibling send's own optimistic enqueue — installs a **new** queue object, so the identity check fails and the stale content-based rollback is skipped. The optimistic entry is then no longer ours to match by content: the server (or the sibling write) is authoritative, and any genuinely-failed entry self-heals on the next `queue_update`. Only when nothing has touched the queue since we enqueued — the object reference is still exactly what we set — does the content-based `lastIndexOf` rollback run, which is correct for the common single-entry and distinct-text cases.

This is strictly stronger than keying on `queue_update` alone: because *any* replacement of the queue object invalidates the rollback, it also protects a later identical send and a message re-enqueued after `clearQueue()`, not just the server-reconcile path.

## Verification

- `packages/react-pi`: `pnpm vitest run` — all passing (the existing single-entry rollback test still passes: nothing replaces the queue object ⇒ identity holds ⇒ rollback still runs).
- New test "does not remove a surviving identical message when an earlier send fails": queues "hello" twice, lets the second succeed, emits a `queue_update` reconciling to the one surviving "hello", then fails the first send, and asserts the surviving message is untouched. Fails on `main` (verified — the surviving message is deleted) and passes with the fix.
- New snapshot-path regression test: applies an authoritative snapshot containing the surviving identical entry between enqueue and failure, and asserts the earlier failed send does not delete it (exercises the `applySnapshot` replacement, not just `queue_update`).
- `tsc --noEmit` adds no new errors (pre-existing test-file type errors, #6801, are unrelated); `pnpm lint` clean.

## Public surface

- `@assistant-ui/react-pi`: behavior fix only, no API changes (public `queue` shape unchanged). Changeset included (patch).

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/7501?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.rsyI117SfFAST4tnfk8_KXUz5AX5LsO2xUaH7eLJdPzg_83Vequ11YvoRFk?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>


Kinfe123:
Addressed the converging feedback from all three reviewers in 5212487.

**Snapshot reconciliation path (rupic, @claude, @cubic-dev-ai — P1):** you're right that `queue_update` isn't the only wholesale queue writer — `applySnapshot` rebuilds the queue on every (re)connect snapshot, `refreshInBackground`, and user `refresh()`, so my `queueGeneration`-on-`queue_update` counter left the bug reachable via the snapshot path. Rather than enumerate event types (and re-chase this whenever a new one touches the queue), I switched to gating on the **queue object's reference identity**, as @claude suggested: the rollback now runs only while `this.state.queue` is still the exact object the optimistic append set. Any reconciliation — `queue_update`, snapshot, or clear — replaces that object, so the stale content-based removal is skipped and Pi's authoritative state stands; a later `queue_update` self-heals. This also covers the clear path (@cubic-dev-ai P1 @494) for free.

Added a second regression test that reconciles via a **snapshot** event (not `queue_update`); both tests fail on `main` and pass with the fix.

On @cubic-dev-ai @496 (two identical sends pending, earlier fails before any reconciliation): the identity gate biases toward not-removing there too, so it never deletes the surviving entry — the failed entry lingers briefly until the next `queue_update` heals it, which is the benign direction (@claude's framing on #7499). Removed the queueGeneration counter entirely, and trimmed the changeset to the user-observable behavior (@cubic-dev-ai P3).

Kinfe123:
Thanks — fixed the description drift.

**Blocking item (description ↔ code mismatch):** you're right, the description narrated a private `queueGeneration` counter that was never in the code. The implementation actually guards on **optimistic-queue object identity** — `sendQueued` captures the `optimisticQueue` object it installs, and on failure rolls back only while `this.state.queue === optimisticQueue`. I've rewritten the **Root cause**, **Change**, and **Verification** sections to describe that mechanism (and noted it's strictly stronger than keying on `queue_update` alone, since any queue-object replacement — `queue_update`, `applySnapshot`, `clearQueue()`, or a sibling send — invalidates the rollback).

**Line 616 / snapshot path:** matches your Round 2 table — the object-identity guard already covers `applySnapshot` because a snapshot installs a new queue object, and the snapshot regression test exercises exactly that (authoritative snapshot with a surviving identical entry between enqueue and failure). No code change needed there.

No functional change in this update — description only.

PR #7610, opened 2026-09-16
fix(react-opencode): reconcile pending requests on reconnect

Kinfe123:
## Problem

An approval or question that is answered while this client is disconnected still shows as pending after the stream reconnects. The prompt stays on screen with nothing behind it, and answering it replies to a request the server has already resolved.

Reproduced on main by the new regression test:

```
AssertionError: expected { id: 'permission_1', … } to be undefined
```

## Root cause

On reconnect the controller lists the server's permissions and questions and dispatches `*.asked` for anything not already pending. The sync is additive only: nothing ever removes a local pending request that the server has stopped reporting, which is exactly the shape of a request resolved during the disconnect.

## Change

The reconnect sync now reconciles in both directions. It records which requests were already pending when each list call was issued, and after the response drops the ones the server did not report back.

Only that captured set is eligible for removal. A request that arrives live while the list is in flight is newer than the response and is left alone, so a genuinely new prompt is never swallowed by an older list.

Withdrawal is a distinct event from a reply (`permission.withdrawn` / `question.withdrawn`): the reply the request actually received happened elsewhere and is not knowable from here, so the entry is dropped from `pending` rather than fabricated into `resolved`.

## Verification

- New `drops a pending request the server no longer reports after reconnect`: one request is pending before the reconnect, a second is asked while the list call is in flight, and the server answers with neither. The first is dropped, the second survives. Fails on main with the diff above.
- The existing reconnect tests still pass, so an additive sync with no missing requests is unaffected.
- `pnpm -C packages/react-opencode test`: 143 passed (12 files).
- `pnpm -C packages/react-opencode build` passes.

## Public surface

`OpenCodeStateEvent` gains two members, `permission.withdrawn` and `question.withdrawn`. The union is additive, so no existing member changes. Changeset: patch for `@assistant-ui/react-opencode`.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/7610?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.0KUS7VdZQvqivgGxzapQfsDUzXiKtqP90gbbiQfZclcpscs5XR4GpR5XVGA?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>


Kinfe123:
Closing as a duplicate of #7609, which fixes the same bug (interactions settled while the event stream was disconnected) without widening the public event surface. Consolidating there; the in-flight-reply race raised in review is being verified against that branch.

PR #7676, opened 2026-09-17
fix(docs): prerender the changelog instead of rendering it per request

Yonom:
## Problem

`/changelog` and `/changelog?page=N` are the slowest pages on www.assistant-ui.com. Ahrefs flags them as slow (TTFB 1.5 to 2 s in the crawl), and every hit is a Vercel cache MISS.

## Root cause

The page reads `searchParams` for `page` and `pkg`, which makes the route dynamic. Every request re-reads the release pages from the data cache and re-renders about 1.6 MB of markdown-derived HTML plus a 1 MB RSC payload on the server.

## Change

Pagination and the package filter move from query strings to path segments on an optional catch-all route: `/changelog`, `/changelog/2`, `/changelog/@assistant-ui/react`, `/changelog/@assistant-ui/react/2`. The unfiltered pages are prerendered through `generateStaticParams` and revalidate hourly, matching the `fetch` revalidation the release fetcher already uses; package pages render on first request and are cached the same way. Unknown packages and out-of-range pages 404, and `/changelog/1` 404s so the first page has one URL.

The old query URLs redirect permanently to the new paths from `proxy.ts`. A `next.config` redirect with `has` query matchers would carry the matched query along onto the destination, which is why the redirect lives in the proxy.

## Verification

Against `next dev`:

- `/changelog`, `/changelog/2`, `/changelog/@assistant-ui/react`, `/changelog/@assistant-ui/react/2`, `/changelog/assistant-stream` render with path-based prev/next links.
- `/changelog/1`, `/changelog/nope`, `/changelog/999`, `/changelog/@assistant-ui/react/999` return 404.
- `/changelog?page=2`, `/changelog?pkg=@assistant-ui/react`, `/changelog?pkg=@assistant-ui/react&page=2`, `/changelog?page=1` return 308 to the path form with no query string.
- `pnpm -C apps/docs typecheck` and oxlint pass on the changed files.

## Public surface

`apps/docs` only (private, no changeset). The sitemap entry for `/changelog` is unchanged.

https://claude.ai/code/session_018UDw8eG1MVXNPp9rSgHLGR

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/7676?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.U9AJ9DOF3yleADz3GRIcltGH4uI3EgOLG2b0MRydcu_r1xezivvTCAbBFs8?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>


Yonom:
Follow-up commit addresses the review points:

- Pagination and package `notFound()` are now gated on a non-empty release fetch, so a degraded GitHub response renders the existing "Unable to load releases" page instead of a cached 404.
- The proxy matcher only fires when `pkg` or `page` is in the query, so the bare `/changelog` never hits the function.
- `pkg` is validated against an npm package-name pattern before it becomes a path segment, and the page number is normalised through `Number`, so `?pkg=../admin` and `?page=02` redirect to `/changelog` and `/changelog/2`.
- The `decodeURIComponent` stays: in the catch-all, `@` arrives as `%40` and the package route 404s without it. A malformed escape such as `/changelog/%25` is rejected by Next before the page runs (`/blog/%25` fails the same way on main), so the second decode cannot throw on input that reaches it.
- `url.search = ""` is intentional: the redirect target is a static page and the canonical drops the query anyway.

PR #7679, opened 2026-09-17
fix(react): reset stall detection between running messages

Kinfe123:
## Problem

In @assistant-ui/react 0.15.19, stall detection can inherit the previous message's inactivity window when a different running message has equal content. The new message can therefore be reported as stalled too early.

Minimal reproduction: render running message `assistant-a`, wait 1.5 seconds, replace it with `assistant-b` containing the same text, and use a 2-second stall timeout.

## Root cause

The activity signature included running state and content values but not message identity. Two different messages with identical content were treated as one uninterrupted activity period.

## Change

Include the active message ID in the stall-detection activity signature. Content updates retain their existing behavior, while a new message always starts its own inactivity window.

## Verification

- `vitest run src/tests/toolCallTiming.test.tsx` in `packages/react` (9 tests, type errors checked by the suite)

The new regression confirms the replacement message is not stalled after one second and becomes stalled only after its own two-second timeout.

## Public surface

- Affects `unstable_useMessageStallDetection` behavior in `@assistant-ui/react`
- Adds a patch changeset

<a href="https://rupic.dev/s/s2.xnWrPJOpq_Rff1GEAo5pINMPVOXiybQCBfOooMeD8dNdXTjqkUIRcGAGojs?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>


Kinfe123:
I rechecked the regression against the original implementation and it still passes without the message ID change. The shipped message renderers remount the scoped hook when the ID changes, so the stall window already resets. This does not confirm a production bug, and the added test cannot prove the proposed fix. Closing this instead of shipping behavior without a failing reproduction.

PR #7680, opened 2026-09-17
fix(safe-content-frame): size cache salts from encoded paths

Kinfe123:
## Problem

The cache-salt helper allocates its pathname segment from JavaScript string length but writes UTF-8 bytes into it. Browsers normally expose percent-encoded ASCII in `location.pathname`, so this is latent correctness hardening for custom hosts or test environments that supply a raw non-ASCII pathname rather than a confirmed browser crash.

## Root cause

UTF-16 code-unit length is not a valid capacity for UTF-8 data. If the helper receives a multibyte pathname, `Uint8Array.set` can exceed the allocated range.

## Change

Encode the pathname once, allocate from its encoded byte length, and reuse those bytes when building the cache salt. ASCII browser paths remain byte-identical.

## Verification

- `vitest run src/index.test.ts` in `packages/safe-content-frame` (14 tests)
- `tsc -p packages/safe-content-frame/tsconfig.json --noEmit`

The regression supplies a raw `/café` pathname to exercise the helper invariant and passes with byte-length allocation.

## Public surface

- Hardens `safe-content-frame` cache-salt construction for non-browser pathname providers
- Adds a patch changeset


Kinfe123:
@rupic-app please re-review. The description is now human-authored and frames this as pathname-provider hardening, and the test now directly asserts that rendering resolves.

PR #7683, opened 2026-09-17
fix(react-mcp): normalize duplicate persisted server IDs

Kinfe123:
## Problem

In @assistant-ui/react-mcp 0.1.17, an app-supplied MCPStorage can return persisted custom-server records with duplicate IDs. Hydrating those records creates ambiguous lookups and reaches the manager's duplicate-ID assertion instead of recovering the stored configuration.

Minimal reproduction: return two custom-server records with ID `docs` from `loadCustomServers()`.

## Root cause

Storage hydration trusted the persisted array as already unique even though server ID uniqueness is required by lookups, OAuth routing, and generated tool prefixes.

## Change

Normalize persisted custom servers during hydration with deterministic first-record-wins semantics. Keep the normalization in memory so mounting the manager never deletes records from app-owned storage. Connector/custom-server ID collisions remain rejected because they are configuration conflicts rather than duplicates within the persisted list.

## Verification

- `vitest run src/resources/McpManagerResource.test.ts` in `packages/react-mcp` (21 tests)
- `tsc -p packages/react-mcp/tsconfig.json --noEmit`

The regression verifies the deterministic in-memory first-record result without mutating storage.

## Public surface

- Affects persisted custom-server hydration in `@assistant-ui/react-mcp`
- Adds a patch changeset

## Reproduction

This is reproducible with `@assistant-ui/react-mcp@0.1.17` by supplying duplicate records from the public storage extension point:

```ts
const docs = {
  id: "docs",
  name: "Docs",
  url: "https://example.com/mcp",
  auth: { type: "none" as const },
  createdAt: 1,
};

const storage = McpCustomStorage({
  loadCustomServers: async () => [
    docs,
    { ...docs, name: "Duplicate Docs", createdAt: 2 },
  ],
  loadAuthState: async () => null,
  saveAuthState: async () => {},
  clearAuthState: async () => {},
});

const root = mount([], storage);
await vi.waitFor(() =>
  expect(root.getValue().getState().isHydrated).toBe(true),
);
```

Against base commit `eac99fe61f61a1e95eddb6da08cf79fd5e47b4ba`, mounting reaches the duplicate-ID assertion. The regression in `McpManagerResource.test.ts` passes on this branch and keeps storage untouched.

<a href="https://rupic.dev/s/s2.ykDowMaWB2MiYrvsxmehqZcXiDhv-zb8AEgZMujtLlvvE9YfmS77dgW8DnI?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>


Kinfe123:
@rupic-app please re-review. The description is human-authored, the Cubic appendix was removed, hydration no longer writes the normalized list back, duplicate IDs are reported, and the test asserts storage is untouched.

Kinfe123:
@rupic-app please re-review. The description now states the affected version and includes an exact-version runnable reproduction plus the base commit where it fails.

PR #7685, opened 2026-09-17
fix(react-mcp): preserve removals during hydration

Kinfe123:
## Problem

In @assistant-ui/react-mcp 0.1.17, removing a custom MCP server before storage hydration finishes can appear to work and then be undone. When the pending load resolves, the persisted server is merged back into memory and can remain in storage.

Minimal reproduction: delay `loadCustomServers()`, call `removeServer("docs")`, then resolve the load with the persisted `docs` record.

## Root cause

Pre-hydration additions were tracked as pending mutations, but removals of records not yet present in memory had no identity-level tombstone. The hydration merge therefore could not distinguish a deliberately removed persisted record from an untouched one.

## Change

Record IDs removed while hydration is pending, filter them from every hydration result for that resource, and persist the filtered list. All removals continue through the existing server removal or fallback cleanup path. Once cleanup succeeds, the shared removal callback records the tombstone and the existing persistence fence preserves storage ordering; failed cleanup does not commit the removal.

## Verification

- `vitest run src/resources/McpManagerResource.test.ts` in `packages/react-mcp` (21 tests)
- `tsc -p packages/react-mcp/tsconfig.json --noEmit`

The regression delays hydration, removes the unseen record, and verifies that it is neither restored in memory nor written back to storage.

## Public surface

- Affects custom-server hydration behavior in `@assistant-ui/react-mcp`
- Adds a patch changeset

## Reproduction

This is reproducible with `@assistant-ui/react-mcp@0.1.17` by delaying the public storage adapter's load:

```ts
let resolveLoad!: (records: MCPCustomServerRecord[]) => void;
const load = new Promise<MCPCustomServerRecord[]>((resolve) => {
  resolveLoad = resolve;
});

const root = mount([], McpCustomStorage({
  loadCustomServers: () => load,
  loadAuthState: async () => null,
  saveAuthState: async () => {},
  clearAuthState: async () => {},
}));

await root.getValue().removeServer("docs");
resolveLoad([docsServer]);
await vi.waitFor(() =>
  expect(root.getValue().getState().isHydrated).toBe(true),
);
expect(root.getValue().getState().customServers).toHaveLength(0);
```

Against base commit `eac99fe61f61a1e95eddb6da08cf79fd5e47b4ba`, the last assertion receives one restored server. The same regression passes on this branch.

<a href="https://rupic.dev/s/s2.tBZA95-Ta_CjrkIVfzHhFgyz8_bPOM2HYGmMp2CrZ9WUlSruAR31_5JhgdQ?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>


Kinfe123:
@rupic-app please re-review. The description is human-authored and the generated appendix was removed. The extra pre-hydration removal branch was removed so normal cleanup and failure atomicity are preserved.

Kinfe123:
@rupic-app please re-review. The description now includes the affected version, runnable delayed-hydration snippet, base commit, and the failing result without the fix.

PR #7729, opened 2026-09-18
fix(core): ignore malformed AssistantFrame messages

Kinfe123:
## Problem

AssistantFrame host and provider listeners trust the nested `message` payload after checking only the event source, origin, and channel. A configured frame can therefore crash either listener with payloads such as `message: null` or malformed message fields.

## Root cause

The listeners cast the cross-frame payload directly to `FrameMessage` without runtime validation. TypeScript types do not apply across the `postMessage` boundary.

## Change

Validate every AssistantFrame message variant before dispatching it. Invalid payloads are ignored, while the existing compatibility behavior for guests that serialize an absent tool error as `null` is preserved.

## Verification

- `pnpm vitest run src/model-context/frame/host.test.ts src/model-context/frame/provider.test.ts --reporter=dot`
- `pnpm exec oxlint packages/core/src/model-context/frame/types.ts packages/core/src/model-context/frame/validate.ts packages/core/src/model-context/frame/host.ts packages/core/src/model-context/frame/provider.ts packages/core/src/model-context/frame/host.test.ts packages/core/src/model-context/frame/provider.test.ts`
- `pnpm changesets:check`

The added cases fail on the previous implementation because malformed messages throw before dispatch and pass with runtime validation.

## Public surface

- `@assistant-ui/core`: `FrameMessage` now reflects the already-supported nullable tool-result error value.
- Patch changeset included.
- No documentation or template changes.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/7729?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.SPDfKjx637sanybESUF5GJFHcZBGcrynczM2uX7poK6GkiXU5iFjaDUtQf8?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>


Kinfe123:
Addressed the review findings. Required `parameters`, `args`, and successful `result` fields are now checked by presence, the tests assert malformed payloads cannot mutate context or execute tools, and the public `FrameMessage` type remains unchanged.

Kinfe123:
Corrected the over-strict tool-result validation. Result-less responses remain valid and resolve to `undefined`, while invalid error types are ignored and then covered by a valid follow-up response. I also restored the missing-type and invalid tool-cancel rows. The focused suite now passes 59 tests, typecheck passes, and the core size budget remains within range.

PR #7730, opened 2026-09-18
fix(react): stop stale hover updates after unmount

Kinfe123:
## Problem

When a message root initially matches `:hover`, it schedules a microtask to synchronize hover state after hydration. If that message unmounts before the microtask runs, cleanup sets hover to false but the queued callback can restore it to true on the stale message runtime. The React and Vue primitives shared this race.

## Root cause

The queued hover callback outlived ref or scope cleanup and had no way to tell that its element was no longer mounted.

## Change

Track whether each hover binding is still active and ignore its queued synchronization after cleanup. The React and Vue implementations use the same lifecycle guard. Mouse enter, mouse leave, and mounted hydration synchronization remain unchanged.

## Verification

- `pnpm --filter @assistant-ui/react exec vitest run src/primitives/message/MessageRoot.test.tsx`
- `pnpm --filter @assistant-ui/vue exec vitest run src/__tests__/primitives-structural.test.ts`
- `pnpm --filter @assistant-ui/react typecheck`
- `pnpm --filter @assistant-ui/vue typecheck`
- focused oxlint checks
- `pnpm changesets:check`

The React tests use the real runtime and ref composition. They confirm hydration sets hover while mounted and that unmount remains the final update. The Vue regression covers the same unmount race.

## Public surface

- `@assistant-ui/react` behavior changes and includes a patch changeset.
- The private `@assistant-ui/vue` package receives the parity fix.

<a href="https://rupic.dev/s/s2.Kb0P8YAqDH76i9BaXtDK0z2I5A3CsvOlDnSZKY8iVtF9BxJrThsBHCC7tO0?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>


Kinfe123:
Addressed the review feedback. The React regression now runs through the real runtime and real ref composition, covers both mounted hover synchronization and unmount cleanup, and the Vue primitive now applies the same stale-microtask guard with its own regression test.

Kinfe123:
Reworked both regressions so the message scope and a sibling state reader stay mounted while only the message root is removed. I also added file-level mock restoration in the Vue suite.

I verified the tests discriminate the fix by temporarily reverting only the two active guards:

- React: 1 failed, 1 passed. The cleanup case received `data-hovering="true"` instead of `"false"`.
- Vue: 1 failed, 5 passed. The cleanup case received `data-hovering="true"` instead of `"false"`.

After restoring the guards, React is 2/2 passing and Vue is 6/6 passing, with both package typechecks passing.

Kinfe123:
Narrowed the Vue microtask control to the single hover callback. The spy is installed immediately before mount with `mockImplementationOnce`, restored immediately after mount, and normal microtask behavior is active again before the component update and final unmount. The Vue suite remains 6/6 passing and typecheck passes.

Kinfe123:
Confirmed again at the current commit after strengthening the callback capture. With only the Vue active guard reverted, the focused suite is 1 failed and 5 passed: the cleanup test receives data-hovering="true" instead of "false". Restoring the guard gives 6/6 passing. The test now captures every microtask queued during mount, asserts that at least one callback was captured, and replays them after unmount, so it no longer assumes the hover callback is the first queued microtask.

PR #7731, opened 2026-09-18
fix(react): relock reasoning after reparenting

Kinfe123:
## Problem

Reasoning content can move between scroll containers through portals or layout changes. After its first animation, `useScrollLock` keeps locking the original container, leaving the current container unlocked and modifying the wrong element.

## Root cause

The hook caches the first scrollable ancestor for its entire lifetime and never invalidates that cache when the animated element is reparented.

## Change

Resolve the nearest scrollable ancestor whenever a lock begins. The previous lock is still cleaned up first, so styles and listeners move cleanly from the old container to the current one.

## Verification

- `pnpm vitest run src/primitives/reasoning/useScrollLock.test.tsx --reporter=dot`
- `pnpm exec oxlint packages/react/src/primitives/reasoning/useScrollLock.ts packages/react/src/primitives/reasoning/useScrollLock.test.tsx`
- `pnpm changesets:check`

The added regression test locks one container, reparents the animated element, and confirms the old container is restored while the new container receives the lock. It fails with the previous permanent ancestor cache.

## Public surface

- Patch changeset included.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/7731?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.uqYj7KcSWhlyiKxHGfWdKFFzG3kvrWEMGoJxSpRsQGAruM0r9VSOJagaGr4?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>


Kinfe123:
Updated the changeset wording so it covers both reparenting and changes to the mounted element’s ancestor chain.

Kinfe123:
Handled this as well. When the animated element no longer has a scrollable ancestor, the stored cleanup is now cleared after restoring the old container. The added regression mutates the old container after that transition and confirms unmount does not replay stale styles. The focused suite passes 8/8 and the React package typecheck passes.

PR #7732, opened 2026-09-18
fix(react): collapse compact composer after text shrinks

Kinfe123:
## Problem

A compact composer expands when wrapping text reaches multiple visual rows, but it stays expanded after that text shrinks back to one nonempty row. It only becomes compact again after the input is completely emptied.

## Root cause

The textarea height callback sets the compact context's multiline state to `true` when it detects more than one row, but never reports `false` for a later one-row height.

## Change

Report the calculated multiline boolean on every nonempty height change. Existing empty-input handling remains unchanged.

## Verification

- `pnpm vitest run src/primitives/composer/ComposerInput.test.tsx --reporter=dot`
- `pnpm exec oxlint packages/react/src/primitives/composer/ComposerInput.tsx packages/react/src/primitives/composer/ComposerInput.test.tsx`
- `pnpm changesets:check`

The regression test drives the real height callback from two rows back to one and confirms the compact context receives `false`. It fails on the previous one-way multiline latch.

## Public surface

- Patch changeset included.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/7732?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.iwIeWHgBgD180wUB9sjdOXkxY0eKZmVk0qkb_JwjfYJ9Nf-n8nVipcnVe7w?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>


Kinfe123:
Closing this after validating the existing compact-mode contract. The multiline state is intentionally a one-way latch until the input is cleared because compact and expanded layouts can give the textarea different widths. Clearing the latch from the height callback can make the input alternate between one and two rows and trigger a layout update loop. The current behavior is documented in the composer guide, so this change would be a regression rather than a bug fix.

PR #7733, opened 2026-09-18
fix(react): clear disabled dropzone drag state

Kinfe123:
## Problem

If an attachment dropzone becomes disabled while a file drag is active, the old drag state can return when the dropzone is re-enabled. Disabled drag-leave and drop handlers return early, so no later event clears the latch.

## Root cause

Drag state was only cleared by enabled drag-leave and drop events. A prop transition to `disabled` left `isDragging` set even though the disabled UI hid the attribute.

## Change

Suppress `data-dragging` immediately while disabled and clear the drag latch when the disabled transition commits. Re-enabling starts from a clean state. If the browser drag is still in flight, a later drag event can correctly highlight the enabled dropzone again.

## Verification

- `pnpm --filter @assistant-ui/react exec vitest run src/primitives/composer/ComposerAttachmentDropzone.test.tsx`
- `pnpm --filter @assistant-ui/react typecheck`
- focused oxlint checks
- `pnpm changesets:check`

The regression test begins a file drag, disables and re-enables the mounted dropzone, confirms the stale highlight stays cleared, then confirms a new drag event can activate it again.

## Public surface

- Patch changeset included.


Kinfe123:
Addressed both comments without adding an effect or a lint exception. Drag state is now tied to a memoized session identity that changes whenever `disabled` changes, so an old drag can never become visible again after re-enabling. This removes the render-phase state update and makes the `data-dragging` condition single-purpose. The production React build and the focused 10-test suite pass. I also checked the emitted file: this anonymous `forwardRef` callback does not receive a compiler memo cache on either main or this branch, so the original line was not newly causing a compiler bailout.

Kinfe123:
Replaced the memoized identity mechanism with the requested effect. The `!disabled` render guard hides the attribute in the disabling commit, and the effect clears the boolean before a later re-enable can revive it. The repository normally rejects synchronous state resets in effects, so this has one narrowly scoped lint suppression with the browser drag-session reason beside it. Focused lint, 10 tests, React typecheck, and the production package build all pass.

PR #7741, opened 2026-09-19
fix(react): restore the packages/react typecheck and format baseline

Kinfe123:
## Problem

`packages/react` does not typecheck or format-check on `main`:

```
src/unstable/webmcp/convertTools.test.ts(210,25): error TS2339: Property 'unit' does not exist on type 'WeatherArgs'.
packages/react/src/primitives/actionBar/ActionBarExportMarkdown.test.tsx — format issues found
```

Both are invisible on `main` because "Typecheck Changed Packages" and the format check only run against the packages a PR changes, and nothing re-runs them for `packages/react` after a merge. They surface on the next PR that touches the package, which is why #7738, #7390 and #7392 all went red on failures they did not introduce.

## Root cause

`convertTools.test.ts` declares `type WeatherArgs = { city: string }` and pins its fixtures to `Tool<WeatherArgs, unknown>`. The test added in #7068 supplies a two-field schema and destructures `unit` in `execute`, which `WeatherArgs` does not have.

Widening `WeatherArgs` alone is not enough. `FrontendTool` declares `parameters: StandardSchemaV1<TArgs> | JSONSchema7`, and `StandardSchemaV1<TArgs>` pins a schema's input and output to one type. The schema under test uses `z.enum(["c", "f"]).default("c")`, so its input (`unit` optional) and its parse output (`unit` present) genuinely differ, and no single `TArgs` satisfies both.

The format failure is an unformatted `mockImplementation` callback that landed with #7520.

## Change

- Widen `WeatherArgs` with an optional `unit` so `execute` can destructure it.
- Cast that one schema to `FrontendTool["parameters"]`, with a comment recording why the type cannot express it.
- Run `oxfmt` on `ActionBarExportMarkdown.test.tsx`.

Test-only; no source or public surface change.

## Verification

- `pnpm exec tsc --noEmit` in `packages/react` — clean (fails on `main` with the error above)
- `pnpm exec vitest run src/unstable/webmcp/convertTools.test.ts src/primitives/actionBar/ActionBarExportMarkdown.test.tsx` — 42 passed, no type errors
- `oxlint` — no errors; `oxfmt --check .` — all 4432 files correctly formatted (1 file fails on `main`)

## Public surface

None. Two test files; no changeset, since the Changeset Semver Check exempts tests.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/7741?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.LW6FCwa8wDpIc8nlKll51LOFMYF3QyWWuwPcC44Sj_dU5m6cn8RkkIUXpm8?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>


Kinfe123:
Added a third baseline repair: `@assistant-ui/core ./internal` was recorded at 32918 gzip but main's built dist measures 33614 (+2.1%, past the 256 B / 2% tolerance).

Building `packages/react` rebuilds `core` as a dependency, so the size gate evaluates that entry and fails on every PR that touches react — the same latency as the other two defects here. `pnpm size:update:all` moved exactly one entry, so this is isolated drift rather than a broad re-baseline.

`pnpm size:check` now exits 0.

Kinfe123:
Good catch, and you were right — the assertion was unnecessary and I have removed it in 7dc5650.

The real cause was not that `Tool` cannot express the schema. It was `exactOptionalPropertyTypes`: under it, `unit?: "c" | "f"` excludes `undefined`, while the schema's inferred input is `unit?: "c" | "f" | undefined`. Those are different types, so assignment failed and I misread the union error (TS reported the `JSONSchema7` arm, complaining about Zod's `required` method) as the schema being inexpressible.

Widening the optional to include `undefined` is enough:

```ts
type WeatherArgs = { city: string; unit?: "c" | "f" | undefined };
```

`parameters` now takes the Zod schema directly, the input relationship reaches `execute` as `StandardSchemaInput` intends, and nothing is masked. `tsc --noEmit` on `packages/react` is clean and the file's 40 tests pass.

PR #7794, opened 2026-09-20
test(x-buildutils): exercise declared import guards through real builds

### LINKED ISSUE #7753: x-buildutils: the undeclared-import guard wiring has no fixture build test
## problem

#7752 covers the pure pieces of the undeclared-import guard (`packageSpecifierName`, `declaredImports`, `undeclaredTypeReferences`) under `node --test`. Nothing exercises the wiring in `packages/x-buildutils/src/index.ts`: the `deps.onlyImport` option handed to tsdown and the post-build declaration walk. Dropping either call site leaves the unit suite green, and a tsdown bump that changes `onlyImport` semantics (exempting builtins on `platform: "neutral"`, visiting fewer node kinds) would widen the guard with every package still building.

## why it is not a quick add

A fixture package in a tmpdir cannot resolve a planted import, so tsc reports TS2307 before the guard sees anything; the hand verification on #7746 and #7752 planted imports that resolve but are undeclared (`hast` through the hoisted `@types/hast`, `vitest` through a sibling's node_modules). A fixture therefore has to live where those packages resolve, inside `packages/x-buildutils` but outside `files`, and outside the `packages/*` workspace glob, and run the real `bin/aui-build.js` against it.

## direction

One fixture per guard shape (an import statement, an inline `import("pkg").Type`, a `/// <reference types="pkg" />`), each expected to exit 1 with the guard's message, plus a clean fixture expected to build, run from the same `node --test` suite with a generous timeout since each `aui-build` takes a few seconds.

<a href="https://rupic.dev/s/s2.IE_XwMba_N5PRbvmj3imh4gIjHRyVdCVYc7umWUrxtmP2UOydLd5dbQLfRA?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

Gujiassh:
## Problem

Closes #7753. The existing helper tests do not exercise the guard wiring in `aui-build`; disconnecting either check leaves those tests green.

## Change

Add four integration tests that invoke the real CLI against disposable packages with a local, resolvable dependency:

- Undeclared statement imports must fail with the `onlyImport` diagnostic.
- Undeclared inline import types and triple-slash references must fail in the emitted-declaration check.
- Declaring the dependency must produce working JavaScript and preserve both declaration reference forms.

Negative cases reject missing-module errors as false positives. Fixtures require no downloads and are cleaned up after each test.

## Verification

- `node --test "packages/x-buildutils/src/**/*.test.ts"`: 15 passed, including all four CLI fixtures.
- `node packages/x-buildutils/node_modules/typescript/bin/tsc6 -p packages/x-buildutils/tsconfig.json --noEmit`: passed.
- oxlint, oxfmt and `git diff --check`: passed.
- Temporarily removing `onlyImport` fails the statement case; removing the final declaration check fails the inline and reference cases. Restoring production code passes all 15 tests.
- Package dry-run: 13 entries, no tests or fixtures included.

Validated on Windows with Node 24.19.0. Full monorepo and Linux CI have not been run locally.

## Public surface

None. Production sources and manifests are unchanged; the existing test exclusion keeps this file out of the published package. No changeset is required for this test-only addition.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/7794?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.uAQLaUNyzoesqUQ6YeGdt5c5qAsMJqGwGGIpltCleFa2ET8B37LPetYzlPs?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>


okisdev (approved):
approving, this covers both call sites for real. i ran the suite in a worktree with its own install (15 passed, the four fixture builds under a second each on macOS) and mutation tested the wiring: dropping `onlyImport: allowedImports` fails only the statement case, dropping the trailing `assertDeclaredTypeReferences()` fails only the inline and reference cases, restoring either brings all 15 back. the Test Changed job log shows the same four tests ran on Linux.

planting the dependency inside the fixture's own `node_modules` also disproves the premise in #7753 that a tmpdir fixture cannot resolve a planted import, so the in-tree fixture and the `assertDeclaredTypeReferences(dir, allowed)` lift the issue sketched are both unnecessary now. the `preserve="true"` on the reference directive is load bearing (without it `assertPreservedDirectives` rejects the fixture before the guard runs), and the `doesNotMatch(/TS2307|Cannot find module/)` assertion is what keeps the negative cases honest.

resolved the two outdated threads from the first push since 4f30f71 addressed them.

PR #7795, opened 2026-09-20
fix(ai-sdk): keep host approval answers with the chat

### LINKED ISSUE #7646: ai-sdk: a host approval answer does not survive a runtime remount over the same chat
follow-up to #7079.

with `onRespondToToolApproval` set, `useAISDKRuntime` holds a host answer in runtime state and never writes it into the `useChat` messages, because an `approval-responded` part there satisfies `lastAssistantMessageIsCompleteWithApprovalResponses` and gets posted to the chat route. that state lives exactly as long as the runtime. a runtime mounted again over the same `Chat` starts empty, so the request renders open until the chat itself records the resolution, and a second answer reaches the host.

first-party case: `AISDKThreads` without `cloud` mounts only the visible thread and keeps each `Chat` in its map, so switching away from a thread and back after a host answer reopens the request. a caller-owned `Chat` passed to `useAISDKRuntime` behind a conditional render takes the same path.

the docs currently tell hosts to bring the resumed run back into the chat and to refuse a second resume. a fix keeps the answer with the `Chat` instead of the runtime: for `AISDKThreads`, on the per-thread entry that already carries `messageRepositoryInstance`; for a caller-owned `Chat`, through a store keyed by that chat which the runtime can reach.

<a href="https://rupic.dev/s/s2.bXuOtt0SqTCk0rl8PfPGR-6SsEp9LobMX0B3FL11vaZSlBMfRmV_p9MVY_o?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

Kinfe123:
## Problem

With `onRespondToToolApproval` set, a host answer is deliberately kept out of the `useChat` messages, because an `approval-responded` part there satisfies `lastAssistantMessageIsCompleteWithApprovalResponses` and gets posted to the chat route. Nothing in the chat records the answer, and the runtime held it in `useState`/`useRef`, so it lived exactly as long as the runtime.

A runtime mounted again over the same `Chat` therefore started empty: the request rendered open again and a second answer reached the host. This is the first-party `AISDKThreads` path without `cloud`, which mounts only the visible thread and keeps each `Chat` in its map, so switching away from a thread and back reopened an answered request.

Closes #7646 (follow-up to #7079).

Reproduced on main by the new regression, which fails with the answer gone after the remount:

```
AssertionError: expected { id: 'approval-1' } to match object { id: 'approval-1', approved: true }
```

## Root cause

`toolApprovalResponses` and `hostApprovalIdsRef` are runtime state. The answer has no home outside the runtime, and the chat is the thing that actually outlives it.

## Change

The answer is keyed on the object that owns it. A module-level `WeakMap` maps an owner to its answers; `useAISDKRuntime` seeds its state from that map and writes through to it. `useChatThread` passes the `Chat` it already holds as the owner, alongside the `unstable_messageRepositoryInstance` it already passes for the same reason — per-thread state that has to survive the runtime.

A `WeakMap` keyed on the chat rather than a map keyed on `chat.id`: two different `Chat`s can share an id (a chat torn down and recreated under the same id is the common case), and an id-keyed module map would both hand the new chat the old chat's answers and never release them. Keyed on the instance, the answers are collected with the chat.

Without an owner the behavior is unchanged, so a caller-owned chat passed straight to `useAISDKRuntime` keeps today's runtime-lifetime semantics until it opts in.

## Verification

- New `keeps a host answer when a runtime remounts over the same chat`: one `Chat`, a host answer, a real unmount, then a fresh mount over that same chat. It fails with the write-through removed and passes with it.
- `pnpm -C packages/ai-sdk test`: 324 passed (29 files).
- `tsc --noEmit` and `aui-build` both pass.

## Public surface

`@assistant-ui/ai-sdk`: adds the optional `unstable_hostApprovalOwner` to the runtime adapter. Additive and `unstable_`-prefixed, matching `unstable_messageRepositoryInstance` next to it. Changeset: patch.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/7795?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.879lzJ_L6zrMouZ_6khM15y6CdrvlSWk9EBUNaY1WRLYgxpaUPp23XNI9zE?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>


Kinfe123:
Fixed in acbded2a. All four reviewers found the same blocking bug and were right: `useChatThread` passed the `useChat` helpers object, which is re-minted every render, so the `WeakMap` key was dead by the next render and the production path was not actually fixed. Line 254 of that file already holds the real instance as `externalChat ?? ownedChat`; the owner now uses it.

My regression passed only because it hand-passed `chatInstance` straight to `useAISDKRuntime`, so it proved the mechanism and never touched the wiring this PR changes. That is the more useful half of the review, so there is now a second regression that mounts through `useChatThread` itself, the path `AISDKThreads` takes. It fails with the owner reverted to the helpers object and passes with the instance.

Also fixed cubic's P1 on a runtime kept mounted across a change of owner: `hostApprovalIdsRef` and the response state are both reseeded from the new owner, so a reused approval id in a different chat is no longer rendered as already answered.

Full suite 325 passed (29 files), tsc and aui-build clean.

Separately, the `cancelledMessages` field in the same file has this identical lifetime defect and is now filed as #7796 rather than folded in here.

One note on `Template Sync`: it failed in toolchain setup with `pnpm runtime set node 24 -g exited with code 1`, before running any sync logic, and this diff touches only `packages/ai-sdk`. GitHub refused a rerun of that workflow, so it may need a maintainer.

Kinfe123:
Both remaining points are in 4b865b3b.

**Retirement.** A stored answer is only needed while the chat has no record of the outcome, so it is now dropped as soon as the chat stops reporting the request as `approval-requested`, with `hostApprovalIdsRef` and the response state synchronised in the same pass. That closes the two edges you named: `convertMessage.ts` can no longer apply a retained answer by id against a resolution the chat already owns, and the per-chat record no longer grows without bound.

Regression: `retires a stored answer once the chat records the outcome`. It answers through the host, flips the chat's tool part to `output-available`, remounts, and asserts nothing is re-applied. With the retirement pass disabled it fails with `expected true to be undefined`, which is the stale answer leaking through.

The approval record itself stays on the part, since that one belongs to the chat; what is retired is our copy of the answer, so the assertion is on `approval.approved` rather than on `approval`.

**Contract.** The JSDoc and `runtimes/ai-sdk/v7.mdx` both still described the pre-#7795 behaviour, that the answer lasts as long as the runtime and a remount reopens the request. That was my omission in the original commit: I changed the behaviour and left the contract asserting the old one. Both now state that with an owner the answer belongs to the chat and is retired once the chat reports the outcome, and that without an owner the previous runtime-lifetime behaviour is unchanged. The v7 page keeps its `resumeStream` guidance, which still applies to a reload, since that starts from the chat alone.

Full suite 326 passed (29 files), tsc and aui-build clean.

Kinfe123:
Fixed in the push above. You were right twice over, and the second one is the part I had wrong.

My previous attempt at scoping was a no-op. `applyResponse` closed over `ownedApprovals` from its own render, so `ownedApprovals === startedWith` compared a value with itself and was always true. There is now an `ownedApprovalsRef` tracking the record the runtime currently renders, which is the only thing that can answer whether the owner moved.

The deeper one your regression request surfaced: rejection cleanup was gated on `hostApprovalIdsRef.current.has(approvalId)`, and that ref is reseeded when the owner changes. After a switch the guard read false, so the rollback did not run at all and the original owner kept an answer it had refused. The response now tracks its own applied state locally, so the captured record is always corrected, while the id ref and the rendered map are only touched while that owner is still the one on screen.

The regression you asked for is `rolls a rejected answer back to its own owner after the owner changes`: it holds the handler open, switches the runtime to a different owner, rejects, then remounts over the original chat and asserts the answer is gone. It fails against the previous guard with `expected true to be undefined`.

Full suite 327 passed (29 files), pnpm typecheck and aui-build clean.

Kinfe123:
Fixed in 50f0f712f. Both mechanism comments were right, and they turned out to be the same root cause.

**The owner record was shared but not observable.** A response captured `startedWith` and then decided whether to touch this runtime's state with `ownedApprovalsRef.current !== startedWith`. That gate asks "is the record still the one on screen", which is the wrong question when two runtime lifetimes share one record: after an unmount and remount over the same owner, both runtimes see the *same* `Map`, so the gate passes and the rollback writes to the unmounted runtime's refs and `setToolApprovalResponses`. The remounted runtime, seeded once at mount, kept the optimistic answer and its `hostApprovalIdsRef` still held the id, so a retry threw `is not waiting for a response`. The same gap is what left a second runtime mounted over one owner showing stale state.

Rather than add a second guard, the record now announces its writes:

```ts
const hostApprovalListenersByChat = new WeakMap<object, Set<() => void>>();
```

`applyResponse` corrects the captured record and then calls `notifyHostApprovals(startedOwner)`; every mounted runtime re-reads `hostApprovalIdsRef` and `toolApprovalResponses` from the record. The per-runtime direct writes remain only for the no-owner case, where the answer genuinely belongs to that runtime alone. `ownedApprovalsRef` is gone — the question it existed to answer no longer needs asking.

That is the "make shared owner mutations observable" option rather than documenting the topology as unsupported, because the rejection race is reachable in the single-runtime remount flow this PR targets, not only in a concurrent-mount topology.

**Regression test** `reopens the request on the remounted runtime when the handler rejects after the remount`: it holds the handler open, answers, unmounts, remounts (asserting the remount shows it answered), then rejects and asserts the request reopens on the runtime now on screen. It fails against the previous source with the rest of the file green, so it isolates this change rather than the file's existing behavior.

**Test strengthening.** The remount case now also mounts a *different* `Chat` carrying the same id `"chat-remount"` and asserts its request begins unanswered, so an id-keyed record would fail it and the object-identity guarantee is actually pinned.

I have not yet addressed the two remaining test notes on the deferred owner-switch and rollback cases (owner B never gets its own approval state; the reopened runtime is read without waiting for a mounted owner-B runtime). Both are real — say the word and I will strengthen those two as well.

`pnpm --filter @assistant-ui/ai-sdk exec vitest run`: 332 passed (29 files). Typecheck (both programs), oxlint and oxfmt clean. #7805 has been rebased onto this and still passes at 334.

Kinfe123:
Both remaining test notes addressed in 0c5500c92.

**Owner B now holds real state.** The case previously created `ownerB` as a bare `{}` that never answered anything, so it could only show owner A reopening — it could not detect a rollback deleting owner B's same-id answer, which is the thing the case exists to protect. The handler is now owner-aware: A's answer hangs so it can be rejected after the switch, and B's resolves, so B holds a genuine answer under the same `approval-1`. The rollback then has a same-id neighbour to get wrong.

**The mounted owner-B runtime is now observed.** The reopened state was read synchronously off `result.current` with no `waitFor`, and the on-screen owner was never asserted at all. The case now settles each assertion and separates the two owners:

- while still mounted on B, B's answer survives A's rollback (`approved === false`)
- remounted over A, the request is reopened (`approved === undefined`)
- remounted over B, B's answer is still there

**These assertions have teeth**, which the previous ones did not. I mutated the rollback to delete the approval id from every owner record rather than only the one the response started under, and the case fails with the other twelve green. Under the old version that mutation passed unnoticed.

Full suite 332 passed (29 files), typecheck both programs, oxlint and oxfmt clean. #7805 rebased on top and green at 334.

Kinfe123:
Latest round in 23baa1009.

**Duplicate answer after a remount** (`:135`). The first remount case now uses a `vi.fn` handler and asserts that answering again after the remount throws `not waiting for a response` and that the handler was called exactly once — so the case pins that the remounted runtime rebuilt its answered-id set, not merely that the rendered approval looks answered.

On the other half of that note: the first-party `useChatThread` wiring is exercised, just by the sibling case `keeps a host answer across a useChatThread remount over one chat`, which mounts through `useChatThread` and is what catches an unstable helper object being used as the owner. The manual-`chatInstance` case is deliberately the narrower unit. The duplicate-rejection assertions requested on `:284` are already in that sibling case (handler spy, `toThrow`, `toHaveBeenCalledOnce`) — that thread looks stale against the current file.

**cubic P2 on `:364` — the premise does not hold.** The comment states the stored answer is never removed because "`applyResponse(false)` only runs on the error or `respondViaAISDK` path". There is a third remover: the retirement effect, which runs on every `chatHelpers.messages` change and drops any stored answer whose request the chat no longer reports as `approval-requested`:

```ts
for (const approvalId of [...ownedApprovals.keys()]) {
  if (stillRequested.has(approvalId)) continue;
  ownedApprovals.delete(approvalId);
  hostApprovalIdsRef.current.delete(approvalId);
  retired = true;
}
```

So an answered approval is retired as soon as the chat records its outcome, and the record does not grow for the chat's lifetime. `retires a stored answer once the chat records the outcome` covers this.

The residual sliver is narrower than described: a *new* request re-issuing the *same* approval id while the old entry is still live. That needs the provider to mint a colliding approval id across runs — ids are provider-generated and unique per request, so the auto-apply cannot fire for an unrelated request under any provider that does not reuse ids. If you want belt-and-braces I would correlate the stored answer with its `toolCallId` and drop it when the id is reattached to a different tool call; that is a record-shape change I would rather not fold into this PR at this point, and #7805 stacks on the same structure. Happy to open it separately if you want it.

332 passed (29 files), typecheck both programs, oxlint and oxfmt clean. #7805 rebased on top, green at 335.

Kinfe123:
Addressed in af87883ab. This one was right and it was a regression, not a nit.

**Retiring on absence was wrong.** The pass keyed on the approval id still appearing as `approval-requested`, but absence is a normal transient state here: `switchToBranch` calls `chatHelpers.setMessages(<other branch>)`, so the answered part simply leaves `messages`. The answer was then deleted from the owner record, from `hostApprovalIdsRef` and from `toolApprovalResponses` — and switching back rendered the request open, with `respondViaHost`'s guard passing, so the host handler could answer the same request twice. `onDelete` and any host `setMessages` that scopes the list are the same shape, and it was on by default because `useChatThread` always passes an owner.

**Now it retires on a recorded outcome.** Each stored answer carries the `toolCallId` it belongs to (`requested.toolCallId` was already in hand), and an entry is retired only when a part with that tool call is present *and* no longer `approval-requested`:

```ts
if (isToolUIPart(part) && part.state !== "approval-requested") {
  resolvedToolCallIds.add(part.toolCallId);
}
...
if (!resolvedToolCallIds.has(entry.toolCallId)) continue;
```

Matching on the tool call rather than the approval id also covers `completePendingToolCalls` stripping `approval` when it rewrites a part to `output-error`, as you noted. An entry whose message is deleted outright now lingers for the chat's lifetime, which is what the `WeakMap` already bounds — the trade you described.

**The non-blocking point is fixed too:** the retirement pass now ends with `notifyHostApprovals(approvalOwner)` instead of setting only its own state, so a second runtime over the same owner sees the retirement. That removes the contradiction with the invariant the rest of the commit establishes.

**Regression test** `keeps a host answer while the answered part is out of the visible messages`: answer, `setMessages([])`, `setMessages(answered)`, then assert the approval is still answered, a second answer throws, and the handler was called exactly once. It fails against the absence-based rule (the previous behavior on this branch) and passes now.

One correction to my earlier reply to cubic on the same effect: I defended the retirement pass as correct because it *did* remove entries. It removed them on the wrong signal, which is what you caught. The `toolCallId` correlation cubic suggested as belt-and-braces is now in, for this reason rather than the id-collision one.

333 passed (29 files); typecheck both programs, oxlint and oxfmt clean. #7805 rebased on top and green at 336.

Kinfe123:
Worked through all eight open threads. Two were live; six are stale against the current branch. Evidence for each below so you can spot-check rather than take my word.

**Live — the subscribe window** (`useAISDKRuntime.ts:380`). You were right: state is seeded during render, the subscriber attaches in a passive effect, and a rollback resolving in that gap was never seen. Fixed by reconciling when the subscription attaches rather than only on later notifications:

```ts
const unsubscribe = subscribeToHostApprovals(approvalOwner, sync);
// The state was seeded during render, so a write landing between then and
// this subscription would otherwise never be seen.
sync();
```

`sync` bails out through the updater (`return unchanged ? prev : next`) when the record already matches, so the common mount path still commits nothing extra — the suite's render-count contracts are unchanged. I used reconcile-on-subscribe rather than a layout subscription because it fixes the same window without moving this work into the commit phase.

**Live — the timing case** (`test.tsx:283`), and here I owe you an honest answer rather than a green test. I could not construct that window deterministically: `renderHook` flushes passive effects inside its own `act`, so between the seed and the subscription there is no point a test can inject the rejection. Rejecting while nothing is mounted does not exercise it either — the next mount seeds from the already-corrected record and passes with or without the fix, so that assertion would have no teeth. I would rather say that than add a case that looks like it pins the window and does not. The reconcile closes it by construction; if you know a way to drive that gap under RTL I will add it.

**Stale — already in the branch.** Each of these was addressed in an earlier round; the line numbers moved, which is likely why the threads did not auto-outdate:

- `test.tsx:135` and `:351` (spy + duplicate-rejection on both remount cases) — `not waiting for a response` is asserted 4 times, with 5 `vi.fn<ApprovalHandler>` spies.
- `test.tsx:356` (fresh `Chat`, same id) — the `twin` chat, asserting its approval begins unanswered.
- `test.tsx:493` (owner B needs real state) — the handler is owner-aware now; B answers the same `approval-1`, and the case asserts B survives A's rollback, A reopens, and B is still answered on remount.
- `useAISDKRuntime.ts:685` (shared ledger not observable) — `notifyHostApprovals` / `subscribeToHostApprovals`.
- `useAISDKRuntime.ts:375` (cubic P2) — resolved by the same change the blocking review asked for: answers now carry `toolCallId` and retire on a recorded outcome, which is exactly the "correlate to the request rather than the approval id" option that comment suggested.

333 passed (29 files), typecheck both programs, oxlint and oxfmt clean. #7805 rebased on top, green at 337.

I am resolving the six stale threads and leaving both live ones open — the second one for your call on whether "closed by construction" is acceptable there.

Issue #7796, opened 2026-09-20
ai-sdk: a cancelled message loses its mark when a runtime remounts over the same chat


Kinfe123:
## problem

`cancelledMessages` in `packages/ai-sdk/src/runtime/useAISDKRuntime.ts` is runtime state scoped by chat id:

```ts
const [cancelledMessages, setCancelledMessages] = useState<{
  chatId: string;
  ids: ReadonlySet<string>;
} | null>(null);
```

Nothing in the `useChat` messages records that a response was stopped, so the mark lives exactly as long as the runtime. A runtime mounted again over the same `Chat` starts empty and the stopped response renders as a complete answer instead of a cancelled one.

This is the same lifetime defect as #7646, one field over. `AISDKThreads` without `cloud` mounts only the visible thread and keeps each `Chat` in its map, so switching away from a thread and back loses which messages were cancelled.

It is a weaker symptom than #7646, which is likely why it has gone unreported: the state is visual, not a second host callback.

## mechanism

`cancelledMessageIds` feeds the converter through `toThreadMessages`, which is what turns the stopped assistant message into an `incomplete` / `cancelled` status. The three sites are the `useState` above, `retractCancellation`, and the `onCancel` handler that adds the id.

## proposal

#7646 introduced a `WeakMap` keyed on the `Chat` for host approval answers, with `useChatThread` supplying the owner. Cancelled ids have the same owner and the same lifetime, so they belong in the same per-chat record rather than a second mechanism.

## note on the test

I attempted this while fixing #7646 and backed it out because I could not get the regression to drive the code: a transport whose stream only enqueues `start` and `start-step` leaves `chat.status` at `submitted` and never reaches `streaming`, so the cancel path was not exercised. Two things the next attempt should carry over:

- the first version of the test leaked a floating promise and an unclosed stream, which broke six sibling tests through cross-test contamination; closing the stream and awaiting the send fixed it
- the owner must be read before the `cancelledMessages` `useState`, or seeding it hits a TDZ `ReferenceError`

The implementation itself was straightforward; only the harness needs work.

[...]

PR #7805, opened 2026-09-20
fix(ai-sdk): keep cancellations with the chat

Kinfe123:
## Problem

Nothing in the `useChat` messages records that a response was stopped. The mark lived in runtime state:

```ts
const [cancelledMessages, setCancelledMessages] = useState<{
  chatId: string;
  ids: ReadonlySet<string>;
} | null>(null);
```

so it ended with the runtime. A runtime mounted again over the same `Chat` started empty and the stopped response rendered as a complete answer instead of a cancelled one. `AISDKThreads` without `cloud` mounts only the visible thread, so switching away from a thread and back lost which messages were cancelled.

Closes #7796. This is the same lifetime defect as #7646, one field over.

## Root cause

`cancelledMessageIds` feeds the converter, which is what turns the stopped assistant message into an `incomplete` / `cancelled` status. With the ids held per runtime, that input is empty on the next mount.

## Change

Stacked on #7795, which introduced a per-chat record keyed by the `Chat` through `unstable_hostApprovalOwner`. Cancelled ids have the same owner and the same lifetime, so they join that record rather than getting a second mechanism: the record becomes `{ approvals, cancelledIds }`, the state seeds from it, and the two mutation sites (`retractCancellation`, and the `onCancel` handler that adds the id) write through. The existing pruning of ids no longer present in the chat is applied to the owned set too, so it does not grow across turns.

Without an owner the behaviour is unchanged.

## Verification

- New `keeps the stopped output cancelled when a runtime remounts over the same chat`, built on the existing `createChatHelpers` harness: stop a streaming response, confirm the cancelled status, unmount, mount again over the same owner. It fails with the seed removed and passes with it.
- `pnpm -C packages/ai-sdk test`: 326 passed (29 files).
- `tsc --noEmit` and `aui-build` both pass.

## Public surface

None beyond #7795's `unstable_hostApprovalOwner`, which this reuses. Changeset: patch.

> Note: stacked on `fix/ai-sdk-host-approval-owner` (#7795) because it builds on that owner record. Retarget to `main` once #7795 lands.

<a href="https://rupic.dev/s/s2.tKfSvjJsnWIOpcjPg19Fle7Hu8xXYSf3bOIAkRCKcXB2hDoBHxFFpSJkvwE?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/7805?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Kinfe123:
Fixed in c67df27fb. All three reviews found the same gap and were right.

`cancelledMessages` was seeded only in the `useState` initializer, so a runtime kept mounted across a change of `unstable_hostApprovalOwner` never read the new owner's `cancelledIds`. React preserves the old state, and the `chatId` filter then either keeps exposing the previous owner's set or exposes nothing for the new chat — so a response the new chat had stopped rendered as a complete answer, which is the exact loss this PR exists to prevent.

The fix mirrors the `lastApprovalOwnerRef` branch that already re-seeds the approval state, because the two are seeded from the same owner record and have to be re-read together:

```ts
setCancelledMessages(
  owned && owned.cancelledIds.size > 0
    ? { chatId: chatHelpers.id, ids: new Set(owned.cancelledIds) }
    : null,
);
```

Regression test `adopts the new owner's cancelled output when the owner changes while mounted`. It seeds owner B's set from a separate mount that stops a run, then renders one runtime on owner A and swaps it to owner B via `rerender` without unmounting, asserting the message is not cancelled before the swap and is after it. It fails on the branch without the `setCancelledMessages` call (`expected undefined to match object { type: 'incomplete', reason: 'cancelled' }`) and passes with it.

`pnpm --filter @assistant-ui/ai-sdk exec vitest run`: 333 passed (29 files). Typecheck (both programs), oxlint and oxfmt clean.

Kinfe123:
Inverse transition covered in the rebased branch.

The previous regression only swapped from an uncancelled owner A to an owner B that had a mark, so it proved the mark appears but not that it clears. `clears the previous owner's cancelled output when the owner changes while mounted` now does the other direction: it cancels under owner A, asserts the message reads cancelled, then swaps the mounted runtime to an owner B that recorded nothing and asserts the message is no longer cancelled.

It fails without the `setCancelledMessages` re-seed (the stale mark from owner A survives the swap) and passes with it, so both halves of the defect are now pinned.

335 passed (29 files), typecheck both programs clean.

Kinfe123:
Found and fixed a second instance of the same defect class in this PR, while re-auditing after the `main`-branch retirement fix on #7795.

**The owner's cancelled ids were pruned by visibility.** `cancelRun` built `liveIds` from `chatHelpers.messages` and deleted any stored id not currently in it:

```ts
const liveIds = new Set(chatHelpers.messages.map((m) => m.id));
for (const id of owned.cancelledIds) {
  if (!liveIds.has(id)) owned.cancelledIds.delete(id);
}
```

Absence is not resolution here either. A branch switch rewrites `messages` without resuming anything, so cancelling on branch B deleted branch A's mark from the owner record — permanently, since the owner record is the durable copy. Switching back to A then rendered a stopped response as a complete answer, which is the exact loss this PR exists to prevent.

The local `prev.ids` filter had the same shape, so the mark was lost on screen as well as in the record.

**Fixed by not pruning on absence.** The mark now survives until the chat itself resolves it (the existing retraction path still clears a mark when the provider resumes that response). The set is bounded by the chat's lifetime through the `WeakMap`, which is the same trade accepted for approvals on #7795.

**Regression test** `keeps a branch's cancelled mark when another branch is cancelled`: cancel on branch A, switch to B and cancel there, then remount over the same owner with A restored and assert A still reads cancelled. It fails with the old pruning restored and passes without it.

337 passed (29 files); typecheck both programs, oxlint and oxfmt clean — all verified on exit codes rather than piped output.

Kinfe123:
All three threads here were addressed in earlier rounds; the line numbers shifted, which is likely why they did not auto-outdate. Evidence so you can spot-check:

- **The two `useAISDKRuntime.ts:380` threads** (`cancelledMessages` seeded only in the `useState` initializer). The owner-change branch now re-seeds it alongside the approval state:
  ```ts
  setCancelledMessages(
    owned && owned.cancelledIds.size > 0
      ? { chatId: chatHelpers.id, ids: new Set(owned.cancelledIds) }
      : null,
  );
  ```
- **`useAISDKRuntime.test.ts:312`** (inverse transition unprotected). `clears the previous owner's cancelled output when the owner changes while mounted` cancels under owner A, asserts the mark shows, swaps the mounted runtime to an owner B holding no ids, and asserts the message is no longer cancelled. It fails without the re-seed above.

Since then this branch also picked up a third case from a re-audit: `keeps a branch's cancelled mark when another branch is cancelled`. `cancelRun` had been pruning the owner's ids against the currently visible messages, so cancelling on one branch deleted another branch's mark from the durable record — the same "absence is not resolution" defect as the retirement pass on #7795. That is fixed and pinned too.

337 passed (29 files), typecheck both programs clean, rebased on the current #7795 head.

PR #7813, opened 2026-09-21
fix(react-native): honor registered fallback data renderers

Kinfe123:
## Problem

A fallback data renderer registered through `setFallbackDataUI` is ignored on React Native. A `data` message part whose `name` has no registered renderer falls through to nothing, while the same registration renders on web and ink.

This is the path `useLangGraphRuntime({ ui: { fallback: MyGenUI } })` uses, so a LangGraph `ui` message with an unrecognised name renders on web and disappears on React Native.

Reproduction — identical runtime state and identical registration, rendered twice, once through core's renderer as a control and once through React Native's `MessageContent`:

```
CORE_TEXT: "FALLBACK:7"     <- core honours the fallback
RN_TEXT:   ""               <- React Native drops it
```

## Root cause

`MessageContent` resolved a data part only through the named registry:

```tsx
const Render = useAuiState((s) => {
  const renders = s.dataRenderers.renderers[part.name];
  if (Array.isArray(renders)) return renders[0];
  return renders;
});
```

`dataRenderers.fallbacks` is never read, so a registered fallback can never be selected. Core resolves named renderer → `fallbacks[0]` → inline fallback (`getDataRenderer`), and ink does the same.

The `Array.isArray` branch is also dead: `DataRenderersState.renderers` is typed `Record<string, DataMessagePartComponent[]>`, so the value is always an array and the non-array return is unreachable.

## Change

Resolve the way core does — named renderer first, then the registered fallback:

```tsx
const Render = useAuiState(
  (s) =>
    s.dataRenderers.renderers[part.name]?.[0] ?? s.dataRenderers.fallbacks[0],
);
```

The inline `Fallback` prop still applies after both, so the existing `renderData` behavior is unchanged; this only adds the tier that was missing between them.

**Test fixture correction.** The change surfaced that `MessageContent.test.tsx` modelled a state shape core never produces: its `dataRenderers` had no `fallbacks` key at all, and one case pinned `{ chart: DataRender }` — a bare component where the registry always stores an array. Four tests failed against the corrected resolution for that reason alone. The fixture is aligned with `DataRenderersState` (adds `fallbacks: []`, wraps the pinned renderers in arrays) rather than weakening the fix to accommodate an unreal shape.

## Verification

```
pnpm exec turbo test --filter=@assistant-ui/react-native   # 37 files, 308 passed
pnpm exec oxlint <changed files>                            # clean
pnpm exec oxfmt <changed files>                             # clean
```

Two tests added to the data-parts block: a registered fallback renders for an unregistered name, and a named renderer still wins over a registered fallback. Confirmed the first fails without the source change — `RN_TEXT: ""` against the control's `CORE_TEXT: "FALLBACK:7"` — and passes with it.

The control matters here: my first attempt registered through `useAssistantDataUI`, which takes `{ name, render }` and early-returns without a `name`, so it registered nothing and the empty output proved nothing. Rendering the same state through core's renderer is what established that the registration was real and only React Native ignored it.

## Public surface

`@assistant-ui/react-native` behavior only. No exports or signatures changed. Patch changeset included.

<a href="https://cubic.dev/pr/assistant-ui/assistant-ui/pull/7813?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<a href="https://rupic.dev/s/s2.Gt4bpAp2iCTfAbblPqNwmY_Z-xGlYT4DjdSgBt8QUIuH_vlZYel14T5-XlI?ref=github" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://rupic.dev/badges/track-dark-v1.svg">Track in Rupic</picture></a>


Kinfe123:
Superseded: this landed on `main` independently.

`DataUIDisplay` in `packages/react-native/src/primitives/message/MessageContent.tsx` now resolves the named renderer and falls back to `dataRenderers.fallbacks[0]`, which is the same fix this PR made. The accompanying test coverage is there too — the fixture carries `fallbacks`, and `uses dataRenderers.fallbacks[0] before renderData when no named renderer matches` plus `prefers a named data renderer over dataRenderers.fallbacks` cover exactly what the two tests here asserted.

Nothing in this branch is missing from `main`, so there is no reason to keep it open. The bug was real — the differential repro was `CORE_TEXT: "FALLBACK:7"` vs `RN_TEXT: ""` on identical runtime state — it just got fixed from another direction first.

